PR #6, opened 2022-06-07
Update the README

nevets963:
I went through all the feedback within Discord.

Known knowns:
- I kept the BTC RPC Explorer example as the Uptime Kuma Dockerfile is IMO more complicatd the BTC RPC Explorer one, plus it's a good example where I can explain dependencies and how env. vars. get shared
- A repeat of section 3.1 where I have both instructions for a cloud VM and umbrel-dev
- Is the url for the install script correct?

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,12 +70,11 @@ ___
 
 ## 2. ☂️&nbsp;&nbsp;Packaging the app for Umbrel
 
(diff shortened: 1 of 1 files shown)


mayankchhabra:
Nice work, @nevets963! Made some tiny tweaks. Everything looks great.

PR #10, opened 2022-06-07
Get the Tor HS for the frontend last to prevent race condition

nevets963:


lukechilds:
Merged as https://github.com/getumbrel/umbrel-apps/commit/55d4eaacf2d4fcbfe7d4f8d2440a6af61682e06b

PR #13, opened 2022-06-08
Remove dbcache and maxmempool from Bitcoin bin. args.

nevets963:
Removing `dbcache` and `maxmempool` from the Bitcoin binary arguments so that these rely on the default values and to allow a user to change them in a `bitcoin.conf` config. file

### DIFF
--- a/bitcoin/umbrel-app.yml
+++ b/bitcoin/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: bitcoin
 category: Finance
 name: Bitcoin Node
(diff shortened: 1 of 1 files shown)


bitcoin/umbrel-app.yml
```diff
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: bitcoin
 category: Finance
 name: Bitcoin Node
-version: "22.0"
+version: "22.0-build-1"
```
lukechilds, line 5:
```suggestion
version: "22.0-build-2"
```

PR #15, opened 2022-06-08
Fix Bitcoind Tor port

kroese:
Currently, Bitcoin is configured with `listen=1` to allow for incoming connections. But as no `external-ip` address is specified, this has no effect, because it does not advertise its address to its peers.

This pull adds the `external-ip` to the parameters with the right Tor address, so that incoming connections become possible because Bitcoin knows which address to advertize.

This allows Umbrel nodes to use eachother to sync and fixes https://github.com/getumbrel/umbrel/issues/1393


lukechilds:
Hey @kroese appreciate you putting in the time to work on this! Unfortunately I don't think this makes sense as a default option to enable for all users, I've replied in more detail on the related issue: https://github.com/getumbrel/umbrel/issues/1393#issuecomment-1161592831

lukechilds:
@kroese hey sorry for lack of response on this, just to let you know we're happy to merge this, we have an update to the Bitcoin app planned shortly so we're planning to bundle this change into that, thanks for your contribution!

nevets963:
Closing in favour of PR #109

lukechilds:
Thanks for working on this @kroese. We bundled your changes into https://github.com/getumbrel/umbrel-apps/commit/df38830bc9c75f38f84f9076fcc2da0659da2e41 and gave you commit credits.

PR #19, opened 2022-06-10
App Submission: LN-Visualizer

MaxKotlan:
# App Submission

### App name
LnVisualizer

### 256x256 SVG icon
LN Visualizer Logo

### Gallery images
ss1
ss2
ss3
ss4
ss5

### I have tested my app on:


ln-visualizer/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
...
+    networks:
+      default:
+        ipv4_address: "${APP_LNVISUALIZER_WEB_IP}"
+
+  api:
+    image: maxkotlan/ln-visualizer-api:v0.0.20@sha256:a0ab930436a3e223d4eee11114d3714b1bc21bb13fcb3c0280e2d8daca78719a
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - "${APP_LNVISUALIZER_API_PORT}:5647"
```
nevets963, line 24: Is this by design, that you expose the APIs port so that other people can communicate with it, or is it only the `web` container that will talk to this container?

ln-visualizer/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
...
+    volumes:
+      - "${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro"
+    environment:
+      LND_CERT_FILE: "/lnd/tls.cert"
+      LND_MACAROON_FILE: "/lnd/data/chain/bitcoin/${BITCOIN_NETWORK}/readonly.macaroon"
```
nevets963, line 29: `$BITCOIN_NETWORK` will get deprecated in a future version of Umbrel. Please can you use `$APP_BITCOIN_NETWORK`

ln-visualizer/exports.sh
```diff
@@ -0,0 +1,4 @@
+export APP_LNVISUALIZER_WEB_IP="10.21.21.98"
```
nevets963, line 1: As your app id is `ln-visualizer`, can you use this prefix for your env. vars. `APP_LN_VISUALIZER`.

ln-visualizer/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: $APP_LNVISUALIZER_WEB_IP
+      APP_PORT: 80
+
+  web:
+    image: maxkotlan/ln-visualizer-web:v0.0.20@sha256:ab897a98a5b3640f8e26d423ef505537042a1ebd2804ef540d0ecbc88dc562c4
```
nevets963, line 10: The digest here appears to be for a specific arch. If you run `docker pull maxkotlan/ln-visualizer-web:v0.0.20` then at the bottom it will give the digest for multiple architectures. In this case it should be: `sha256:9e4fff7600666ce27d1c835fa66e4a3fb34aabd7018506d749134baf9e8bc481`

ln-visualizer/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
...
+    restart: on-failure
+    stop_grace_period: 1m
+    environment:
+      LN_VISUALIZER_API_URL: "http://${APP_LNVISUALIZER_API_IP}:5647"
+    networks:
+      default:
+        ipv4_address: "${APP_LNVISUALIZER_WEB_IP}"
+
+  api:
+    image: maxkotlan/ln-visualizer-api:v0.0.20@sha256:a0ab930436a3e223d4eee11114d3714b1bc21bb13fcb3c0280e2d8daca78719a
```
nevets963, line 20: The digest for this image should be: `sha256:cfea255f3343dbacf96c5470dcd442f1cc034337dfb46ecfbbc6d5c1978da1a4`

nevets963 (commented):
This looks awesome! And is a really great PR here! Considering the app structure has recently changed for Umbrel, this is very well put together!

I've left some minor feedback.

ln-visualizer/docker-compose.yml
MaxKotlan: After taking a second look at this, I think there is no real reason to expose this port.
The web communicates with the api container through a reverse proxy to the internal ip/port.
So I removed the environment variable APP_LNVISUALIZER_API_PORT and exposing the port externally.

ln-visualizer/docker-compose.yml
MaxKotlan: Good catch. I updated it to APP_BITCOIN_NETWORK

ln-visualizer/exports.sh
MaxKotlan: I agree that makes more sense. I've been pretty inconsistent with the naming so far, so thanks for calling this out.
I changed it to APP_LN_VISUALIZER

ln-visualizer/docker-compose.yml
MaxKotlan: Ah that makes more sense. I wasn't sure which to use.
Changed to `sha256:9e4fff7600666ce27d1c835fa66e4a3fb34aabd7018506d749134baf9e8bc481`

ln-visualizer/docker-compose.yml
MaxKotlan: Changed to `sha256:cfea255f3343dbacf96c5470dcd442f1cc034337dfb46ecfbbc6d5c1978da1a4`

ln-visualizer/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      LN_VISUALIZER_API_URL: "http://${APP_LN_VISUALIZER_API_IP}:5647"
+    networks:
+      default:
+        ipv4_address: "${APP_LN_VISUALIZER_WEB_IP}"
+
+  api:
+    image: maxkotlan/ln-visualizer-api:v0.0.20@sha256:cfea255f3343dbacf96c5470dcd442f1cc034337dfb46ecfbbc6d5c1978da1a4
+    restart: on-failure
+    stop_grace_period: 1m
+    user: 1000:1000
```
MaxKotlan, line 23: Also just wanted to mention that I just added this user line I saw in other umbrel apps.
I can't add it to the web right now because the web needs to override an nginx config, but figured it made sense to add to the api.

MaxKotlan: @kroese oh it does? I was thinking it did the opposite haha. Thought it ran as a lower privileged user or something. Guess I'll be removing that...

MaxKotlan: Ah okay. I think I will keep it for now then. Appreciate you checking.

nevets963 (approved):
Tested on AMD64 and ARM (Pi) :)

@MaxKotlan Approved from me!

@mayankchhabra Could you prepare the app store graphics?

@lukechilds Could you have a final check over?

MaxKotlan:
Awesome. Thanks for looking it over and testing it out @nevets963. Let me know if I need to make any changes before the final release.

lukechilds:
Tested out and looks really cool!

The WebSocket connection keeps dropping for me though. I get this error in the page:

Screenshot 2022-06-27 at 16 34 04

The WebSocket connection is made successfully, requests `initsync` and receives some events:

Screenshot 2022-06-27 at 16 34 58

Then after  a few events are sent gets disconnected. Here is the disconnect event that's logged if it's helpfull:

Screenshot 2022-06-27 at 16 34 09

Are you guys experiencing this?

@nevets963 I noticed in the logs the app proxy is claiming the client disconnected:

```
api_1         | ::ffff:100.111.172.88, 10.21.0.106 requesting initsync
app_proxy_1   | Validating token: b68894131ba05365a82838bc15d1a77c115b60d324d360a9ed399fd73f9a3e8a
web_1         | 10.21.0.106 - - [27/Jun/2022:09:36:23 +0000] "GET /assets/lnvisualizer_fin.svg HTTP/1.1" 200 4513 "http://umbrel-mainnet:5646/03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.5005.115 Safari/537.36"
web_1         | 10.21.0.106 - - [27/Jun/2022:09:36:59 +0000] "GET /api/ HTTP/1.1" 101 57506527 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.5005.115 Safari/537.36"
app_proxy_1   | [HPM] Client disconnected
```

But according to the error show to the user the client seems to think the backend is disappearing. Potentially an issue with the proxy dropping the connection for seam reason?

@MaxKotlan do you have any ideas what else could be causing this?

lukechilds:
I just tested exposing the LnVisualizer port directly without going through Umbrel's app proxy and I hit the same issue so it seems the issue is unrelated to Umbrel's app proxy.

lukechilds:
Same issue in Brave and Firefox so seems unlikely to be a browser issue too.

MaxKotlan:
Thanks for testing this @lukechilds.

Hmm... Looks like I should have done more testing over tor.
I think it might be an issue with how the app manages initial sync requests, and large requests over tor being dropped.

Currently, with the way the app is architected, it downloads the entire lightning network graph data to the browser. This is a fairly large request which in total results in about 100MB or so of data being downloaded.
It works well locally, and is okay when directly connected to a remote host, however over tor, it seems to have issues.

I think tor might be dropping the connection halfway through due to the amount of data being sent. (Network might see it as spam).
After tor drops the connection halfway through an initial sync, the app doesn't seem to reconnect properly.

I think for now I can modify the app so it reconnects properly and continues the initial sync when the connection is dropped.

I do also have plans to change the initial sync process in the future. As the lightning network grows in size, I think the app will have to move to more server-sided processing. Currently everything is client-side which is why all the data is being downloaded at once.

I have plans for this but it might take some time to get to that point.

nevets963:
> I think tor might be dropping the connection halfway through due to the amount of data being sent.

Ah! That explains why I didn't encounter this when I tested LnVisualizer. There is currently a significant DDoS attack against the Tor network: https://status.torproject.org/issues/2022-06-09-network-ddos/

From what I see atm, Tor is pretty slow and brittle so I think a retry mechanism is a good idea.

lukechilds:
@MaxKotlan when you say "over Tor" do you mean accessing the app UI via Tor Browser or when the lightning node is connected to the rest of the network via Tor?

lukechilds:
Just tested again and seems to be working great! It didn't seem to get disconnected this time as I only saw one request to `/api` but either way it's working well.

Unless I missed something it looks like `APP_LN_VISUALIZER_WEB_PORT` is unused so would be great to remove that if it's not needed. Otherwise this looks good to merge!

@mayankchhabra this is ready for gallery assets.

mayankchhabra:
Okay the gallery assets are live at [getumbrel/umbrel-apps-gallery](https://github.com/getumbrel/umbrel-apps-gallery/tree/master/ln-visualizer).

Great work on the app, Max! If you need to update the gallery images or the app icon in the future, feel free to submit a PR to the above repo. Cheers and looking forward to the launch!

lukechilds (approved):

lukechilds:
Ok @MaxKotlan your app is now live on the Umbrel App Store, congrats! 🚀

ln-visualizer/exports.sh
```diff
@@ -0,0 +1,3 @@
+export APP_LN_VISUALIZER_WEB_IP="10.21.21.102"
+export APP_LN_VISUALIZER_WEB_PORT="5646"
```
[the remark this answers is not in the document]
MaxKotlan: I think you're right about APP_LN_VISUALIZER_WEB_PORT being an unused variable. That was left over from the previous version of umbrel. Good catch.

mayankchhabra:
Yay! Congrats on the launch!

PR #30, opened 2022-06-12
Update Bitfeed to v2.3.4

mononaut:
Changes since v2.2.1:

- Visualize transactions replaced/dropped from the mempool.
- Block explorer functionality: look up past blocks and transactions.
- Substantially reduced CPU load on start up.
- Smoother, fancier graphics


lukechilds:
Looks great, thanks @mononaut!

PR #35, opened 2022-06-13
Upgrade LND to 0.15.0 beta

kroese:
I know it is not the final release, but I had a lot of force-closures on my channels due to a bug in 0.14, so I cannot afford to wait for it any longer.

### DIFF
--- a/lightning/docker-compose.yml
+++ b/lightning/docker-compose.yml
@@ -36,7 +36,7 @@ services:
         ipv4_address: $APP_LIGHTNING_IP
   
   lnd:
-    image: lightninglabs/lnd:v0.15.0-beta.rc6@sha256:cdc00ba130b8a2eac1befa4be0173529647434c5ddcb847d5bf590631af84ae7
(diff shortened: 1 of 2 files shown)


nevets963:
@kroese Thanks for submitting several PRs!! We will need to rigorously test the updates before we're able to release these to end-users. I noted that the digests are wrong in the PRs and are for a specific architecture (amd64).

To get a multi-arch digest, you can run: `docker pull <image>` and the it will display the digest at the bottom. In this case for example, it should be: `sha256:0a6fabd235eac34db2fef3203fceefb705290742b050c399228a3a4cca465334`

lukechilds:
Thanks for updating to the release image @kroese!

Tested [f4dec7b](https://github.com/getumbrel/umbrel-apps/pull/35/commits/f4dec7b4bae3d87a2fc6f798e1974d019f0bcb9e) on mainnet amd64, just syncing a new testnet arm node to test.

lukechilds:
Working great for me on arm, thanks again @kroese!

PR #38, opened 2022-06-17
Node-Red App Submission

### LINKED ISSUE #36: unable to install node-red which has dependency of btc-node
unable to install node-red due to missing packages dependency (when not plan to install btc-node)

截圖 2022-06-14 下午10 36 14
d

nevets963:
A standalone (vanilla) Node-Red app for Umbrel.

Tested on AMD64 and ARM (Pi).

Only the app store graphics needs to be taken care of.

Closes #36 

@lukechilds I'll create another PR to update Node-Red (Bitcoin version)


node-red-standalone/umbrel-app.yml
```diff
@@ -0,0 +1,29 @@
+manifestVersion: 1
+id: node-red-standalone
+category: Automation
+name: "Node-RED (Standalone)"
```
lukechilds, line 4: This can just be Node-RED:

```suggestion
name: "Node-RED"
```

lukechilds:
>@lukechilds I'll create another PR to update Node-Red (Bitcoin version)

@nevets963 you can do it in this PR since they are essentially part of the same change.

lukechilds:
Oh sorry my bad I thought you meant just the name update, I realise now you meant you were gonna update the version separately. Yeah that should be a separate PR but nw I'll pull that commit out seperately.

node-red-standalone/umbrel-app.yml
```diff
@@ -0,0 +1,29 @@
+manifestVersion: 1
+id: node-red-standalone
+category: Automation
+name: "Node-RED"
+version: "2.2.2-12"
+tagline: Wire together the Internet of Things
+description: >-
```
lukechilds, line 7: Lets add a note here that this variant of Node-RED has integration with Bitcoin, Lightning and electrs.

node-red/umbrel-app.yml
```diff
@@ -1,8 +1,8 @@
 manifestVersion: 1
 id: node-red
```
lukechilds, line 2: There are two flows in the config for the Bitcoin variant. One is a demo flow accessing Bitcoin Core RPC which we don't want in the standalone variant. Another is a flow for changing the default password which is not currently in this variant but should be.

lukechilds:
Tested and working well, @nevets963 are you able to apply the feedback in the two comments?

@mayankchhabra do we want to ship this with the same gallery assets as the Bitcoin version?

node-red/umbrel-app.yml
nevets963: Shall we instead consider having the App Proxy protect the app instead of using an app password?

lukechilds: One issue with that is you can trigger a Node-RED flow via an HTTP request. That functionality would be broken if we put it behind auth. It doesn't look like they use a predictable path prefix for that we can whitelist like `/api/*` or something it looks like routes are served from the root: https://cookbook.nodered.org/http/create-an-http-endpoint

lukechilds: Just noticed we already do that and are breaking it. I setup a URL `/foo` that returns `Hello world` and I get:

```
$ curl http://umbrel-mainnet:1880/foo
Found. Redirecting to http://umbrel-mainnet:2000/?origin=host&app=node-red&path=%2Ffoo⏎    
```

Ok I guess since we're already breaking it and no ones complained it's fine to continue like this. We could maybe whitelist `/public/*` and explain in the app repo that routes are by default protected but if you prefix your route with `/public/` it will work without auth. Seems like that gives us the best of both worlds, what do you think?

lukechilds: I'm also thinking, what about if we add an optional HTTP user/pass combo that bypasses the proxy. Just as like a more generic solution for this. so e.g if you pinged `http://node-red:16b38f420a21337c0fb7d482050d3d5721ff71120e0967aea7e9ad2c9932f5ee@umbrel-mainnet:1880/foo` the app proxy would let the request through without auth.

Not super high priority but seems like it could be a potential solution for the client app issue if we can't easily whitelist routes.

nevets963: Ok, let's go with whitelisting `/public/*`. I just tested that and is working nicely as expected. I checked whether there are assets already being served from `/public` but I can't find any. As you say, we'll get the best of both worlds.

The only problem I see is; the app update process will not update `settings.json` to disable auth for current installations - but perhaps that is fine as that may contain user-defined settings and we don't want to touch that anyways...?

lukechilds: You mean so we can't remove auth for existing installs? Yeah I think that's ok.

nevets963:
@lukechilds Those changes have been made and tested.

lukechilds:
Was it intentional to leave auth enabled for the Bitcoin variant?

nevets963:
I like the consistency too but then I realised that if `defaultUsername` and `defaultPassword` become empty strings, those existing installs would lose the (default) credentials shown in the app store and users may not know them and then can't login.

PR #39, opened 2022-06-18
App Submission: Squeaknode - core-lightning version

yzernik:
This adds the core-lightning version of Squeaknode to the app store.

It works exactly the same as the existing Squeaknode app, but uses core-lightning as the lightning node.

I also added an additional environment variable for `APP_CORE_LIGHTNING_RPC_FILE_PATH` in the core-lightning app.


nevets963:
Hey @yzernik Thanks for the submission to support Squeaknode with Core Lightning!

Could you change the id to: `squeaknode-core-lightning` and then change the env. variable names accordingly?

Also, to trigger an app update for core-lightning, can you update its app manifest file (`core-lightning/umbrel-app.yml`) to have the version: `0.11.1-build-4`.

Once that's done I'll give it a test :)

Thanks!
Steve

core-lightning-squeaknode/docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+
+  web:
+    image: ghcr.io/squeaknode/squeaknode:v0.3.4@sha256:c5b8b4326b20e487ec0366a4e8af27f8aa894421d267a8e6d78916c8be765e7f
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - "$APP_CORE_LIGHTNING_SQUEAKNODE_GRPC_PORT:$APP_CORE_LIGHTNING_SQUEAKNODE_GRPC_PORT"
+    volumes:
+      - ${APP_DATA_DIR}/sqk:/sqk
+      - ${APP_CORE_LIGHTNING_RPC_FILE_PATH}:/c-lightning/lightning-rpc
```
lukechilds, line 17: This env var won't exist if the user hasn't updated to the newer version of Core Lightning where we've added this new env var to it's exports file.

This will cause issues because it'll resolve to `.` which will mean Squeaknode will essentially break out of the app sandbox and get access to a lot more stuff than it should be allowed to.

Also generally mounting files into Docker as volumes is not a good idea because if the file doesn't exist Docker will automatically create a directory in the same place. I'm not 100% sure but I'd guess the `lightning-rpc` file doesn't exist when Core Lightning starts and it would create it as part of it's startup process. If that's the case this is going to create a race condition on each Umbrel start where if Squeaknode starts before Core Lightning has already started and created it's `lightning-rpc` file Docker will create an empty directory in it's place and Core Lightning will probably have issues starting because of that.

I'm honestly not sure of the best way to resolve this. One solution to resolve the race condition would be to mount the entire Core Lightning dir in, but that's suboptimal since then Squeaknode gets access to Core Lightning private keys. We also don't fix the missing env var issue.

Is Squeaknode able to interact with Core Lightning's new gRPC port? Maybe that would be a cleaner way to interface. I believe It's the same protocol as the unix socket but just exposed over gRPC.

If Squeaknode could interface with Core Lightning over it's network interface then that would resolve both the race condition and the sandbox issue for old versions.

yzernik: @lukechilds  Another option is to use the `–rpc-file` option when you start core-lightning to specify a different location for the rpc file outside of that directory.

yzernik: @lukechilds I updated the PR so now instead of `APP_CORE_LIGHTNING_RPC_FILE_PATH` it exports `APP_CORE_LIGHTNING_RPC_FILE_DIR`.

I added the `--rpc-file` config to core-lightning, and created a new rpc directory to hold the rpc file with a .gitkeep.

yzernik:
@nevets963 App name and core-lightning manifest version updated.

lukechilds:
@yzernik sorry for the delay on this, will review asap. Is there a reason you closed the PR?

yzernik:
@lukechilds I agree that it would be preferable to use gRPC for the connection to core-lightning. But I am currently using the [pyln-client](https://pypi.org/project/pyln-client/) library in the Squeaknode app, which doesn't look like it supports gRPC yet.

core-lightning/docker-compose.yml
```diff
@@ -57,9 +57,10 @@ services:
       - --bind-addr=${APP_CORE_LIGHTNING_DAEMON_IP}:9735
       - --addr=statictor:${TOR_PROXY_IP}:29051
       - --tor-service-password=${TOR_PASSWORD}
+      - --rpc-file=/data/.lightning/rpc/lightning-rpc
```
lukechilds, line 60: Let remove this and use the default path which ends up on the host as `${UMBREL_ROOT}/app-data/core-lightning/data/lightningd/bitcoin/lightning-rpc`. That way we don't need to also update core-lightning and force users to update it before installing squeaknode-core-lightning to avoid issues.

squeaknode-core-lightning/docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+
+  web:
+    image: ghcr.io/squeaknode/squeaknode:v0.3.4@sha256:c5b8b4326b20e487ec0366a4e8af27f8aa894421d267a8e6d78916c8be765e7f
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - "$APP_SQUEAKNODE_CORE_LIGHTNING_GRPC_PORT:$APP_SQUEAKNODE_CORE_LIGHTNING_GRPC_PORT"
+    volumes:
+      - ${APP_DATA_DIR}/sqk:/sqk
+      - ${APP_CORE_LIGHTNING_RPC_FILE_DIR}:/c-lightning/rpc
```
lukechilds, line 17: With the above mentioned change this should become:

```
"${UMBREL_ROOT}/app-data/core-lightning/data/lightningd:/c-lightning
```

squeaknode-core-lightning/docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+    environment:
+      # Bitcoin Core connection details
+      SQUEAKNODE_BITCOIN_RPC_HOST: $APP_BITCOIN_NODE_IP
+      SQUEAKNODE_BITCOIN_RPC_PORT: $APP_BITCOIN_RPC_PORT
+      SQUEAKNODE_BITCOIN_RPC_USER: $APP_BITCOIN_RPC_USER
+      SQUEAKNODE_BITCOIN_RPC_PASS: $APP_BITCOIN_RPC_PASS
+
+      # lightning environment variables
+      SQUEAKNODE_LIGHTNING_BACKEND: "clightning"
+      SQUEAKNODE_LIGHTNING_CLIGHTNING_RPC_FILE: "/c-lightning/rpc/lightning-rpc"
```
lukechilds, line 28: With the above mentioned change this should become

```
SQUEAKNODE_LIGHTNING_CLIGHTNING_RPC_FILE: "/c-lightning/bitcoin/lightning-rpc"
```

lukechilds:
Some suggested changes on how we can push this without running into issues having to force a CLN app update first to avoid broken state. Its sub optimal because the entire CLN dir gets mounted into Squeaknode but there's no significant difference in security since Squeaknode already has total control of CLN via the RPC socket.

In the future we should transition away from allowing direct filesystem access to CLN and instead require apps to use the CLN rpc port.

nevets963:
Hey @yzernik Just a friendly ping to hear your thoughts on @lukechilds' suggested changes?

yzernik:
Sure @nevets963 @lukechilds!

I'm busy until the start of next week, I'll have a look then. Thanks for reviewing this.

nmfretz:
Hey @yzernik is this PR still something you would like to keep open and working on?

nmfretz:
Hi @yzernik I am closing this PR for now due to inactivity, but please don't hesitate to reopen if/when you want to work on this again!

PR #42, opened 2022-06-22
Update Urbit app to 0.4.1

mopfel-winrux:
Updatd the urbit binry to v1.9 and added some extra instruction on how to use the app

### DIFF
--- a/urbit/docker-compose.yml
+++ b/urbit/docker-compose.yml
@@ -8,7 +8,6 @@ services:
   manager:
     image: mopfelwinrux/urbit-umbrel:v0.4.1@sha256:083db8cdb2109f29c65e186b47eeecdacd5d8021ce1c4593cf8a9f945f646171
     ports:
-      - "${APP_URBIT_PORT}:8090"
       - "${APP_URBIT_P2P_PORT}:34343"
     volumes:
       - ${APP_DATA_DIR}/piers:/data/piers
--- a/urbit/docker-compose.yml
+++ b/urbit/docker-compose.yml
@@ -1,5 +1,10 @@
 version: "3.3"
 services:
+  app_proxy:
+    environment:
+      APP_HOST: $APP_URBIT_IP
+      APP_PORT: 8090
+
   manager:
     image: mopfelwinrux/urbit-umbrel:v0.4.1@sha256:083db8cdb2109f29c65e186b47eeecdacd5d8021ce1c4593cf8a9f945f646171
     ports:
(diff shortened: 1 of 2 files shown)


nevets963 (commented):
@mopfel-winrux Thanks for the app update!

Is there some reason that you've removed the app_proxy service from Urbit?

urbit/docker-compose.yml
```diff
@@ -6,8 +6,9 @@ services:
       APP_PORT: 8090
 
   manager:
-    image: mopfelwinrux/urbit-umbrel:v0.3.0@sha256:09ee4550be66894cc2b763d5cff9a4ed15b13a3c2a298519a857c6542069d8c7
+    image: mopfelwinrux/urbit-umbrel:v0.4.1@sha256:083db8cdb2109f29c65e186b47eeecdacd5d8021ce1c4593cf8a9f945f646171
     ports:
+      - "${APP_URBIT_PORT}:8090"
```
nevets963, line 11: @mopfel-winrux Ok, no worries re `app_proxy` service. This line will also need to be dropped and then I'll give this a test :)

mopfel-winrux: Done!

nevets963 (approved):
@mopfel-winrux Approved from me! Would you consider dropping Basic HTTP auth in favour of Umbrel's App Proxy (released in v0.5) which automatically now protects apps  using the user's Umbrel password (and 2FA where applicable)? This would create a smoother experience not having 2 sets of credentials. Not critical but a suggestion.

Tested on AMD64 and ARM (Pi).

@lukechilds Can you have a check over too?

mopfel-winrux:
@nevets963 How would I go about using Umbrel's App Proxy?

nevets963:
@mopfel-winrux The App Proxy is already turned on by default for most apps including Urbit. If for example, you're not logged into the Umbrel dashboard, and then access directly Urbit hosted at eg. http://umbrel.local:8090, you're now presented with an Umbrel provided authentication screen.

There is nothing required to do in terms of getting this benefit, however we could improve the UX if Urbit dropped HTTP Basic Auth (as this is no longer required). Perhaps if you want to keep the functionality but have it off for Umbrel; you check if the `APP_PASSWORD` env. var. is defined in your container, and then apply the HTTP Basic Auth, and if not, have no auth. and therefore rely on our App Proxy?

lukechilds (approved):

urbit/umbrel-app.yml
```diff
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: urbit
 category: Networking
 name: Urbit
-version: "0.3.0"
+version: "0.4.0"
```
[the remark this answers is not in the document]
lukechilds:
```suggestion
version: "0.4.1"
```

PR #43, opened 2022-06-23
App submission: ESPHome

ShonP40:
# App Submission

### App name
ESPHome

### 256x256 SVG icon

https://svgshare.com/i/iYU.svg

### Gallery images

dashboard_empty

dashboard_states

hero

### I have tested my app on:


esphome/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: $APP_ESPHOME_IP
+      APP_PORT: 6052
+      PROXY_AUTH_ADD: "true"
```
nevets963, line 8: By default, this is true, so we can drop this.

esphome/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+
+  server:
+    image: esphome/esphome:2022.6.2@sha256:69d382af56b10006a833eb90d29339b12efdddfb93382764a379c17fe252eda6
+    volumes:
+      - ${APP_DATA_DIR}/data:/config
+      - /etc/localtime:/etc/localtime:ro
```
nevets963, line 14: Out of interest, how come this gets mounted in?

esphome/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+      # Needed for flashing new ESPs
+      - /dev:/dev
+    networks:
+      default:
+        ipv4_address: $APP_ESPHOME_IP
+    restart: on-failure
+    # Needed for flashing new ESPs
+    privileged: true
```
nevets963, line 22: Have you tried any cgroup configuration that would grant access to the hardware without 'breaking the sandbox' using `privileged: true` (which essentially whitelists all hardware).

Here is an example where a container accesses a Bitcoin hardware wallet via USB that if I recall correctly works: https://github.com/getumbrel/umbrel/pull/1240/files#diff-c82306ea6e81e5a7b6b58ee6c5dc339fe222eda0936d2c21a22d1d1c5038de5aR24

esphome/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: $APP_ESPHOME_IP
+      APP_PORT: 6052
+      PROXY_AUTH_ADD: "true"
+
+  server:
+    image: esphome/esphome:2022.6.2@sha256:69d382af56b10006a833eb90d29339b12efdddfb93382764a379c17fe252eda6
```
nevets963, line 11: This appears to have the digest for architecture ARM64. The digest should be: `sha256:1fac6334b85e76c7fd936036891b7a7ee761ee73fd119662b2cb564c3faf818f`

You can get a multi-arch. digest by running: `docker pull <image>` and then you'll see the digest at the bottom.

nevets963 (commented):
Hey @ShonP40 Thanks so much for the app submission! I have a bag of ESP32s (a model with a camera on it) that I've wanting to do something with! haha

I've left some feedback on your PR which I hope we can address :)

esphome/docker-compose.yml
ShonP40: Never really checked tbh.
Just saw that ESPHome's recommended docker-compose had it in

https://esphome.io/guides/getting_started_command_line.html

esphome/docker-compose.yml
ShonP40: I remember trying it a while ago but it wasn't working properly.

I can give it another try though.

ShonP40: Seems to be working

I'll push it in a couple minutes

ShonP40: nope, it detects the ESP, but fails to flash
image

ShonP40: I guess that we only have two options:

1. Keep privileged mode enabled
2. Use the dashboard as is, without the ability to directly flash new ESPs through USB

nevets963: As of right now, we don't want to allow applications to have `privileged: true`. We can have this approved and figure out the flashing behaviour later?

Btw, what did you try in terms of trying to get this to work, as I think it should be possible with similar (ish) cgroup config that I posted above.

esphome/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+  app_proxy:
+    environment:
+      APP_HOST: $APP_ESPHOME_IP
+      APP_PORT: 6052
+
+  server:
+    image: esphome/esphome:2022.6.2@sha256:1fac6334b85e76c7fd936036891b7a7ee761ee73fd119662b2cb564c3faf818f
+    volumes:
+      - ${APP_DATA_DIR}/data:/config
+      - /etc/localtime:/etc/localtime:ro
```
nevets963, line 13: I don't like mounting this in from the host. It looks like this is only used for logging: https://github.com/esphome/issues/issues/398

There is a semi-interesting point, that this approach will also constantly make syscalls.

Perhaps as an alternative (as suggested), we set the `TZ` env. var. to `UTC` ?

ShonP40: Alright, I'll remove it

Also, `UTC` should already be the default

esphome/docker-compose.yml
ShonP40: I tried group `189`, but apparently it's `188` (my bad).

It works now, going to remove `privileged: true` and push.

esphome/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    environment:
+      APP_HOST: $APP_ESPHOME_IP
+      APP_PORT: 6052
+
+  server:
+    image: esphome/esphome:2022.6.2@sha256:1fac6334b85e76c7fd936036891b7a7ee761ee73fd119662b2cb564c3faf818f
+    volumes:
+      - ${APP_DATA_DIR}/data:/config
+      # Needed for flashing new ESPs
+      - /dev:/dev
```
nevets963, line 14: @ShonP40 Just another thought/idea. Could you check if it is still working using `/dev/bus/usb:/dev/bus/usb` to lock this down further?

And can you also check that it is still working when you plug in a USB device **after** the container has started?

ShonP40: ESPHome flashes ESPs through a tty connection though
e.g. `/dev/ttyUSB0`

nevets963 (approved):
Hey @ShonP40, thanks again for the app submission and for also working through my suggestions/feedback. This is approved from me!

Tested on AMD64 and ARM64 (Pi). I don't have a ESP32 device to hand atm - so I trust that you've tested the USB flashing. It would be cool if we could get HTTPS working locally to use `WebSerial` - that sounds pretty cool!

@mayankchhabra Could you help prepare graphics with @ShonP40 for the App Store?

@lukechilds Could you also check this over and set/correct the container IP if needed?

ShonP40:
Any updates on this?

lukechilds:
Hey @ShonP40 sorry for dropping the ball on this, testing your app out right now.

esphome/docker-compose.yml
lukechilds, line 14: Passing all of `/dev` directly through to the container makes me feel a little uneasy. We just need access to `/dev/ttyUSB0` right? Maybe we can lock it down to:

```
devices:
  - "/dev/ttyUSB0:/dev/ttyUSB0"
```

Alternatively, do we need the flashing functionality of ESPHome to work directly on Umbrel hardware? Umbrel is normally run on a server hidden away somewhere and you'd access the web ui on a laptop or something. So having to USB connect a device to your server to flash it through a web UI on your laptop seems like an anti pattern.

If you could plug the device into your laptop and flash it via WebUSB or something it seems like that would be a more logical approach. Thoughts?

lukechilds: Ok so just testing the app out looks like it does support this behaviour by default although WebUSB access is disabled due to no SSL. However it sends the user to https://web.esphome.io/?dashboard_wizard which then does allow flashing a bootstrap esphome firmware via WebUSB (since this page has SSL) which will allow the device to be discoverable on the network and the setup can then be continued via the local ESPHome install.

I don't have an ESP32 device to test this flow on but it seems like the best option to me. @ShonP40 do you agree with this? If we can remove the `/dev` access and we can go with this bootstrap option I think this is good to be merged.

If you have an ESP32 device you could test this flow with it would be great if you could confirm.

lukechilds: One thing that definitely would be worth testing is that after flashing the bootstrap firmware the Umbrel can detect the device with the current compose config. Often network discovery stuff is done via UDP scanning which I don't think is possible from a container unless you add `network_mode: "host"`.

ShonP40: I can limit it to `ttyUSB0` but then it'll only detect one ESP at a time.

And about discovery, you are right, it doesn't detect ESPs that are already on the network and doesn't update the online indicator.
However, this only breaks the auto adopt functionality, all the other functions continue to work normally.

So it's up to you guys if we should make this run on the host network or not.

ShonP40: If you do want auto adopt to work (thus requiring this container to run on the host network), I'm guessing that users will want Home Assistant on the host network as well (so it'll be able to detect these ESPs automatically)

lukechilds: >And about discovery, you are right, it doesn't detect ESPs that are already on the network and doesn't update the online indicator.
However, this only breaks the auto adopt functionality, all the other functions continue to work normally.

So does this mean you can for example update the firmware once the device is setup? Or are you saying it's not possible to actually complete the setup process?

Basically I'm just trying to figure out (without having an ESP32 device to test) if this is in a usable state without host networking and hardware access or not. If it is, I think we should ship it like that, then maybe revisit adding more permissions if there are significant requests from users for it.

>If you do want auto adopt to work (thus requiring this container to run on the host network), I'm guessing that users will want Home Assistant on the host network as well (so it'll be able to detect these ESPs automatically)

As a general rule we don't want to allow host networking for apps since it also violates the app sandbox. There is a workaround for mDNS where we just run some sort of DNS proxy on the host that resolves mDNS requests to local IPs and forwards on normal DNS requests. Then configure app containers to use this for DNS. That should allow apps like ESPHome and Home Assistant to resolve .local domain names without requiring host networking access. We don't have this implemented yet however.

We could potentially ship host networking as a short term workaround until that's implemented but it causes a lot of other problems. We have an authentication proxy sitting in front of all Umbrel app web UIs that verify the user is logged in to the Umbrel before proxying traffic through. This breaks if the app is directly listening on host ports. It also means we don't get explicit control over which ports are being opened on the host to prevent port collisions between apps.

In your opinion is ESPHome useful without host networking or hardware access? If so I think we should ship like that with the goal of enabling mDNS resolution in the future at the Umbrel level without having to give apps host networking. So the app will just suddenly be even better in the future once we solve the mDNS problem and the auto discovery logic will start working.

ShonP40: Yea, ESPHome is completely useful without host networking.
You can update devices OTA.

The only caveat would be that you'll need to make it upload using the device's IP instead of its .local domain by adding
`use_address: ESP IP` to your ESPHome config (for each device)

lukechilds: Ok sweet, that's the same way Home Assistant works atm, you need to manually enter an IP to pair a device. Lets ship like this and we'll try to get a good solution for mDNS discovery implemented in Umbrel.

Thanks for your help @ShonP40!

ShonP40:
It will be better to keep access to `/dev/ttyUSB0` though...

esphome/exports.sh
```diff
@@ -0,0 +1,2 @@
+export APP_ESPHOME_PORT="8093"
```
lukechilds, line 1: Sorry @ShonP40 just one last thing, ESPHome dashboard is served on it's default port of `6052` in the container but we map that to port `8093` on the host. Is there a reason for this? No other app is using `6052` so it's available, shall we use `6052` on both the container and host so the user accesses ESPHome on it's default port port?

lukechilds (approved):

lukechilds:
Ok this is looking great! Thanks so much @ShonP40!

@mayankchhabra this is ready for gallery assets.

ShonP40 (commented):
It will be better to keep access to `/dev/ttyUSB0` though...
Especially for initial flashing

ShonP40:
Not having direct hardware flashing kinda makes this PR useless for my use case

lukechilds:
Ok no worries @ShonP40 now we've locked down to `/dev/ttyUSB0` it's ok to ship that.

It's live now, you should see it available to install in your Umbrel's app store in the next 15 minutes. Thanks again!

ShonP40:
Alright, thank you!

ShonP40:
@lukechilds I just tested flashing with the `/dev/ttyUSB0` mount and it doesn't really seem to like it.
image
It worked when I mounted the entire `/dev` directory though

ShonP40:
```
volumes:
      - /dev/ttyUSB0:/dev/ttyUSB0
```
Needs to be replaced with:
```
devices:
  - "/dev/ttyUSB0:/dev/ttyUSB0"
```

ShonP40:
Fixed this in #86

lukechilds:
@ShonP40 thanks. Can you try sshing into your Umbrel and running: `sudo ~/umbrel/scripts/repo update && sudo ~/umbrel/scripts/app update esphome`.

ShonP40:
Yea, noticed your fix after submitting that PR lol.
Alright, let me force pull onto my Umbrel instance

lukechilds:
Hmmn, this is giving me:

```
ERROR: for server  Cannot start service server: error gathering device information while adding custom device "/dev/ttyUSB0": no such file or directory
ERROR: Encountered errors while bringing up the project.
```

lukechilds:
Maybe this means the app can only successfully start if a serial device is attached to the host and exists at `/dev/ttyUSB0`?

ShonP40:
Never tried to mess with connected devices on docker to this extent, I was always just mounting the entire `/dev` directory

lukechilds:
Let me just quickly revert the device before it propagates to Umbrel users since this will fail on install. Then we can figure out how to pass in the device and push an app update.

ShonP40:
Alright.

Also, I'm getting this when the ESP is connected on startup
`ERROR: for server  Cannot start service server: error gathering device information while adding custom device "/dev/ttyUSB0": not a device node`

lukechilds:
>Is there any way to only mount it when its actually plugged in?

I guess by mounting in the `/dev` directory like we were before. Then ESPHome will see the device later on when it exists and it doesn't have to exist at startup to satisfy Docker volume mappings since we're mounting in the parent directory.

My issue with that was I was thinking it was pretty much the same as `privileged: true` since the container can interact with raw low level devices like RAM and block devices. However I'm not totally sure that's actually correct or if there's some cgroups stuff that's preventing that from happening.

ShonP40:
Exposing `/dev` isn't enough for it to be able to access a device, thus requiring the cgroup I added.
Without it, it can only detect a device being connected to it, but can't actually communicate with it.

ShonP40:
(The cgroup rule I added allows it to communicate with `tty` devices)

lukechilds:
Could this work with `/dev/bus/usb`? I'm not sure if the serial device also exists on the USB bus or if some Linux serial USB logic jumps in and it only ever exists at `/dev/ttyUSB0`.

But potentially we can pass in `/dev/bus/usb` and the appropriate cgroup rule and ESPHome will be able to find the serial device from that.

ShonP40:
I tried that before making this PR and it didn't work.

Also, the only way a container could access a device in `/dev` is if it has privileged mode enabled, or has the correct cgroup for the specific type of devices it needs to access.

So by using `c 188:* rmw`, I'm only allowing it to access `tty` devices, it can't do anything with other devices it may detects in `/dev`

lukechilds:
>Also, the only way a container could access a device in /dev is if it has privileged mode enabled, or has the correct cgroup for the specific type of devices it needs to access.
>
>So by using c 188:* rmw, I'm only allowing it to access tty devices, it can't do anything with other devices it may detect in /dev

That sounds like we should be able to allow this for now. Do you have a source I can read more on this? I'm not totally familiar with how cgroups work. We can't really allow something like this long term since it seems to break down the app sandbox, e.g two apps with access to `/dev` could probably communicate through `/dev` via a unix socket or something. Relying on cgroups alone to prevent hardware access feels like kind of a weak protection over having no access to `/dev` at all.

But yeah I think it's probably ok to allow this as a short term workaround until we have a safer solution figured out for giving apps access to host hardware. Could you just point me to some docs or something if you know of any that confirm this is definitely safe?

ShonP40:
https://docs.docker.com/engine/security/

ShonP40:
@lukechilds Any updates on this?

lukechilds:
@ShonP40 apologies for the delay.

@nevets963 are you able to double check this? If you can confirm that passing in `/dev` along with the cgroup rule only gives access to TTY devices and nothing else it's ok to add back in.

ShonP40:
@nevets963 ?

Issue #44, opened 2022-06-23
Does suredbits-wallet require mempool to be defined as a dependency?


nevets963:
Does suredbits-wallet require mempool to be defined as a dependency within `umbrel-app.yml` as it uses `APP_MEMPOOL_PORT` ?

https://github.com/getumbrel/umbrel-apps/blob/a43beb42b06ae0d5ccd033a340d941838a8e2a3e/suredbits-wallet/docker-compose.yml#L20

nevets963:
Closing as this was resolved by https://github.com/getumbrel/umbrel-apps/pull/45 and an issue opened on the app: https://github.com/bitcoin-s/bitcoin-s-ts/issues/157

PR #45, opened 2022-06-24
2022 06 24 Suredbits Wallet / Krystal Bull 1.9.2

Christewart:
Updates Suredbits apps to the [1.9.2](https://github.com/bitcoin-s/bitcoin-s/releases/tag/1.9.2) release of bitcoin-s.

On a platform note, we are capping our max memory usage now on the Suredbits Wallet. Previously we had umbrel users note that SB wallet consumes a lot of memory due to the jvm.


Christewart:
@nevets963 This should be good to merge. 

For what it's worth, there seem to be a connectivity issue between the UI docker image and the backend docker image on Krystal Bull on my Umbrel. For whatever reason, this went away after ~1 hour. We don't have any networking changes for the release with Krystal Bull, the app is pretty simple. Suredbits Wallet is much more complex.

nevets963:
@Christewart That sounds wierd and wasn't my experience testing just now on both AMD64 and ARM. Both apps installed  and were ready to go immedaitely.

Do you have any thoughts on this? https://github.com/getumbrel/umbrel-apps/issues/44
In short, whether mempool is a hard requirement for Suredbits Wallet and requires the user to have that installed before installing Suredbits Wallet?

Christewart:
> Do you have any thoughts on this? #44 In short, whether mempool is a hard requirement for Suredbits Wallet and requires the user to have that installed before installing Suredbits Wallet?

So from talking with @user411 this change will cause a breaking change at the UI level when we display a link to a transaction on a block explorer. We will file an issue to fix this on our repos, but I think this is a small enough bug that we can merge this

Christewart:
@nevets963 On 8ccc38cf53915a89ba9f595f8ffcdfad7df26484 we now comment out the url entirely, as we handle the case properly when the url isn't defined.

More generally, how are we suppose to figure out which apps the user has installed so we could support this feature in the future? It would be nice from a privacy perspective to use the local mempool instance if the app is installed

nevets963:
@Christewart Currently we have no way to tell an app which apps a user has already installed (but we're thinking this through atm as we have a similar case for another app).

You can remove this feature as it could be broken for some (but perhaps working for most), keep the same as before, or we make mempool a requirement (a dependency). FWIW, it's now a pretty nice experience to install dependencies; here is when I install Lightning (LND) for example:
Screenshot 2022-06-24 at 23 44 43


Either way, I'm happy to approve the PR as I've tested on AMD64 and ARM (Pi) and both apps are working nicely.

@lukechilds Do you have some thoughts on this PR?

nevets963 (approved):
Tested on AMD64 and ARM (Pi). Approved by me.

lukechilds:
Looks great!

PR #47, opened 2022-06-24
Added PROXY_AUTH_ADD: "false"

jensgertsen:
Service needs to be publicly accessible to return LNURLS - thus disabling umbrel default password protection setting PROXY_AUTH_ADD: "false" - still uses the password generated by umbrel.

### DIFF
--- a/sparkkiosk/docker-compose.yml
+++ b/sparkkiosk/docker-compose.yml
@@ -3,7 +3,8 @@ version: "3.7"
 services:
   app_proxy:
     environment:
-      APP_HOST: $APP_SPARKKIOSK_IP
+      PROXY_AUTH_ADD: "false"
+	  APP_HOST: $APP_SPARKKIOSK_IP
       APP_PORT: 21214
 
   web:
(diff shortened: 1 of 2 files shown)


nevets963:
Hey @jensgertsen I'm not intimately familiar with sparkkiosk, however I just took a look at the source. Another potential solution is the whitelist feature within the App Proxy, that way we keep the authentication through App Proxy and whitelist your APIs/paths. The App Proxy creates a better experience once apps drop secondary authentication systems (with separate app passwords, etc), and instead rely on user's existing credentials (and where applicable 2FA).

e.g. I see you have defined `GET /lnurl/:id` and you could whitelist this using:
```
PROXY_AUTH_WHITELIST: "/lnurl/*"
```

Could something like the above work instead?

jensgertsen:
I would rather make the minimal change now to get the app woring in 0.5, and then rework the auth. to use the new scheme along with some other improvements next.

lukechilds:
>I would rather make the minimal change now to get the app woring in 0.5, and then rework the auth. to use the new scheme along with some other improvements next.

Sounds good @jensgertsen, sorry we must have overlooked this during testing! This is merged in now so will go out to all Umbrel users with the fix over the next 15 minutes.

jensgertsen:
You guys are awesome. I love that you can push app updates outside the OS updates, this platform keeps getting better!

PR #48, opened 2022-06-25
Updated simpletorrent to 1.3.9

satsbaba:
Updated the docker image to 1.3.9

### DIFF
--- a/simple-torrent/docker-compose.yml
+++ b/simple-torrent/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: $APP_SIMPLE_TORRENT_PORT
 
   server:
-    image: boypt/cloud-torrent:1.3.9@sha256:8b0a0bb1c4ae376f8ca611b7d32c446049714c415ac010f31b5370a3b2fdcde8
+    image: boypt/cloud-torrent:1.3.9@sha256:90cc62869ebaabbdba31535cdff4c66ede98b341956d2ebcd6650610a37e9819
     user: "1000:1000"
     restart: on-failure
     command: >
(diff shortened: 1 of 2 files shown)


simple-torrent/docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       APP_PORT: $APP_SIMPLE_TORRENT_PORT
 
   server:
-    image: boypt/cloud-torrent:1.3.4@sha256:a1355e9ac319d523ce1fd6e8d34ab8e4bf0b5476c52bfe244e37fea64f6f4228
+    image: boypt/cloud-torrent:1.3.9@sha256:8b0a0bb1c4ae376f8ca611b7d32c446049714c415ac010f31b5370a3b2fdcde8
```
nevets963, line 10: This has the digest for ARM64. To get a multi-arch digest, you can run: `docker pull <image>` and at the bottom it will say the digest.

In this case it should be: `sha256:90cc62869ebaabbdba31535cdff4c66ede98b341956d2ebcd6650610a37e9819`

nevets963 (commented):
Thanks @satsbaba for the PR to update Simple Torrent!

I've left some feedback about the Docker image's digest.

Can you also bump the version number within the app's manifest file (`umbrel-app.yml`), once done, I'll give this a test :)

simple-torrent/docker-compose.yml
satsbaba: Done!!

nevets963 (approved):
@satsbaba Thanks for making those changes. This is approved from me!

@lukechilds Do you want to have a final check over?

lukechilds:
Looks great, thanks @satsbaba!

PR #51, opened 2022-06-27
lightning-terminal: bump version to v0.7.0-alpha

guggero:
https://github.com/lightninglabs/lightning-terminal/releases/tag/v0.7.0-alpha

### DIFF
--- a/lightning-terminal/docker-compose.yml
+++ b/lightning-terminal/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: $APP_LIGHTNING_TERMINAL_PORT
 
   web:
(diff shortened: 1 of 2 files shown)


nevets963 (approved):
@guggero Thanks for the app update! Approved by me!

Would you consider removing the requirement for a password with Lightning Terminal? In Umbrel v0.5 and onwards, we automatically password protect (including 2FA where applicable) all installed apps. If you remove the password requirement, we can improve the user experience as they then don't need a secondary password for Lightning Terminal.

I tried dropping the `command` arg `--uipassword_env`, but I got this error:
```
could not load config: could not read UI password: mandatory password for UI not configured. specify either a password directly or a file or environment variable that contains the password
```

@lukechilds Do you want to check this too?

guggero:
The plan is to eventually remove the password, yes. But we're not quite there yet. But the improvements in this version actually get us quite a bit closer towards that goal.

lukechilds (approved):

lukechilds:
Looks great, thanks @guggero!

PR #60, opened 2022-07-05
App submission: Oak Node

ok300:
# App Submission

### App name
Oak Node

### 256x256 SVG icon

https://oak-node.net/doc/tip/www/static/oak-logo.svg

### Gallery images

Screenshot 2022-07-05 at 12-06-09 Oak

Screenshot 2022-07-05 at 14-32-38 Oak

Screenshot 2022-07-05 at 12-05-52 Oak

### I have tested my app on:

---

Looking forward to your feedback. Thanks!

### DIFF
--- a/oak-node/exports.sh
+++ b/oak-node/exports.sh
@@ -1,2 +1,2 @@
-export APP_OAK_NODE_IP="10.21.21.99"
+export APP_OAK_NODE_IP="10.21.21.105"
 export APP_OAK_NODE_PORT="8100"
--- a/oak-node/exports.sh
+++ b/oak-node/exports.sh
@@ -1,0 +1,2 @@
+export APP_OAK_NODE_IP="10.21.21.98"
+export APP_OAK_NODE_PORT="8100"
--- a/oak-node/exports.sh
+++ b/oak-node/exports.sh
@@ -1,2 +1,2 @@
-export APP_OAK_NODE_IP="10.21.21.98"
+export APP_OAK_NODE_IP="10.21.21.99"
 export APP_OAK_NODE_PORT="8100"
--- a/oak-node/docker-compose.yml
+++ b/oak-node/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: $APP_OAK_NODE_PORT
 
   web:
-    image: oak-node.net/oak:v0.3@sha256:d4bf366d60235c0aa8149ed7741900cc26ad6884a4bd77b36196f3934dbb8dac
+    image: oak-node.net/oak:v0.3.1@sha256:f78f19763705dfb2237fd8a5c78f27052a848a5e56d89b3ffbcdc1edef0ba57e
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/oak-node/docker-compose.yml
+++ b/oak-node/docker-compose.yml
@@ -14,7 +14,7 @@ services:
     volumes:
       - $APP_LIGHTNING_NODE_DATA_DIR:/lnd-dir:ro
       - ${APP_DATA_DIR}/data:/data
-      - ${APP_DATA_DIR}/data:/oak/log
+      - ${APP_DATA_DIR}/log:/oak/log
     environment:
       ROCKET_PORT: $APP_OAK_NODE_PORT
       ROCKET_ADDRESS: 0.0.0.0
--- a/oak-node/docker-compose.yml
+++ b/oak-node/docker-compose.yml
@@ -1,0 +1,27 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: $APP_OAK_NODE_IP
+      APP_PORT: $APP_OAK_NODE_PORT
+
+  web:
+    image: oak-node.net/oak:v0.3@sha256:d4bf366d60235c0aa8149ed7741900cc26ad6884a4bd77b36196f3934dbb8dac
+    user: "1000:1000"
+    restart: on-failure
+    stop_grace_period: 1m
+    volumes:
+      - $APP_LIGHTNING_NODE_DATA_DIR:/lnd-dir:ro
+      - ${APP_DATA_DIR}/data:/data
+      - ${APP_DATA_DIR}/data:/oak/log
+    environment:
+      ROCKET_PORT: $APP_OAK_NODE_PORT
+      ROCKET_ADDRESS: 0.0.0.0
+      OAK_DATA_DIR: /data
+      OAK_LND_REST_API_URL: https://$APP_LIGHTNING_NODE_IP:$APP_LIGHTNING_NODE_REST_PORT
+      OAK_LND_MACAROON_PATH: /lnd-dir/data/chain/bitcoin/$APP_BITCOIN_NETWORK/admin.macaroon
+      OAK_LND_CERT_PATH: /lnd-dir/tls.cert
+    networks:
+      default:
+        ipv4_address: $APP_OAK_NODE_IP
--- a/oak-node/umbrel-app.yml
+++ b/oak-node/umbrel-app.yml
@@ -1,6 +1,6 @@
 manifestVersion: 1
 id: oak-node
-category: Automation
+category: Finance
 name: Oak Node
 version: "0.3.1"
 tagline: Do more with your LND node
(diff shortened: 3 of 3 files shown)


nevets963 (approved):
Hey @ok300 Thanks a lot for your app submission! It's really impressive to see pretty much a perfect PR considering we changed the app structure recently! Awesome work!

I've tested the app on both AMD64 and ARM64 (Pi) and they're working perfectly.

1 thing I noted on my regtest node is this tiny error by 'Next Run':
Screenshot 2022-07-05 at 22 14 22

I can see why and I don't think it's a blocking issue; just an observation.

I also noted that you're not caching any of the static assets (css, js, etc files). I think this is OK for local network access, but if the app is accessed over Tor (which is really slow) or Tailscale, then the latency could have some impact. Of course, I understand you then need to think about cache invalidation when you update these assets, but something to think about. I won't block the PR on that either...

I would also consider a much punchy description e.g.
```
Oak Node is a self-hosted a Scheduled Payments module built for LND that lets you support your favorite content creators by regularly sending them tips to their Lightning Address.

It also includes an optional bot module, which lets you interact with your Oak Node from a chat interface.
```
But again, I won't block on that, your call :)

This is approved from me!

@mayankchhabra Could you help with preparing assets for the App Store?

@lukechilds Could you also test Oak Node and check/fix up the container IP once you're happy to merge?

ok300:
@nevets963 : thanks for the kind words and detailed feedback :)

About the things you found:

- the "Next Run" issue you saw on regtest: I've seen it in some tests where LND is not fully up and running. Not a bug (doesn't affect functionality) but yes it is annoying. Added to the list, will address it in the next release.

- the caching of static assets: unfortunately its a limitation of a library I'm using. As soon as they add support for that, I'll integrate it.

As for the description, you're right it needed some polishing. I already updated it to be shorter and crispier.

Any more feedback is always welcome :)

ok300:
FYI: I've bumped the app version, which fixes the 2 minor issues we discussed above.

Tested it on an Umbrel VM.

oak-node/exports.sh
```diff
@@ -0,0 +1,2 @@
+export APP_OAK_NODE_IP="10.21.21.99"
```
lukechilds, line 1:
```suggestion
export APP_OAK_NODE_IP="10.21.21.105"
```

lukechilds (approved):
Tested and working great!

lukechilds:
Nice work @ok300! This is ready to ship.

@mayankchhabra are you able to put together some release assets?

ok300:
No worries @lukechilds, thanks for the review and IP fixes.

mayankchhabra:
Really nice work on Oak Node (and packaging it!) @ok300. The app store gallery images and app icon are now live via [getumbrel/umbrel-apps-gallery](https://github.com/getumbrel/umbrel-apps-gallery/tree/master/oak-node).

Good luck with the launch in the app store! :)

PS: We added a splash of color to the app icon. In case you'd like us to revert it, just let us know. And also if you need to update the gallery images or the icon in the future, feel free to submit a PR to the above repo. Cheers!

lukechilds:
Woop it's all live. 🚀

Congrats @ok300, Oak Node should propagate to all Umbrel app stores within the next 15 minutes.

ok300:
Thank you both! I can see it in the appstore, and yes the logo did need a splash of color :)

plazareff:
It seems it no longer works because Oak :

	•	wrongly builds LNURL callback URLs
	•	incorrectly parses LNbits callback
	•	tries to decode outdated LND JSON responses

Maybe it should be removed from appstore.

nmfretz:
Hey @plazareff, thanks for the report. Can you provide the specific steps that result in Oak Node failing for you so I can try and recreate? I just tested it out (hadn't before) and it worked great for me using a classic Lightning Address for the destination. For example, Wallet of Satoshi LN Address that returns a BOLT11 invoice.

@ok300 hasn't released an update in a while, so newer schemes like Bolt 12, BIP353 dns addresses, etc likely won't work.

plazareff:
The problems I encountered all had in common the way url's are constructed when a schedule is set as I found several '?' where there should be '&' like in :

`...?webhook_data=xxxx?amount=21000` 

instead of :

`...?webhook_data=xxxx&amount=21000`

And also many errors when parsing responses like:

`Failed to parse lnurl-pay invoice: missing field 'pr'` or `Failed to call LND send_payment_sync: error decoding response body: missing field 'payment_error'`

I don't know about the newer schemes, I just installed in umbrel with lnd and I couldn't schedule a payment, nor send one to the creator, even though the url seemed right:

`oak-node_web_1        | 2025-11-25 20:32:09 - ln_sdk::lnurl INFO: [lnurl-pay 2/5] Requesting invoice at https://oak-node.net/.well-known/lnurlp/carlos?amount=21000&comment=Hi, I don't know if this form will get through as when i try a onetime send to any address, i get : Failed to call API: Failed to pay invoice: Failed to call LND send_payment_sync: error decoding response body: missing field payment_error at line 1 column 63`

Which resulted in:

`oak-node_web_1        | 2025-11-25 20:32:27 - oak::model::api ERROR: GenericError("Failed to call API: Failed to parse lnurl-pay invoice: missing field pr at line 1 column 42") caused by Failed to parse lnurl-pay invoice: missing field pr at line 1 column 42`

So I gave up thinking that something was quite broken and turned to zap planner...

nmfretz:
Thanks for the extra details @plazareff! I hope zap planner is working for your needs in the meantime.

If you ever feel like trying Oak Node again, it’s possible that the errors you’re seeing are due to LNURL callback issues with the specific services/addresses you’re using, rather than Oak itself.

Just to show you what I mean:

I can successfully do one-time payments and scheduled payments (e.g. every block / every 3 blocks) to a plain old Lightning Address from Wallet of Satoshi.

In those cases, Oak:
- resolves the Lightning Address,
- hits the LNURL-pay callback,
- gets back a normal JSON response with a pr field (BOLT11 invoice)

But for the specific example you mentioned with:

> https://oak-node.net/.well-known/lnurlp/carlos?amount=21000...

it actually looks like the LNURL server for that address is what’s broken right now, not Oak. You can see this without using Oak at all:

Open this in your browser:
https://oak-node.net/.well-known/lnurlp/carlos
You should see a normal LNURL-pay metadata JSON (status OK, min/max amounts, callback URL, etc.).

Now try calling the callback with an amount. For example open this in your browser:
https://oak-node.net/.well-known/lnurlp/carlos?amount=21000

That returns a JSON error (status "ERROR" / “Server Error”) and no pr field.

When Oak asks that URL for an invoice, it gets the same error JSON back and then fails with:

> Failed to parse lnurl-pay invoice: missing field pr at line 1 column 42

which is Oak trying to parse an invoice out of a response that doesn’t contain one. So in this specific case, the failure is upstream, on the carlos@oak-node.net side.

On the URL construction bit you mentioned (`...?webhook_data=xxxx?amount=21000`), that indeed looks strange. But again I think it's likely due to how the Lightning Address / LNURL service you're using is constructing the callback url. Could be Oak though I guess. If you don't mind, what service are you trying to use? That could help explain the issue.

plazareff:
Thanks for the reply @nmfretz !

It's true I had not double checked for Carlos' url as I was using the integrated support/contact form, I did not imagine it could fail. My prejudice ;-)

For the rest, there are 2 issues I can identify after some more investigation:

For the double ? it is because one is inside the callback URI given to Oak by the LNBit Pay Link extension which is like `/lnurlp/api/v1/lnurl/cb/5yZENh?webhook_data=alias%3D%27webhook_data%27+extra%3D%7B%7D`. Oak does not check if there is one ? or not and adds ?amount=x where, if it checked, that should be &amount=x

I could not get the Pay Link extension not to generate the webhook_data part of the callback url.

But there is another issue with services like getalby which gives callback urls like https://getalby.com/lnurlp/user/callback, so no double ? issue there. But for those, I get the error triggered by `ln_sdk::lnurl INFO: [lnurl-pay 4/5] Attempting to pay it`

So it maybe is not Oak but ln_sdk which I see is called in src/lnd.rs:
`use ln_sdk::client_v2::lnd::{LndClient, LndClientOps, LndConfig};`

For the extra ? I could trace the issue:
https://oak-node.net/cgit/ln-sdk.git/tree/src/lnurl.rs

`let mut invoice_lookup_resp = format!(
        "{}?amount={amount_msats}",
        onetime_pay_request.request.callback
    );`

For the Failed to call LND send_payment_sync error I receive with GetAlby, I'm afraid this is beyond my competence.

If you're in touch with Carlos, it may be worth updating his ln-sdk ?

PR #62, opened 2022-07-06
Update Kollider to 1.0.6

djkazic:


nevets963 (approved):
@djkazic Thank you for the Kollider app update! Approved from me!

Would you consider dropping the usage of `$APP_PASSWORD`? In Umbrel v0.5, we now protect all apps using user's password (and where applicable 2FA). If we remove this secondary password for Kollider, we can dramatically improve the user's experience.

@lukechilds Do you also want to test?

djkazic:
I think we can reasonably accommodate that change in 1.0.7

nevets963:
Ok nice! Yh, Itchysats has an open issue to also remove the usage of `$APP_PASSWORD`: https://github.com/itchysats/itchysats/issues/2314

PR #63, opened 2022-07-06
Update tailscale to 1.26.2

willcosgrove:


nevets963:
@willcosgrove Thanks for pushing an app update for Tailscale! It looks like Tailscale 1.26.2 is already out: https://tailscale.com/changelog/#2022-07-05-client

Would you mind updating your PR to that version? Unless there is some issue with 1.26.2 to not update to that version...

Thanks!
Steve

willcosgrove:
Yeah no problem! Sorry I didn't notice that when I made this PR. Updated to 1.26.2

nevets963 (approved):
Thanks @willcosgrove for the PR! Approved from me!

@lukechilds Do you want to test too?

PR #65, opened 2022-07-08
App Submission - lndboss

niteshbalusu11:
# App Submission

### App name
LndBoss

### 256x256 SVG icon

https://svgshare.com/s/iwg
...

### Gallery images

...
BOX-remix2 (1)

Screen Shot 2022-07-08 at 2 57 49 PM
Screen Shot 2022-07-08 at 2 58 07 PM
Screen Shot 2022-07-08 at 2 59 37 PM

### I have tested my app on:
- [X] [Umbrel dev environment](https://github.com/getumbrel/umbrel-dev)
- [X] [Custom Umbrel install on Linux](https://github.com/getumbrel/umbrel#-installation)

### DIFF
--- a/lndboss/.bosgui/umbrel/credentials.json
+++ b/lndboss/.bosgui/umbrel/credentials.json
@@ -1,0 +1,5 @@
+{
+  "cert_path": "/home/node/.lnd/tls.cert",
+  "macaroon_path": "/home/node/.lnd/data/chain/bitcoin/mainnet/admin.macaroon",
+  "socket": "10.21.21.9:10009"
+}
--- a/lndboss/.bosgui/umbrel/credentials.json
+++ b/lndboss/.bosgui/umbrel/credentials.json
@@ -1,5 +1,0 @@
-{
-  "cert_path": "/home/node/.lnd/tls.cert",
-  "macaroon_path": "/home/node/.lnd/data/chain/bitcoin/mainnet/admin.macaroon",
-  "socket": "10.21.21.9:10009"
-}
--- a/lndboss/umbrel-app.yml
+++ b/lndboss/umbrel-app.yml
@@ -1,0 +1,25 @@
+manifestVersion: 1
+id: lndboss
+category: Lightning Node Management
+name: LndBoss
+version: "1.12.0"
+tagline: A GUI for BalanceOfSatoshis
+description: LndBoss is a GUI for BalanceOfSatoshis.
+  It is a tool that makes it easy to run your favorite
+  bos commands and helps manage your lightning node.
+  You can schedule jobs to automatically rebalance channels,
+  integration with amboss to post updates and much more.
+developer: Nitesh Balusu
+website: https://github.com/niteshbalusu11
+dependencies:
+  - lightning
+repo: https://github.com/niteshbalusu11/lndboss
+support: https://t.me/lndboss
+port: 8055
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
+path: ""
+defaultUsername: ""
+defaultPassword: ""
--- a/lndboss/umbrel-app.yml
+++ b/lndboss/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: lndboss
 category: Lightning Node Management
 name: LndBoss
-version: "1.15.2"
+version: "1.16.4"
 tagline: A GUI for BalanceOfSatoshis
 description: LndBoss is a GUI for BalanceOfSatoshis.
   It is a tool that makes it easy to run your favorite
--- a/lndboss/umbrel-app.yml
+++ b/lndboss/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: lndboss
 category: Lightning Node Management
 name: LndBoss
-version: "1.15.1"
+version: "1.15.2"
 tagline: A GUI for BalanceOfSatoshis
 description: LndBoss is a GUI for BalanceOfSatoshis.
   It is a tool that makes it easy to run your favorite
--- a/lndboss/umbrel-app.yml
+++ b/lndboss/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: lndboss
 category: Lightning Node Management
 name: LndBoss
-version: "1.12.0"
+version: "1.15.1"
 tagline: A GUI for BalanceOfSatoshis
 description: LndBoss is a GUI for BalanceOfSatoshis.
   It is a tool that makes it easy to run your favorite
--- a/lndboss/.bosgui/config.json
+++ b/lndboss/.bosgui/config.json
@@ -1,1 +1,0 @@
-{ "default_saved_node": "umbrel" }
--- a/lndboss/.bosgui/config.json
+++ b/lndboss/.bosgui/config.json
@@ -1,0 +1,1 @@
+{ "default_saved_node": "umbrel" }
--- a/lndboss/docker-compose.yml
+++ b/lndboss/docker-compose.yml
@@ -1,0 +1,24 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: $APP_LNDBOSS_IP
+      APP_PORT: $APP_LNDBOSS_PORT
+
+  web:
+    image: niteshbalusu/lndboss:v1.12.0@sha256:98443ae685da7fd30282a0e0d56c7bf8341afef34cf52b28400b632c773a02c3
+    restart: on-failure
+    stop_grace_period: 1m
+    volumes:
+      - ${APP_LIGHTNING_NODE_DATA_DIR}:/home/node/.lnd:ro
+      - ${APP_DATA_DIR}/.bosgui:/home/node/.bosgui
+    extra_hosts:
+      - "umbrel.local:${APP_LIGHTNING_NODE_IP}"
+      - "localhost:${APP_LIGHTNING_NODE_IP}"
+    environment:
+      NODE_ENV: "production"
+      PORT: $APP_LNDBOSS_PORT
+    networks:
+      default:
+        ipv4_address: $APP_LNDBOSS_IP
--- a/lndboss/docker-compose.yml
+++ b/lndboss/docker-compose.yml
@@ -5,21 +5,18 @@ services:
     environment:
       APP_HOST: $APP_LNDBOSS_IP
       APP_PORT: $APP_LNDBOSS_PORT
-
   web:
-    image: niteshbalusu/lndboss:v1.15.1@sha256:df0e99fac53df68204420c2ad40baf247e274dbf77264c1f0701645a26f9cea0
+    image: niteshbalusu/lndboss:v1.15.2@sha256:c6fecdb6a4a0c1960d6ae22eba9ad73abaed45224d3c3cfd3d6f1d04f52a1c36
     restart: on-failure
     stop_grace_period: 1m
     volumes:
       - ${APP_LIGHTNING_NODE_DATA_DIR}:/home/node/.lnd:ro
       - ${APP_DATA_DIR}/.bosgui:/home/node/.bosgui
-    extra_hosts:
-      - "umbrel.local:${APP_LIGHTNING_NODE_IP}"
-      - "localhost:${APP_LIGHTNING_NODE_IP}"
     environment:
       BOS_DATA_PATH: "/home/node/.bosgui"
       NODE_ENV: "production"
       PORT: $APP_LNDBOSS_PORT
+      BOS_DEFAULT_LND_SOCKET: $APP_LIGHTNING_NODE_IP:$APP_LIGHTNING_NODE_GRPC_PORT
     networks:
       default:
         ipv4_address: $APP_LNDBOSS_IP
--- a/lndboss/docker-compose.yml
+++ b/lndboss/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: $APP_LNDBOSS_PORT
 
   web:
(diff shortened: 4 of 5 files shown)


niteshbalusu11:
This is ready for testing, there is no default password. Please register when you open the app.

lndboss/.bosgui/umbrel/credentials.json
```diff
@@ -0,0 +1,5 @@
+{
+  "cert_path": "/home/node/.lnd/tls.cert",
+  "macaroon_path": "/home/node/.lnd/data/chain/bitcoin/mainnet/admin.macaroon",
+  "socket": "10.21.21.9:10009"
```
nevets963, line 4: Is it possible to rely on configuration passed in via environment variables? Should Bitcoin's container IP change in the future (or become a dynamic IP), then lndboss will break.
This also goes for where you have `mainnet` hardcoded above. This will break in perhaps an unexpected way for users (like myself) who run a node using `regtest`.

nevets963 (commented):
Hey @niteshbalusu11 Thanks a lot of the app submission! I'm glad my help on TG was helpful :)

This PR looks really good. Considering we changed the app structure in Umbrel v0.5, it's very impressive.

I've left some feedback about some hardcoded variables you have which I hope you can address. After that, I'll test lndboss on my Umbrel nodes.

lndboss/.bosgui/umbrel/credentials.json
niteshbalusu11: Hey Steve,

These values can be changed from the UI, the app supports running multiple nodes from a single UI, after logging in, you can click the menu on the top left and authenticate a new node. 

Without doing adding that credentials.json file, the app can read LND creds directly, the only problem is with the socket. It reads the LND_IP/hostname from lnd.conf file, however in the latest version of Umbrel, those settings were removed from the conf file, to get around it, I could do something like this in the docker-compose file:
```
extra-hosts:
 - localhost:${UMBREL_LND_IP}
 - umbrel.local:${UMBREL_LND_IP}
```
And it resolves the IP to localhost. 
This works on a custom Umbrel install on AWS, idk if it works on a raspberry Pi though. 


Let me know what you prefer.

niteshbalusu11:
Hey Steve, thanks for reviewing. I left a reply on the review. 

I had a question about the image being pulled in the docker-compose file, does Umbrel allow NOT setting the SHA256 digest and just set the tag to `niteshbalusu/lndboss:latest` to always just pull the latest version from dockerhub considering I have to submit a new PR for app updates. This way, if someone wants to update, they could just docker-compose down and up.

lndboss/.bosgui/umbrel/credentials.json
nevets963: I see - you use this `credentials.json` file as a place to store the connection details... As we currently don't have a clean DNS setup for containers, I think using the extra hosts is the best idea atm.

Could you perhaps also write `credentials.json` on first boot of lndboss and write out state like the bitcoin network (ie check that the file doesn't exist and if not, use env. vars. and write them out). That way for users using eg regtest, when they install lndboss, 'it just works'.

@lukechilds Do you have any ideas here?

nevets963:
@niteshbalusu11 There are a couple of reasons we don't do/allow this; 1) is a security, where if you Docker hub account is breached, it could allow someone to push something malicious. 2) is predictability, if an update just goes out, it may not obvious to the user what version they're running or even if they want to run the latest. 3) is quality control, where we test app submissions/updates for specific versions to try our very best that users have a good experience when running these apps.

lndboss/.bosgui/umbrel/credentials.json
niteshbalusu11: Yeah, that file is used to point the app to the creds if it can't find itself or control a node remotely as it can also take base64 macaroon and cert. 

I think the extra hosts way works ok, if you could test it on a Pi. 

I could see if I can do it another way in a future update maybe, if that's ok? Since there is a workaround to update the creds path via the UI if it's necessary.

niteshbalusu11:
> @niteshbalusu11 There are a couple of reasons we don't do/allow this; 1) is a security, where if your Docker hub account is breached, it could allow someone to push something malicious. 2) is predictability, if an update just goes out, it may not obvious to the user what version they're running or even if they want to run the latest. 3) is quality control, where we test app submissions/updates for specific versions to try our very best that users have a good experience when running these apps.

Cool, sounds good. Thanks!

niteshbalusu11:
Bumped version.

niteshbalusu11:
Bumped version (bug fixes)

lndboss/.bosgui/umbrel/credentials.json
lukechilds: Yeah these values need to be read from environment variables. It isn't safe to hardcode the IPs we currently use and just assume they won't change in the future.

Ideally LndBoss would accept env vars for LND connection details like Ride the Lightning does here:

https://github.com/getumbrel/umbrel-apps/blob/642a4bd52a2cfbb2431a84b1e9ea1b9a8b9d2273/ride-the-lightning/docker-compose.yml#L28-L31

Alternatively as a quick fix you could add an entrypoint to your container that reads these values from the env vars and writes them out to the config file, then starts LndBoss.

lukechilds:
Tested this on my node and it's working great, nice job @niteshbalusu11!

We just need to resolve the hardcoded config issue before we can merge: https://github.com/getumbrel/umbrel-apps/pull/65/files#r933121263 

Also a nice improvement would be if there was some sort of flag we could enable to disable authentication like `LNDBOSS_DISABLE_AUTH=true` or something like that. Umbrel already implements authentication in front of all apps so requiring a user to create an account and remember a secondary password to access the app is redundant. But that's totally up to you.

niteshbalusu11:
> Tested this on my node and it's working great, nice job @niteshbalusu11!
> 
> 
> 
> We just need to resolve the hardcoded config issue before we can merge: https://github.com/getumbrel/umbrel-apps/pull/65/files#r933121263 
> 
> 
> 
> Also a nice improvement would be if there was some sort of flag we could enable to disable authentication like `LNDBOSS_DISABLE_AUTH=true` or something like that. Umbrel already implements authentication in front of all apps so requiring a user to create an account and remember a secondary password to access the app is redundant. But that's totally up to you.


Thanks for reviewing @lukechilds 
The hard coding thing, I think it won't be an issue because all those values can be changed from the UI itself without anybody ever touching the command line. Do you think that's alright? 

Regarding disabling authentication, I cannot do that at the moment because the password is needed for a getting JWT auth token which is passed along when the client makes calls to the server. (The App has both a client and a server running on the same port).

lukechilds:
>Thanks for reviewing @lukechilds
The hard coding thing, I think it won't be an issue because all those values can be changed from the UI itself without anybody ever touching the command line. Do you think that's alright?

No I think we need to load this dynamically from the environment variables. We pass `$APP_LIGHTNING_NODE_IP` to apps which is the IP of the LND container on the internal private virtual network. If we ever change this in the future to be something different or assigned dynamically, Umbrel users will have no idea what this value is to enter it into your app. If it's dynamically assigned it also wouldn't be trivial to find, it would involve SSHing into the Umbrel and running advanced Docker commands that many Umbrel users wouldn't be capable of doing. They could potentially have to redo this each time the app is restarted. If LndBoss could read this information provided by Umbrel in environment variables this wouldn't be an issue. This is how all other apps connecting to LND on Umbrel currently work.

>Regarding disabling authentication, I cannot do that at the moment because the password is needed for a getting JWT auth token which is passed along when the client makes calls to the server. (The App has both a client and a server running on the same port).

A simple workaround for this would be that when `LNDBOSS_DISABLE_AUTH=true` is set the server will just return a valid JWT to anyone without requiring a password. This isn't insecure on Umbrel because only authenticated users are ever allowed through the proxy to hit that endpoint. That's not a major issue for shipping the app though, you can add that in an update any time in the future if you want to. Only the env var issue needs to be resolved before we can ship.

lukechilds:
`extra_hosts` isn't something we allow apps to use. It could also break stuff in the future if we add support for mDNS for containers. Also I think forcing `localhost` to resolve to the LND container IP could have a lot of strange side affects.

Are you able to pass that data through in environment variables like:

```yaml
environment:
  LN_SERVER_URL: "https://$APP_LIGHTNING_NODE_IP:$APP_LIGHTNING_NODE_REST_PORT" 
  MACAROON_PATH: "/lnd/data/chain/bitcoin/$APP_BITCOIN_NETWORK" 
  CONFIG_PATH: "/lnd/lnd.conf" 
```

or similar, however you want them formatted, and then read them in your app?

niteshbalusu11:
@lukechilds  Alright, I bumped the version and added support for pulling socket via ENV variable. `BOS_DEFAULT_LND_SOCKET: $APP_LIGHTNING_NODE_IP:$APP_LIGHTNING_NODE_GRPC_PORT` this in the docker-compose file should do it.

lndboss/umbrel-app.yml
```diff
@@ -0,0 +1,25 @@
+manifestVersion: 1
+id: lndboss
+category: Lightning Node Management
+name: LndBoss
+version: "1.12.0"
```
lukechilds, line 5:
```suggestion
version: "1.15.1"
```

lndboss/umbrel-app.yml
```diff
@@ -0,0 +1,25 @@
+manifestVersion: 1
+id: lndboss
+category: Lightning Node Management
+name: LndBoss
+version: "1.15.1"
```
lukechilds, line 5:
```suggestion
version: "1.15.2"
```

niteshbalusu11:
Version in the manifest needs to be 1.15.2, please. I missed fixing that.

lukechilds (approved):

lukechilds:
Looking great, thank you @niteshbalusu11!

Would be nice if the cert/macaroon could be passed in via env vars too. This would allow the app to work in regtest/testnet installs. But that's not super critical, I'll leave it up to you to add in an app update if you want to.

This is all good to g live! 🚀

@mayankchhabra are you able to put some gallery assets together?

niteshbalusu11:
> Looking great, thank you @niteshbalusu11!
> 
> 
> 
> Would be nice if the cert/macaroon could be passed in via env vars too. This would allow the app to work in regtest/testnet installs. But that's not super critical, I'll leave it up to you to add in an app update if you want to.
> 
> 
> 
> This is all good to go live! 🚀
> 
> 
> 
> @mayankchhabra are you able to put some gallery assets together?


It looks for all possible network versions when we the pass the lnd directory so that part is good 👍

mayankchhabra:
Solid work on LndBoss @niteshbalusu11! The app store gallery images and app icon are now live via [getumbrel/umbrel-apps-gallery](https://github.com/getumbrel/umbrel-apps-gallery/tree/master/lndboss).

Good luck with the launch in the app store! 🚀

PS: If you'd like to update the gallery images or the app icon in the future, feel free to submit a PR to the above repo. Cheers!

niteshbalusu11:
Thanks Mayank!

nevets963:
Hey @niteshbalusu11 

I was doing some final testing on my Pi-based Umbrel before a merge, but then I noticed the app takes a very long time to install (in my case about 4 minutes as I have a few apps installed) due to the Docker image extraction. The image is about 1.2GB which seems very big for a web app.

This creates a bad user experience due to the install time and burns a CPU for a fair amount of time doing the extraction. Can you look at why this is, and optimise the image? Potentially you’re including some large files that are actually not needed…

Steve

niteshbalusu11:
Hey @nevets963 - the app is the same size of Thunderhub for example which extracts to about 1.2GB. The app runs both web client and also a server on the same port which probably could be why the install time is taking a little longer, however once it starts up it does not take much resources to run. 
I don't think I'm including any unnecessary files but I can double check.

niteshbalusu11:
@nevets963 I figured out how to reduce the image size, it's now 600MB on my machine, can you test again?
Thank you for pointing it out, learning Docker slowly.

niteshbalusu11:
> 

Yeah, I tried keeping it, I noticed the Alpine image is smaller for my app than buster slim. The Alpine image is about 70MB smaller for me.

lukechilds:
Taking a quick look, it looks like the reason your Docker image is so large is because you have 87.9MB of source files included that aren't needed and 738.7MB of dependencies, many of which are not runtime deps and therefore aren't needed.

```
docker run -it --entrypoint sh niteshbalusu/lndboss:v1.16.4 -c 'du -h -d 1 .'
87.9M   ./src
2.1M    ./dist
738.7M  ./node_modules
828.7M  .
```

If I nuke that `src` dir the container still runs fine, I don't think it's needed at all.

Re the 738.7MB of Node.js dependencies (that's a lot!) looking at your project you have a lot of frontend dependencies added to the `dependencies` block of your package.json ([here](https://github.com/niteshbalusu11/lndboss/blob/f54bf9231954c49f799bda9a7461faab857e05cf/package.json#L38-L81)) which means these get included minified and treeshaken in the built source that gets served in your frontend (good) but also replicated in their raw unminified untreeshaken form in your servers node_modules directory (bad).

If you split these dependencies out into frontend build deps and backend runtime deps and ensure only runtime deps are included in the final Docker images you could probably reduce the image size to around 25% it's current size.

niteshbalusu11:
@lukechilds thanks for taking a look. In the latest version I only included the built front end in the src directory, (that's just my directory setup, front end build in the src folder and backend build in the root dir) so I dont think I can nuke that, the app would still start because the backend is serving it but I doubt anything would open in the browser.🤔

Regarding tree shaking, won't that depend on if the libraries used in the front end actually support it?  

I'm looking at improving build times and speed in future versions but it's going to take a while.

niteshbalusu11:
Spent a lot of hours this weekend to see how much bundle size I can decrease, there seems to be a NextJs limitation if you use a custom server to start production web app, so for now I have to include all client side deps as well until they address the issue. 
It is what it is for now, image size is at 630mb. 

There is another way is split up client and server and run them on separate ports, but that's adding a lot of complexity which is not worth it IMO.

PR #66, opened 2022-07-09
feat: update samourai-server app to 1.16.0

louneskmt:


nevets963 (commented):
Hey @louneskmt ! Thanks for pushing an update for Samourai Server!

I've managed to test on my mainnet running on ARM64 (a Pi) and everything appears to be working.
Atleast when I tested the Samourai Wallet app on my Android, the setup differed a tiny bit t the instructions here:
Screenshot 2022-07-10 at 21 14 38

e.g. As the connection here is via Tor, it's important that they click 'Enable Tor' during the setup process and this might not be obvious for everyone.

I couldn't test on AMD64 instance as it runs regtest and nginx crashes due to missing conf file. Perhaps there is a way to support regtest, or maybe provide a conf file for regtest that simply returns a static message that regtest is not supported (so that atleast the app starts) - just an idea.

@lukechilds Could you also test? I think you said you have a AMD64 mainnet node right?

lukechilds:
Yep running on amd64 mainnet. I already have the app setup and paired with clients, just tested a rescan.

lukechilds:
@louneskmt do you wanna add an entry for the new `releaseNotes` property in `umbrel-app.yml` for this PR?

louneskmt:
Sure thing, will do in a hour or so

lukechilds:
Thanks @louneskmt!

PR #68, opened 2022-07-10
LNDg v1.2.1

cryptosharks131:
Updates LNDg to v1.2.0
https://github.com/cryptosharks131/lndg/pkgs/container/lndg/28046226?tag=v1.2.0
https://github.com/cryptosharks131/lndg/releases/tag/1.2.0

### DIFF
--- a/lndg/docker-compose.yml
+++ b/lndg/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: $APP_LNDG_PORT
 
   web:
-    image: ghcr.io/cryptosharks131/lndg:v1.1.1@sha256:fd975e4d6ee2de6da932eb8e93648b13ad7de9b71ec2a35eac19d2845d7a5f31
+    image: ghcr.io/cryptosharks131/lndg:v1.2.0@sha256:f43af90d521eb696d22cc35fd197fc7510b686c14fe2025ec0949b7d709a7670
     restart: on-failure
     stop_grace_period: 1m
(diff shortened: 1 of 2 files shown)


nevets963 (approved):
Hey @cryptosharks131, thank you for the LNDg app update! Approved from me!

@cryptosharks131 Umbrel v0.5 introduced the 'App Proxy' which now protects all apps using the user's Umbrel password (and where appicable 2FA). This means we can drop secondary passwords for apps and improve the overall user experience. Reading the docs (https://github.com/cryptosharks131/lndg/blob/master/README.md#setup-lndg-initializepy-options) it looks like a password is mandatory atm. Would you consider making an app password optional, and then not using this with Umbrel?

@lukechilds Do you also want to check this over?

cryptosharks131:
The password is mandatory at the moment but I can look into the possibility of making it an optional system setting in a future update.

I have another build `v1.2.1` I will release in the next day or so to fix a small issue found with `v1.2.0`. It may be best to just wait for that build to be ready.

cryptosharks131:
Updated images to v1.2.1 and requested for approval again.
https://github.com/cryptosharks131/lndg/pkgs/container/lndg/28159541?tag=v1.2.1
https://github.com/cryptosharks131/lndg/releases/tag/1.2.1

nevets963 (approved):
LNDg v1.2.1 tested on AMD64 and ARM64. Approved!

Will merge now.

PR #81, opened 2022-07-22
Update LnVisualizer to v0.0.25

MaxKotlan:
new features:

- initial sync performance improvements
- can filter by node feature bits
- filter dropdown for node network type (onion vs clearnet)
- new statistics window. Will calculate min/max/total/average for both the entire network and for subgraph being viewed.
- new mode to normalize channel color to min/max range of subgraph

### DIFF
--- a/ln-visualizer/docker-compose.yml
+++ b/ln-visualizer/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 80
 
   web:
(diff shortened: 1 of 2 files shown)


lukechilds:
Nice work @MaxKotlan! We now have a new `releaseNotes` property in `umbrel-app.yml` btw which might be nice to make use of to communicate these new features. Users will see the `releaseNotes` text in the app update UI before they click update.

MaxKotlan:
Thanks @lukechilds. 
Cool feature! I added the release notes section. Will make sure to update that from now on.

lukechilds (approved):
Nice update, working great for me!

PR #82, opened 2022-07-22
Upgrade ItchySats to `0.5.0`

scratchscratchscratchy:
Long was the wait - finally a new ItchySats version!

Note: Users on versions `0.4.x` are still supported for the time being, `0.5.0` is not breaking. However, support for previous protocol version will eventually be discontinued, upgrading is encouraged :)

For detailed changes please refer to the [release changelogs](https://github.com/itchysats/itchysats/releases/).

### DIFF
--- a/itchysats/docker-compose.yml
+++ b/itchysats/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8000
 
   web:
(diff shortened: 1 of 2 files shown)


lukechilds:
Thanks for this @scratchscratchscratchy! We actually have a new `releaseNotes` property in `umbrel-app.yml` that will be shown in the app update view to users. This would be a great place for you to communicate the upgrade recommendation.

lukechilds (approved):
Looks great, thanks!

PR #85, opened 2022-07-26
Add environment variable for ItchySats

scratchscratchscratchy:
Unfortunately this one slipped through because we closed a previous upgrade PR that included it.
The missing variable is not critical, i.e. the application works without it, but reports may be incorrect.

### DIFF
--- a/itchysats/docker-compose.yml
+++ b/itchysats/docker-compose.yml
@@ -17,6 +17,8 @@ services:
       - --umbrel-seed=$APP_SEED
       - $APP_BITCOIN_NETWORK
       - --electrum=tcp://$APP_ELECTRS_NODE_IP:$APP_ELECTRS_NODE_PORT
+    environment:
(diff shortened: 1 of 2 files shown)


lukechilds:
@scratchscratchscratchy do you wanna bump the version too so all existing 0.5.0 installs get prompted to update?

We can just change `0.5.0` [here](https://github.com/getumbrel/umbrel-apps/blob/581247366599f2c2d9dc132580a315d3d9a64ba8/itchysats/umbrel-app.yml#L5) to `0.5.0-build-2` or something to that effect. Whatever you prefer.

lukechilds:
Would be good to update the release notes too if so.

scratchscratchscratchy:
Thanks for the hints! Sorry I missed the ping here 😬

lukechilds (approved):

lukechilds:
Looks great!

PR #91, opened 2022-07-30
Bump Oak Node to v0.3.2

ok300:
Bump to next version, which brings a Nostr bot and ability to send to onion LN addresses.

Tested on an Umbrel VM.

### DIFF
--- a/oak-node/docker-compose.yml
+++ b/oak-node/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: $APP_OAK_NODE_PORT
 
   web:
-    image: oak-node.net/oak:v0.3.1@sha256:f78f19763705dfb2237fd8a5c78f27052a848a5e56d89b3ffbcdc1edef0ba57e
+    image: oak-node.net/oak:v0.3.2@sha256:ed6dc77c36032b036b7182b02305b76312fa6c89c7cafc6047cb59d32ab1e507
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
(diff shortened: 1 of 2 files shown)


lukechilds:
Woop woop Nostr!

Btw what format should the Nostr pubkey be in? I tried adding my npub but I get an `Invalid pubkey` error.

lukechilds (approved):
Will wait for confirmation from your on if the Nostry pubkey thing is a bug or if I did something wrong.

Other than that, all tested and working well and ready to ship!

ok300:
@lukechilds thanks for the quick review!

Interesting finding, haven't seen an npub key before. Maybe its a new addition to the protocol?

The way I tested was:

* I went to a branle instance, which is a browser-based Nostr client
  * like https://nostr.rocks or https://branle.netlify.app
* in the welcome popup, I clicked "Generate"
* Then clicked "Proceed" 
* A new popup then showed the generated nostr private and public keys
* I saved the private key (to later sign-in to other Nostr instances) and pasted the pubkey to the Oak Node UI

This is not the most secure way to generate keys, as its browser-based. But it is the fastest. There are local clients which can generate keys locally too.

Does this work for you?

Also could you tell me how you got the npub key? The tool you used or the process? Might have to support that as well.

lukechilds:
I'm using Damus (https://github.com/damus-io/damus) by @jb55. Unsure if npub is a Damus specific thing or part of the spec. Yep just tried with nostr.rocks with the hex format and it works. Very cool!

lukechilds:
@ok300 this is seriously badass.

Screenshot 2022-07-31 at 18 47 02

ok300:
Glad you like it :)

Yes the Damus pubkey might not be a Nostr pubkey per se.

I'll take that as feedback and add better instructions on the bot page re: the type of pubkey, how to get one, etc.

jb55:
> I'm using Damus (https://github.com/damus-io/damus) by @jb55. Unsure if npub is a Damus specific thing or part of the spec. Yep just tried with nostr.rocks with the hex format and it works. Very cool!

yeah it's not spec'd yet but you can convert the new key format to an older hex key format via https://damus.io/key/

jb55:
btw this is dope, love to see nostr integrations!

lukechilds:
Niceee, cheers @jb55, https://damus.io/key worked great!


@ok300 just fyi, there's no way to change the Nostr pubkey from the UI that I can see. I had to uninstall and then reinstall the app to do a hard reset and add my real Nostr pubkey.

ok300:
> Niceee, cheers @jb55, https://damus.io/key worked great!

Looking good, didn't know it could do that!


> @ok300 just fyi, there's no way to change the Nostr pubkey from the UI that I can see. I had to uninstall and then reinstall the app to do a hard reset and add my real Nostr pubkey.

Good point. Just implemented it, available with the next release.

PR #93, opened 2022-07-30
fix: samourai-server app depends on electrs

louneskmt:
Samourai Server depends on electrs (local indexer).

https://github.com/getumbrel/umbrel-apps/blob/642a4bd52a2cfbb2431a84b1e9ea1b9a8b9d2273/samourai-server/docker-compose.yml#L80-L84

### DIFF
--- a/samourai-server/umbrel-app.yml
+++ b/samourai-server/umbrel-app.yml
@@ -20,6 +20,7 @@ developer: Samourai
 website: https://samouraiwallet.com
 dependencies:
   - bitcoin
+  - electrs
 repo: https://github.com/louneskmt/umbrel-samourai-dojo/tree/v1.16.1-umbrel
 support: https://t.me/SamouraiWallet
 port: 3005


louneskmt:
How can we handle cases where people previously downloaded Samourai Server without installing electrs? They currently have a broken install.

louneskmt:
@nevets963 @lukechilds How can we handle cases where people previously downloaded Samourai Server without installing electrs? They currently have a broken install.

nevets963:
@louneskmt We can’t at this very moment as we only check app dependencies when an app is installed and not when an app is updated. A small oversight in v0.5. A hacky solution could be if we put some code into `exports.sh` to install electrs… Ideally we should try to solve this in the next Umbrel update.

lukechilds:
Just quickly merged this so we don't have the issue on any new installs. Good catch lounes!

Re fixing old installs, have there been reports of issues with Samourai Server that have turned out to not have electrs? We know anyone that updated from pre 0.5 would have automatically got electrs installed as part of the migration. The number of users that have installed a fresh Umbrel since 0.5 was released and manually installed Samourai Server but not electrs is probably very low. Potentially the simple solution is to tell these users to also install electrs if they contact us. If they just tried re-installing the app on their own it should now resolve the issue so it probably wouldn't even come to that.

If we want to deal with it automatically we could add a post update hook to samourai-server that attempts to install electrs if it isn't installed. I think that's overkill though.

louneskmt:
I think it's overkill too. There is one user facing this, and I already told him to install electrs.

I'm not sure what happens if you're running this app without electrs. I know that when checking on the DMT, there is no info regarding the local indexer, but not sure if the fact that the indexer is missing is actually displayed somewhere, and if the user can successfully peer their SW with the Dojo. If they can, and can send coins to a generated address, it could be a pain to then reinstall Dojo and reimport the xpubs.

PR #98, opened 2022-08-02
Remove Lightning Shell.

ibz:
I got really busy recently building Plebeian Market and I don't find any more spare time to update this app.

I think it is better to remove it, since the whole point of is was having recent versions of all included apps.

Thanks!


nevets963:
Hey @ibz Does the above sound like a reasonable solution for now?

PR #102, opened 2022-08-06
Lnbits: Fix public exposition

johnongit:
This commit fix the following issue when we try to expose lnbits through a reverse proxy. 
https://github.com/lnbits/lnbits-legend/issues/696

Currently, lnbits doesn't trust x-fowarded-proto: https header and umbrel APP proxy erase the header

### DIFF
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -6,6 +6,7 @@ services:
       APP_HOST: $APP_LNBITS_IP
       APP_PORT: $APP_LNBITS_PORT
       PROXY_AUTH_ADD: "false"
+      PROXY_TRUST_UPSTREAM: "true
   web:
     image: lnbitsdocker/lnbits-legend:0.8.0@sha256:3426e326bac347d09b4e018780cb1ecd8e5eac35851383d476076769ab4a9b2e
     user: 1000:1000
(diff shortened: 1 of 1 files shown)


lukechilds:
This looks like it could introduce issues where anyone would be able to ping an LNbits instance running on Umbrel with `X-Forwarded-For` set to some random IP to spoof the IP in a way LNbits would trust. Unsure if this could have significant security implications for LNbits but certainly seems like not the right choice for the default config. This is only safe to do when running Umbrel's app proxy behind a trusted proxy, not when exposing the Umbrel app proxy directly which is the common use case. 

`PROXY_TRUST_UPSTREAM="true"` can be set on a per user basis for users running reverse proxies. See example relating to BTCPay here: https://github.com/getumbrel/umbrel/issues/1428#issuecomment-1184584190

johnongit:
Thx for your quick answer. 
Lnbits seems to be a useless extension in local and Tor context (lnurl on Tor is badly supported by wallet). 

I have created a little script to update docker-compose file for users who want to expose their lnbits extension. 

If this PR is refused, I will just propose a simple tutorial like btcpay. 

Unfortunately, each lnbits app update will erase the "fix" (this why I have submitted this PR)

nevets963:
@johnongit As @lukechilds has suggested, we can't have apps blindly trusting the upstream as it makes it trivial to spoof certain potentially important HTTP headers. Our best solution to circumvent our protections is documented here: https://github.com/getumbrel/umbrel/issues/1428#issuecomment-1184584190

Unless we can come up with another idea; shall I close this PR?

PR #103, opened 2022-08-06
Update jam v0.0.10

theborakompanioni:
This version includes:
- Jam: [v0.0.10](https://github.com/joinmarket-webui/jam/releases/tag/v0.0.10)
- joinmarket-clientserver: [v0.9.6](https://github.com/JoinMarket-Org/joinmarket-clientserver/releases/tag/v0.9.6)

All notable changes of the UI can be seen in the [Jam v0.0.10 changelog](https://github.com/joinmarket-webui/jam/blob/v0.0.10/CHANGELOG.md)

Updates to the image:
- opt-in for basic authentication
- run local tor daemon

### DIFF
--- a/jam/docker-compose.yml
+++ b/jam/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 80
 
   jam:
-    image: ghcr.io/joinmarket-webui/joinmarket-webui-standalone:v0.0.9-clientserver-v0.9.6@sha256:f36da8ed2f75b8db8dca34783b1e1ed11d0048270140b1c2cc072b15191e6b6c
+    image: ghcr.io/joinmarket-webui/joinmarket-webui-standalone:v0.0.9-clientserver-v0.9.6@sha256:4287ebb5d7d9bda7acdc77662df7adf0b6e7a5974a39a314c3ec0e85336ec1d5
     restart: on-failure
     stop_grace_period: 1m
     init: true
--- a/jam/docker-compose.yml
+++ b/jam/docker-compose.yml
@@ -17,8 +17,6 @@ services:
       RESTORE_DEFAULT_CONFIG: "true"
       REMOVE_LOCK_FILES: "true"
       ENSURE_WALLET: "true"
-      APP_USER: umbrel
-      APP_PASSWORD: "${APP_PASSWORD}"
       jm_network: $APP_BITCOIN_NETWORK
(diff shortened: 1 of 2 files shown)


jam/docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       APP_PORT: 80
 
   jam:
-    image: ghcr.io/joinmarket-webui/joinmarket-webui-standalone:v0.0.9-clientserver-v0.9.6@sha256:f36da8ed2f75b8db8dca34783b1e1ed11d0048270140b1c2cc072b15191e6b6c
+    image: ghcr.io/joinmarket-webui/joinmarket-webui-standalone:v0.0.9-clientserver-v0.9.6@sha256:4287ebb5d7d9bda7acdc77662df7adf0b6e7a5974a39a314c3ec0e85336ec1d5
```
nevets963, line 10: The digest for a multi-arch image should be `sha256:f36da8ed2f75b8db8dca34783b1e1ed11d0048270140b1c2cc072b15191e6b6c` here.

nevets963 (commented):
@theborakompanioni Thanks for the Jam update and I'm glad my help was useful!

It looks like the digest for the image is wrong or a specific arch. You can can get the multi-arch digest using `docker pull <image>` and then it will display it at the bottom.

Re, removing `defaultUsername`, `defaultPassword` and `deterministicPassword`- you're correct in that these should become empty strings or `false` for `deterministicPassword`. **But**, there is currently a bug in Umbrel's frontend where if we remove them now it will remove these fields for current installations of Jam and those users will not see/know the credentials to login. A bugfix for this goes out in the next Umbrel version, so I propose in your next Jam update PR we remove/correct these properties then. Of course this isn't a bullet proof solution, but it appears that most users update their Umbrels quite quickly.

jam/docker-compose.yml
theborakompanioni: Oh man, I forgot to adapt the version : /
Well.. that's embarrassing. Fix now.  Thank you.

theborakompanioni:
> @theborakompanioni Thanks for the Jam update and I'm glad my help was useful!

Of course. You are doing an amazing job. Can recommend : -)

> [...], so I propose in your next Jam update PR we remove/correct these properties then. Of course this isn't a bullet proof solution, but it appears that most users update their Umbrels quite quickly.

Okay, yes, thanks for the suggestion. Re-added setting up basic authentication.

PR #108, opened 2022-08-11
Elements: Bump to latest block

tiero:
https://blockstream.info/liquid/block/d38e10daf597ae3a4d27ada44d9bca592ac0302446fd8b7c9c2b0844e23fbaab

Please @lukechilds review

### DIFF
--- a/elements/docker-compose.yml
+++ b/elements/docker-compose.yml
@@ -20,7 +20,7 @@ services:
       - -txindex=1
       # Current Elements Core requires >8GB of RAM to verify confidential proofs during IBD
       # We skip historical block verification for now, as we don't have a way to set the memory limit
-      - -assumevalid=d4046e2c27e32e8582af39e402528e5df6ede1e1dd3c551764699e02e50bbf1e
+      - -assumevalid=d38e10daf597ae3a4d27ada44d9bca592ac0302446fd8b7c9c2b0844e23fbaab
       - -validatepegin=1
       - -fallbackfee=0.000001
       # Attach to Bitcoin network


lukechilds:
Hey @tiero, thanks for the PR!

I'm not sure this actually does what we originally thought it does. The goal was to skip validation to attempt to reduce memory usage right? I've noticed Elements uses ~6GB RAM on my machine, even after skipping validation. Maybe it makes more sense to drop this option and add a line to the app description that warns the user they need at least ~6GB of free RAM to run this app? What do you think?

tiero:
You are correct, memory usage spiked last month. Hopefully, the upcoming Elements release will be less resource hungry.

> add a line to the app description that warns the user they need at least ~6GB of free RAM to run this app? What do you think?

ok lemme do that

tiero:
@lukechilds It uses 600MB looking in a elements node on a AWS machine of mine

image

lukechilds:
Hmmmn that's odd, here's mine:

```
CONTAINER ID   NAME              CPU %     MEM USAGE / LIMIT     MEM %     NET I/O           BLOCK I/O        PIDS
b60a8ba1effe   elements_node_1   200.07%   4.306GiB / 30.69GiB   14.03%    97.9MB / 2.11MB   4.38GB / 729MB   28
```

That's after a fresh start, I haven't had it running for a while.

Does your AWS node have the same config we're using in Umbrel?

tiero:
Only difference may be the validatepegin=0

Other thing is that my node is always up, so no need to verify bulk of new blocks in a row, but only one per minute which makes it more manageable

lukechilds:
Maybe it's `validatepegin=0` that's causing the high mem usage? I'm also running on an amd64 AWS box in the above example. Even once sync has caught up it still sits around 4GB-6GB.

nevets963:
@tiero Were you able to work out if `validatepegin=0` makes a difference in order to progress this PR?

PR #112, opened 2022-08-16
Upgrade LND to v0.15.1-beta.rc1

kroese:


lukechilds:
Thanks for the submission @kroese. Appreciate you taking the time to put this together but we don't normally ship release candidates, we wait until there is a proper release.

PR #117, opened 2022-08-18
App Submission: Whoogle Search

ceramicwhite:
# App Submission

### App name
Whoogle Search
...

### 256x256 SVG icon
https://svgshare.com/i/kDM.svg

...

### Gallery images

whoogle-1
whoogle-2
whoogle-3
whoogle-4

...

### I have tested my app on:


nevets963 (commented):
Hey @ceramicwhite Thanks so much for the PR to add Whoogle Search to the Umbrel App Store! I just pushed a couple of minor changes to the PR and tested on my `amd64` and `arm64` Umbrel servers and it's working perfectly.

This is good to go from me 🚀 

@mayankchhabra Could you prepare the app store assets?

Once that's done, I'll patch up the container IPs and get this merged :)

whoogle-search/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: $APP_WHOOGLE_SEARCH_IP
+      APP_PORT: $APP_WHOOGLE_SEARCH_PORT
+      PROXY_AUTH_ADD: "false"
```
nevets963, line 8: @ceramicwhite Is there a specific reason you don't want to password protect Whoogle? We try to ensure that all user's applications are protected and not just open to usage (as potentially some users have deployed Umbrel in the cloud).

ceramicwhite: Like I mentioned on the Libreddit submission, I just think this is an app similar to the way I use my self-hosted mempool.space. I want to jump straight in and not need to type my umbrel password for an app that doesn't contain technically sensitive information.
In terms of hosting in the cloud I have run an Umbrel on one of my VPS and you need to have subdomains for each app you install as the dashboard is not useable for launching apps and auth complicates this even more.
I completely understand the want to blanket protect all apps but I can say as a user for things like my explorer I would be turning it off. 
Maybe something for the Umbrel team to think about would be to potentially offer a toggle setting in the Umbrel UI to allow users to turn the auth off for certain apps that don't pertain sensitive data after users install them.

nevets963: Yh, I understand this POV and although we don't have a UI to control this behaviour, it is possible atm to disable the auth via SSH: https://github.com/getumbrel/umbrel/issues/1412#issuecomment-1149022283
In terms of hosting in the cloud; I thought a potential attack vector could be, someone/something malicous searching for incriminating terms - I don't know how likely, but seems plausible.

Could you clarify how the dashboard is not useable? Is it because some cloud vendors have firewalls that automatically block all ports?

ceramicwhite: Hmm yeah that's fair concern. 
Idk maybe there's a middle ground in the future, like an option for the app-proxy that is just a pin code, not super secure but a decent basic prevention measure for apps that don't contain sensitive data.
If you are using a VPS and probably have 2FA on, it would be really annoying to type your password and code in just to make a search.
If you setup a reverse proxy for the dashboard, apps will just show a "Starting" state, I remember seeing discussions around this pre the app store, only workaround I ever found is using [damanic/socatorssl](https://github.com/damanic/socatorssl) to get around this.
Example:
```
docker run \
    -p 443:443 \
    -e "PUBLIC_PORT=443" \
    -e "TOR_SITE=mgjh7ed75krjnem4xolgplj76fbnx66gxv3gwhvw34xzfkfjd7vqz2id.onion" \
    -e "TOR_SITE_PORT=9051" \
    -e "SSL_CERT=fullchain.pem" \
    -e "SSL_KEY=privkey.pem" \
    --name umbrelProxyDashboard \
    --mount type=bind,source=/etc/letsencrypt/archive/YOUR-SITE.com,target=/etc/socatorssl/certs/ \
    damanic/socatorssl
```

mayankchhabra:
Awesome work packaging Whoogle @ceramicwhite! The app store gallery images are [now live](https://github.com/getumbrel/umbrel-apps-gallery/tree/master/whoogle-search).

Excited for the launch in the app store! 🚀

PS: If you'd like to update the gallery images or the app icon in the future, feel free to submit a PR to the above repo. Cheers!

PR #119, opened 2022-08-22
Add umbrel-hosted Mempool block explorer to Thunderhub

aphex3k:
This PR replaces part of https://github.com/getumbrel/umbrel/pull/1213 and will add support for a locally hosted mempool instance increasing the privacy of the node and its users.


nevets963 (commented):
Hey @aphex3k Although I want to allow users to rely/use their local mempool instances for exploring blocks - I think this PR has 1 too many problems. If we force mempool as a dependency, then that will enforce electrs be also installed. And Thunderhub is perfectly capable of running without mempool (ie mempool is not a hard requirement).

Additiionally, the url for the env. var. `MEMPOOL_URL` would also be wrong in some contexts such as if the user accesses over Tor or via Tailscale. `$APP_DOMAIN` is likely to resolve to `umbrel.local` which is only suitable for access via the LAN.

I wish I had a better way atm to add this capability.

aphex3k:
I understand. I have it working for me on my local branch with a docker-compose.override (see https://github.com/getumbrel/umbrel/pull/1487). Maybe there is a chance in the future where the local mempool would be used when it is already installed anyways and the URL question has been resolved...

Issue #97, opened 2022-08-02
Application request: Plex Media Server


johnpc:
Would be awesome if Plex was available as an umbrel app. This is their docker image: https://github.com/plexinc/pms-docker

aphex3k:
PMS is now available in the app store: https://github.com/getumbrel/umbrel-apps/tree/master/plex

thashm:
No casting devices can be found on umbrel, yet on windows plesk media server casting devices like chromecast TV, etc can be found. How to fix this?

PR #121, opened 2022-08-24
app submission: plex media server

johnpc:
# App Submission

I was surprised that Plex was not already available on umbrel, as I saw chatter in an old reddit thread that it was coming soon. I have been using Plex on my umbrel device and it works seamlessly with SimpleTorrent downloads. I have been using my roku, Safari on my mac, and the iOS app clients and have not experienced any lag or issues on raspberry pi 4.

### App name

Plex

### 256x256 SVG icon

plex

...

### Gallery images

plex1
plex2
plex3
...

### I have tested my app on:

### DIFF
--- a/plex/docker-compose.yml
+++ b/plex/docker-compose.yml
@@ -1,0 +1,21 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: $APP_PLEX_IP
+      APP_PORT: $APP_PLEX_PORT
+
+  server:
+    image: ghcr.io/linuxserver/plex:version-1.28.1.6104-788f82488
+    user: "1000:1000"
+    restart: on-failure
+    environment:
+      - PUID=1001
+      - PGID=100
+      - VERSION=docker
+    volumes:
+      - ${APP_DATA_DIR}/downloads:/downloads
+    networks:
+      default:
+        ipv4_address: $APP_PLEX_IP
--- a/plex/umbrel-app.yml
+++ b/plex/umbrel-app.yml
@@ -1,0 +1,22 @@
+manifestVersion: 1
+id: plex
+category: Files
+name: Plex
+version: "1.28.1.6104-788f82488"
+tagline: Watch anytime, anywhere with the Plex app.
+description: >-
+  Plex is an American streaming media service and a client–server media player platform, made by Plex, Inc. The Plex Media Server organizes video, audio, and photos from a user's collections and from online services, and streams it to the players. The official clients and unofficial third-party clients run on mobile devices, smart TVs, streaming boxes, and in web apps.
+developer: Plex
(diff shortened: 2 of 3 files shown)


plex/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+  server:
+    image: ghcr.io/linuxserver/plex:version-1.28.1.6104-788f82488
+    user: "1000:1000"
+    restart: on-failure
+    environment:
+      - PUID=1001
+      - PGID=100
+      - VERSION=docker
+    volumes:
+      - ${APP_LIGHTNING_NODE_DATA_DIR}/downloads:/downloads
```
aphex3k, line 18: This is most likely not the correct volume you want to mount. This should probably look like this:
```
 - ${APP_DATA_DIR}/downloads:/downloads
```

plex/umbrel-app.yml
```diff
@@ -0,0 +1,23 @@
...
+category: Files
+name: Plex
+version: "1.28.1.6104-788f82488"
+tagline: Watch anytime, anywhere with the Plex app.
+description: >-
+  Plex is an American streaming media service and a client–server media player platform, made by Plex, Inc. The Plex Media Server organizes video, audio, and photos from a user's collections and from online services, and streams it to the players. The official clients and unofficial third-party clients run on mobile devices, smart TVs, streaming boxes, and in web apps.
+developer: Plex
+website: https://plex.tv
+dependencies:
+ - simple-torrent
```
aphex3k, line 12: This dependency is not strictly required for local file playback

aphex3k (changes requested):

plex/docker-compose.yml
aphex3k: Umbrel convention would actually be `- ${APP_DATA_DIR}/data:/downloads` but 🤷🏻

aphex3k:
This would resolve app request https://github.com/getumbrel/umbrel-apps/issues/97.

plex/docker-compose.yml
johnpc: Ah, I intended to point it to simple-torrent's data_dir. But to your point people might want to use plex without simple torrent data since plex does have other features

plex/umbrel-app.yml
johnpc: true, this is just how I use Plex but it's not necessary

aphex3k (approved):

lukechilds:
Hey @johnpc thanks for working on this!

We'd been working on some media apps (including Plex) in secret, we just made them live. We gave you commit credits in our final merged commit (https://github.com/getumbrel/umbrel-apps/commit/60878f278d544b204d8e7c96240c797f43a9b319) since you did some similar work over here. Thanks again!

PR #126, opened 2022-08-29
App Submission: Libreddit

ceramicwhite:
# App Submission

### App name
Libreddit

### 256x256 SVG icon
https://svgshare.com/i/m89.svg

...

### Gallery images

...

### I have tested my app on:


nevets963 (commented):
Hey @ceramicwhite Thanks for another app submission and for another fantastic PR! This is pretty much perfect!

There is just 1 question re protecting the Libreddit. By default we should have this on - you can think of this like how Android/iOS requires a passcode to access your apps.

@mayankchhabra Please can you prepare the App Store graphics?

If we can resolve the app proxy auth then is good to go 🚀

libreddit/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: $APP_LIBREDDIT_IP
+      APP_PORT: $APP_LIBREDDIT_PORT
+      PROXY_AUTH_ADD: "false"
```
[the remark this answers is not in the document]
ceramicwhite: No specific reason. Like using mempool.space having the auth there would be a pain more than a desire. I just feel like for users to want to use it, the auth would be a nuisance for apps like this but totally understand the want to protect it.

heysagnik:
mock 1
mock 2
mock 3

heysagnik:
suggestions appreciated

libreddit/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: $APP_LIBREDDIT_IP
+      APP_PORT: $APP_LIBREDDIT_PORT
+
+  web:
+    image: ceramicwhite/libreddit:0.22.9@sha256:00a133da4ae8eb1b7d3ce9af0ca387e2dd929dc6b490736532b5d998a025c362
```
nevets963, line 10: @ceramicwhite Do you have the `Dockerfile`s published somewhere online?

ceramicwhite: The images are untouched, I just pinned all the official arch images together as a list, as there are different Dockerfiles for each arch and the Images are tag separately and no official list was made.
If you inspect my list you can see the digests match the official Images on [docker hub](https://hub.docker.com/r/spikecodes/libreddit/tags)

nevets963: Ok nice 👍

mayankchhabra:
Great work here as well, @ceramicwhite! 🔥

@heysagnik thanks for working on the gallery images! 😍
They turned out quite nice. Are you available to incorporate some feedback in the next 1-2 days? Mainly:

- Change to dark mode browser (so it flushes well with the dark UI). Let me know if you need the Figma file for the browser.
- Change to Inter Font and use the same everywhere.
- Remove Libreddit icon from third image and Umbrel icon from the second image.
- I'm not sure we can use the Reddit logo in the first image due to copyright/trademark, so let's remove it and replace it with an avatar like you have in the first and third image.
- Move browser in the first and second image down a bit so we add a bit padding (like in the third image).
- Use capitalization in text.
- Replace text in the first image to: "Browse Reddit. Without the bloat."
- Replace text in the second image to: "No ads. No JavaScript. No tracking. Pure Reddit."
- Replace text in the third image to: "Blazing fast and feather light experience."

heysagnik:
mock 1
mock 2
mock 3

I tried ! Check these :)

mayankchhabra:
Nice work here @heysagnik! Unfortunately after looking at [Reddit's trademark use policy](https://www.redditinc.com/policies/trademark-use-policy), it seems like we may not have the rights to use their avatars. But thanks for sharing your Figma file. We tweaked the designs and replaced avatars with emojis. Thank you for your work!

The app store gallery images are [now live](https://github.com/getumbrel/umbrel-apps-gallery/tree/master/libreddit).

And with that, we're all set for the launch! 🚀

nevets963:
All round great work everyone! Merging in now 🚀 

This will be live for everyone in about 15 mins.

PR #128, opened 2022-08-30
App Submission: Nitter

ceramicwhite:
# App Submission

### Nitter
...

### 256x256 SVG icon
https://svgshare.com/i/mCx.svg

...

### Gallery images

...

### I have tested my app on:


nitter/data/nitter/entrypoint.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/sh
```
nevets963, line 1: It looks like this entrypoint script is never called/used as the `docker-compose.yml` has `/entrypoint.sh` defined as the `entrypoint` which looks like the default implementation included in the image already (https://github.com/goodtiding5/docker-nitter/blob/01a2e49c7a050edbe4b06bd49bb97b12fcac2a17/Dockerfile#L42)

nitter/data/nitter/nitter.conf
```diff
@@ -0,0 +1,45 @@
+[Server]
```
nevets963, line 1: I don't think this file is needed either as this also already bundled into the image (https://github.com/goodtiding5/docker-nitter/blob/01a2e49c7a050edbe4b06bd49bb97b12fcac2a17/Dockerfile#L43). I can see that you've filled out the values already, but I think we can leverage the env. vars. that the `entrypoint.sh` script is looking to do the templating for us?

nitter/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    # Official Nim image used as base image for nitter doesn't offer Arm variant so using a custom image
+    # https://github.com/zedeus/nitter/issues/399#issuecomment-997263495
+    image: quay.io/unixfox/nitter:build-20220830@sha256:d93da20aea7dbcf117e0f15bb309f5fc55729b19693d203c1c9708a2fe001582
+    restart: on-failure
+    stop_grace_period: 1m
+    init: true
+    # to pleb to figure out how to run non-root
+    #user: "1000:1000"
+    environment:
+      REDIS_HOST: "redis"
```
nevets963, line 17: If we leverage the `entrypoint.sh` script already in the image, then we should only have to populate the env. vars. here for the config. here. Maybe i've missed something as to why this wouldn't work?

nitter/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    volumes:
+      - ${APP_DATA_DIR}/data/nitter:/data
+    entrypoint: /entrypoint.sh
```
nevets963, line 20: Is this needed if the entrypoint is already set in the Dockerfile to use `/entrypoint.sh`?

nevets963 (commented):
Hey @ceramicwhite Thanks a lot for another app submission!! I've made a couple of tiny changes so I can test the app, but I also noted a few things and left some feedback below.

ceramicwhite:
All your comments are on point. Those files are redundant. Although for some reason the env vars don't seem to work when specified in the compose even when removing the entrypoint and .conf.

nevets963:
I think it's potentially because you already had a generated `.conf` file (https://github.com/goodtiding5/docker-nitter/blob/01a2e49c7a050edbe4b06bd49bb97b12fcac2a17/entrypoint.sh#L33). I just added the `REDIS_HOST` env. var. and it's working nicely thus far. I think with a little more polish, this is close to merge 🚀

nitter/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+
+  web:
+    # Official Nim image used as base image for nitter doesn't offer Arm variant so using a custom image w/ patch for umbrel
+    # https://github.com/ceramicwhite/umbrel-patch-ci/blob/master/nitter/umbrel.patch
+    image: ceramicwhite/nitter:build-20220906@sha256:6ddb53fe66846ffcded344d1919bee3149b705f04a9347c37b8eaa562695ce81
+    restart: on-failure
+    stop_grace_period: 1m
+    init: true
+    # app is run with su-exec to drop privileges to a non-root user
+    #user: "1000:1000"
```
nevets963, line 15: I took a look at the source and your patches today. I'm wondering if we can with reletive ease enable an arbitrary user/group ID to be used.

My idea is; we drop these lines: https://github.com/ceramicwhite/umbrel-patch-ci/blob/090560766cd67f141ec4147bb90dc68fcff42d3e/nitter/umbrel.patch#L82-L83

And then we drop the `chown` lines e.g. `chown www-data:www-data ABC` (https://github.com/goodtiding5/docker-nitter/blob/01a2e49c7a050edbe4b06bd49bb97b12fcac2a17/entrypoint.sh#L27).

@ceramicwhite Thoughts on this?

nevets963 (commented):
I've added some polish and I think we're nearly ready with this. I'm just wondering if we can use an arbitrary user/group id.

ceramicwhite:
@nevets963 It's looking looking good 🚀

nitter/docker-compose.yml
ceramicwhite: Yeah you're right it's fine on my end now

nevets963 (approved):
@ceramicwhite Amazing! Thanks for checking, and implementing that!

Tested on `amd64` and `arm64`. Approved! 🚀 

@mayankchhabra Can you please prepare the App Store graphics?

Before merge, I'll correct the container IPs :)

mayankchhabra:
Yooo solid work once again, @ceramicwhite! The app store gallery images are [now live](https://github.com/getumbrel/umbrel-apps-gallery/tree/master/nitter).

Excited for the launch! 🎉

ceramicwhite:
@nevets963 For some reason now it's live the theme is not defaulting to the Twitter one, which I think is a much cleaner theme to have as default.
If you want to fix I rebuilt the image and just set the theme directly in conf:
`ceramicwhite/nitter:build-20220912@sha256:a9ae4a90b9ebedd7cb9c9a7ec0c85c8107807dfc12b2125233958f536b5af733`

nevets963:
@ceramicwhite Now you mention it, I some very strange behaviour with regards to the theme... Using Firefox, I re-installed  nitter (many many times) and I see the `auto_(twitter)` theme. However, when I then open the app in Chrome using private browsing, I then see the `auto` (red) theme.

When I check the page source, I see this broken `link` tag:
```
<link rel="stylesheet" type="text/css" href="/css/themes/auto_.css" />
```

Do you have a diff you can share for that build?

Btw, do you know where the settings data is saved? I thought it would write back out to `nitter.conf`, but I change the theme, I don't see the new theme written in there.

ceramicwhite:
@nevets963 Ahh yep that's the problem, it needs the parentheses in the conf.

If we just change to:
`NITTER_THEME: '"auto_(twitter)"'`
It works fine on my end now.

nevets963:
@ceramicwhite Nice find! I think potentially all the config. values are suppossed to have quotes, but it's probably important here due to the parentheses. Anyways, this looks resolved by changing the env. value. This went live in this commit: https://github.com/getumbrel/umbrel-apps/commit/133b382c76e10dfbecf6d5daeda5788a3573186d

Thanks again!

PR #129, opened 2022-08-30
App Submission: Invidious

ceramicwhite:
# App Submission

### App name
Invidious

### 256x256 SVG icon
[invidious.svg](https://cloudflare-ipfs.com/ipfs/QmXZc15ZXd6Bnrk15RWaMaT6sknrqiBvyrtCXkbnMZNGf4/invidious.svg)

...

### Gallery images

...

### I have tested my app on:


nevets963 (commented):
Hey @ceramicwhite The app submissions keep on coming 🚀 Many thanks for submitting Invidious! I just gave it a test and it works really nicely 😄 

I'm wondering if we can simplify this PR and in general the distribution of Invidious if you included the DB schema/migrations into your `ceramicwhite/invidious` Docker image and then execute them out of there before the invidous binary runs? Basically your app will set itself up without any external files. I think it's potentially also a simplification when deploying to Kubernetes too.

Another reason (although this may change soon) is Umbrel's app update process only updates a few files such `docker-compose.yml`, `umbrel-app.yml`, etc. So if you add another extra DB table, then it becomes a little awkward vs simply including these files inside your Docker image.

What are your thoughts on the above?

ceramicwhite:
Yeah that makes sense. 
Those are the official images, I just pinned the amd & arm ones together. The migration files I can see are included in the image, I'm not sure how you would refence them in the initialization script besides as a declared volume:

```
invidious:
    volumes:
      - config:/invidious/config/sql
  postgres:
    volumes:
      - config:/config/sql
      - ${APP_DATA_DIR}/data/config/init-invidious-db.sh:/docker-entrypoint-initdb.d/init-invidious-db.sh
volumes:
    config:
```

Otherwise I would just copy them into the postgres base image: 
`ceramicwhite/postgres-invidious:v14@sha256:9a7fdb1fd71cc2f9e9ea1e9aacd6ce63fb906c36060a22ea69079cd790e77d44`

If there is a simple way to do this I'm all ears

ceramicwhite:
I tried and was unable to add into the entrypoint script and just went for using a Postgres image and adding in the migration files. It's using the same base as btcpayserver incase of cache at least.

nevets963 (commented):
@ceramicwhite I've made a few changes and everything seems to be working nicely! I've tested on `amd64` and `arm64`.

When I open Invidious, I see the Popular section is first open, but I don't see any videos. Do you know what might be wrong? If we can resolve this somehow, then this is good to ship 🚀  It's just not a great a UX to install, open the app and then see nothing.

ceramicwhite:
Totally agree and also wondered what was wrong, so I ran the official compose and then realized that's actually the search tab(default). Let me have a look maybe there is a var to start on the trending page instead.

ceramicwhite:
@nevets963 Aight default home is set to search.

Also a custom Postgres image is not needed as long as we use v13.x or lower and have ` check_tables: true`

nevets963 (approved):
@ceramicwhite Nice find with that config! It's also nice that we use stock postgres Docker image. Really great job here and thanks for working on this to really polish this PR! 🤩 

Tested on `arm64` and `amd64`. And approved from me! 🚀 

@mayankchhabra Would you able to create the app store graphics please?

mayankchhabra:
Keep up coming, @ceramicwhite! Thanks for opening the floodgates, haha. The app store gallery images are [now live](https://github.com/getumbrel/umbrel-apps-gallery/tree/master/invidious).

All set for the for launch! 🚀

pynixis (commented):
What's the status of this getting pushed to production?

nevets963:
@pynixis Invidious has been merged already and is available through the Umbrel App Store to download.

pynixis:
Correct. 

Current version: 2022.09.08 I believe latest is 2022.12.02.

ceramicwhite:
> Correct.
> 
> Current version: 2022.09.08 I believe latest is 2022.12.02.

Updated https://github.com/getumbrel/umbrel-apps/pull/317 to https://github.com/iv-org/invidious/pull/3426 @pynixis

TimothyShimmin:
Can this get an update?

ceramicwhite:
@TimothyShimmin 
the manifest is always up to date on docker hub. send the pr:

`ceramicwhite/invidious:build-eda7444@sha256:adb27aa2b7acda32c8eb006d1e8ec56b7669c3f9724f65bc86fcc2d526781c30`

PR #131, opened 2022-08-30
Update lnd to v0.15.1-beta

aphex3k:
1. Update LND to v0.15.1
2. Enable database migration


lightning/exports.sh
```diff
@@ -15,10 +15,6 @@ BIN_ARGS+=( "--rpclisten=0.0.0.0:${APP_LIGHTNING_NODE_GRPC_PORT}" )
 BIN_ARGS+=( "--restlisten=0.0.0.0:${APP_LIGHTNING_NODE_REST_PORT}" )
 BIN_ARGS+=( "--tlsautorefresh" )
 
-# Enable database migration for v0.15.1+
-# See https://github.com/lightningnetwork/lnd/releases/tag/v0.15.1-beta
-BIN_ARGS+=( "--db.prune-revocation" )
-
```
aphex3k: Any reasoning why this was disabled?

nevets963: I had a quick chat with the team. Ideally we don't want to have too many features on by default (ie be too opinionated) . There is potentially a reason why lnd doesn't just have this on by default. This flag will enable for everyone and make it difficult for people to turn it off. The people who gain most from this change are those with active/bigger LND nodes who can simply add this line to their `lnd.conf` if they wish.

aphex3k: Alright, thats understandable. Although I don't know if this is available as option in `.conf` because the release notes specifically call out it is a cli argument only.

nevets963: It looks like it can be set in the `.conf` file. I just found this in their sample: https://github.com/lightningnetwork/lnd/blob/9d04b0c3d9577b2ab8c253e140220e80181619b8/sample-lnd.conf#L1215

PR #132, opened 2022-09-01
Upgrade ItchySats to `0.6.0`

scratchscratchscratchy:
Changelog `0.6.0`: https://github.com/itchysats/itchysats/releases/tag/0.6.0

PR #133, opened 2022-09-01
Upgrade ItchySats to `0.6.1`

scratchscratchscratchy:
https://github.com/getumbrel/umbrel-apps/pull/132 merged super fast ❤️🚀

When testing on Umbrel we noticed a minor UI bug and fixed some log issues. 
Quick patch release `0.6.1`, for details see:
https://github.com/itchysats/itchysats/releases/tag/0.6.1

I kept the release notes for `0.6.0` in the release notes for `0.6.1` so users get informed about the new feature and dropping legacy support :)

### DIFF
--- a/itchysats/umbrel-app.yml
+++ b/itchysats/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: itchysats
 category: Finance
 name: ItchySats
-version: "0.6.0"
+version: "0.6.1"
 tagline: Peer-2-peer derivatives on Bitcoin
 description: >-
   ItchySats enables peer-2-peer CFD trading on Bitcoin using DLCs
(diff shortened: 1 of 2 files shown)


itchysats/umbrel-app.yml
```diff
@@ -60,5 +60,6 @@ defaultUsername: itchysats
 deterministicPassword: true
 torOnly: false
 releaseNotes: >-
-  This release enables ETHUSD CFDs. You can now trade BTCUSD and ETHUSD using your Bitcoin.
-  With this release support for the legacy network stack is dropped; versions prior to 0.5.4 are not supported anymore and have to upgrade to maintain compatibility.
\ No newline at end of file
+  This release fixes a minor UI bug for 0.6.0.
```
nevets963, line 63: 1 tiny thing (which was also there in the previous release) is; these lines won't appear on new lines so all the text is simply concatenated. With the 3 points; can you add an empty between each line? I would do this myself, but for some reason I can push to your branch.

nevets963 (commented):
Tested on `amd64` and `arm64`. Just 1 piece of feedback now that the release notes are a little bigger.

itchysats/umbrel-app.yml
scratchscratchscratchy: Done, thanks for the hint!

PR #144, opened 2022-09-13
Upgrade Pi-Hole to 2022.09.2

mccolister1017:
getumbrel/umbrel-apps/issues/142  Docker Tag 2022.09.2


nevets963 (approved):
@mccolister1017 Thank you for the Pi-hole app update PR!

I've also set the `path` to `/admin/` to automatically take the user to the correct page when click they app within Umbrel.

This has been tested on `amd64` and `arm64`. Approved! 🚀

Issue #145, opened 2022-09-14
Pi-Hole can't run as a DHCP Server


0x346e3730:
Looking at [pi-hole's docker-compose.yaml](https://github.com/getumbrel/umbrel-apps/blob/master/pi-hole/docker-compose.yml), this part is missing for it to be able to run as a DHCP server :
```yaml
cap_add:
      - NET_ADMIN
```

Everything is documented here https://github.com/pi-hole/docker-pi-hole, is it ok to do a PR to add this by default ? If so I've opened https://github.com/getumbrel/umbrel-apps/pull/146 to make it possible for pi-hole to act as a DHCP server

PR #146, opened 2022-09-14
Pi-hole: Add DHCP support

0x346e3730:
Add the possibility to pi-hole to act as a DHCP server


pi-hole/docker-compose.yml
```diff
@@ -24,3 +25,5 @@ services:
     networks:
       default:
         ipv4_address: $APP_PI_HOLE_IP
+    cap_add:
+      - NET_ADMIN
```
nevets963, line 29: Do you know why this capability needs to be added?

nevets963 (commented):
@0x346e3730 Thanks for the PR! Have you managed to test this within your network?

pi-hole/docker-compose.yml
0x346e3730: Yes, It's documented on https://github.com/pi-hole/docker-pi-hole as I explained in https://github.com/getumbrel/umbrel-apps/issues/145.

0x346e3730:
> @0x346e3730 Thanks for the PR! Have you managed to test this within your network?

I have not no as I have no test environment and cannot break my install

nevets963:
@0x346e3730 I've looked into the usage of `CAP_NET_ADMIN` and although I see it's a requirement for the DHCP server, it also unlocks a number of other privileges such as being able to modify network interface configuration (full details here: https://man7.org/linux/man-pages/man7/capabilities.7.html), which we have some reluctance to allow. As we also don't see other requests for this, I'm inclided to keep this PR open and allow other users to upvote, and then we can revisit this feature request again later.

BeauAgst:
This feature is required for anyone who is using Virgin Media in the UK, as you need to allow Pi-hole to take over DHCP. Without it, Pi-hole is entirely useless and does not work.

For context, Virgin Media has a 17% market share in the UK.

unlimited01:
This one bit me as I was migrating my Pihole server to an Umbrel deploy.  I had to eventually roll back.
the open port (`udp67`) and `NET_ADMIN` is how I run my existing pi-hole container now.  If I can't have y pi-hole instance run my DHCP, it's a blocker for me.

SolidRhino:
Please add this option. Ziggo (internet provider in the Netherlands) don't allow setting the DNS to an internal server.

nmfretz:
Thanks @0x346e3730, we've decided to move forward with adding DHCP functionality to the app!

> I have not no as I have no test environment and cannot break my install

I will test this evening when others on my network aren't using the internet. I have merged in the most recent changes from Master and tested that the app does not break when updated, so if you want to test this you do not need to worry about your install breaking.

Here are the steps to update your app if you decide you want to test:

1) Connect to your Umbrel device by SSH on a Raspberry Pi or by accessing a terminal on your Linux machine.

 2) Run `sudo ~/umbrel/scripts/repo checkout https://github.com/0x346e3730/umbrel-apps.git#pihole-dhcp` to change the apps repo on your umbrel so that you can update the pi-hole version from this PR.
 
 3) Either update pi-hole through the app store in the browser UI, or by running the following command:
 `sudo ~/umbrel/scripts/app update pi-hole`
 
 After you are done testing you can reset the apps repo to the official umbrel-apps repo:
 `sudo ~/umbrel/scripts/repo checkout https://github.com/getumbrel/umbrel-apps.git`
 
 and restore the previous version of pi-hole through the app store in the browser UI, or by running:
 `sudo ~/umbrel/scripts/app update pi-hole`

nmfretz:
Leaving this here for our reference:

[Docker DHCP docs](https://docs.pi-hole.net/docker/dhcp/#docker-dhcp-and-network-modes)

0x346e3730:
Since then I've bought a router so that even if I change my internet provider or my box I don't hade to re-do my whole install so I won't be able to test it neither but I'm glad it's getting added !

nmfretz:
Going live! Thanks for the feedback everyone.

PR #151, opened 2022-09-15
App Submission: Tdex Provider

tiero:
# App Submission

### App name

TDEX Provider

### 256x256 SVG icon

TDEX_240_icon

### Gallery images

- 
Screenshot 2022-10-11 at 13 54 52
Screenshot 2022-10-11 at 13 55 55
Screenshot 2022-10-11 at 13 56 40

### I have tested my app on:


nevets963 (commented):
Hey @tiero. Thank you for the app submission! Let me know if this is ready for review and I'll give this a test 😄

tiero:
I added the SVG Icon for the App! Give some time to prepare screenshot as well.

> Let me know if this is ready for review and I'll give this a test

I think is ready, we just need to figure out one thing: we have two services, one for the daemon/backend and one for the dashaboard/frontend. 

The frontend wants a "TDEXD Connect URL" as pairing mechanism, to fetch it you can visit the backend endpoint which will show an HTML page showing the connection string.

tdex/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
+version: "3.8"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: $APP_TDEX_CONNECT_IP
+      APP_PORT: $APP_TDEX_CONNECT_PORT
+  
+  tdexd:
+    container_name: "tdexd"
```
nevets963, line 10: We can drop this, and rely on docker-compose to name the container automatically for us.

tdex/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+    image: docker pull ghcr.io/tdex-network/tdexd:v0.8.15@sha256:e79aff49e5220d55df7c94d01e404e11b7e5cf094b459b3e6fbb918a7f2a40d6
```
nevets963, line 11: This looks like a typo. This should be: `ghcr.io/tdex-network/tdexd:v0.8.15@sha256:e79aff49e5220d55df7c94d01e404e11b7e5cf094b459b3e6fbb918a7f2a40d6`

tdex/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+      # $DEVICE_DOMAIN_NAME - A .local domain name for the Umbrel server (e.g. "umbrel.local")
+      #
+      # Tor proxy environment variables
+      # $TOR_PROXY_IP - Local IP of Tor proxy
+      # $TOR_PROXY_PORT - Port of Tor proxy
+      #
+      # App specific environment variables
+      # $APP_HIDDEN_SERVICE - The address of the Tor hidden service your app will be exposed at
+      # $APP_PASSWORD - Unique plain text password that can be used for authentication in your app, shown to the user in the Umbrel UI
+      # $APP_SEED - Unique 256 bit long hex string (128 bits of entropy) deterministically derived from user's Umbrel seed and your app's ID
```
nevets963, line 48: Shall we cleanup all these dead comments?

tdex/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+  dashboard:
+    container_name: dashboard
+    image: ghcr.io/tdex-network/dashboard:latest
```
nevets963, line 51: This needs to be pinned to a specific version along with a digest for security reasons

tdex/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+    restart: unless-stopped
```
nevets963, line 52: We usually default to `on-failure` so that the container only restarts with a non-zero status code (which indicates an error).

tdex/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+    environment:
+      USE_PROXY: "false"
```
nevets963, line 54: What is this used for? As we will 'front' the application with a http proxy (the app proxy), so I'm wondering if this relevant.

tdex/exports.sh
```diff
@@ -0,0 +1,2 @@
+export APP_TDEX_CONNECT_IP="x.x.x.x"
+export APP_TDEX_CONNECT_PORT="9999"
```
nevets963, line 2: Don't worry about these details for now, I can patch this up for you

tdex/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
...
+
+  A small provider's fee can be taken out of each trade and added to the reserves. While the BASE_ASSET-QUOTE_ASSET reserve ratio is constantly shifting, fees make sure that the total combined reserve size increases with every trade. Guaranteed arbitrage opportunities from price fluctuations should push a steady flow of transactions through the system and increase the amount of fee revenue generated. 
+
+  A liquidity provider has full control over the market making strategy to apply needed to calculate the market rate at which to accept trades. That being said, there is a possibility to apply an automated market-making relying only on the reserves balances and the amount requested by the trader, without the need to connect to an external price feed. The default strategy of the alpha daemond is the constant product market-making. In short, this model generates a full order-book based on an initial price for the market. Every transaction that occurs on this market will adjust the prices of the market accordingly. It's a basic supply and demand automated market making system.
```
nevets963, line 12: This all looks really nicely filled out! The detail here is great!

nevets963 (commented):
Hey @tiero I've left some feedback to help get this PR into a testable state. With a few small changes, I can try this out on my Pi! 😄 

Re

> The frontend wants a "TDEXD Connect URL" as pairing mechanism, to fetch it you can visit the backend endpoint which will show an HTML page showing the connection string. Any suggestion on best to do it? ie. to redirect users from one page to another?

We suggest only having 1 Web UI that is exposed to the user because each app is protected by the App Proxy which protects all apps with the user's password (and where applicable 2FA). If you need to share some state, this can be done programmatically by having a folder mounted in both containers that can then share some state.

If this is not possible, then alternatively you can run a small reverse proxy (e.g. caddy) that exposes your UIs on 1 port. For example, `/` to route to your dashboard and `/daemon` to the daemon.

tdex/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
+manifestVersion: 1
+id: tdex
+category: Finance
+name: TDEX Provider
+version: "0.8.15"
+tagline: Provider liquidity on the TDEX Network, an open protocol to trade Liquid Network assets.
```
lukechilds, line 6: This should be provide, right?

```suggestion
tagline: Provide liquidity on the TDEX Network, an open protocol to trade Liquid Network assets.
```

tdex/docker-compose.yml
tiero: This is a app related variable, we use it to tell we are not running as Tauri packaged app and therefore to not use the grcp-web to grpc proxy embedded

tiero:
Ok fixed most of the comments. I should be able to test on raspberry as well tomorrow

nevets963:
@tiero Would you be able to grant me access to that I can push a few changes to your PR/branch?

> This can be onion hidden service, but must be accessible from anybody outside. Any ideas on how to expose this?

Do you mean you want to expose the `tdexd` service via a Tor hidden service? If so, you can define a `torrc.template` file. E.g: https://github.com/getumbrel/umbrel-apps/blob/f6d132c786b65ff4a7d1f09dbe080ac34aac4e41/electrs/torrc.template#L1-L7 And then inside of `exports.sh` read the Tor Onion hostname:
https://github.com/getumbrel/umbrel-apps/blob/f6d132c786b65ff4a7d1f09dbe080ac34aac4e41/electrs/exports.sh#L6-L7 

Which can then be fed into application as an environment variable.

tiero:
Yes granted write access to the repo.

> Which can then be fed into application as an environment variable.

Seems perfect! Thanks

tiero:
> do you think is correct to pass the service name (ie. `tdexd:9090`) as the URL of the daemon/backend to the frontend


I realized that this way accessing from remote will not work, so we need to pass the onion hidden service, I am right?

Would this make mandatory to use Tor Browser also when reached from LAN?

nevets963:
@tiero I can help you with the `torrc.template` file :) I'll patch this up as soon as I can.

Re how the dashboard communicate with the daemon. I see a problem here; you won't be able to use `tdexd:9090` as a hostname that the dashboard can access. This hostname is only accessible inside of the Docker bridge network (this is useful when containers talk to other containers). But here, the browser won't understand what `tdexd:9090` means.

What I would suggest is; you run a HTTP proxy (e.g Caddy), have `/` defined as your dashboard and perhaps `/api` or `/daemon` for your daemon. And then have your dashboard automatically set the hostname for daemon using `window.location.host`. This would then remove setup burden from the user, but also work in any environment/context, whether it's over `umbrel.local`, access using Tailscale or Tor. Does what I've described make sense as a possible solution?

tiero:
Thanks @nevets963 this was the right call!

We added this feature to read from `window.location` in packaged mode, which will be soon available in the next version of the dashboard. 

I made a preparatory commit, adding the caddy reverse proxy, please let me know if it's what you intended (I tested locally the solution and works great)

nevets963 (approved):
@tiero I've patched up a couple of things and putting this all together is working really nicely now! I've managed to glide through the setup process, create a wallet and create a market. I've tested this over Tor too - that is working. I've also tested on `arm64` and `amd64`.

This is approved from me! 🚀 

@mayankchhabra Could you help prepare the app store graphics please?

tdex/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+      TDEX_LOG_LEVEL: 5
+      TDEX_EXPLORER_ENDPOINT: https://liquid.network/liquid/api
+      TDEX_OPERATOR_LISTENING_PORT: 9090
+      TDEX_TRADE_LISTENING_PORT:  9090
+      TDEX_NO_OPERATOR_TLS: "true"
+      # these vars are used to generate the connect URL
+      # we assume to always show the external endpoint
+      # reachable remotely by other dashboards via Tor
+      TDEX_CONNECT_ADDR: "$APP_HIDDEN_SERVICE"
+      TDEX_CONNECT_PROTO: http
```
nevets963, line 26: @tiero As the dashboard now automatically detects the correct connection url. Potentially these 2 variables are no longer needed?

tiero: These variables are needed for the daemon to also generate a html page used to retrieve the TDEX CONNECT URL  (it should be visibile navigating to `/api/` with the browser) 

Useful if users wants to connect from outside using the tdex dashboard as dekstop app.

tiero:
> And then inside of exports.sh read the Tor Onion hostname:

This may be the last missing thing: I know we now proxy daemon on /api, so one hidden service may suffice, but need to be sure TDEX trader's clients can connect to provider's URL that contains a path in the endpoint.

Otherwise,would each tor onion endpoint be visible on the Umbrel app page somehow or I need to feed it back in the app and show it there?

tdex/docker-compose.yml
tiero: > we can grab the hostname for the daemon specifically and feed that in?

What do you mean here? how you can "introspect" the hostname, with a bash script or something?

tiero: > Tor Hidden Service url for the daemon specifically. ie xyz.onion:9090 

How could the user know about the two difference onion services that are running? ie. one to use to navigate to the dashaboard and the other to give to traders?

tiero: I know Tor is terrible, but it's the only way for TDEX Daemon operators to expose the public trade API and allow TDEX App and other clients to do NAT traversal and reach the daemon running on their Umbrel.

An alternative would be to have services such as ngrok?

nevets963: Yh, unfortunately at this time it's a little tricky in terms of remote access. Potentially Tailscale is another option (that we have in the app store), but that also has a few challenges.

tiero: I bumped the dashboard with this new ENV VAR to feed the hidden service.

I created APP_TDEX_TRADER_HIDDEN_SERVICE, hope this works.

Can you help me to create the exports.sh file?

tiero:
I bumped the dashboard with this new ENV VAR to feed the hidden service.  https://github.com/getumbrel/umbrel-apps/pull/151/commits/30e3c016c9af92f646a2d96afabea628725dc20c

I created APP_TDEX_TRADER_HIDDEN_SERVICE, hope this works.

Can you help me to create the exports.sh file?

nevets963:
Ok nice! I created and tested the `exports.sh` file: https://github.com/getumbrel/umbrel-apps/pull/151/files#diff-57845f7651a52c8b2e41e0d60018bdb1f6f8386b5329d4b764c7e75b89d2f6aeR1-R2 😄

tiero:
@nevets963 amazing! Will test now on a custom umbrel install on a Linux VM

tiero:
ah last thing: what about whitelist/blacklist? Since we exposing the tor hidden service directly of the APP_TDEX_TRADER_HIDDEN_SERVICE, there is no need to whitelist /api via PROXY_AUTH_WHITELIST right?

The Umbrel password will be asked only if reached via browser or the **main** hidden service?

nevets963:
This depends on the setup you want to provide. If you 'advertise' the Tor Hidden Service url for the daemon in the dashboard, then when users access this over Tor, they will talk directly to the daemon. Given the url is a random string plus (as I understand it), the daemon has it's own auth (macaroon) system, then I suspect this ideal for your usecase.

However, instead, you could 'advertise' the hidden service for caddy which is protected by our App Proxy and have users access this via Tor (which is a different Tor Hidden Service url) and go to `/api`. From what I see, this doesn't seem ideal, and what I said above seems more aligned to your/your user's usecase?

tiero:
> the daemon has it's own auth (macaroon) system, then I suspect this ideal for your usecase.

Yes exactly we do not need the daemon to be authenticated on top, since operator APIs have macaroon auth and the trader APIs (public) must be accessed publically.

nevets963:
Makes sense to me!

tiero:
Installed umbrel on a VM, I followed the guide to install the app using `~/umbrel/scripts` but then I cant launch the app, what should be the right port?


image

tiero:
@nevets963 we discovered a bug in out implementation, so bear with us until we fix the dashboard.

tiero:
We fixed the latest bug, so now the dashboard should be able to  connect with daemon nicely.

I tested in a Linux VM instaling the Umbrel there etc.. all works nice.

One thing: the injected `APP_TDEX_DAEMON_HIDDEN_SERVICE` variable seems to return an onion address `xxx.onion`, but apparently seems not to be reachable via Tor Browser.

(connecting via clearnet to the IP of my VM, with the given Umbrel cookie, works just fine, so daemon is running correctly I assume)

What I could check? 

I can also give you access to the VM if you may want to check this out live.

nevets963:
@tiero Hmm, potentially you're missing the port number from the .onion url which I set to `9090` (https://github.com/getumbrel/umbrel-apps/pull/151/files#diff-a100f2f28349dfdbcda3df3dafa8388677ca7bf17d677c939b6202128e6bb863R3). It's working OK here, where it's asking me to enter a password. You could change this so that the .onion is listening on port 80 (standard HTTP) and then forwards this to the daemon (listening on `9090`). Potentially user's expect to use port `9090` - not sure; feel free to change this.

Let me know if the above was the problem.

tiero:
thanks @nevets963 that was it 🎉 

I decided to listen on port 80, mostly because most TDEX liquidity providers in tdex network currently expose on port 80, introducing anything different in Umbrel will clearly mark them as Umbrel users.

Please check again, but I am good on my side.

I will try to test on an actual Raspberry today, to make the final test, and then we are good on my side.
Will send screenshot as well

tiero:
Tested on Rpi and uploaded screenshots in the PR description! 

We just want to make the latest review and changes to copy in the dashboard, then we are good for merging!

nevets963:
@tiero Just a friendly ping to see how things are getting on from your side?

tiero:
> @tiero Just a friendly ping to see how things are getting on from your side?

Apologies for the late response! We are tweaking the last things in terms of copy and layout, we plan to push it beginning of January.

tiero:
Hey guys! We are ready to publish the app in the store! Thank you for waiting

tdex/docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8080
   
   tdexd:
-    image: ghcr.io/tdex-network/tdexd:v0.8.19@sha256:5a2c0db9f74812e67aa98f50f396e81edd23ef49cb6797b084027ba6124db707
+    image: ghcr.io/tdex-network/tdexd:v0.0.9@sha256:77729e6c0019752059cc5bf624675e261abd8b856e0a7e8f59889b7ec0a35eb3
```
nevets963, line 10: @tiero Is there a typo here as I can't pull down `v0.0.9`, but I see `v0.9.0` available?

tiero: Ups, yes typo before commit

nevets963 (commented):
@tiero Looks like we're now really close! Due to some changes in Umbrel related to Tor, we need to make a few small changes which I've detailed below. I did just test Tdex in it's current form and it appears to be workly nicely (outside of Tor access to the daemon).

Let me know if I can help if with anything.

tdex/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: tdex_caddy_1
+      APP_PORT: 8080
+  
+  tdexd:
+    image: ghcr.io/tdex-network/tdexd:v0.9.0@sha256:77729e6c0019752059cc5bf624675e261abd8b856e0a7e8f59889b7ec0a35eb3
+    tty: true
```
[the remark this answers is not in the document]
tiero: We have a CLI embedded which has some interactive commands. I do think is necessary yes

tdex/torrc.template
```diff
@@ -0,0 +1,7 @@
+# TDEX Public Trader Interface Hidden Service
+HiddenServiceDir /data/app-$APP_ID-daemon
+HiddenServicePort 80 tdex_tdexd_1:9090
```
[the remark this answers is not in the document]
tiero: did this, not clear to me how to invoke these hooks? or is umbrel handling on top?

tdex/torrc.template
```diff
@@ -0,0 +1,7 @@
...
+
+# Dashboard Frontend Hidden Service
+HiddenServiceDir /data/app-$APP_ID
+HiddenServicePort 80 $APP_PROXY_HOSTNAME:$APP_PROXY_PORT
```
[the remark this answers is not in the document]
tiero: How to modify the file the?

tdex/torrc.template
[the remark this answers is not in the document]
tiero: @nevets963 It seems to me you have write access to the fork already https://github.com/tdex-network/umbrel-apps/tree/tdex

nmfretz:
Hey @tiero I'm taking over for Steve and can help you with the final stages of getting Tdex Provider into the App Store.

Taking a look now.

tdex/torrc.template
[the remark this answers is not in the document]
nmfretz: The only thing you have to do here now is delete the `Dashboard Frontend Hidden Service` code, since it is no longer needed.

tdex/torrc.template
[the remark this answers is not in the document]
nmfretz: All you need to do is edit the pre-start hook file appropriately for you app. There's nothing you need to do on your end to invoke these hooks. Umbrel will automatically run this pre-start hook when the app is started. See here for example:
https://github.com/getumbrel/umbrel/blob/d7c9f4d16317d52eef2193458935dc673790f4db/scripts/app#L362-L380

tdex/hooks/pre-start
```diff
@@ -0,0 +1,26 @@
...
+	if [[ -f "${HIDDEN_SERVICE_FILE}" ]]; then
+		echo "App: ${APP_ID} - Hidden service file created successfully!"
+		break
+	fi
+	sleep 0.1
+done
+
+if [[ ! -f "${HIDDEN_SERVICE_FILE}" ]]; then
+	echo "App: ${APP_ID} - Hidden service file wasn't created"
+fi
```
nmfretz, line 26: This file just needs to be edited slightly so that it refers to your app components and not the Lightning Node app. Right now this file is using the grpc container for the hidden service file and then booting up the lnd container (both containers are part of the Lightning Node app).

```suggestion
#!/usr/bin/env bash

# Delay booting TDEX until the Tor Hidden Service is ready

HIDDEN_SERVICE_FILE="${TOR_DATA_DIR}/app-${APP_ID}-daemon/hostname"

if [[ -f "${HIDDEN_SERVICE_FILE}" ]]; then
	exit
fi

"${UMBREL_ROOT}/scripts/app" compose "${APP_ID}" up --detach tdexd
"${UMBREL_ROOT}/scripts/app" compose "${APP_ID}" up --detach tor

echo "App: ${APP_ID} - Generating Tor Hidden Service..."

for attempt in $(seq 1 100); do
	if [[ -f "${HIDDEN_SERVICE_FILE}" ]]; then
		echo "App: ${APP_ID} - Hidden service file created successfully!"
		break
	fi
	sleep 0.1
done

if [[ ! -f "${HIDDEN_SERVICE_FILE}" ]]; then
	echo "App: ${APP_ID} - Hidden service file wasn't created"
fi
```

nmfretz:
Alright, looks like some small changes related to Tor are needed (see my review) and then I can do some final testing.

tiero:
Is anything missing from this submission? @nmfretz

nmfretz:
Sorry for the delay @tiero. Testing this now.

nmfretz:
Hey Marco, I am having some issues when first installing the app where in Advanced Settings, the `Tor Hidden Service of Trader Interface` field shows up as `notyetset.onion` indicating that whatever env var is being accessed to display the onion address was passed into a docker container before the hidden service was created. A restart of the app after installation fixes this and an onion address shows in Advanced Settings.

To help fix this can you please walk me through how the dashboard is displaying this onion address? Looking at the `docker-compose.yml` I see that you have `$APP_TDEX_DAEMON_HIDDEN_SERVICE` being passed as an env var to both `tdexd` (as TDEX_CONNECT_ADDR) and `dashboard` (as TRADER_HIDDEN_SERVICE). Is it using one of these?

The way the `pre-start` hook works right now is that the dashboard container is not started until the hidden service file exists. This means `TRADER_HIDDEN_SERVICE` will be set correctly. However, the variable `TDEX_CONNECT_ADDR` in the `tdexd` container will be set to `notyetset.onion` on install.

tiero:
Hey @nmfretz  we should have fixed it. We tested it on an online machine with Umbrel installed on it.

Can you double-check if everything checks out now?

nmfretz:
Thanks @tiero! I will look at this very soon. Here is how your app logo and gallery images are looking. Can you please let me know:
1) What text you would like to see on each of the images
2) If you would like any changes to be made to the logo or gallery images

TDEX_Provider

tiero:
Will get back to you very soon

tiero:
These are the copy

1. Provide liquidity for multiple TDEX market

2. Non-custodial - participate in markets and withdraw with your own keys 

3. Detailed insights for each market

nmfretz:
Thanks @tiero, here are the images and icon:







icon

tiero:
LFG 🚀

tdex/umbrel-app.yml
```diff
@@ -0,0 +1,28 @@
...
+version: "0.8.15"
+tagline: Provide liquidity on the TDEX Network, an open protocol to trade Liquid Network assets.
+description: >-
+  A Provider holds Liquid reserves of both a BASE_ASSET-QUOTE_ASSET in his non-custodial Liquid hot-wallet, running automated market-making strategies, either with or without an oracle. Providers are incentivized to be always on and need to expose a public reachable endpoint either via clearnet or using a Onion hidden service.
+  A small provider's fee can be taken out of each trade and added to the reserves. 
+  A liquidity provider has full control over the market making strategy to apply needed to calculate the market rate at which to accept trades. That being said, there is a possibility to apply an automated market-making relying only on the reserves balances and the amount requested by the trader, without the need to connect to an external price feed. The default strategy of the alpha daemond is the constant product market-making. In short, this model generates a full order-book based on an initial price for the market. Every transaction that occurs on this market will adjust the prices of the market accordingly.
+releaseNotes: >-
```
nmfretz, line 11: We can remove the entire releaseNotes section for the initial app release. This gets used once we start doing app updates.

tdex/exports.sh
```diff
@@ -0,0 +1,7 @@
+export APP_TDEX_PORT="9090"
+
+daemon_hidden_service_file="${EXPORTS_TOR_DATA_DIR}/app-${EXPORTS_APP_ID}-daemon/hostname"
+export APP_TDEX_DAEMON_HIDDEN_SERVICE="$(cat "${daemon_hidden_service_file}" 2>/dev/null || echo "daemon_not_yet_set.onion")"
+
+# Necessary for the tdexd daemon to be able to write to the volume
+sudo chmod -R 777 "${UMBREL_ROOT}/app-data/tdex/tdex-data"
```
nmfretz, line 7: Did you want to give full read/write/executable permissions to user, group, and others here? 
or does giving read/write/execute to user and read/execute to group and others work for you... for example, by running:
`sudo chmod -R 755 "${UMBREL_ROOT}/app-data/tdex/tdex-data"`

nmfretz (commented):
I've tested and it's working well @tiero! I checked and the `${APP_TDEX_DAEMON_HIDDEN_SERVICE}` env var is being set correctly inside the `tdexd` and `dashboard` containers.

I have left a couple small comments below. Let's get this shipped 🚀!

tdex/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+  dashboard:
+    image: ghcr.io/tdex-network/dashboard:v0.1.55@sha256:dfa7cdb75117c97e671285fc9754a6cabafc7f74760c314e6625c33067509beb
+    restart: on-failure
+    environment:
+      USE_PROXY: "false"
+      IS_PACKAGED: "true"
+      TRADER_HIDDEN_SERVICE:  "${APP_TDEX_DAEMON_HIDDEN_SERVICE}"
+
+  caddy:
+    image: caddy:2.6.4-alpine@sha256:5acae6b87b26a591d6d14ec2704b7f2cd94b888ad62af16a02356d3124198ede
```
nmfretz, line 36: This needs to be the multi-arch digest. I think this is the one:
`sha256:4dfec6c3b22c36b63ea4a3633c7cdbdaa9926d1324c27db2b0e2b70ef9cd105a`

tdex/exports.sh
nmfretz: I'm okay with this going out as-is and then we can consider handling this another way in the future to avoid giving such relaxed permissions to your app's data directory.

I quickly looked at the [dockerfile for tdex](https://github.com/tdex-network/tdex-daemon/blob/master/Dockerfile) and noticed that the user and group are set to `tdex`. One way you might be able to solve this issue is by having tdex run as user/group 1000.

nmfretz:
@tiero I pushed some changes:
- added correct multi-arch digest for caddy
- removed the release notes, which you can add back in the first time you update the app
- added some whitespace to the app description so that is shows up nicely in the App Store
- updated the submitter field to show that you are a TDEX developer. We're trying to make it obvious in the app store that an app has been packaged by the developer of the app, vs a community member, vs Umbrel.

I tested on amd64 and arm64 from an umbrel set-up standpoint. The app launches properly and has Tor config and env vars are set correctly.  We are good to launch as soon as you give me the word! 🚀

For future updates you may want to look into running tdexd as user 1000 to avoid having to `chmod` the data directory on the host https://github.com/getumbrel/umbrel-apps/pull/151#discussion_r1188091047

Another thing to consider for future updates is you could change the app description to something super simple and general and link to more details on [your site](https://tdex.network/) As the app developer, this is completely up to you!

nmfretz:
Merging now based on Telegram convo.

Thanks again for all your hard work bringing TDEX to Umbrel @tiero!

PR #156, opened 2022-09-20
Update HA to Version 2022.9.5

ertl:


nevets963 (approved):
@ertl Thank you for pushing the HA app update!

Tested on `amd64` and `arm64`. Approved! 🚀

PR #161, opened 2022-09-26
Update robosats to v0.2.0-alpha

aphex3k:
This is the first alpha available for robosats after the MVP version.


nevets963 (approved):
@aphex3k Thanks for another app update!

Tested on `arm64` and `amd64`. Approved!  🚀 

Merging now.

PR #166, opened 2022-09-28
Update Pi-hole to 2022.09.4

mccolister1017:
9 days old update .


nevets963 (commented):
@mccolister1017 Many thanks again for the Pi-Hole app update! Would you be able to give me write access or create these PRs on separate branches; that way I can make some small modifications. I think the release notes here are a little verbose.

PR #168, opened 2022-09-28
skip node auth for whoogle searches

gustavobrossi:
IMO doesn't make sense to have to authenticate into my node to run searches


gustavobrossi:
I've checked the history and that was discussed before, and it was decided to leave it with auth enabled

PR #171, opened 2022-09-28
Update Pi-hole to 2022.09.4 

mccolister1017:
Release notes https://github.com/pi-hole/docker-pi-hole/releases/tag/2022.09.4


nevets963 (commented):
Hey @mccolister1017 It looks like I still don't have write access for me to be able to add the release notes. I suspect it's because this PR is coming from your `master` branch. If you create a branch or grant me write access then I should be able to do this.

PR #173, opened 2022-09-29
Update Pi-hole to 2022.09.4

mccolister1017:
Release notes https://github.com/pi-hole/docker-pi-hole/releases/tag/2022.09.4

@nevets963  hopefully this is good and i think you can edits it .


nevets963 (approved):
@mccolister1017 Perfect! I was able to add some release notes. In this case, it looks like mainly internal improvements from the Pi-hole project.

Tested on `arm64` and `amd64`. Approved! 🚀 

Thanks again! Merging now :)

PR #176, opened 2022-09-29
RTL Update on CLN

ShahanaFarooqui:
RTL Update on CLN

### DIFF
--- a/core-lightning-rtl/docker-compose.yml
+++ b/core-lightning-rtl/docker-compose.yml
@@ -7,12 +7,12 @@ services:
       APP_PORT: 3000
 
   rtl:
-    image: shahanafarooqui/rtl:0.12.3@sha256:b1327ecfddae5f40c4c985e31c2bc158c4697a274b61bf84acfb27486d05c5bf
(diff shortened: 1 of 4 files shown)


nevets963 (commented):
@ShahanaFarooqui Just before I review this, we can perhaps also update the release notes here. I looked at the changelog for 0.13.0 and I see specific updates for Core Lightning. So perhaps here, we can use this text for the release notes? However, open to suggestions :)

```
- Liquidity Ads interface for viewing ads, researching nodes and buying channels
- Liquidity Ads configuration to publish ads on the network
- Regtest support
- Dashboard performance improvements
```

ShahanaFarooqui:
Hi, thank you for release notes suggestions. I think they covered all major features. We can move forward with these.

nevets963 (commented):
@ShahanaFarooqui I just noticed the update for Core Lightning itself. Is v0.9.0 of `c-lightning-rest` a hard requirement to work with RTL v0.13?

ShahanaFarooqui:
0.9.0 has liquidity ads APIs. So RTL’s liquidity ads feature will not work without this update but it will not break anything else.

PR #186, opened 2022-10-07
Update Jam to v0.1.1

theborakompanioni:
This version includes:
- Jam: [v0.1.1](https://github.com/joinmarket-webui/jam/releases/tag/v0.1.1)
- joinmarket-clientserver: [v0.9.8](https://github.com/JoinMarket-Org/joinmarket-clientserver/releases/tag/v0.9.8)

All notable changes of the UI can be seen in the [Jam v0.1.1 changelog](https://github.com/joinmarket-webui/jam/blob/v0.1.1/CHANGELOG.md)

Updates to the image:
- feat: randomize mandatory fee config values [#66](https://github.com/joinmarket-webui/jam-docker/pull/66)

### DIFF
--- a/jam/docker-compose.yml
+++ b/jam/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 80
 
   jam:
-    image: ghcr.io/joinmarket-webui/jam-standalone:v0.1.0-clientserver-v0.9.8@sha256:4067e86e97b4342ee27021b31de8b31394d640e066f850acbcf4fc9e72c56302
+    image: ghcr.io/joinmarket-webui/jam-standalone:v0.1.1-clientserver-v0.9.8@sha256:0b17e193c12b44323e8a815994457f69e0e8ee1299702e6fca7e8547352eafea
     restart: on-failure
     stop_grace_period: 1m
     init: true
--- a/jam/docker-compose.yml
+++ b/jam/docker-compose.yml
@@ -25,8 +25,6 @@ services:
       jm_rpc_user: $APP_BITCOIN_RPC_USER
       jm_rpc_password: "${APP_BITCOIN_RPC_PASS}"
       jm_rpc_wallet_file: jam_default
-      jm_max_cj_fee_abs: 300000 # in sats
-      jm_max_cj_fee_rel: 0.0003 # 0.03%
     networks:
       default:
         ipv4_address: $APP_JAM_IP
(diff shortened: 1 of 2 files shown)


nevets963 (commented):
@theborakompanioni Thanks for Jam app update! I gave it a test on my `arm64` and `amd64` Umbrel servers and they seem to run OK.

Except for `amd64` where I run `regtest` and I now get this new error message:
```
Unhandled Thrown Error!
404 Not Found

💿 Hey developer 👋

You can provide a way better UX than this when your app throws errors by providing your own errorElement props on <Route>
```

If you think this is nothing serious then I'm happy to approve this.

theborakompanioni:
> @theborakompanioni Thanks for Jam app update! I gave it a test on my `arm64` and `amd64` Umbrel servers and they seem to run OK.
> 
> Except for `amd64` where I run `regtest` and I now get this new error message:
> 
> ```
> Unhandled Thrown Error!
> 404 Not Found
> 
> 💿 Hey developer 👋
> 
> You can provide a way better UX than this when your app throws errors by providing your own errorElement props on <Route>
> ```
> 
> If you think this is nothing serious then I'm happy to approve this.

Thanks you so much for the heads-up. This should definitely not happen. Would you possibly be so kind as to try again after waiting a few seconds for the backend to come up and with browser cache cleared?

Got this too during development (https://github.com/joinmarket-webui/jam/pull/519#discussion_r985961844) but only for brief moments, subsequently redirecting to the index page. Looking at the code, it seems this might happen if the backend is down (or still booting - there is a small delay), so from this point of view I suggest it is safe to merge.
However, this is not a nice experience and will be addressed in the next release.

nevets963:
I just tried again after having it up for >12 hours and cleared my browser cache but I still get the error. I'm not sure how to interpret this:
> it is safe to merge only if the backend comes up normally without any errors on amd64

Does that mean you want to patch this case up? Or because it's likely only an issue on `regtest`, then we're good to merge?

theborakompanioni:
> Does that mean you want to patch this case up? Or because it's likely only an issue on `regtest`, then we're good to merge?

After chatting, the issue has been identified (the [infamous descriptor wallet bug](https://github.com/joinmarket-webui/jam/issues/502)). 
Next release will have a proper error message, but this PR is safe to merge!

Thank you for your help and your patience @nevets963 :orange_heart:

PR #187, opened 2022-10-07
Update Lnvisualizer to 0.0.26

MaxKotlan:
A few visible changes:
  - new channel color mapping modes. can map channel color by fee-rate, base-fee, cltv-delta, max-htlc, min-htlc
  - new color range ui menu

Also better container security. Only include cert and readonly macaroon in container.


nevets963 (approved):
@MaxKotlan Many thanks for the app update! I just tested on `arm64` and `amd64`.

Merging now 🚀

PR #190, opened 2022-10-10
Update lnd to 0.15.2-beta

MaxKotlan:
Fixes bug that causes lnd to crash on startup.

PR #192, opened 2022-10-10
Bugfix for v0.15.1-beta

aphex3k:


aphex3k:
duplicating #190 and #191

PR #199, opened 2022-10-15
App Submission: Trilium Notes

IMPranshu:
# App Submission

### Trilium Notes
...

### 256x256 SVG icon
https://svgshare.com/i/nPJ.svg
...

### Gallery images
trilium-note2
trilium-note5
trilium-note4
trilium-note3
trilium-note1

...

### I have tested my app on:

### DIFF
--- a/trilium-note/docker-compose.yml
+++ b/trilium-note/docker-compose.yml
@@ -11,9 +11,6 @@ services:
     restart: on-failure
     hostname: "${DEVICE_HOSTNAME}"
     environment:
-      - PUID=1000
-      - PGID=1000
-      - VERSION=docker
       - TRILIUM_DATA_DIR=/data
     volumes:
       - ${APP_DATA_DIR}/data:/data
--- a/trilium-note/docker-compose.yml
+++ b/trilium-note/docker-compose.yml
@@ -1,6 +1,11 @@
 version: "3.7"
 
 services:
+  app_proxy:
+    environment:
+      APP_HOST: trilium-note_server_1
+      APP_PORT: 8080
+
   server:
     image: zadam/trilium:0.55.1@sha256:9bf48a49359cd567bb8e9ce87c1d310713ba98ac8d517d75b07b8de907525077
     restart: on-failure
@@ -10,8 +15,5 @@ services:
       - PGID=1000
       - VERSION=docker
       - TRILIUM_DATA_DIR=/data
-    ports:
-      - 8080:8080
     volumes:
       - ${APP_DATA_DIR}/data:/data
-
--- a/trilium-note/docker-compose.yml
+++ b/trilium-note/docker-compose.yml
@@ -9,9 +9,9 @@ services:
       - PUID=1000
       - PGID=1000
       - VERSION=docker
-      - TRILIUM_DATA_DIR=/home/node/trilium-data
+      - TRILIUM_DATA_DIR=/data
     ports:
       - 8080:8080
     volumes:
-      - ${UMBREL_ROOT}/data/storage/trilium:/home/node/trilium-data
+      - ${APP_DATA_DIR}/data:/data
 
--- a/trilium-note/docker-compose.yml
+++ b/trilium-note/docker-compose.yml
@@ -1,0 +1,17 @@
+version: "3.7"
+
+services:
+  server:
+    image: zadam/trilium:0.55.1@sha256:9bf48a49359cd567bb8e9ce87c1d310713ba98ac8d517d75b07b8de907525077
+    restart: on-failure
+    hostname: "${DEVICE_HOSTNAME}"
+    environment:
+      - PUID=1000
+      - PGID=1000
+      - VERSION=docker
+      - TRILIUM_DATA_DIR=/home/node/trilium-data
+    ports:
+      - 8080:8080
+    volumes:
+      - ${UMBREL_ROOT}/data/storage/trilium:/home/node/trilium-data
+
--- a/trilium-note/umbrel-app.yml
+++ b/trilium-note/umbrel-app.yml
@@ -29,7 +29,7 @@ website: https://github.com/zadam/trilium
 dependencies: []
 repo: https://github.com/zadam/trilium
 support: https://github.com/zadam/trilium
-port: 8080
+port: 3779
 gallery:
   - 1.jpg
   - 2.jpg
--- a/trilium-note/umbrel-app.yml
+++ b/trilium-note/umbrel-app.yml
@@ -1,0 +1,40 @@
+manifestVersion: 1
+id: trilium-note
+category: Productivity
+name: Trilium Note
+version: "0.55.1"
+tagline: Build your personal knowledge base with Trilium Notes
+description: >-
+  Notes can be arranged into arbitrarily deep tree. Single note can be placed into multiple places in the tree (see cloning)
+  Rich WYSIWYG note editing including e.g. tables, images and math with markdown autoformat
+  Support for editing notes with source code, including syntax highlighting
+  Fast and easy navigation between notes, full text search and note hoisting
+  Seamless note versioning
+  Note attributes can be used for note organization, querying and advanced scripting
+  Synchronization with self-hosted sync server
+  there's a 3rd party service for hosting synchronisation server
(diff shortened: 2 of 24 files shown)


trilium-note/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: "3.7"
+
+services:
+  server:
+    image: zadam/trilium:latest
```
nevets963, line 5: Here we'll need to pin the image to a particular tag along with its digest.

trilium-note/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: "3.7"
+
+services:
+  server:
```
nevets963, line 4: Is there a reason you've not used/included the app proxy here?

trilium-note/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+    environment:
+      - PUID=1000
+      - PGID=1000
+      - VERSION=docker
+      - TRILIUM_DATA_DIR=/home/node/trilium-data
+    ports:
+      - 8080:8080
+    volumes:
+      - ${UMBREL_ROOT}/data/storage/trilium:/home/node/trilium-data
+    networks:
```
nevets963, line 17: We no longer require apps to define container IPs, so this section can be dropped.

trilium-note/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+    hostname: "${DEVICE_HOSTNAME}"
+    environment:
+      - PUID=1000
+      - PGID=1000
+      - VERSION=docker
+      - TRILIUM_DATA_DIR=/home/node/trilium-data
+    ports:
+      - 8080:8080
+    volumes:
+      - ${UMBREL_ROOT}/data/storage/trilium:/home/node/trilium-data
```
nevets963, line 16: Is there some reason to use Umbrel's shared storage folder as opposed to having the storage be part of the app install? For example: https://github.com/getumbrel/umbrel-apps/blob/8b7c1ce80ba516bea973434c77adddc6c94d2cc6/code-server/docker-compose.yml#L14

trilium-note/umbrel-app.yml
```diff
@@ -0,0 +1,42 @@
...
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
+path: /#root
+defaultUsername: ""
+defaultPassword: ""
+torOnly: false
+permissions:
+  - STORAGE_DOWNLOADS
```
nevets963, line 42: If shared storage is not required, then this should be dropped.

nevets963 (commented):
@IMPranshu Many thanks for the Trilium app submission!

Before I go deeper, I've left some feedback and questions.

trilium-note/docker-compose.yml
IMPranshu: Yes. The main reason the app was breaking was the app_proxy tag. As the is being used as a server it is now running fine. Been stuck on it for weeks then saw the Plex app is facing the same issue on Umbrel and I took the reference from them.

trilium-note/docker-compose.yml
IMPranshu: But I tested it quite a few times and the app was going into a different location sometimes when I removed this. So I tried to put it here as it was then working fine almost all of the times.

trilium-note/docker-compose.yml
IMPranshu: Yes. Upon stopping and restarting the app the app data was getting removed and also all the data with it. This method fixed that issue.

trilium-note/umbrel-app.yml
IMPranshu: But shared storage helps in multi-device support which makes it unique and is the Notion-like feature which makes it stand out. I would love to keep this here.

trilium-note/docker-compose.yml
IMPranshu: Yup will do that here.

trilium-note/docker-compose.yml
nevets963: Stopping/restarting an app won't remove the app's data. It only happens if you uninstall the app. To me, this looks like it should be using the location defined using `$APP_DATA_DIR` which is the app install directory.

trilium-note/umbrel-app.yml
nevets963: What do you mean by multi-device support?

The variable `STORAGE_DOWNLOADS` isn't actually implemented/handled by Umbrel (yet), so this doesn't actually do anything atm.

trilium-note/docker-compose.yml
nevets963: Can you elaborate a little more on what was broken?

trilium-note/docker-compose.yml
IMPranshu: Ok. Will try and see maybe it I was doing something in a worng way.

trilium-note/docker-compose.yml
IMPranshu: YES. I can share screenshots of the error.

image

this was the compose file
```
version: "3.7"

services:
  app_proxy:
    environment:
      APP_HOST: $APP_TRILIUM_NOTE_IP
      APP_PORT: 8080

  server:
    image: zadam/trilium:latest
    restart: on-failure
    user: "1000:1000"
    volumes:
      - ${APP_DATA_DIR}/home/node/trilium-data
    networks:
      default:
        ipv4_address: $APP_TRILIUM_NOTE_IP
        ```

trilium-note/docker-compose.yml
nevets963: I'm not sure what you're referring re 'going into a different location'. What I mean is; this `networks` block can be removed because we longer need to set container IPs and this simplifies deploying apps on Umbrel. You'll see many apps are still doing this, but this was before we stopped adding this section.

IMPranshu: Ok cool. I followed the default template from the umbrel docs that's why it's there. I will remove and push the changes. Thanks

IMPranshu: I have made the necessary changes and pushed the commit.

trilium-note/docker-compose.yml
IMPranshu: I have double-checked it now. The volumes part has to remain that only because if that is in the app directory after stopping and starting umbrel the app data is getting reset which is a big issue. Keeping it in Umbrel Root storage keeps the user data safe and reliable for long-term use. I don't know if this issue is for this particular app of something else but that is the case for now.

trilium-note/docker-compose.yml
IMPranshu: Plex has the same issue I have read while going through their repo. Maybe it's the same issue here

nevets963: The error you've shared looks like an issue related to yaml formatting. I've just pushed a commit to add the app proxy and it appears to be working fine.

trilium-note/docker-compose.yml
nevets963: I am unable to replicate this. I've set the volume to: `- ${APP_DATA_DIR}/data:/data` and it appears to be working as expected. I can restart the app and all the data is persisted within the app's install directory.

nevets963 (approved):
Hey @IMPranshu I've managed to patch up a few things in this PR and I think we're now ready to launch! 🚀 

@mayankchhabra Could you prepare the app assets for Trilium Notes please?

nevets963 (commented):
@IMPranshu Thanks for working on another app submission. Could you open a separate PR for it so we don't pollute this PR with another app?

IMPranshu:
Oo sorry. I was just testing out. I didn't realised. Will make a different branch for this

nevets963:
The graphics for the App Store are now available here: https://github.com/getumbrel/umbrel-apps-gallery/pull/20

I will do 1 last test and this merge this 🚀

IMPranshu:
Thanks, @nevets963.

IMPranshu:
Thanks, @nevets963.

PR #200, opened 2022-10-16
Update Tailscale  to v1.33.19

mccolister1017:


nevets963 (approved):
@mccolister1017 Thanks again for the app update!

Merging now! 🚀

PR #217, opened 2022-10-25
App Submission: Kdenlive

IMPranshu:
# App Submission

### KdenLive
...

### 256x256 SVG icon
https://svgshare.com/i/ndj.svg
...

### Gallery images
kde5
kde4
kde3
kde2
kde1

...

### I have tested my app on:

### DIFF
--- a/kdenlive/docker-compose.yml
+++ b/kdenlive/docker-compose.yml
@@ -5,6 +5,7 @@ services:
     environment:
       APP_HOST: kdenlive_server_1
       APP_PORT: 3000
+  
   server:
     image: linuxserver/kdenlive:421.12.3@sha256:c9fd0cb367e32f74259eb3293674da8f6107a8e6d6b5136965001d5c8e0f486b
     restart: on-failure
--- a/kdenlive/docker-compose.yml
+++ b/kdenlive/docker-compose.yml
@@ -1,6 +1,10 @@
 version: "3.7"
 
 services:
+  app_proxy:
+    environment:
+      APP_HOST: kdenlive_server_1
+      APP_PORT: 3000
   server:
     image: linuxserver/kdenlive:421.12.3@sha256:c9fd0cb367e32f74259eb3293674da8f6107a8e6d6b5136965001d5c8e0f486b
     restart: on-failure
@@ -10,5 +14,3 @@ services:
       - SUBFOLDER=/
     volumes:
       - ${APP_DATA_DIR}/data/config:/config
-    ports:
-      - 3000:3000
--- a/kdenlive/docker-compose.yml
+++ b/kdenlive/docker-compose.yml
@@ -1,0 +1,14 @@
+version: "2.1"
+
+services:
+  server:
+    image: linuxserver/kdenlive:421.12.3
+    restart: on-failure
+    environment:
+      - PUID=1000
+      - PGID=1000
+      - SUBFOLDER=/
+    volumes:
+      - ${APP_DATA_DIR}/data/config:/config
+    ports:
+      - 3000:3000
--- a/kdenlive/docker-compose.yml
+++ b/kdenlive/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   server:
-    image: linuxserver/kdenlive:421.12.3
+    image: linuxserver/kdenlive:421.12.3@sha256:c9fd0cb367e32f74259eb3293674da8f6107a8e6d6b5136965001d5c8e0f486b
     restart: on-failure
     environment:
       - PUID=1000
--- a/kdenlive/docker-compose.yml
+++ b/kdenlive/docker-compose.yml
@@ -1,4 +1,4 @@
-version: "2.1"
+version: "3.7"
 
 services:
   server:
--- a/kdenlive/umbrel-app.yml
+++ b/kdenlive/umbrel-app.yml
@@ -1,0 +1,33 @@
+manifestVersion: 1
+id: kdenlive
+category: Media
+name: KdenLive
+version: "421.12.3"
+tagline: Free and open source video editor.
+description: >-
+  Kdenlive is an open source video editor. The project was started around 2003. Kdenlive is built on Qt and the KDE Frameworks libraries. Most of the video processing is done by the MLT Framework, which relies on many other open source projects like FFmpeg, frei0r, movit, ladspa, sox, etc…
+
+  Our software was designed to answer most needs, from basic video editing to professional work.
+
+  Kdenlive is developed by a small team and new contributors are welcome.
+
+  Multi-track video editing - Kdenlive allows you to use and arrange several audio and video tracks, each one can be locked or muted to your convenience.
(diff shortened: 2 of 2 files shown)


kdenlive/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: "2.1"
+
+services:
+  server:
+    image: linuxserver/kdenlive:421.12.3
```
nevets963, line 5: Could you add a multi-arch Docker image digest here? You can get this by running: `docker pull linuxserver/kdenlive:421.12.3`

kdenlive/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: "2.1"
+
+services:
+  server:
```
nevets963, line 4: Could you add the Umbrel App Proxy docker service here?

kdenlive/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: "2.1"
```
nevets963, line 1: For all of our apps, we've tried to standardise on version `3.7`, could you change this please?

kdenlive/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+
+services:
+  server:
+    image: linuxserver/kdenlive:421.12.3
+    restart: on-failure
+    environment:
+      - PUID=1000
+      - PGID=1000
+      - SUBFOLDER=/
```
nevets963, line 10: I can't find the docs for this Docker image on `linuxserver.io` to understand the env. vars. Would you know where to find more info? I'm particularly interested in what the `SUBFOLDER` var. does.

kdenlive/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    volumes:
+      - ${APP_DATA_DIR}/data/config:/config
+    ports:
```
nevets963, line 13: Once the app proxy is added, we can remove this block

kdenlive/umbrel-app.yml
```diff
@@ -0,0 +1,33 @@
...
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+torOnly: false
+submitter: Pranshu Agrawal
+submission:
```
nevets963, line 33: Could you add the submission url, which will be the url of this PR?

kdenlive/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+services:
+  server:
+    image: linuxserver/kdenlive:421.12.3
+    restart: on-failure
+    environment:
+      - PUID=1000
+      - PGID=1000
+      - SUBFOLDER=/
+    volumes:
+      - ${APP_DATA_DIR}/data/config:/config
```
nevets963, line 12: Are both the app's config and video files stored inside of this `/config` location? Seems like they might be separate - but not sure without docs.

kdenlive/umbrel-app.yml
```diff
@@ -0,0 +1,33 @@
+manifestVersion: 1
+id: kdenlive
+category: Media
+name: KdenLive
+version: "421.12.3"
+tagline: Free and open source video editor.
+description: >-
```
nevets963, line 7: Due to the use of `>-`, a new line is achieved using 2 new lines. It looks like a few of the sections are missing newlines (spacing between each).

nevets963 (commented):
@IMPranshu Thanks again for pushing another app into the Umbrel App Store!

I've tested some feedback which I hope you can help with. After this, I'll then test this on my Umbrel servers.

Thanks!

kdenlive/docker-compose.yml
IMPranshu: That is something that can be used with "Reverse Proxy" which I am planning to test in the coming days.

kdenlive/docker-compose.yml
IMPranshu: done changes for this

kdenlive/docker-compose.yml
IMPranshu: done changes for this

kdenlive/umbrel-app.yml
IMPranshu: done changes for this

kdenlive/docker-compose.yml
IMPranshu: Using `app_proxy` didn't work for me. It was showing the port already existed at the time of making the PR. Will try again now.

kdenlive/umbrel-app.yml
IMPranshu: done changes for this

kdenlive/docker-compose.yml
IMPranshu: seems like my fault will push changes

kdenlive/docker-compose.yml
IMPranshu: Are you talking about the `ports` section?

kdenlive/docker-compose.yml
nevets963: Ok cool. Can you share the link to the docs?

kdenlive/docker-compose.yml
IMPranshu: pushed changes for this as well. Sorry. My bad

kdenlive/docker-compose.yml
nevets963: Yes; the `ports` section for the `server` service can be dropped once we use the app proxy.

IMPranshu: pushed changes for this

kdenlive/docker-compose.yml
nevets963: This could be because the app proxy and the `server` service were trying to open the same port.

kdenlive/docker-compose.yml
IMPranshu: https://hub.docker.com/r/linuxserver/kdenlive

kdenlive/docker-compose.yml
IMPranshu: Yeah. Right

kdenlive/docker-compose.yml
nevets963: Can you explain a bit more how `SUBFOLDER` is used with a reverse proxy?

IMPranshu: Yes. So when a reverse proxy is configured the `SUBFOLDER` tag specifies which of the folders can be accessed remotely i.e. limit the access to that folder and not others when accessing from outside the LAN. 

I have kept it here for now so that we can refer and update it later on. If you want we can comment that out as well, no issues. Won't be a problem in running the app.

nevets963:
@IMPranshu Thanks again for the app submission PRs! We are however seeing a lack of interest/need from the community and so I'll close many of these PRs. If we were to accept them, we then carry the burden of keeping them up-to-date and this already a challenge with our current apps.

We do appreciate the effort, but we also need to carefully manage the apps so that users get the best possible experience.

PR #221, opened 2022-10-25
App Submission: MyLar3

IMPranshu:
# App Submission

### MyLar3
...

### 256x256 SVG icon
https://svgshare.com/i/neP.svg
...

### Gallery images
Screenshot from 2022-10-25 20-55-16
Screenshot from 2022-10-25 20-55-41
Screenshot from 2022-10-25 20-56-17
Screenshot from 2022-10-25 20-56-01

...

### I have tested my app on:

### DIFF
--- a/mylar3/docker-compose.yml
+++ b/mylar3/docker-compose.yml
@@ -3,7 +3,7 @@ version: "3.7"
 services:
   app_proxy:
     environment:
-      APP_HOST: trilium-note_server_1
+      APP_HOST: mylar_server_1
       APP_PORT: 8090
   server:
     image: linuxserver/mylar3:0.6.9@sha256:d07a8e6311fa24770c59c4f4690857b833ba2e9c087ec70b3c6684ea14e774ca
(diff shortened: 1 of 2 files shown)


IMPranshu:
This App is now working on Umbrel. Had to remove `app-proxy`. I have verified it on both of my devices. Sorry!

nevets963:
@IMPranshu What was the issue you had when using the App Proxy?

IMPranshu:
When I was adding the `app proxy` the apps were stuck at `starting` in the Umbrel dashboard. This was happening with all of my apps. Don't know if it was a local issue but I removed the `app proxy` and the app worked fine.

nevets963:
@IMPranshu I'm not sure what issue you could be having with the App Proxy. I just added it and it's working fine (https://github.com/getumbrel/umbrel-apps/pull/221/commits/029fdd911f77fe9bd42451bbba7b34d572fc7f8a). It's potentially because you were trying to open the same port twice, ie the app proxy and the `server` container were trying to open the same port on the host.

IMPranshu:
I think the same was cached and was causing the error in all of my apps.

PR #227, opened 2022-10-27
Update LN Markets to v1.2.5

kibotrel:
Patchnote: [v1.2.5](https://github.com/ln-markets/umbrel/releases/tag/v1.2.5)

Tested on `arm64` !


nevets963 (approved):
@kibotrel Thanks for the app update!

Merging now 🚀

PR #235, opened 2022-10-30
Update BTCPay Server to use direct postgres nbx connection

Kukks:
A few versions ago, we introduced optimizations to btcpay to directly connect to the nbx db for faster querying and better dashboard functionality.


nevets963 (approved):
@Kukks Thanks for submitting the PR! I didn't know about this flag 😆 

Merging now 🚀

Issue #236, opened 2022-10-30
BTCPayServer Thinks It's in HTTP Context


leesalminen:
When attempting to use BTCPayServer on Umbrel in clearnet (say using Cloudflare Tunnel or Tailscale), BTCPayServer still thinks it's being connected to over HTTP (as opposed to HTTPS). Upon attempting to log in, you are presented with this error message from BTCPayServer: "You cannot login over an insecure connection. Please use HTTPS or Tor."

This is happening due to a lack of proper headers being passed by the app_proxy. Namely, the `X-Forwarded-Proto` header.

Proposed solution: introduce an nginx container to the BTCPayServer app that sits between `app_proxy` and `btcpayserver` so it can set the appropriate headers.

nevets963:
@leesalminen The App Proxy will forward headers from its upstream if told to do so for security reasons. We addressed this issue in Umbrel v0.5.1 and provided some instructions here: https://github.com/getumbrel/umbrel/issues/1428#issuecomment-1184584190

leesalminen:
Confirmed this issue is not valid, closing.

PR #237, opened 2022-10-30
App Update: BTCPayServer fixes

leesalminen:
This PR resolves issue #236. 

We introduce a new `nginx` container to the BTCPayServer app, which sets the proper headers for BTCPayServer to operate in a HTTPS context, conditionally, based on whether the user is accessing BTCPayServer via a clearnet domain name.


nevets963:
@leesalminen I'm wondering if my comment here: https://github.com/getumbrel/umbrel-apps/issues/236#issuecomment-1296709110 satisfies your usecase? And therefore whether this PR is still applicable?

leesalminen:
Yes, it does. This PR is no longer applicable. Closing out.

PR #243, opened 2022-11-02
Update Electrs to v0.9.9

nevets963:

### DIFF
--- a/electrs/docker-compose.yml
+++ b/electrs/docker-compose.yml
@@ -24,7 +24,7 @@ services:
         ipv4_address: $APP_ELECTRS_IP
   
   electrs:
-    image: getumbrel/electrs:v0.9.4@sha256:b1590ac6cfb0e5b481c6a7af7f0626d76cbb91c63702b0f5c47e2829e9c37997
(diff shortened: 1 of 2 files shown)


gustavobrossi:
🙏🏻

lestephane:
@nevets963 Is there a doc somewhere of the step required to do the same thing with v0.9.10. I'd be happy to chip in, not knowing where the docker images are built, I created a PR at https://github.com/getumbrel/docker-electrs/pull/13 but it is almost certainly not the only step. I need more info on contributing such updates, and testing, like you did. Any info is welcome...

nevets963:
@lestephane There are some docs in this repo's readme about how app updates get rolled out on Umbrel (https://github.com/getumbrel/umbrel-apps)

In short, we'll need:
- A PR (like this one) that updates the `docker-compose.yml` and `umbrel-app.yml`. Most notable details are: Docker images (we'll need to merge in https://github.com/getumbrel/docker-electrs/pull/13 first), the version and release notes updated
- Have your new PR tested on `arm64` and `amd64`.

We don't always release software immediately after it becomes available - just so that it allows late bugs to be discovered and fixed before it gets rolled out to a large number of users.

PR #244, opened 2022-11-02
Updated to 1.10 urbit binaries

mopfel-winrux:
I updated the urbit binary app to v1.10

### DIFF
--- a/rotki/docker-compose.yml
+++ b/rotki/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 80
 
   web:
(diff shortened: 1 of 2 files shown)


nevets963 (commented):
@mopfel-winrux Thanks for the PR to update Urbit!

I'm just wondering about the versioning with Urbit because we now have a field available inside of the `umbrel-app.yml` called `releaseNotes` which will show the user some notes about the app update. Would you be able populate this? Does `0.5.0` refer to this release? https://github.com/urbit/urbit/releases/tag/urbit-v1.10

nevets963:
@mopfel-winrux Just a friendly ping re https://github.com/getumbrel/umbrel-apps/pull/244#pullrequestreview-1167169947

PR #249, opened 2022-11-02
Enable RPC middleware in LND

jamaljsr:
The next major release of Lightning Terminal will require the `--rpcmiddleware.enabled` flag to be set in order to use some new features we've been working on. `litd` will not startup without this flag being set. I am preemptively opening this PR before that release in order to have the LND dependency in order.

More details on what this flag does can be found at https://github.com/lightningnetwork/lnd/pull/5101.

### DIFF
--- a/lightning/exports.sh
+++ b/lightning/exports.sh
@@ -21,6 +21,9 @@ BIN_ARGS+=( "--tlsautorefresh" )
 # lnd.conf for existing users via a migration.
 BIN_ARGS+=( "--accept-amp" )
 
+# Lightning Terminal (litd) now requires this flag to be set or it will not startup
+BIN_ARGS+=( "--rpcmiddleware.enable" )
+
 # [Bitcoind]
 BIN_ARGS+=( "--bitcoind.rpchost=${APP_BITCOIN_NODE_IP}" )
 BIN_ARGS+=( "--bitcoind.rpcuser=${APP_BITCOIN_RPC_USER}" )
--- a/lightning/data/lnd/lnd.conf
+++ b/lightning/data/lnd/lnd.conf
@@ -6,3 +6,6 @@ accept-amp=1
 
 [Bitcoin]
 bitcoin.defaultchanconfs=2
+
+[rpcmiddleware]
+rpcmiddleware.enable=true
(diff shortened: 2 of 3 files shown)


lightning/exports.sh
```diff
@@ -21,6 +21,9 @@ BIN_ARGS+=( "--tlsautorefresh" )
 # lnd.conf for existing users via a migration.
 BIN_ARGS+=( "--accept-amp" )
 
+# Lightning Terminal (litd) now requires this flag to be set or it will not startup
+BIN_ARGS+=( "--rpcmiddleware.enabled" )
```
nevets963, line 25: Potentially a typo here? My `lnd` binary says the flag is: `--rpcmiddleware.enable`

nevets963 (commented):
@jamaljsr Thansk for the PR to enable rpcmiddleware within LND. Is there some reason that isn't enabled by default?

We should also update the version within the app manifest (`umbrel-app.yml`) so that when this merges in, it will register as an update within Umbrel.

jamaljsr:
> Is there some reason that isn't enabled by default?

It's not enabled by default because not all nodes need this additional functionality, but we will be making use of the extensibility point in litd going forward. We will update the litd installation instructions to make users aware that they need to set this LND flag when installing manually. Since Umbrel users are using a pre-packaged LND app, they won't be able to set this flag themselves. This is why we'd like to have this flag included in the app.

> We should also update the version within the app manifest (`umbrel-app.yml`) so that when this merges in, it will register as an update within Umbrel.

Thanks for this info I wasn't aware of. I used the version `0.15.4-beta-build-2` following what was done in [7d7b37f](https://github.com/getumbrel/umbrel-apps/commit/7d7b37f02e2b7ebdc70b7ad4460151f410fe4fd6#diff-12711e89dd4daf198f6bb8e95a7f923c3c2c127d15023db6ae1089ecc1568815R5). Is this ok?

lightning/exports.sh
jamaljsr: Ah, typo. Thanks.

Fixed in https://github.com/getumbrel/umbrel-apps/pull/249/commits/7002ebc92a59ec128518671cc391dd1e85e8105b

aphex3k (approved):

nevets963 (approved):
Tested on my Umbrels with real sats.

Merging now 😄

PR #250, opened 2022-11-03
App Submission: Nginx Proxy Manager

IMPranshu:
# App Submission

### Nginx Proxy Manager
...

### 256x256 SVG icon
https://github.com/IMPranshu/svg-logo/blob/main/npm.svg
...

### Gallery images
npm6
npm1
npm2
npm3
npm4
npm5

...

### I have tested my app on:

### DIFF
--- a/nginx-proxy-manager/docker-compose.yml
+++ b/nginx-proxy-manager/docker-compose.yml
@@ -1,0 +1,13 @@
+version: "3"
+
+services:
+  server:
+    image: jc21/nginx-proxy-manager:latest
+    restart: on-failure
+    ports:
+      - 81:81
+      - 8180:80
+      - 8443:443
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${APP_DATA_DIR}/letsencrypt:/etc/letsencrypt
--- a/nginx-proxy-manager/docker-compose.yml
+++ b/nginx-proxy-manager/docker-compose.yml
@@ -1,7 +1,12 @@
-version: "3"
+version: "3.7"
 
 services:
+  app_proxy:
+    environment:
+      APP_HOST: nginx_proxy_manager_app_proxy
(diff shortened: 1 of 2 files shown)


nevets963 (commented):
@IMPranshu I'm thinking about the usage of this proxy manager. As I understand it, you would typically have this hosted accepting incoming HTTP/HTTPS traffic, and then this proxy does the SSL termination and will forward the traffic to internal services?

If so, I see a few challenges:
- The Umbrel Web UI is currently listening on port 80, so we would have to use a non-standard port for incoming HTTP traffic. And if you're running Umbrel on a public cloud server, that could abit problematic
- Most Umbrel users are running their server in their own home, which means to expose their apps to the outside world, they'll need need setup port forwarding on their router or use a third-party tunneling service

Any thoughts on the usage model here?

IMPranshu:
> * The Umbrel Web UI is currently listening on port 80, so we would have to use a non-standard port for incoming HTTP traffic. And if you're running Umbrel on a public cloud server, that could abit problematic

Maybe in the future release, Umbrel can provide an option to run Web UI in a different port, which can be toggled by the user as per their choice which will make things easier.



> * Most Umbrel users are running their server in their own home, which means to expose their apps to the outside world, they'll need need setup port forwarding on their router or use a third-party tunneling service

Yes. Port forwarding has to be set up. But think of the possibilities when anyone can access their umbrel via their own domain name and their websites/apps hosted on their own Umbrel from anywhere it will be just like a "Closed Circle" of sorts.

Nginx running on Umbrel will allow Running Umbrel Apps via their own Domain name. That is what I am trying to achieve with this and future app additions.

Would love to hear your thoughts regarding/against this.

IMPranshu:
This App is now working on Umbrel. Had to remove `app-proxy`. I have verified it on both of my devices. Sorry!

PR #256, opened 2022-11-06
App Update: Photoprism to 221105

Xosten:
### Note
After this version, Photoprism started to separate their images based on OS/Arch. Unless we use the `latest` tag for the image I don't really know how we could proceed to keep this application up-to-date. 

See Docker Hub : https://hub.docker.com/r/photoprism/photoprism/tags

### DIFF
--- a/photoprism/docker-compose.yml
+++ b/photoprism/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       # More details here: https://docs.photoprism.app/user-guide/sync/webdav/#server-url
       PROXY_AUTH_WHITELIST: "/originals/*,/import/*"
   web:
-    image: photoprism/photoprism:20220121@sha256:8dd9e8e5290c8bf648d0e91ea8ef81038ffc8800ebf6322417682a0342e04065
+    image: photoprism/photoprism:221105-bullseye@sha256:52f2cfb9f46aec44dd2dec87c78ca121cc6fd174510db1ac371072c527f86558
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: "1m"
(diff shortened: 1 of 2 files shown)


nevets963:
@Xosten We and Umbrel users really appreciate the PRs to keep the apps up to date!

Looking at the available tags, I see they build a few different variations. On the surface, this tag looks fine to me: https://hub.docker.com/layers/photoprism/photoprism/221105-bullseye/images/sha256-bc2429c4e23a570fb8c6b5ffa86acb323c01c72012a5bf94c3c47207f070a767?context=explore. Of course, this will need testing.

The `-bullseye` simply means that's the base image they're using within the Docker image (as shown here: https://github.com/photoprism/photoprism/blob/7a295cab4931d15d685e272b9363c734cfe78c0f/docker/photoprism/bullseye/Dockerfile#L21-L22). As opposed to meaning this the OS that the Docker image will run on.

Xosten:
Thanks Steven for taking a look. I didn't understand the same thing when I read through the Dockerhub. With your recommendation, I tested out the `-bullseye` tag on my end. I uploaded some pictures with the current Photoprism version, then updated Photoprism to `20221105-bullseye` without a problem. I have tested it on a Linux VM.

https://github.com/Xosten/umbrel-apps/tree/Update-Photoprism-221105-bullseye

nevets963 (approved):
@Xosten Thanks for submitting the PR!

I made some minor changes; like changing the versions simply to `221105` as the `-bullseye` is just an internal detail that users can ignore. And I fixed up the release notes.

Merging now 🚀

Xosten:
Awesome! Thanks Steven!

PR #257, opened 2022-11-06
App Update: Nextcloud to 25.0.1

Xosten:


nevets963:
@Xosten Have you tested with this an existing (and older) Nextcloud installation?

We're aware of an issue where if you upgrade an older Nextcloud instance to something more recent, then it seems to corrupt the Nextcloud installation which is then completely inaccessible 😨

Xosten:
> @Xosten Have you tested with this an existing (and older) Nextcloud installation?
> 
> We're aware of an issue where if you upgrade an older Nextcloud instance to something more recent, then it seems to corrupt the Nextcloud installation which is then completely inaccessible 😨

Hi Steven,

This is on me as when I did my research, it didn't come up. Sorry about that...

Upgrading from 22.2.10 (current version) to 25.0.1 is blocking indeed the instance into the maintenance state. When I force to stop the maintenance mode, it gave me the option to upgrade all the plugins. After that, it gave me another error saying that I was jumping to many versions. I guess we can't jump over many versions....

So I created a new branch on my end, with the 23.0.11 version and managed to upgrade my Nextcloud instance without a problem. Could you try it on your end and confirm it works? If it works, we could upgrade to 23 and could envision upgrading to 24 in a few months maybe?

https://github.com/Xosten/umbrel-apps/tree/Update-Nextcloud-23.0.11

nevets963:
@Xosten Ok, don't worry about this as we know about issue and I'm working internally to build a robust update process for this. Essentially stepping through the intermediate versions until we get to the latest version.

The problem with simply waiting months is; if a user doesn't install that intermediate version, then they'll end up with the same issue, and I have no idea how easy it is to recover from that...

Xosten:
Ok, no problem. So should we leave this app update pending in that case?

nevets963:
I'll close it as they'll be another PR that will address this

PR #259, opened 2022-11-07
App Update: lnplus (Lightning Network+) - Fix broken 'Edit Account' form

lnplus:
Important update: Fixes the error when loading the page: `/edit_account`


lnplus:
I'm not sure why all the commits are visible, but only the very last one is the actual change: 2c8bd68

nevets963 (approved):
Tested on `arm64` and `amd64`,

Approved 🚀

PR #260, opened 2022-11-07
Update Jellyfin ports in docker-compose.yml

JimSB:
proposing adding additional 1900 port for Jellyfin, as some users are having trouble with scanning media libraries DLNA/UPnP

### DIFF
--- a/jellyfin/docker-compose.yml
+++ b/jellyfin/docker-compose.yml
@@ -20,6 +20,7 @@ services:
     ports:
       # Service auto-discovery
       - 7359:7359/udp
+      - 1900:1900/udp
     networks:
       default:
         ipv4_address: $APP_JELLYFIN_IP


nevets963 (commented):
@JimSB Thank you for the PR!

Although I agree with the change, the problem is; currently the Plex app is set to listen on `1900/UDP`, which means it then won't be possible to have both installed. And if we approved this, assuming that users would (typically) install 1 or the other. Should they then attempt to install the other, the app will get stuck in the Umbrel UI showing 'Starting...' as we don't have proper error handling atm for this kind of situation.

@lukechilds What are your thoughts here?

lukechilds:
Thanks for the PR @JimSB, really appreciate it!

>@lukechilds What are your thoughts here?

Is there any benefit to exposing the DLNA port on the host as 1901 or does it have to be 1900 to work?

If it can be useful as 1901, lets do that. If not then unfortunately we don't have a good way to handle port conflicts right now. Port conflicts in apps are something we definitely want to handle gracefully in Umbrel but it'll require some changes to Umbrel internally that we don't have ready just yet.

lukechilds:
Also @JimSB just to offer you a (kind of) solution. You can get this working on your personal Umbrel by SSHing in and editing `~/umbrel/app-data/jellyfin/docker-compose.yml` to include the port as you have in your PR. You'd also need to comment out the 1900 port in Plex's compose file if you have Plex installed too. It's important you only edit the files contained in `~/umbrel/app-data/` and not anywhere else.

Then restart your Umbrel and Jellyfin will have it's DLNA port exposed. You'll have to re-apply these changes after a Jellyfin app update. I appreciate this is not an ideal solution and we definitely plan to handle this better in a future release but unfortunately the things we need in place to support this aren't quite ready yet.

nmfretz:
@lukechilds @JimSB are you both okay if we close this PR?

PR #261, opened 2022-11-09
Add proxy auth whitelist for nolooking

nickfarrow:
From what I can tell, this auth is adding some protection to the payjoin endpoints.

Results in wallets giving messages like:
```
2022-11-10 00:59:48,027 ERROR Error parsing received PSBT
com.sparrowwallet.drongo.psbt.PSBTParseException: Provided string is not a PSBT
```

I think we should remove this auth for now, to fix the app, and inform users to keep their onion private? (Worst they can do is open channels)

### DIFF
--- a/nolooking/docker-compose.yml
+++ b/nolooking/docker-compose.yml
@@ -5,6 +5,7 @@ services:
     environment:
       APP_HOST: nolooking_server_1
       APP_PORT: 4444
+      PROXY_AUTH_WHITELIST: "/pj"
   
   server:
     image: satsale/nolooking:0.1.2@sha256:0e26450f6be181d2392bb51b7ecc69446d782d778622e5634c85f54c62a2fdd2
--- a/nolooking/umbrel-app.yml
+++ b/nolooking/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: nolooking
 category: Lightning Node Management 
 name: nolooking
-version: "0.1.0"
+version: "0.1.0-build-2"
 tagline: Use PayJoin to open channels for your lightning node
 description: >-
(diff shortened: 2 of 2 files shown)


nevets963:
@nickfarrow Just before we completely disable this. A potential solution might be to whitelist certain endpoints inside nolooking. See here for an example: https://github.com/getumbrel/umbrel-apps/blob/39bdd2e73f17f386cab9ed894ea05c1ebffdffe1/photoprism/docker-compose.yml#L10

nickfarrow:
Much better! Thanks for the tip

nolooking/docker-compose.yml
```diff
@@ -5,6 +5,7 @@ services:
     environment:
       APP_HOST: nolooking_server_1
       APP_PORT: 4444
+      PROXY_AUTH_WHITELIST: "/pj"
```
nevets963, line 8: Just wanted to check if this is not the equivalent as turning off the authentication (in the App Proxy) all together? If I recall, the whole app was hosted at `/pj` right?

nickfarrow: Yeah it is, I think this is OK for now (since I believe the only risk of leaked onion is that someone could send you money and open channels). We are separating endpoints out in an upcoming PR
https://github.com/chaincase-app/nolooking/pull/55

nevets963: Ok understood, do you want to wait for that to be ready to avoid pushing another update. Or would you like this merged ASAP?

nickfarrow: Let's do it now since the app is broken until then, and as long don't another update PR sometime next week. Thanks!

nevets963:
@nickfarrow I just bumped the `version` property in `umbrel-app.yml` so that it registers as an update within Umbrel and also added release notes.

Merging now 😄

nickfarrow:
Thank you @nevets963!!

PR #273, opened 2022-11-10
App Submission: WikiJS

IMPranshu:
# App Submission

### WikiJS
...

### 256x256 SVG icon
https://github.com/IMPranshu/svg-logo/blob/main/wiki.svg

...

### Gallery images
wiki
Screenshot from 2022-11-11 01-24-20
Screenshot from 2022-11-11 01-25-55
Screenshot from 2022-11-11 01-26-36

...

### I have tested my app on:

### DIFF
--- a/wikijs/docker-compose.yml
+++ b/wikijs/docker-compose.yml
@@ -5,6 +5,7 @@ services:
     environment:
       APP_HOST: wikijs_server_1
       APP_PORT: 3000
+      PROXY_AUTH_ADD: "false"
   
   server:
     image: linuxserver/wikijs:2.5.294@sha256:e9f79a619fb0da03dc2e20d1543a02eb655f09a1a112c6998ebd584ada533d66
--- a/wikijs/docker-compose.yml
+++ b/wikijs/docker-compose.yml
@@ -3,13 +3,12 @@ version: "3.7"
 services:
   app_proxy:
     environment:
-      APP_HOST: wikijs
+      APP_HOST: wikijs_server_1
       APP_PORT: 3000
+  
   server:
-    container_name: wikijs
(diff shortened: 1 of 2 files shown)


nevets963 (approved):
@IMPranshu Thanks for submitting WikiJS! We appreciate your efforts that went into this PR!

@mayankchhabra the App Proxy's default configuration will lock-down this app and therefore this will behave like a private Wiki (which maybe desirable). However, given this has it's own auth. system and this type of application is usually about sharing data/info, it seems like we should the App Proxy off by default for WikiJS. Thoughts?
We'll help need your help here again with App Store graphics :)

However in its current form, this is approved from me.

Once graphics assets are merged, I'll run another test and then merge this in.

mayankchhabra:
Thanks again for another solid app submission, @IMPranshu!

> the App Proxy's default configuration will lock-down this app and therefore this will behave like a private Wiki (which maybe desirable). However, given this has it's own auth. system and this type of application is usually about sharing data/info, it seems like we should the App Proxy off by default for WikiJS.

Sounds good, agree with your reasoning! We can turn it off.

nevets963:
Gallery assets are located here: https://github.com/getumbrel/umbrel-apps-gallery/pull/28 Will start on final tests today before merging.

nevets963:
Great work all round everyone! And Happy Holidays! Merging now 🚀

kszere:
How update Wiki.JS to v2.5.300?

Niclassslua:
image

<br><br>

Compared to other App Icons the WikiJS icon is very low quality.

I would suggest using the one I uploaded. It's 900x900.

<br><br>

wikijs

PR #282, opened 2022-11-15
Update Pi-hole to v5.14.1

mccolister1017:


nevets963 (approved):
@mccolister1017 Thanks again for the PR to update Pi-Hole!! 😄 

Merging now 🚀

PR #296, opened 2022-11-18
Update nolooking to v0.2.0-alpha

DanGould:
Built for docker here: https://github.com/DanGould/nolooking/actions/runs/3509289629, but renamed so the hash doesn't match. From here on out we're building on tag @chaincase-app/nolooking

### DIFF
--- a/nolooking/docker-compose.yml
+++ b/nolooking/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_WHITELIST: "/pj"
   
   server:
-    image: chaincase/nolooking:v0.2.1-alpha@sha256:905b6f8cb89a292e5f0d6cad3b5a093751725cb3d4981e046f9aac67fc201c1b
+    image: chaincase/nolooking:v0.2.1-alpha@sha256:3e1c0d32d86762e07074bdf27b117c61d5400aa1336663f91dd302ff72fd7671
     restart: on-failure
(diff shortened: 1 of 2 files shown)


nickfarrow:
Awesome, thanks for updating @DanGould

DanGould:
Pushed a build with amd64 & arm64

nickfarrow:
Tested!

nickfarrow:
@nevets963 I see a different digest on 
https://hub.docker.com/layers/chaincase/nolooking/v0.2.1-alpha/images/sha256-905b6f8cb89a292e5f0d6cad3b5a093751725cb3d4981e046f9aac67fc201c1b

Both worked with `sudo ./scripts/app install nolooking` though 🤔

nevets963:
@nickfarrow That is the image digest for `arm64` which if we use means that won't then work on `amd64`. If you do a `docker pull <image>`, it will return the multi-arch digest at the bottom.

nevets963 (commented):
I just tested on `arm64` and `amd64` and this looks nearly good to go! Would you mind updating the release notes (`releaseNotes`) within `umbrel-app.yml` and then I'll get this merged in 🚀

nickfarrow:
Added release notes. Thanks!

nevets963 (approved):
Everything looks great!

Merging now 🚀

PR #297, opened 2022-11-19
Pihole 2022.11.1 with FTL 5.19.2

mccolister1017:
Update is 2 days old .


nevets963 (commented):
@mccolister1017 I'm just wondering if you can clarify this update as I don't see a new release for Pi-Hole (https://github.com/pi-hole/pi-hole/releases). So is it just that the Docker image includes an update for the FTL component but Pi-Hole remains on 5.14.1?

nevets963 (approved):
@mccolister1017 Thanks for the clarification (and the PR!) 😄 

Merging now 🚀

PR #300, opened 2022-11-22
App Update: Uptime kuma v1.18.5

wiwatsrt:
Update Uptime kuma v1.17.1 to v1.18.5


nevets963 (approved):
@wiwatsrt 1000 thanks for the PR to update Uptime Kuma!

I corrected the Docker image digest as it was for a specific platform. You can get multi-arch digests by running `docker pull <image>` and then it will display the digest at the end.

Merging now 🚀

PR #302, opened 2022-11-23
App Submission: Audiobookshelf

ceramicwhite:
# App Submission

### App name
Audiobookshelf

### 256x256 SVG icon
https://cloudflare-ipfs.com/ipfs/QmaqAsckuH7mxS791uvLXnUGn3c4sSQuakzsSCkFGbmHLt/audiobookshelf.svg

...

### Gallery images
LibraryStream
series
library-stats

### I have tested my app on:


nevets963 (commented):
@ceramicwhite Many many thanks for submitting Audiobookshelf into the Umbrel App Store! This is a great addition along with the other media apps 😄 

Tested on `arm64` and `amd64` and is working pretty nicely for me.

@mayankchhabra Just checking in here re the location(s) of the audiobooks and podcasts. Here the mounts are for the Umbrel downloads directory which makes sense to me - also good with you too?
If so, I'll add a script to this PR so that these directories get setup properly when audiobookshelf gets installed.

ceramicwhite:
@nevets963 Yeah, Audiobookshelf is fire. My friends, family, and I use my server daily, it's become a mini-audible.
I put the global downloads folder otherwise users would need to ssh in and copy over the audiobooks or ftp in and upload books. At least this way user can just torrent from Transmission straight into a mapped folder for Audiobookshelf.
Otherwise maybe worth considering adding something like [doublecommander](https://github.com/linuxserver/docker-doublecommander/) that has view of all the media apps folders on Umbrel

nevets963:
Gallery assets are located here: https://github.com/getumbrel/umbrel-apps-gallery/pull/22 Will start on final tests today before merging.

nevets963:
@ceramicwhite The gallery assets have been merged in and I did some final testing. Many thanks! Merging now 🚀 and Happy Holidays! 🎄

anthony-robin:
To start, I would like to thank you all very much for submitting this awesome app to the Umbrel store ! 🤩

Then I would like to escalate an issue I'm facing with the [mobile Android application](https://github.com/advplyr/audiobookshelf-app): indeed, it seems impossible to connect to the self hosted app on my Umbrel as shown in the following screenshot:

Screenshot_20221225_145153_audiobookshelf

After investigating, I discovered that the self hosted app is protected by the Umbrel authentication first which prevent the mobile app to connect the server:

Capture d’écran 2022-12-25 à 15 00 21

The question is: should authentication be disabled on Umbrel side or the issue addressed in the mobile app side ?

To end, thanks again and have a merry christmas ! 🎄 🥂

nevets963:
@anthony-robin Thanks for reporting this issue. We've seen this before with other apps - as we by default lockdown apps using our App Proxy Authentication which is what you see above.

There are a few ways to solve this:
- Do you know if Audiobookshelf has an API (that has its own auth. system) which is used by these mobile apps? If this is the case, we can whitelist that with the App Proxy.
- A manual override can be set (using something similar to these instructions: https://github.com/getumbrel/umbrel/issues/1412#issuecomment-1149022283)

anthony-robin:
@nevets963 Thank you for your reply !

I'm not aware on how Audiobookshelf works under the hood, so I just opened an issue in the repository (reference: https://github.com/advplyr/audiobookshelf-app/issues/484) to get more details.

Awaiting it, I followed your second point to disable the proxy for the app and it worked like a charm 💯

ceramicwhite:
@anthony-robin I've sent a PR to bump version and add whitelist for proxy.

@nevets963 There is an [API](https://api.audiobookshelf.org/#login) but the mobile pings root, so no way around this as far as I know without whitelisting the root domain.
Tbh, if people are exposing their umbrel and/or running on a VPS they should be setting up the admin user right after install anyways.
An option to protect users before they setup their own admin user would be to skip the init process of the server by including a file at:
`app-data/audiobookshelf/data/config/users/data/data.1.json`

`{"id":"root","username":"umbrel","pash":"$2a$08$zdi8XLuSvxHgsYu2JHCubuCpBikD1VOeyym65rCDU7SID1XendDXq","type":"root","token":"","mediaProgress":[],"seriesHideFromContinueListening":[],"bookmarks":[],"isActive":true,"isLocked":false,"lastSeen":1672263579487,"createdAt":1672263569887,"settings":{"mobileOrderBy":"recent","mobileOrderDesc":true,"mobileFilterBy":"all","orderBy":"media.metadata.title","orderDesc":false,"filterBy":"all","playbackRate":1,"bookshelfCoverSize":120,"collapseSeries":false},"permissions":{"download":true,"update":true,"delete":true,"upload":true,"accessAllLibraries":true,"accessAllTags":true,"accessExplicitContent":true},"librariesAccessible":[],"itemTagsAccessible":[]}`

This would set a default admin user, with username: `umbrel` & password: `moneyprintergobrrr`

nevets963:
@ceramicwhite Thanks for the detail! Would it be possible to whitelist: `/` and `/api/*`, could that work? Potentially `/login` would be needed too?

ceramicwhite:
@nevets963 whitelisting `/` will work.
This also worked:
```
      PROXY_AUTH_WHITELIST: "/"
      PROXY_AUTH_BLACKLIST: "/login,/api,/config"
```

calm108:
**ERROR: for web Get "https://ghcr.io/v2/" Install fails on Umbrel**

I cannot install AudioBookShelf on Umbrel. I have dome multiple clean installs of Umbrel. installed AudioBookShelf first etc. I always get the same error:
ERROR: for web Get "https://ghcr.io/v2/": dial tcp: lookup ghcr.io on 192.168.1.254:53: no such host
Get "https://ghcr.io/v2/": dial tcp: lookup ghcr.io on 192.168.1.254:53: no such host

I CAN however install AudioBookShelf successfully using Portainer on Umbrel pulling the image from docker.io This leads me to suspect there is possibly something wrong with the image or where it is being pulled form for Umbrel? my guess only.

Can anyone help resolve or shed some light on this? I prefer to have AudioBookShelf fully managed under Umbrel and not in a Portainer container unless I have to.

PR #307, opened 2022-11-23
RTL on LND v0.13.2

ShahanaFarooqui:
RTL on LND v0.13.2


nevets963 (approved):
@ShahanaFarooqui Thanks for the 2nd PR for RTL with LND!

Tested `arm64` and `amd64`.

Merging now 🚀

PR #310, opened 2022-11-28
Remove hard-coded container IPs

nevets963:
This PR removes hard-coded container IPs which simplifies the deployment of an app on Umbrel.

Most apps also do not need to 'export' any variables - so where unneccesary, the `exports.sh` scripts have also been removed.


RoflWolke7:
@nevets963: This commit breaks addons!
At least Home Assistant and Pi hole do not work anymore.

There is a mixup between dashes and underscores!
e.g.
HomeAssistant is configured with a network alias "home-assistant-server-1", but in docker-compose of HomeAssistant the APP_HOST env variable for app_proxy was set to: "home-assistant_server_1". 

So the app proxy is logging:

Error waiting for port: "The address 'home-assistant_server_1' cannot be found"
Retrying...

nevets963:
@RoflWolke7 Thanks for sharing this finding and for the link!

Can I ask how you have this version of `docker-compose` installed? On both my Pi-based Umbrel (setup by flashing an SD card with Umbrel OS) and my Intel-based Umbrel (setup using this [bash script](https://umbrel.sh)), I have version `1.29.2` installed where I don't see any issue.

RoflWolke7:
You are right - my "fault".
A while ago (I nearly forgot it already) after an upgrade the containers did not start anymore.
And after short trouble shooting I recognized that docker-compose was missing.
So I installed it by my own.

I was not aware that there is a version "incompatibility".
But I suppose this is a fact good to know for you too.

I fixed it on my umbrella setting the env variable "COMPOSE_COMPATIBILITY=true" in my umbrel/.env

nevets963:
@RoflWolke7 Perfect, glad to hear adding that flag to your .env file works as I was about to suggest that. Thanks again for sharing this!

santochibtc:
I have the same problem as @RoflWolke7, in my case with the samourai app
The whirlpool GUI can't authenticate
"Authentication failed: Unable to connect to wallet backend: http://samourai-server_nginx_1:80/v2/"
That string samourai-server_nginx_1 is not being translated to the local IP

PR #314, opened 2022-11-29
Access docker.sock from within container.

dskvr:
Most common problem with Umbrel is `lnd` crashes. Enable the ability to check `lnd` uptime and generally most useful feature in Uptime Kuma (docker container monitoring)

### DIFF
--- a/uptime-kuma/docker-compose.yml
+++ b/uptime-kuma/docker-compose.yml
@@ -12,6 +12,7 @@ services:
     stop_grace_period: 1m
     volumes:
       - ${APP_DATA_DIR}/data/uptime-kuma:/app/data
+      - /var/run/docker.sock:/var/run/docker.sock
     networks:
       default:
         ipv4_address: $APP_UPTIME_KUMA_IP


uptime-kuma/docker-compose.yml
```diff
@@ -12,6 +12,7 @@ services:
     stop_grace_period: 1m
     volumes:
       - ${APP_DATA_DIR}/data/uptime-kuma:/app/data
+      - /var/run/docker.sock:/var/run/docker.sock
```
nevets963, line 15: We can't allow an app to have access to the Docker unix socket for security reasons, as they could take over and manipulate any running container on the host.

As an alternative, you can have Uptime Kuma ping the hostname `lightning_lnd_1` for the LND container to check if it's running or not.

aphex3k:
> Most common problem with Umbrel is `lnd` crashes.

I can not second that opinion, however, I have added [support for docker-compose.override.yml](https://github.com/getumbrel/umbrel/pull/1487) to my umbrel installation and this is one of the modifications that I'm running as well.

PR #316, opened 2022-12-02
Update Gitea to 1.17.3

pynixis:
Updating docker images to latest.

### DIFF
--- a/gitea/docker-compose.yml
+++ b/gitea/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   server:
-    image: gitea/gitea:latest
+    image: gitea/gitea:1.17.3
     user: "1000:1000"
     restart: on-failure
     ports:
--- a/gitea/docker-compose.yml
+++ b/gitea/docker-compose.yml
@@ -34,7 +34,7 @@ services:
         ipv4_address: $APP_GITEA_IP
 
   db:
-    image: mariadb:latest
+    image: mariadb:10.9.4
     user: "1000:1000"
     restart: on-failure
     volumes:
--- a/gitea/docker-compose.yml
+++ b/gitea/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   server:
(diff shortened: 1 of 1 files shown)


gitea/docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   server:
-    image: gitea/gitea:1.16.8-rootless@sha256:8a6ad22694c76122ae075af1153a544577724c45cb45af44963aa78dc179cc46
+    image: gitea/gitea:latest
```
nevets963, line 11: We can't accept the usage of `latest` as a Docker image tag because this would allow updates to simply get rolled out without at least some check from us.

pynixis (commented):
Let me know. If we need a different version for each image.

gitea/docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   server:
-    image: gitea/gitea:1.16.8-rootless@sha256:8a6ad22694c76122ae075af1153a544577724c45cb45af44963aa78dc179cc46
+    image: gitea/gitea:1.17.3
```
nevets963, line 11: Yes exactly, we need to use a specific tag for a specific version of the app. We're just missing the Docker image digest. You can get this by running: `docker pull <image>` and it will display at the end.

gitea/docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
         ipv4_address: $APP_GITEA_IP
 
   db:
-    image: mariadb:10.5.12@sha256:dfcba5641bdbfd7cbf5b07eeed707e6a3672f46823695a0d3aba2e49bbd9b1dd
+    image: mariadb:10.9.4
```
nevets963, line 37: Are you sure it's safe for the database to update from `10.5.12` to `10.9.4`?

pynixis: No. The mariadb image needs to be backtested.

nevets963: @pynixis Would you be able to kindly test this? I would suggest you install the current version of gitea and test this app update with the newer version of mariadb.

nevets963:
Closing to due to no activity. Happy to re-open if we can get the newer DB container tested.

PR #322, opened 2022-12-07
Update nolooking to v0.2.2-alpha

nickfarrow:


nevets963 (approved):
@nickfarrow Thanks for the app update PR!

Merging now 🚀

PR #339, opened 2022-12-21
Bump Torq to v0.15.4

maxwedwards:
Apologies that we only released a week ago but we got a few nice bug fixes in and a new network toggle for changing between mainnet and the various test nets which keeps the data segregated for those who use Torq with a testnet node as well as a main net node.

Tested on a real RaspberryPi

### DIFF
--- a/torq/docker-compose.yml
+++ b/torq/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 7028
 
   web:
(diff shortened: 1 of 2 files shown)


mayankchhabra:
Hey Max! Thanks for the submitting update. 🚀

No need to apologize, you can ship updates as quickly as you'd like!

Let's wait for @nevets963 to give this a green light!

nevets963 (approved):
@maxwedwards No worries at all - we totally understand that teams want to move quickly to ship features and fix bugs. Happy holidays! 🎄 

Merging now 🚀

PR #341, opened 2022-12-23
Pihole to v5.18

mccolister1017:
full changelog here https://github.com/pi-hole/docker-pi-hole/releases


nevets963 (approved):
@mccolister1017 Many thanks for the PR 😄  I just touched up the release notes as they were slightly mangled.

Merging now 🚀

PR #344, opened 2022-12-28
Update Audiobookshelf to 2.2.11

ceramicwhite:
- Bump Image to 2.2.11
- Added  `PROXY_AUTH_WHITELIST` for mobile app login


audiobookshelf/docker-compose.yml
```diff
@@ -5,9 +5,12 @@ services:
     environment:
       APP_HOST: audiobookshelf_web_1
       APP_PORT: 80
+      PROXY_AUTH_WHITELIST: "/"
+      PROXY_AUTH_BLACKLIST: "/login,/api,/config"
```
nevets963, line 9: @ceramicwhite I think some adjustment is still needed here. I've been testing with the iOS app and I am getting `Unknown Error` when trying to login. Without digging deeper - perhaps the app (like the webapp) is using `/login` for auth too? But here it's marked a blacklisted (ie protected by the App Proxy).

ceramicwhite: Hmm strange, I just tried it again with a fresh install worked fine. On my main ABS server I tell friends to click login twice in the mobile app because first time it usually throws an error, fails handshake or something. Maybe that's what happened?

nevets963: This is what I'm seeing:
IMG_1772

If I remove `/login` from the App Proxy blacklist, then it works - I'm happy with that as a solution because audiobookshelf already has it's own auth. system. But I do see that on first install, that `/login` is used to set the inital `root` user password...

ceramicwhite: sounds good!

nevets963 (approved):
@ceramicwhite Thanks for working with me to find a resolution!

Merging now 🚀

PR #351, opened 2023-01-03
Update Kollider to v1.0.7

djkazic:
This fixes a bug on the arm version where pyzmq cannot connect to a hostname. Due to the recent change from hardcoded IPs to hostnames, this broke the Kollider app for some users.

### DIFF
--- a/kollider/docker-compose.yml
+++ b/kollider/docker-compose.yml
@@ -20,7 +20,7 @@ services:
       LND_ZMQ_SUB_ADDRESS: "tcp://kollider_ws_1:5556"
 
   ws:
-    image: kolliderhq/kollider-ws-client:v1.0.7@sha256:7136a6b25b452c93ed2762f5ab00e6a6d9a76d586c1eb1bdf962bee684135cd5
+    image: kolliderhq/kollider-ws-client:v1.0.7@sha256:8a68f15a8f15ce4957629f02346d3202e685781f2af026ab2c1c028179b42830
     init: true
     user: 1000:1000
     restart: on-failure
@@ -35,7 +35,7 @@ services:
       KOLLIDER_ZMQ_HEDGER_SUB_ADDRESS: "tcp://kollider_backend_1:5559"
 
   web:
-    image: kolliderhq/kollider-lite-app:v1.0.7@sha256:fd8b116d5b6e82686fd71f5d41206157e5a311b96587ff5ec14f01c24a13fd94
+    image: kolliderhq/kollider-lite-app:v1.0.7@sha256:e459ae26485c52fb2967d7e9c2afeca19b51845c8f4739d02feda9e01db3037a
     init: true
     user: 1000:1000
     restart: on-failure
--- a/kollider/docker-compose.yml
+++ b/kollider/docker-compose.yml
@@ -5,9 +5,9 @@ services:
     environment:
       APP_HOST: kollider_web_1
       APP_PORT: 3000
-  
+
   backend:
-    image: kolliderhq/kollider-lite-backend:v1.0.6@sha256:5558948e716fe31956ef1e3985aede704b0bcd9b3af2532395c8774cffdf1666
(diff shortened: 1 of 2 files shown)


nevets963 (commented):
@djkazic Thanks for the PR to fix this - going foward we wanted to simplify app deployment on Umbrel and therefore we've opted for hostnames over hard-coded and static container IPs.

I'm currently getting an error when trying to pull 2 out of the 3 Docker images on 2 completely separate Umbrel servers I have - would you be able to take a look at this?

Would you be able to add release notes to app manifest (`umbrel-app.yml`)? This somewhat new feature will show details to the user when they update. I just tried to push myself, but I unfortunately don't have write access. Here is an example: https://github.com/getumbrel/umbrel-apps/blob/9e2756aee9088d2639a95c94eafc9cb9c896594a/electrs/umbrel-app.yml#L32-L39

djkazic:
Forgot to update the SHA256 hashes. I'll do that now.

djkazic:
Requested changes have been made, please review this when you can @nevets963

kollider/docker-compose.yml
```diff
@@ -18,9 +18,9 @@ services:
     environment:
       LND_IP: $APP_LIGHTNING_NODE_IP
       LND_ZMQ_SUB_ADDRESS: "tcp://kollider_ws_1:5556"
-  
+
   ws:
-    image: kolliderhq/kollider-ws-client:v1.0.6@sha256:05f3d1ddf9949b377788e442a1050457559291a50510f5d6dfa66a28c283a80a
+    image: kolliderhq/kollider-ws-client:v1.0.7@sha256:7136a6b25b452c93ed2762f5ab00e6a6d9a76d586c1eb1bdf962bee684135cd5
```
nevets963, line 23: It looks like this image is not built for `arm64` (https://hub.docker.com/layers/kolliderhq/kollider-ws-client/v1.0.7/images/sha256-7136a6b25b452c93ed2762f5ab00e6a6d9a76d586c1eb1bdf962bee684135cd5?context=explore).

I am getting this error:
```
[FATAL tini (7)] exec docker-entrypoint.sh failed: Exec format error
[FATAL tini (7)] exec docker-entrypoint.sh failed: Exec format error
[FATAL tini (6)] exec docker-entrypoint.sh failed: Exec format error
[FATAL tini (6)] exec docker-entrypoint.sh failed: Exec format error
```

kollider/docker-compose.yml
```diff
@@ -33,9 +33,9 @@ services:
       KOLLIDER_ZMQ_SUB_ADDRESS: "tcp://kollider_backend_1:5557"
       KOLLIDER_ZMQ_HEDGER_ADDRESS: "tcp://kollider_backend_1:5558"
       KOLLIDER_ZMQ_HEDGER_SUB_ADDRESS: "tcp://kollider_backend_1:5559"
-  
+
   web:
-    image: kolliderhq/kollider-lite-app:v1.0.6@sha256:f5bbbea86ed4d03c9a7c66527edbcda05228986ba5721e65c08a885edb74151a
+    image: kolliderhq/kollider-lite-app:v1.0.7@sha256:fd8b116d5b6e82686fd71f5d41206157e5a311b96587ff5ec14f01c24a13fd94
```
nevets963, line 38: And the same for this, re `arm64` support.

djkazic:
Hmm okay. They're supposed to be identical to 1.0.6, but I'll just rebuild them.

I pulled the 1.0.6 images and retagged them to 1.0.7 on my workstation but I guess that only pulled the arch matching my workstation.

djkazic:
@nevets963 rebuild complete

nevets963 (approved):
Thanks for the work to sort out the few issues here!

Merging now 🚀 

Do you think it would be possible to drop the requirement for the `umbrel.local` domain? Currently, it's not possible to access Kollider via Tor or Tailscale.

PR #352, opened 2023-01-03
Update Urbit to v1.15.0

mopfel-winrux:
Update to use the official urbit bianries. This will allow users to autoupdate

### DIFF
--- a/urbit/docker-compose.yml
+++ b/urbit/docker-compose.yml
@@ -2,18 +2,16 @@ version: "3.3"
 services:
   app_proxy:
     environment:
-      APP_HOST: $APP_URBIT_IP
+      APP_HOST: urbit_manager_1
       APP_PORT: 8090
+  
   manager:
     image: mopfelwinrux/urbit-umbrel:v1.15.0@sha256:f0f908dc1069b606e5fb45b407f9729f84b1f031932b4ddfcc7ca998d0fa90bc
     ports:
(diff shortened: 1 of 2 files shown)


nevets963 (commented):
@mopfel-winrux Thanks for the app update!

On looking closer, we can't accept an app that has auto-update functionality like what is described in the feature notes because it would circumvent our review process to check that the app is functional on both `arm64` and `amd64`, and perhaps more importantly, if the location of the binary (on the Internet) gets hacked, it means a compromised update could get rolled out without users knowing.

mopfel-winrux:
the autoupdate functionality was approved for `amd64` prior to this, and the pulling of the binary has been the core functionality for both `amd64` and arm64. Do I need to submit a PR request every time the urbit binary is updated?"

nevets963:
Yes. For each new version that you wish to roll out, we require a PR which would contain up to date Docker images for functionality that you wish to release.

mopfel-winrux:
I modified this to not autoupdate

nevets963 (approved):
@mopfel-winrux Thanks for reverting to versioned Docker images for Urbit updates.

I just patched up the `docker-compose.yml` file and added some release notes.

Tested on`arm64` and `amd64`.

This is approved from me. If you're happy with my changes, then I'll merge this in 🚀

mopfel-winrux:
looks good to me

Issue #346, opened 2022-12-31
New volume mapping gets "root" ownership when updating an app


woofbotapp:
This issue actually belongs to the "umbrel" repo (not "umbrel-apps"), but I'm writing it here because it is relevant to app-development.

I wanted to update my app to save its log files into a docker volume, so I could read it easily from outside (without `docker -it ... /bin/bash`). So I added the following files to my next version `docker-compose.yml`:

```
volumes:
  - ${APP_DATA_DIR}/data/server-logs:/app/logs
```

and also added a `data/server-logs/.gitkeep` file so the directory will be read for mounting.

When I updated the app in the dashboard UI (without uninstalling and re-installing it), the `data/server-logs` dir was created with "root:root" ownership, and was inaccessible inside the docker (the NodeJs process running under "node:node" user and saw its ownership as "root:root").

However, when I uninstalled the app and reinstalled it, the `data/server-logs` dir was created with "umbrel:umbrel" ownership and was accessible inside the docker (the NodeJs process saw its ownership as "node:node" like the other files).

It could be a problem with my Dockerfile or configurations, but I believe that the update mechanism at some point runs as root and does not "chown" the new files correctly (like a fresh installation).

al-lac:
Closing this as this is a currently known limitation that can be solved with a pre-start hook, see here for an example: https://github.com/getumbrel/umbrel-apps/blob/master/wikijs/hooks/pre-start

PR #353, opened 2023-01-04
Update WoofBot to v0.5.2

woofbotapp:
Tested on my raspberry pi.

This is a minor issue for those who update instead of uninstalling and reinstalling:
https://github.com/getumbrel/umbrel-apps/issues/346

(Minor issue, those logs are only needed for debugging purposes).

### DIFF
--- a/woofbot/docker-compose.yml
+++ b/woofbot/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8080
   
   web:
-    image: woofbot/woofbot:v0.4.6@sha256:dd6cf833c0f5ffc3003550671228ba610e66c300cb7889b72923e6070f67d5dc
+    image: woofbot/woofbot:v0.5.2@sha256:55647a3ea53433047f86f2ed2bd69369e935b92609af2840abceecd0a010be10
     restart: on-failure
     stop_grace_period: 1m
     environment:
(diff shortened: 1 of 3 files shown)


nevets963 (commented):
@woofbotapp Thanks for the app update!

Currently we have a whitelist of files/folders that we can copy during an app update and as you say, because `root` will perform the operation, it will create the folder on the host under `root` too. To solve this issue, we can create a `pre-start` hook (here is a rough example: https://github.com/getumbrel/umbrel-apps/blob/9e2756aee9088d2639a95c94eafc9cb9c896594a/bitcoin/hooks/pre-start) that checks for the existence of the `server-logs` folder, and if not, create it and `chown` to `1000:1000`.

I can help here, but I would need write access.

woofbotapp:
@nevets963 Done! The upgrade worked on my device.

Inspired from: https://github.com/getumbrel/umbrel-apps/blob/9e2756aee9088d2639a95c94eafc9cb9c896594a/nextcloud/hooks/pre-start

nevets963 (commented):
@woofbotapp Nice work with the pre-start hook - it works perfectly.

I forgot to mention that support for hooks came in with the manifest version `1.1`. Can you bump that to `1.1` in your `umbrel-app.yml`?

Otherwise, this is tested on `arm64` and `amd64`.

woofbotapp:
bumped

nevets963 (approved):
Tested one more time on my side.

Everything looks good. Merging now 🚀

PR #356, opened 2023-01-05
Migrate contents of Simple torrent downloads directory to Umbrel's shared downloads directory

### LINKED ISSUE #349: Should simple-torrent and transmission write to `UMBREL_ROOT` instead of `APP_DATA_DIR` so that jellyfin can read them?
Apps like simple-torrent download files to `${APP_DATA_DIR}/data/download`: https://github.com/getumbrel/umbrel-apps/blob/master/simple-torrent/docker-compose.yml#L18

But apps like jellyfin read files from `${UMBREL_ROOT}/data/storage/downloads`: https://github.com/getumbrel/umbrel-apps/blob/master/jellyfin/docker-compose.yml#L19

Wouldn't it be easier for jellyfin to consume content from simple-torrent and transmission if they wrote their data to  `${UMBREL_ROOT}/data/storage/downloads` instead of  `${APP_DATA_DIR}/data/download` so that they could share data between these apps?

If not, how am I supposed to share this data? Manually move files from folders accessed by one container to folders accessible to the other?

nevets963:
This PR adds a migration using a `pre-start` hook to migrate existing installations to use Umbrel's shared downloads directory.

Tested the following cases:
- Creating Umbrel's shared downloads directory if it does not exist
- Migrating an existing install with downloads (tested with 10GB dummy data)
- Migrating an existing install with no downloads
- App restart after migration to ensure it does not run again

Closes #349

### DIFF
--- a/simple-torrent/umbrel-app.yml
+++ b/simple-torrent/umbrel-app.yml
@@ -1,4 +1,4 @@
-manifestVersion: 1
+manifestVersion: 1.1
 id: simple-torrent
 category: Networking
 name: SimpleTorrent
@@ -42,5 +42,7 @@ path: ""
 defaultUsername: ""
 defaultPassword: ""
 torOnly: false
+permissions:
+  - STORAGE_DOWNLOADS
 submitter: Umbrel
 submission: https://github.com/getumbrel/umbrel/commit/b12c581ae14255c2087fc8e2f0e7014a351a363b
--- a/simple-torrent/umbrel-app.yml
+++ b/simple-torrent/umbrel-app.yml
@@ -29,7 +29,7 @@ description: >-
 
   - Magnet RSS subscribing supported
 releaseNotes: >-
-  - SimpleTorrent now utilizes the shared downloads folder that can be accessed by other apps on Umbrel, such as Jellyfin, FileBrowser, Transmission, Plex, and more.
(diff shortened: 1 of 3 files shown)


simple-torrent/umbrel-app.yml
```diff
@@ -28,6 +28,8 @@ description: >-
   - Protocol Handler to magnet:
 
   - Magnet RSS subscribing supported
+releaseNotes: >-
+  - Migrate downloads from local (app) directory to Umbrel's shared downloads directory
```
mayankchhabra, line 32:
```suggestion
  - SimpleTorrent now utilizes the shared downloads folder that can be accessed by other apps on Umbrel, such as Jellyfin, FileBrowser, Transmission, Plex, and more.
```

simple-torrent/umbrel-app.yml
```diff
@@ -28,6 +28,8 @@ description: >-
   - Protocol Handler to magnet:
 
   - Magnet RSS subscribing supported
+releaseNotes: >-
+  - SimpleTorrent now utilizes the shared downloads folder that can be accessed by other apps on Umbrel, such as Jellyfin, FileBrowser, Transmission, Plex, and more.
```
mayankchhabra, line 32:
```suggestion
  - SimpleTorrent now utilizes the shared downloads folder that can be accessed by other apps on Umbrel, such as Jellyfin, File Browser, Transmission, Plex, and more.
```

PR #357, opened 2023-01-06
Bump nolooking v0.3.0-alpha: Send PayJoin

DanGould:
- Send PayJoin support

Question: Do we need to re-state envvars in `docker-compose.yml`?


nolooking/docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_WHITELIST: "/pj"
   
   server:
-    image: chaincase/nolooking:v0.2.2-alpha@sha256:e4a95443a05b44659f55a20c592142d59a4d8f54830a71f4c3d22a3a6bf88801
+    image: chaincase/nolooking:v0.3.0-alpha@sha256:733025b09682e219d06d14b27d487b8af4ae78590e625867460be8229eec8d3e
```
nevets963, line 11: It looks like the Docker image digest is wrong. Here is the digest for `amd64` only. To get the multi-arch digest, you can run `docker pull <image>:<tag>` and it will display the digest at the bottom once complete.

In this case it should be: `sha256:ba879c089e6609cb85869081c18bc36086984276c5cd73bce2a24ff5f53cc1f0`

nolooking/docker-compose.yml
```diff
@@ -23,3 +23,5 @@ services:
       DEVICE_DOMAIN_NAME: $DEVICE_DOMAIN_NAME
       APP_PASSWORD: $APP_PASSWORD
       APP_HIDDEN_SERVICE: $APP_HIDDEN_SERVICE
+      TOR_PROXY_IP: $TOR_PROXY_IP
+      TOR_PROXY_PORT: $TOR_PROXY_PORT
```
nevets963, line 27: > Question: Do we need to re-state envvars in docker-compose.yml?

Yes; Umbrel exposes a number of environment variables, but it's the developer's choice whether they're needed and you would need to be set/use them with the docker services like you've done above.

nevets963 (commented):
@DanGould Thanks for the PR to update nolooking!

I've left some feedback below - once addressed, I'll give this a test.

nolooking/docker-compose.yml
DanGould: 💯

nolooking/docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_WHITELIST: "/pj"
   
   server:
-    image: chaincase/nolooking:v0.2.2-alpha@sha256:e4a95443a05b44659f55a20c592142d59a4d8f54830a71f4c3d22a3a6bf88801
+    image: chaincase/nolooking:v0.3.0-alpha@sha256:ba879c089e6609cb85869081c18bc36086984276c5cd73bce2a24ff5f53cc1f0
     restart: on-failure
```
nevets963, line 12: I had an issue on 1 of my Umbrels with LND causing nolooking to crash. It looks like when you exit, it's exiting with a success status code (`0`) which means the container's restart policy (`no-failure`) will not be applied, and therefore won't automatically restart the container when something goes wrong.

You can see from the error I got, that LND was still booting, and so a auto-restarting container would be quite effective. You could also however, solve this in your app, but having a 'retry loop' for certain errors.
```
Running nolooking
Error: Lnd(VersionRequestFailed(Status { code: Unknown, message: "the RPC server is in the process of starting up, but not yet ready to accept calls", metadata: MetadataMap { headers: {"content-type": "application/grpc"} }, source: None }))
nolooking exited!
```

DanGould: Aw rats. Must be because the Dockerfile CMD starts nolooking using `run_nolooking` shell script instead of calling the binary directly, which exits with a non-zero code in the scenario you described. Building a new docker image to fix!

nmfretz:
Hey @DanGould. I'm subbing out @nevets963 for review, so please feel free to tag me if/when you start working more on this again.

DanGould:
ok ⚡️

DanGould:
I've updated this PR to draw from the latest release. I believe I've also resolved all of the feedback here:

 I used `docker pull nolooking:0.4.0-alpha` in order to get the multi-arch digest. I believe this also fixes #359 by replacing the Entrypoint with:

```docker
CMD -- --bind-ip=0.0.0.0 --bind-port=4444 --lnd-address=https://$LND_HOST:$LND_GRPC_PORT --lnd-cert-path=$TLS_FILE --lnd-macaroon-path=$MACAROON_FILE --endpoint=http://$APP_HIDDEN_SERVICE --tor-proxy-address=$TOR_PROXY_IP:$TOR_PROXY_PORT
```

nmfretz (commented):
Awesome, thanks for doing this @DanGould. I just tried it out, but the app does not start and I get the following error:

```
$ sudo docker logs -f nolooking_server_1
Error: Invalid configuration: Configuration parameter 'bind_port' not specified.
```

I also left a formatting suggestion below for the description and release notes.

nmfretz:
Hi @DanGould, just checking in to see if you are still working on this and want to keep this PR open?

DanGould:
Yes, working on it

nmfretz:
Awesome, will keep open. Thanks Dan.

tlindi:
Any progress?

This is still pending: https://github.com/getumbrel/umbrel-apps/issues/359

nmfretz:
@DanGould are you okay if I close this old PR for organizational purposes? We can reopen again when you have time to work on it.

DanGould:
Yes, thanks for staying on top of it when I kept pushing it back

nmfretz:
Thanks @DanGould! Please ping me whenever you're ready 🤝

PR #368, opened 2023-01-13
Update Urbit to v1.16

yapishu:
- Updated Urbit to official v1.16 binary: https://github.com/urbit/vere/releases

### DIFF
--- a/urbit/umbrel-app.yml
+++ b/urbit/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: urbit
 category: Networking
 name: Urbit
-version: "1.15"
+version: "v1.16"
 tagline: Run Urbit on your Umbrel
 description: >-
   Urbit is a personal server for self-sovereign personal & networked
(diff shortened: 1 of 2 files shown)


urbit/umbrel-app.yml
```diff
@@ -35,18 +35,4 @@ torOnly: false
...
-  - Improves bit-slice performance (for noun de/serialization and many atom jets) by ensuring that implementation inner loops can be vectorized.
-
-  - Adds the vile command, for exporting the keyfile from a ship
-
-  - Adds the %xray hint, for printing the bytecode of a given expression (ty ~topfet-parmed!)
-
-  - Adds the eval command, for running hoon (from stdin) without booting a ship (ty ~mopfel-winrux!)
-
-  - Adds a better error message when a ship is already running as root (ty ~dinleb-rambep!)
+Updated Urbit binary to v1.16
```
nevets963, line 38: Would you be able to populate the release notes/changes for Urbit v1.16. This would give users some context (in the Umbrel UI) as to what has changed with this app update.

nevets963 (commented):
@yapishu Thanks for PR to update Urbit 😄 

I've left some minor feedback re the release notes.

urbit/umbrel-app.yml
yapishu: Sure thing

nevets963: Could you fix the formatting? In order to be valid yaml, this block needs to be indented. And also due to the usage of `>-` with this blocks, we need empty lines between lines of text to actually get new lines. I would do this myself, but I can't checkout this PR due to not having permissions.

yapishu: No problem 👍

nevets963 (approved):
@yapishu Thanks again for the PR to update Urbit and for the requested changes.

Merging now 🚀

PR #371, opened 2023-01-15
Fix Urbit retrieval URL

yapishu:
- Urbit project moved binary to new GH url (urbit/vere); fixing download URL

### DIFF
--- a/urbit/umbrel-app.yml
+++ b/urbit/umbrel-app.yml
@@ -35,10 +35,9 @@ torOnly: false
 submitter: ~mopfel-winrux
 submission: https://github.com/getumbrel/umbrel/pull/1246
 releaseNotes: >-
-  Updated Urbit binary to v1.16
-  
-  Arvo 417K
-  
-  Vere 1.16
-  
-  This release is functionally equivalent to v1.15 but is built in the new Vere repo using Bazel instead of Nix with a new and improved release pipeline. It marks the successful extraction of the runtime from the Urbit monorepo. It also adds a new AArch64-specific macOS build and removes the Windows build.
+Updated Urbit binary to v1.16
+
+- Fix URL with vere repo
+
+This release is functionally equivalent to [v1.15](https://github.com/urbit/urbit/releases/tag/urbit-v1.15) but is built in the [new Vere repo](https://github.com/urbit/vere) using [Bazel](https://bazel.build) instead of [Nix](https://nixos.org) with a new and improved release pipeline. It marks the successful extraction of the runtime from the [Urbit monorepo](https://github.com/urbit/urbit). It also adds a new AArch64-specific macOS build and removes the Windows build.
+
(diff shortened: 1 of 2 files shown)


urbit/umbrel-app.yml
```diff
@@ -35,10 +35,9 @@ torOnly: false
...
-  Arvo 417K
-  
-  Vere 1.16
-  
-  This release is functionally equivalent to v1.15 but is built in the new Vere repo using Bazel instead of Nix with a new and improved release pipeline. It marks the successful extraction of the runtime from the Urbit monorepo. It also adds a new AArch64-specific macOS build and removes the Windows build.
+Updated Urbit binary to v1.16
+
+- Fix URL with vere repo
+
+This release is functionally equivalent to [v1.15](https://github.com/urbit/urbit/releases/tag/urbit-v1.15) but is built in the [new Vere repo](https://github.com/urbit/vere) using [Bazel](https://bazel.build) instead of [Nix](https://nixos.org) with a new and improved release pipeline. It marks the successful extraction of the runtime from the [Urbit monorepo](https://github.com/urbit/urbit). It also adds a new AArch64-specific macOS build and removes the Windows build.
```
nevets963, line 42: Can you fix the formatting - this needs to be indented.

nevets963 (commented):
@yapishu Inorder for this to register as an app update within Umbrel, you'll need to bump the `version` attribute inside the app manifest (`umbrel-app.yml`). Because this is the same version, I would suggest something like: `v1.16-hotfix1`.

I would also strongly encourage you not to update Docker images with the same tag otherwise we'll have users currently trying to update but it'll fail due to the wrong image digest.

urbit/umbrel-app.yml
yapishu: Apologies! Both fixed

urbit/umbrel-app.yml
```diff
@@ -1,44 +1,43 @@
...
+deterministicPassword: true
+torOnly: false
+submitter: ~mopfel-winrux
+submission: https://github.com/getumbrel/umbrel/pull/1246
+releaseNotes: >-
+  - Updated Urbit binary to v1.16
+
+  - Fix URL with vere repo
+
+  This release is functionally equivalent to [v1.15](https://github.com/urbit/urbit/releases/tag/urbit-v1.15) but is built in the [new Vere repo](https://github.com/urbit/vere) using [Bazel](https://bazel.build) instead of [Nix](https://nixos.org) with a new and improved release pipeline. It marks the successful extraction of the runtime from the [Urbit monorepo](https://github.com/urbit/urbit). It also adds a new AArch64-specific macOS build and removes the Windows build.
```
nevets963, line 42: Markdown is not currently supported and should be plaintext; could you change the release notes to be more friendly when this is plaintext?

Would you also know why git is marking this entire file as changed?

nevets963 (approved):
@yapishu Looks good to me!

Merging now 🚀

PR #373, opened 2023-01-16
Fix urbit binary extraction

yapishu:
In addition to changing the repo, the new releases have a new naming scheme that breaks the previous binary download script ( https://github.com/mopfel-winrux/urbit-umbrel/commits/main/install-urbit.sh ). Don't mean to be annoying! Many things broke with the new release scheme.


nevets963 (approved):
@yapishu This is approved from me.

Merging now 🚀

PR #391, opened 2023-02-02
Update BTCPay Server to 1.7.5

Kukks:


nevets963 (commented):
@Kukks Thank you for the PR to keep BTCPay Server up to date within Umbrel!

Could you also update the `version` property and `releaseNotes` inside the app manifest (`umbrel-app.yml`)? After that, I'll give this a test.

nevets963 (approved):
@Kukks @JimSB Thanks for the PR and efforts to update BTCPay Server.

Merging now 🚀

PR #401, opened 2023-02-08
Samourai server remote access fix

nevets963:


santochibtc:
I have updated the app and now I can't connect the Whirpool GUI
"Authentication failed: Unable to connect to wallet backend: http://samourai-server_nginx_1:80/v2/"

PR #402, opened 2023-02-10
Update BTCPay Server to 1.7.10

Kukks:


nevets963 (commented):
@Kukks Thanks for the app update PR! I'm currently testing this on my Umbrel servers. Strangely on 1 but not the other, I am getting this error: `429 - Too Many Requests` frequently making this not usable. Has something changed with rate limiting/throttling with BTCPay Server?

Kukks:
How do you get that error? I don't believe anything has changed

lukechilds:
Just throwing an idea out there but could this be BTCPay getting throttled/banned by Bitcoin Core P2P and then catching that error and bubbling it up as a 429? I think we can set something like `whitelist=<internal-app-private-subnet>` on Bitcoin Core (or maybe `whitebind` I forget the difference) to make sure Umbrel apps never get banned from the P2P port.

Kukks:
> Just throwing an idea out there but could this be BTCPay getting throttled/banned by Bitcoin Core P2P and then catching that error and bubbling it up as a 429? I think we can set something like `whitelist=<internal-app-private-subnet>` on Bitcoin Core (or maybe `whitebind` I forget the difference) to make sure Umbrel apps never get banned from the P2P port.

Nah we don't do that. Where does the error pop up?  Doies it show the btcpay UI for the error? What page? Would really like to get this shipped this week :)

nmfretz:
Hey @Kukks, I can help get an update out ASAP. There is now another PR open for version 1.7.12 [here](https://github.com/getumbrel/umbrel-apps/pull/426). I have included the `NBXPLORER_BTCHASTXINDEX` and `BTCPAY_TORSERVICES: "btcpayserver:$APP_HIDDEN_SERVICE"` that you added here, as well as the tweaked wording to the description.

Could you please take a look at PR https://github.com/getumbrel/umbrel-apps/pull/426 and let me know if anything jumps out at you that looks incorrect.  I have tested on amd64 and arm64 without issue.

nmfretz:
Thanks for this PR. Closing in favour of new PR for major BTCPay Server release tomorrow.

PR #411, opened 2023-02-12
App Submission: Snort

v0l:
# App Submission

### App name
Snort

### 256x256 SVG icon
https://void.cat/d/QzA8i39F3Tqr986ouuLDKL

### Gallery images

https://void.cat/d/EL7xnxC4Fab123kUVB5w1e
https://void.cat/d/DenpDVYyWhvBKDQhzbMGTd
https://void.cat/d/RxUH2Ve3KKU6BspnCQF3xP

The same images but 2x size:
https://void.cat/d/WP56AV1Zz3qg3RKLbQ4tSd
https://void.cat/d/Ehq6hGmTSEs2RTJLBmv9M7
https://void.cat/d/4iS5jssVAL9sf1bsLxALUX

### I have tested my app on:

### DIFF
--- a/snort/docker-compose.yml
+++ b/snort/docker-compose.yml
@@ -1,0 +1,11 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: snort_web_1
+      APP_PORT: 80
+
+  web:
+    image: ghcr.io/v0l/snort:v0.1.4
+    init: true
--- a/snort/docker-compose.yml
+++ b/snort/docker-compose.yml
@@ -7,5 +7,5 @@ services:
       APP_PORT: 80
 
   web:
-    image: ghcr.io/v0l/snort:v0.1.4
+    image: ghcr.io/v0l/snort:v0.1.4@sha256:734f74393e2560e625bad9e2485ccccfa6af53caab64a67d971380244cfc54dd
     init: true
--- a/snort/umbrel-app.yml
+++ b/snort/umbrel-app.yml
@@ -1,0 +1,39 @@
+manifestVersion: 1
+id: snort
+category: Social
+name: Snort
+version: "v0.1.4"
+tagline: Fast Nostr UI
+description: >-
+  Snort is a client for Nostr that is built using React, designed to be fast and lightweight while still offering a wide range of features. 
+  Our focus is on providing users with an easy-to-use interface to interact with the Nostr network, making it accessible to all types of users. 
+  With its clean and intuitive design, Snort makes it simple to navigate the Nostr network and access its features, allowing users to efficiently manage their data.
+releaseNotes: >-
+  ## Added
+
+  - LNC + LNDHub wallet support
+
+  - Magnet link previews
+
+  - NIP-06 Key generation
+
+  - NIP-58 Badges
+
+  - Wavlake player embeds
+
+  - FastZaps
+developer: Snort
+website: https://snort.social
+dependencies: []
+repo: https://github.com/v0l/snort
+support: https://t.me/snort_social
+port: 52027
+gallery:
+  - 1.png
+  - 2.png
+  - 3.png
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+submitter: v0l
+submission: https://github.com/getumbrel/umbrel-apps/pull/411


snort/umbrel-app.yml
```diff
@@ -0,0 +1,28 @@
+manifestVersion: 1
+id: snort
+category: Social
+name: Snort
+version: "v0.1.1"
+tagline: Fast Nostr UI
+description: >-
+  Snort is a Nostr UI built in React
```
nevets963, line 8: Would you be able to add a little extra text about why Snort is so awesome? 😄

nevets963 (commented):
@v0l Thanks for the amazing speed here to build the Docker images and submit this app submission. We're now super close 🚀 

I've left some minor feedback. Once that is addressed, I'll give this a test on my arm and Intel Umbrel servers and if that is good, I can speak with our graphics designer to get graphics made for the App Store. Does Snort already have a logo we can use? And would you have some preferred screenshots that we should use with the app listing in the store?

v0l:
There is a bunch of stuff which doesnt work very well on non-secure connections, would it be possible to add self-signed certificate? lol

lukechilds:
@v0l just wanted to say great work on Snort, I'm a huge fan! Can't wait to see this in the Umbrel app store!

v0l:
> There is a bunch of stuff which doesnt work very well on non-secure connections, would it be possible to add self-signed certificate? lol

For one our current lib for all crypto operations uses `window.subtle` which is only available on `https`

Syturio:
We need this!

v0l:
@lukechilds do any other apps use self signed certs? I want to keep using the same lib for event signing and hmac functions but the use `window.subtle` which isnt available on http

lukechilds:
@v0l Ahhh, we don't recommend using self signed certs, it's really poor UX (scary initial errors) and encourages poor security best practises to non-technical users (blindly clicking through SSL warnings). Unfortunately there aren't currently any good workarounds for secure communication on local networks that "just work" without those mentioned issues, it's something we have some ideas on how to workaround at the Umbrel level to resolve this for all apps but unfortunately that isn't yet implemented.

Perhaps as workaround that works today with good UX it would be possible for you to detect http and load a webcrypto polyfill in that case? That would hopefully allow you to get Snort working on Umbrel while still keeping your existing crypto implementation.

v0l:
Ok i think it's ready now @lukechilds @nevets963, sorry it took so long to get back to this.

nmfretz:
Amazing, thanks @v0l. I can take over from @nevets963 from here. Taking a look now.

snort/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: snort_web_1
+      APP_PORT: 80
+
+  web:
+    image: ghcr.io/v0l/snort:v0.1.4
```
nmfretz, line 10: I see the newest version tag is v.0.1.3 here: https://github.com/v0l/snort/pkgs/container/snort

Are we just waiting on 0.1.4 to be pushed?

If you want to go with 0.1.3 you can use this sha for multi-arch:

```suggestion
    image: ghcr.io/v0l/snort:v0.1.3@sha256:bab2b88b09e7894152e530b4bee06df532cdff79389a60eedcc197675bc63fb6
```

nmfretz: Ah, I see that 0.1.4 is the umbrel release candidate. I'll wait until it is available.

nmfretz: I have tested new 0.1.4 image on arm64 and amd64. Can you please add the SHA256 hash:

```suggestion
    image: ghcr.io/v0l/snort:v0.1.4@sha256:734f74393e2560e625bad9e2485ccccfa6af53caab64a67d971380244cfc54dd
```

nmfretz:
Alright, I have tested on arm64 and amd64. Had some fun zapping and watching events show up on my Umbrel Nostr Relay.

Once the multi-arch image hash is [added](https://github.com/getumbrel/umbrel-apps/pull/411#discussion_r1139475828) this is good to go. Fantastic work @v0l! And thank you for the high quality gallery images and logo.

nmfretz:
Perfect. Going live now.

nmfretz:
FYI, if you're ever looking for an SVG badge to show that your snort is available on the umbrel app store, you can grab them using these links. If you update the icon at some point, then these will update automatically:

https://apps.umbrel.com/api/app/snort/badge-dark.svg
https://apps.umbrel.com/api/app/snort/badge-light.svg
https://apps.umbrel.com/badge-dark.svg
https://apps.umbrel.com/badge-light.svg

PR #415, opened 2023-02-14
Update Urbit to v1.20

yapishu:
- Updated Urbit to official v1.20 binary: https://github.com/urbit/vere/releases


nevets963 (approved):
@yapishu Thanks for the app update PR!

Merging now 🚀

PR #418, opened 2023-02-15
[GitHub Actions] Auto Updater

ShonP40:
A GitHub action for automatically updating all the apps on this repo daily.

I have a working example of this on my own [Umbrel apps repo](https://github.com/ShonP40/umbrel-apps).

The action fetches the latest version of each app from GitHub’s API, and then uses that data to get its Docker image digest from Docker Hub’s API.
There are failsafes as well in case of an API outage.

Each app will need to have a basic `.env` file in its folder that specifies the ID (same as the folder name), a friendly name, GitHub repo (`org/repo`) and Docker repo (`org/image`).

I will create an env file for each app in this repository once you guys confirm that this action will be added to not waste time (the action can also skip apps that don’t have an `.env` file).

Requirements for app devs for this to work:
- Have a GitHub repo with each update published as a GitHub release & have the latest release flagged as `Latest`
- Have a Docker image with versioned releases (pretty sure that this is already a requirement for Umbrel itself)

### DIFF
--- a/.github/workflows/auto-update.yml
+++ b/.github/workflows/auto-update.yml
@@ -1,0 +1,72 @@
+name: Auto Updater
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  workflow_dispatch:
+
+jobs:
+    generateInputPaths:
+        name: List all app folders
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v3
+
+            - name: Generate matrix with all the apps in this repo
+              id: set-matrix
+              run: echo "matrix=$(ls -d ./*/ | jq --raw-input --slurp --compact-output 'split("\n")[:-1]')" >> $GITHUB_OUTPUT
+        outputs:
+            matrix: ${{ steps.set-matrix.outputs.matrix }}
+
+    update:
+        needs: generateInputPaths
+        runs-on: ubuntu-latest
+        strategy:
+            max-parallel: 1
+            fail-fast: false
+            matrix:
+                path: ${{ fromJson(needs.generateInputPaths.outputs.matrix) }}
+
+        steps:
+            - uses: actions/checkout@v3
+
+            - name: Check file existence
+              id: check_files
+              uses: andstor/file-existence-action@v2
+              with:
+                files: ${{ github.workspace }}/${{ matrix.path }}/.env
+                ignore_case: false
+            
+            - name: Load .env file
+              if: steps.check_files.outputs.files_exists == 'true'
+              uses: xom9ikk/dotenv@v2
+              with:
+                path: ${{ github.workspace }}/${{ matrix.path }}
+
+            - name: Fetch the latest version
+              id: get_latest
+              if: steps.check_files.outputs.files_exists == 'true'
+              run: echo "latest=$(curl -sL https://api.github.com/repos/${{ env.GITHUB }}/releases/latest | jq -r ".tag_name")" >> $GITHUB_OUTPUT
+            
+            - name: Fetch the latest digest
+              id: get_hash
+              if: ${{ steps.get_latest.outputs.latest != 'null' && steps.check_files.outputs.files_exists == 'true' }}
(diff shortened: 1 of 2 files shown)


ShonP40:
1. This action was only really meant for apps that don't have any extra containers bundled in their `docker-compose` file
2. Keeping it on a PR level will only overwhelm maintainers & not provide any real world benefits than just expecting users to manually push PRs to update their apps
3. The point is to always have the latest stable release of each app available to users as an update. Limiting this to only run weekly will skip a few updates per each app (take Home Assistant as an example, they can sometimes push important bug fixes multiple times a week).
4. As for changelogs, Umbrel should implement a system for scraping the latest changelogs for each app from GitHub's API if we want to always keep this up to date (most app submitters don't even bother updating the changelog on their submissions).
5. Consider this repo an App Store, it simply provides downloads for existing apps. It's up to the user to look at changelogs to know about any breaking changes. You wouldn't want to just prevent new users from downloading the latest version of an app because existing users don't want to migrate over.
In the end its the user's choice wether or not to upgrade the app on their own Umbrel instance, updating it on this repo will not force update it on your Umbrel instance.
6. As the main Umbrel repo, I think that all apps here should be kept up to date to not start imposing security risks. If you want to run a specific version of an app, you can always host your own repo with it.

ShonP40:
If your main concern is crypto wallets, we can just not add the config `.env` file to them and they'll just be skipped during the auto update process

BeauAgst:
Hey @ShonP40,

I think you're right about overwhelming the maintainers, I was trying to look at possible alternatives that didn't mean continuous (potentially breaking) updates. Maybe there's a halfway-house here that accommodates your 3rd point as well?
* Major updates provide a PR
* Minors or patches could be automatically handled

This does put some onus on the developer of these apps to not produce any breaking changes still. I think 5 will be the point of contention here. I can't speak on behalf of the Umbrel team, but I wouldn't imagine them wanting to allow that to happen. An app store traditionally doesn't offer updates to apps that could break them, cause data loss or other destructive actions. You may not be forced to update, but there would probably need to be a big old warning label saying "if you update this app, X, Y, Z".

Let's see what the Umbrel team say!

ShonP40:
Just saying, but you don't have to mention me in every message lol. I get notifications from this anyways.

Also, 99% of the time, app devs will not introduce breaking changes to the `docker-compose` file, so simply letting the auto updater do its job will not break anything.

But sure, lets wait and see what the Umbrel team suggests.

nmfretz:
Thanks for this PR @ShonP40. Appreciate the initiative. We definitely want to use GitHub actions to automate some of the app-update process. Things will become increasingly overwhelming as more and more apps are added to the App Store.

At this time we don't want to do a full-blown automatic update to the App Store without (1) manually checking for breaking changes (or deprecations) of apps and associated components like databases, (2) checking that the updates work on both 64-bit ARM and x86 architectures, and (3) doing some manual tests on a production Umbrel instance. One concern with an automatic update to the App Store is that if a malicious actor gains access to a 3rd-party apps repo or if the location of the binary gets hacked, a malicious update could make its way to the App Store without us knowing in time.

That being said, GitHub actions that make the app update process easier would be very beneficial. For example, actions to:

- Check that an app has a unique port on Umbrel
- Check that docker image digests are multi-arch
- Check whether docker images have implicit volumes defined
- Run the previous version of an app on Umbrel (on both 64-bit ARM and x86), and then update it and check to see that it still runs on Umbrel.

nmfretz:
Closing this PR for now. Thanks for all these suggestions, and please feel free to reopen if you have any interest in these:
> That being said, GitHub actions that make the app update process easier would be very beneficial. For example, actions to:
> - Check that an app has a unique port on Umbrel
> - Check that docker image digests are multi-arch
> - Check whether docker images have implicit volumes defined
> - Run the previous version of an app on Umbrel (on both 64-bit ARM and x86), and then update it and check to see that > - it still runs on Umbrel.

ShonP40:
Don’t really have time to work on this at the moment.
So unless anyone else wants to redo the action, this process will probably just stay fully manual.

PR #421, opened 2023-02-18
Update PiHole to 2023.02.2

Xosten:
Hi,

I had noticed that the changelogs and versioning in `umbrel-app.yml` didn't follow the `docker` github repository of PiHole. I was wondering if there would be a reason?

PiHole Docker Repo: https://github.com/pi-hole/docker-pi-hole/

This PR Updates to the latest version and switched the versioning to follow PiHole's Docker Github versioning.
Tested on AMD VM

### DIFF
--- a/pi-hole/docker-compose.yml
+++ b/pi-hole/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   server:
-    image: pihole/pihole:2023.02.0@sha256:65047a58be709cbfee9586ed224e7e1fd49233be4074e1659baee335105093f2
+    image: pihole/pihole:2023.02.2@sha256:9abbf1c218f32a4084e614150a44714f046f73c40a9d2889a0e6edf01ff0a387
     # Pi-hole doesn't currently support running as non-root
     # https://github.com/pi-hole/docker-pi-hole/issues/685
     # user: "1000:1000"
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @Xosten. I have tested on amd64 and arm64 with no issues.

I'm a bit torn on what to do for versioning for Pi-hole going forward, and would be curious to hear your thoughts. It looks like the [docker-pi-hole](https://github.com/pi-hole/docker-pi-hole) versions (e.g., 2023.02.02) are just the dates of the docker image/tags and don't necessarily have anything to do with the versions of Pi-hole Core, FTL, etc. As far as I can tell it looks like the version in the Umbrel app store right now (5.18.4) is referring only to the [AdminLTE dashboard](https://github.com/pi-hole/AdminLTE), which seems like maybe not the right thing to do.

Maybe we should change to [docker-pi-hole](https://github.com/pi-hole/docker-pi-hole) versions as you suggest (2023.02.02) and then in the `releaseNotes` section we indicate which versions of the core Pi-hole components are included. So for example, 2023.02.02 contains core v5.15.5, FTL v5.21, AdminLTE v5.18.4.

Xosten:
Thanks @nmfretz for sharing your thoughts regarding my suggestion, I really appreciate.

I agree with your idea to add in the `releaseNotes` the current/new versions of the Pi-Hole components. I'll update the `umbrel-app.yml` to reflect this.

Thanks for taking the time sharing this :).

nmfretz:
Awesome, thanks for that. I ended up adding an explanation of the version change at the top of the release notes.

This is now live in the app store. Thanks again @Xosten!

Xosten:
Awesome! Thanks man!

PR #426, opened 2023-02-20
Update BTCPay Server to 1.7.12

JimSB:
I believe I formatted everything correctly for umbrel-app.yml and hopefully there are no other issues :sweat_smile:

### DIFF
--- a/btcpay-server/umbrel-app.yml
+++ b/btcpay-server/umbrel-app.yml
@@ -63,6 +63,6 @@ releaseNotes: >
   - Fix typo in error message when vault is opening a wallet from the wrong chain (#4640) @eltociear
 
 
-  For release notes of versions between 1.7.5 and 1.7.12, see https://github.com/btcpayserver/btcpayserver/tags
+  For release notes of versions between 1.7.5 and 1.7.12, see https://github.com/btcpayserver/btcpayserver/releases
 submitter: Umbrel
 submission: https://github.com/getumbrel/umbrel/pull/353
--- a/btcpay-server/umbrel-app.yml
+++ b/btcpay-server/umbrel-app.yml
@@ -14,10 +14,10 @@ description: >-
   Payments with BTCPay Server go directly to your wallet, which increases the privacy and security. Your private keys are never uploaded to your Umbrel. There is no address re-use, since each invoice generates a new address deriving from your xpubkey.
 
 
-  You can not only to attach an unlimited number of stores and use the Lightning Network but also become a payment processor for others. Thanks to the apps built on top of it, you can use BTCPay to receive donations, start a crowdfunding campaign or have an in-store Point of Sale.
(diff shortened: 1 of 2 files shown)


btcpay-server/umbrel-app.yml
```diff
@@ -34,20 +34,24 @@ path: ""
...
-  - Greenfield: Lightning addresses API (#4546) @Kukks
+  Update recommended for shared instances.
 
   
   Bug fixes
 
-  - Fix several HTML injections (#4545) @NicolasDorier
-  - BIP21: Uppercase addresses only in QR, not in payment URL (#4553) @dennisreimann
-  - Checkout v2: UI fixes (#4552) @dennisreimann
+  - Fix second order XSS: Harden file type input (#4635) @NicolasDorier
```
nmfretz, line 42: @JimSB - a newline is needed between each bullet in order to format the releaseNotes section so that there is one bullet per line

nmfretz:
Nice job @JimSB! I have tested both ARM and x86 and everything looks good.

One small thing to change is that newlines are needed between each `-` bullet point in the release notes section so that they show as one bullet per line. Once you make that change we are good to go!

This is what it looks like in the app store without the newlines:

image

btcpay-server/umbrel-app.yml
JimSB: added a new commit with new lines! ty!

nmfretz:
@Kukks - would you recommend that we include any relevant info in the `releaseNotes` section from updates between 1.7.5 and 1.7.12? For example, 1.7.8 that provides a migration path for legacy MySql and SQLite installations.

Kukks:
> @Kukks - would you recommend that we include any relevant info in the `releaseNotes` section from updates between 1.7.5 and 1.7.12? For example, 1.7.8 that provides a migration path for legacy MySql and SQLite installations.

I think it is unrealistic to maintain a dedicated changelog for specific deployments. Is there a way to link to external ones instead?

nmfretz:
> I think it is unrealistic to maintain a dedicated changelog for specific deployments. Is there a way to link to external ones instead?

Right now we can include a url in the release notes to direct people to an external source. It won't hyperlink, but I can look into adding that functionality to reduce the friction for a user to view logs. 

I added a small note at the bottom of the release notes, with a url to [github releases](https://github.com/btcpayserver/btcpayserver/releases). 

image

Kukks:
That's great, thank you!

Kukks:
We will do another release of btcpay most likely tomorrow. Do you think we
should wait or just do another pr tomorrow?

Kukks:
@nmfretz  The release was just delayed till next week. Let's go for this one :)

nmfretz:
Alright, it's live. Thanks again @Kukks and @JimSB!

Kukks:
Awesome, thank you so much for all your help!

PR #427, opened 2023-02-21
Update Urbit to v1.21

yapishu:
- Updated Urbit to official v1.21 binary: https://github.com/urbit/vere/releases


nmfretz:
Thanks for this update @yapishu. Much appreciated! I have tested on both amd64 and arm64 and everything is good.

It's now live!

PR #428, opened 2023-02-22
Update Gitea to 1.18.5

JimSB:
I'm not sure if we need the MariaDB hash as well, I used the 10.11.2 tag from here: [MariaDB Docker Hub](https://hub.docker.com/layers/library/mariadb/10.11.2/images/sha256-dc249cd0b7e41739fc4c5dbfcfda160afc7d6d1e40ce14ffbe2b62c285dd06ef?context=explore)

### DIFF
--- a/gitea/docker-compose.yml
+++ b/gitea/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   server:
-    image: gitea/gitea:1.16.8-rootless@sha256:8a6ad22694c76122ae075af1153a544577724c45cb45af44963aa78dc179cc46
+    image: gitea/gitea:1.18.5-rootless@sha256:40d4a630fd90cb41bea245034a1a86e3ba7220e8f6d7e5e0d0a50583df2e1a16
     user: "1000:1000"
     restart: on-failure
     ports:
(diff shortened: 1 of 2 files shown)


gitea/docker-compose.yml
```diff
@@ -31,7 +31,7 @@ services:
       GITEA__database__PASSWD: "moneyprintergobrrr"
 
   db:
-    image: mariadb:10.5.12@sha256:dfcba5641bdbfd7cbf5b07eeed707e6a3672f46823695a0d3aba2e49bbd9b1dd
+    image: mariadb:10.11.2@sha256:dd0f492b6b6e7bb4aa707181b799d4efe42cb3a9f6012ec3dbaf326d402151e8
```
nevets963, line 34: Using a newer DB version will require careful testing to ensure this version of MariaDB works with existing installations that would have been using version `10.5.12`.

nmfretz: @JimSB and @nevets963 I have tested this on existing gitea installations (amd64 and arm64) and had no issues related to the mariadb update. All my repos were maintained, and I was able to push new commits.

nmfretz:
Thanks for this @JimSB! I have tested this on existing Gitea installations (amd64 and arm64) with no issues related to the MariaDB update. All my repos were maintained, and I was able to push new commits after the update. I also looked through both the Gitea and MariaDB release notes for all the releases between the current umbrel versions and the new release versions and didn't see anything concerning.

Going live now.

PR #437, opened 2023-02-28
Update Urbit to v1.22

yapishu:
- Updated Urbit to official v1.22 binary: https://github.com/urbit/vere/releases


nmfretz:
Thanks for this @yapishu. I have tested on amd64 and arm64. Going live now!

Issue #397, opened 2023-02-05
Outdated Samourai Server app (6 months)


louneskmt:
Hi, the Samourai Server app is (very) outdated: currently 1.16.1 (6 months ago), latest is [1.18.1](https://code.samourai.io/dojo/samourai-dojo/-/releases/v1.18.1). An update would really be appreciated.

mayankchhabra:
Hey, thanks for pointing out, @louneskmt! We'll get this done on a priority 🤝

louneskmt:
Is Whirlpool broken just for me? I can't connect remotely using Whirlpool GUI. I'm thinking it may be #310 that broke it, not sure.

<img width="325" alt="image" src="https://user-images.githubusercontent.com/38882571/218298679-bb44cc89-db3a-4b5c-9a72-1c15e6b8c3c6.png">

<details>
<summary>Whirlpool logs</summary>

```
whirlpool_1   | 2023-02-12 07:30:44.503 ERROR 12 --- [nio-8898-exec-6] c.s.wallet.api.backend.BackendApi        : 
whirlpool_1   | 
whirlpool_1   | com.samourai.wallet.api.backend.beans.HttpException: java.lang.IllegalArgumentException: Invalid URI host: null (authority: samourai-server_nginx_1:80)
whirlpool_1   | 	at com.samourai.http.client.JacksonHttpClient.postUrlEncoded(JacksonHttpClient.java:96) ~[extlibj-0.0.28.jar!/:na]
whirlpool_1   | 	at com.samourai.wallet.api.backend.BackendOAuthApi.oAuthAuthenticate(BackendOAuthApi.java:35) ~[extlibj-0.0.28.jar!/:na]
whirlpool_1   | 	at com.samourai.wallet.util.oauth.OAuthManagerJava.newAccessToken(OAuthManagerJava.java:44) ~[extlibj-0.0.28.jar!/:na]
whirlpool_1   | 	at com.samourai.wallet.util.oauth.OAuthManagerJava.getOAuthAccessToken(OAuthManagerJava.java:29) ~[extlibj-0.0.28.jar!/:na]
whirlpool_1   | 	at com.samourai.wallet.api.backend.BackendApi.computeHeaders(BackendApi.java:241) ~[extlibj-0.0.28.jar!/:na]
[...]

PR #438, opened 2023-03-01
Update Samourai Server (revert using Docker DNS)

nmfretz:
Fixes Whirpool GUI issue noted in Issue https://github.com/getumbrel/umbrel-apps/issues/397

Whirlpool GUI connection/authentication does not appear to do any DNS resolution. The following error is shown in the GUI:

image

and the following error is logged in the whirpool container:
`whirlpool_1   | com.samourai.wallet.api.backend.beans.HttpException: java.lang.IllegalArgumentException: Invalid URI host: null (authority: samourai-server_nginx_1:80)`

Reverting back to using hardcoded container IPs instead of name containers w/ Docker DNS fixes this issue.

### DIFF
--- a/samourai-server/docker-compose.yml
+++ b/samourai-server/docker-compose.yml
@@ -3,10 +3,9 @@ version: "3.7"
 services:
   app_proxy:
     environment:
-      APP_HOST: samourai-server_nginx_1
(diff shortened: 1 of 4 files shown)


nmfretz:
Tested on arm64 and amd64 🗡️

PR #439, opened 2023-03-01
Update lnbits to v0.10.1

aphex3k:
Update LNbits to latest v0.10

This should also enable umbrel installations to become [cashu](https://github.com/cashubtc/cashu) mints.

### DIFF
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -32,3 +32,6 @@ services:
       LNBITS_DEFAULT_WALLET_NAME: "LNbits wallet"
       LNBITS_DISABLED_EXTENSIONS: "amilk"
       LNBITS_ADMIN_EXTENSIONS: "ngrok"
+      
+      # Cashu
+      CASHU_PRIVATE_KEY: ${APP_SEED}
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   web:
-    image: lnbitsdocker/lnbits-legend:0.9.7@sha256:1fccd9611a23b9d9259ca4adff860e3e2d441ed42261d192141e76b261cd3b74
+    image: lnbitsdocker/lnbits-legend:0.10.1@sha256:c94f30fb41a6608aae2a1d0bc47831fccb109a646118100dfc23c967ab13eefe
     init: true
     restart: on-failure
     stop_grace_period: 1m
--- a/lnbits/umbrel-app.yml
+++ b/lnbits/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: lnbits
 category: Finance
 name: LNbits
-version: "0.10"
+version: "0.10.1"
 tagline: Multi-user wallet management system
 description: >-
   LNbits is a simple multi-user and account system for Lightning
--- a/lnbits/umbrel-app.yml
+++ b/lnbits/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: lnbits
 category: Finance
 name: LNbits
-version: "0.9.7"
+version: "0.10"
 tagline: Multi-user wallet management system
 description: >-
   LNbits is a simple multi-user and account system for Lightning
@@ -28,8 +28,16 @@ path: ""
 defaultUsername: ""
 defaultPassword: ""
 releaseNotes:  >-
-  - This minor release updates the Docker images and fixes the dependency versions from 0.9.6.1.
+  - Major Update: Installable extensions
   
-  See the full changelog here: https://github.com/lnbits/lnbits/releases/tag/0.9.7
+  - Update: Clean/format code base
+  
+  - Update: respect https setting in admin ui link
+  
+  - Update: Added cyber theme
+  
+  - Bugfix: hide Admin UI if LNBITS_ADMIN_UI is false #1537
+  
+  See the full changelog here: https://github.com/lnbits/lnbits/releases/tag/0.10
 submitter: Umbrel
 submission: https://github.com/getumbrel/umbrel/pull/372


lnbits/docker-compose.yml
```diff
@@ -32,3 +32,6 @@ services:
       LNBITS_DEFAULT_WALLET_NAME: "LNbits wallet"
       LNBITS_DISABLED_EXTENSIONS: "amilk"
       LNBITS_ADMIN_EXTENSIONS: "ngrok"
+      
+      # Cashu
+      CASHU_PRIVATE_KEY: ${APP_SEED}
```
aphex3k, line 37: > Cashu mint and wallet
Here you can create multiple cashu mints that you can share. Each mint can service many users but all ecash tokens of a mint are only valid inside that mint and not across different mints. To exchange funds between mints, use Lightning payments.

> Important
If you are the operator of this LNbits instance, make sure to set CASHU_PRIVATE_KEY="randomkey" in your configuration file. Do not create mints before setting the key and do not change the key once set.

I would like the private key to be a value derived from entropy already available in the umbrel application environment in order to prevent complication during installation, operation and restore of a mint. I've seen other apps use the APP_SEED or derived value. Is this acceptable in this scenario?

aphex3k: Tagging @callebtc for visibility.

nmfretz:
Thanks for this update @aphex3k. Taking a look now.

aphex3k:
> Thanks for this update @aphex3k. Taking a look now.

It works on my machine :P

lnbits/docker-compose.yml
nmfretz: This is the correct way to access this 👍 . In this case it will be a 256 bit hex string deterministically derived from 128 bits of cryptographically secure entropy. Is that all that is needed  for `CASHU_PRIVATE_KEY`?

aphex3k: According to Cashu documentation it can be an arbitrary string that doesn't have any special requirements as long as it doesn't change for the lifetime of the mint.

nmfretz:
> It works on my machine :P

I checked the docker image digest and it is multi-arch and boots up fine on amd64 and arm64. However, with this newest version, where all extensions are [removed by default](https://github.com/lnbits/lnbits/releases/tag/0.10) , I actually don't see any option to add extensions in the UI. Is this the case for you as well? I think we may need to access the UI as an admin user using `LNBITS_ADMIN_USERS` env var in order to install extensions like cashu.

aphex3k:
Yes, I added my existing wallet user as admin to enable Cashu extension.

Non-admin users work just fine. I have not used other lnbits extensions before or after the update.

aphex3k:
> I think we may need to access the UI as an admin user

There is an option to create and spit out an admin user on the command line as a oneshot but that doesn't work on Umbrel.

aphex3k:
What we would need to do is create a valid admin user string (derived from the seed) and export it to the environment for the docker image to import it?!?

nmfretz:
> There is an option to create and spit out an admin user on the command line as a oneshot but that doesn't work on Umbrel.

In the docker-compose.yml I think we have two options:
https://github.com/lnbits/lnbits/blob/c81d3e06c8c4217c1c38a71e24964df0e7b80d2c/.env.example#L11-L21
1) set `LNBITS_ADMIN_UI=true`, which would allow a user to generate a superuser account through the admin UI on startup... I'm not sure this is a good solution with existing wallets in mind.
2) Add users to `LNBITS_ADMIN_USERS`.

Pinging @lukechilds for some extra help here.

aphex3k:
> 2. Add users to `LNBITS_ADMIN_USERS`.
> 
> Pinging @lukechilds for some extra help here.

Can we add the `LNBITS_ADMIN_USERS` but have the actual wallet value derive from `${APP_SEED}` so that we end up with a deterministic wallet that has admin capabilities and is shown on the umbrel app store page where the passwords are typically shown?

lukechilds:
Sorry guys, missed this!

Just from the linked docs:

```
# Enable Admin GUI, available for the first user in LNBITS_ADMIN_USERS if available
# Warning: Enabling this will make LNbits ignore this configuration file. Your settings will
# be stored in your database and you will be able to change them only through the Admin UI.
# Disable this to make LNbits use this config file again.
LNBITS_ADMIN_UI=false
```

Just generally speaking, this sounds like exactly what we want. Ideally we don't want to have a static config that all Umbrel users need to use. If LNbits now supports dynamic configuration from the UI that's definitely the way for us to go, assuming it still respects all the ENV vars we pass in to setup LND.

But it sounds like if we do that we degrade the functionality of existing users? Not sure I fully understand the issue above re things getting disabled. Are current Umbrel LNbits installs treating all users as admin and if we enable this they become non admins? If that's the issue maybe we could add a hook that lists the current users and exports `LNBITS_ADMIN_USERS` with them included. We'd want to run this hook once only for existing installs and save the result in a file to prevent all users of new installs being upgraded to admins.

Or is the issue that previous Umbrel LNbits installs were auto generating a config on start (or users are updating it via SSH) and as soon as we ship `LNBITS_ADMIN_UI=true` we disable their config and they go back to defaults?

If that's the case potentially we could do something clever where we add a hook that detects old/new installs and sets different env vars on them. So new installs just get `LNBITS_ADMIN_UI=false` (which I think is all they need right?) and old installs get something else to get them working as expected.

nmfretz:
Sorry for the delay on this, @aphex3k.

I see that Calle just posted that Cashu wallet on LNbits is being sunsetted and moved to its own deployment (https://wallet.cashu.me/):
https://snort.social/e/nevent1qqsfun6953jhzrg2yjay784q6pqvtunnpxgu8wy90w340j8w98x295gpz3mhxue69uhhyetvv9ujuerpd46hxtnfdu5d7hvq

We still need to figure out a good way to enable admin users for existing and new deploys, but we should likely not include cashu mint functionality anymore.

nmfretz:
Thanks again for all your work on this @aphex3k. I'm closing this PR in favour of https://github.com/getumbrel/umbrel-apps/pull/522 to address the recent critical bug disclosure (https://twitter.com/lnbits/status/1654416778804580352?s=20)

PR #441, opened 2023-03-02
Update WoofBot to v0.7.0 + App Rename

woofbotapp:
Renamed app to "WoofBot without Lightning"

Here's the new logo:

logo-no-lightning

PR #442, opened 2023-03-02
App Submission: WoofBot LND

woofbotapp:
Just like: https://github.com/getumbrel/umbrel-apps/pull/441 but with LND integration.

Tested on my Raspberry Pi 4.

Logo:

logo-lnd

I asked Lightning Labs for permissions to use the LND logo inside WoofBot logo, and they agreed with the condition that I specify that the app is not affiliated with LND (see app description).

Gallery screenshots can be taken from the original WoofBot app (although it's about time I update both of them, because the dashboard UI has changed).

### DIFF
--- a/woofbot-lnd/docker-compose.yml
+++ b/woofbot-lnd/docker-compose.yml
@@ -1,0 +1,43 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: woofbot-lnd_web_1
+      APP_PORT: 8080
+
+  web:
+    image: woofbot/woofbot:v0.7.0@sha256:36f024668d4c27738f3e2676bc949408b63e2bdffd2c2e450c8c0ae1f0776956
+    user: 1000:1000
+    restart: on-failure
+    stop_grace_period: 1m
+    environment:
+      MONGODB_URI: "mongodb://woofbot-lnd_mongodb_1:27017/woofbot"
+      APP_PORT: "8080"
+      APP_SEED: $APP_SEED
+      APP_BITCOIN_NODE_IP: $APP_BITCOIN_NODE_IP
+      APP_BITCOIN_RPC_USER: $APP_BITCOIN_RPC_USER
+      APP_BITCOIN_RPC_PASS: $APP_BITCOIN_RPC_PASS
+      APP_BITCOIN_RPC_PORT: $APP_BITCOIN_RPC_PORT
+      APP_LIGHTNING_NODE_IP: $APP_LIGHTNING_NODE_IP
+      APP_LIGHTNING_NODE_GRPC_PORT: $APP_LIGHTNING_NODE_GRPC_PORT
+      LND_TLS_PATH: "/lnd/tls.cert"
+      LND_READONLY_MACAROON_PATH: "/lnd/data/chain/bitcoin/$APP_BITCOIN_NETWORK/readonly.macaroon"
+      SERVER_LOGS_FILEPATH: /app/logs/server.log
+    volumes:
+      - ${APP_DATA_DIR}/data/server-logs:/app/logs
+      - "${APP_LIGHTNING_NODE_DATA_DIR}/tls.cert:/lnd/tls.cert:ro"
+      - "${APP_LIGHTNING_NODE_DATA_DIR}/data/chain/bitcoin/${APP_BITCOIN_NETWORK}/readonly.macaroon:/lnd/data/chain/bitcoin/${APP_BITCOIN_NETWORK}/readonly.macaroon:ro"
+    depends_on:
+      - mongodb
+
+  mongodb:
+    # Newer mongo versions don't support Raspberry Pi 4 arm64
+    image: mongo:4.4.6-bionic@sha256:3d0e6df9fd5bc42cbf8ef8bc9e6c4e78f6f26c7157dbd7bdec72d202ab8ebe3a
+    user: 1000:1000
+    restart: on-failure
+    stop_grace_period: 1m
+    volumes:
+      - ${APP_DATA_DIR}/data/db:/data/db
+      - ${APP_DATA_DIR}/data/db/diagnostic.data:/data/db/diagnostic.data
+      - ${APP_DATA_DIR}/data/db/journal:/data/db/journal
--- a/woofbot-lnd/umbrel-app.yml
+++ b/woofbot-lnd/umbrel-app.yml
@@ -1,0 +1,35 @@
+manifestVersion: 1.1
+id: woofbot-lnd
+category: Finance
+name: WoofBot (LND)
+version: "0.7.0"
+tagline: A chatbot for your personal node
+description: >-
+  WoofBot is a chat bot that runs on your personal Bitcoin node and sends telegram notifications
+  based on pre-configured conditions: addresses, transactions, new-blocks and price changes.
+
+  This version of WoofBot integrates with LND lightning api. WoofBot is NOT affiliated with
+  Lightning Labs, the creators of LND. WoofBot only communicates with LND's api.
+developer: WoofBotApp
+website: https://github.com/woofbotapp
+dependencies:
+  - bitcoin
+  - lightning
+repo: https://github.com/woofbotapp/woofbotapp
+support: https://github.com/woofbotapp/woofbotapp/discussions
+port: 8093
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+releaseNotes: >
+  - Notification when a lightning channel is opened
+
+  - Notification when a lightning channel is closed
+
+  - Notification on payment forwarding
+submitter: WoofBotApp
+submission: https://github.com/getumbrel/umbrel-apps/pull/442


woofbot-lnd/hooks/pre-start
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+APP_DATA_DIR="$(readlink -f $(dirname "${BASH_SOURCE[0]}")/..)/data"
+
+[ ! -d "${APP_DATA_DIR}" ] && mkdir -p "${APP_DATA_DIR}" && chown 1000:1000 "${APP_DATA_DIR}"
+
+SERVER_LOGS_DIR="${APP_DATA_DIR}/server-logs"
+
+[ ! -d "${SERVER_LOGS_DIR}" ] && mkdir -p "${SERVER_LOGS_DIR}" && chown 1000:1000 "${SERVER_LOGS_DIR}"
```
nmfretz, line 10: Is this pre-start hook needed? Umbrel already exports an `APP_DATA_DIR` env var automatically for your app, and you have .gitkeep folders under the required directories.

woofbot-lnd/umbrel-app.yml
```diff
@@ -0,0 +1,33 @@
...
+  This version of WoofBot integrates with LND lightning api. WoofBot is NOT affiliated with
+  Lightning Labs, the creators of LND. WoofBot only communicates with LND's api.
+developer: WoofBotApp
+website: https://github.com/woofbotapp
+dependencies:
+  - bitcoin
+  - lightning
+repo: https://github.com/woofbotapp/woofbotapp
+support: https://github.com/woofbotapp/woofbotapp/discussions
+port: 8093
```
nmfretz, line 20: I doubled checked and this is a free port. Thanks!

woofbot-lnd/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: woofbot-lnd_web_1
```
nmfretz, line 6: Nice change to container names for docker DNS! Feel free to do the same thing to Woofbot without Lightning for the next update.

woofbot-lnd/umbrel-app.yml
```diff
@@ -0,0 +1,33 @@
...
+  - 3.jpg
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+releaseNotes: >
+  - Notification when a lightning channel is opened
+  - Notification when a lightning channel is closed
+  - Notification on payment forwarding
+submitter: WoofBotApp
+submission: https://github.com/getumbrel/umbrel-apps/pull/28
```
nmfretz, line 33: Do you want to change this to refer to this pull request since it is technically a different app?
https://github.com/getumbrel/umbrel-apps/pull/442

nmfretz (commented):
Thanks for this new submission @woofbotapp. I just recently played around with the original WoofBot app and it is awesome. Great work! Super excited to have more functionality with LND integration.

I have left some very small review comments. Once addressed this will be good to go!

nmfretz:
Also, same question as https://github.com/getumbrel/umbrel-apps/pull/441#issuecomment-1454271478. Would you like us to help with your app logo by giving it a background or would you like the background to be white?

woofbot-lnd/hooks/pre-start
nmfretz: Ah, I was just browsing GitHub Issues for this repo and came accross https://github.com/getumbrel/umbrel-apps/issues/346. I see what you are doing here now.

woofbotapp: Yes, it's the same code from the original WoofBot. It's needed for the MongoDB database.

woofbot-lnd/umbrel-app.yml
woofbotapp: The original port for WoofBot (without Lightning) was 8092 - I think that you guys allocated it because my original pick was already taken. So 8093 for WoofBot-LND makes sense.

woofbot-lnd/docker-compose.yml
woofbotapp: I'm afraid to change things in the original WoofBot that would break updates...

It will make more sense to have a mechanism for "dynamic permissions", where the user could choose whether to allow access to LND or not during installation (Or maybe not have LND installed at all).
The docker-compose.yml and umbral-app.yml files should be built dynamically, from a script or a template.
For example, if I'll want WoofBot to integrate in the future with the SimpleTorrent app, to get a notification when a download is complete, I'll have to make 4 copies: WoofBot-without-lightning-without-torrents, WoofBot-LND-without-torrents, WoofBot-without-lightning-SimpleTorrent, WoofBot-LND-SimpleTorrent...

woofbot-lnd/umbrel-app.yml
woofbotapp: Fixed

woofbotapp:
I don't mind about the background. Thanks @nmfretz !

woofbot-lnd/umbrel-app.yml
nmfretz: Perfect, thanks

woofbot-lnd/docker-compose.yml
nmfretz: Ya, I hear you on this. If you ever want to change OG WoofBot, just let me know and I can do all necessary changes/testing to make sure it doesn't break on update.

nmfretz: > For example, if I'll want WoofBot to integrate in the future with the SimpleTorrent app, to get a notification when a download is complete, I'll have to make 4 copies: WoofBot-without-lightning-without-torrents, WoofBot-LND-without-torrents, WoofBot-without-lightning-SimpleTorrent, WoofBot-LND-SimpleTorrent...

Agreed, need a way to do this with dynamic permissions. I would love WoofBot with even more functionality.

woofbot-lnd/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
...
+      APP_BITCOIN_RPC_PORT: $APP_BITCOIN_RPC_PORT
+      APP_LIGHTNING_NODE_IP: $APP_LIGHTNING_NODE_IP
+      APP_LIGHTNING_NODE_GRPC_PORT: $APP_LIGHTNING_NODE_GRPC_PORT
+      LND_TLS_PATH: "/lnd/tls.cert"
+      LND_READONLY_MACAROON_PATH: "/lnd/data/chain/bitcoin/$APP_BITCOIN_NETWORK/readonly.macaroon"
+      SERVER_LOGS_FILEPATH: /app/logs/server.log
+    volumes:
+      - ${APP_DATA_DIR}/data/server-logs:/app/logs
+      - "${APP_LIGHTNING_NODE_DATA_DIR}/tls.cert:/lnd/tls.cert:ro"
+      - "${APP_LIGHTNING_NODE_DATA_DIR}/data/chain/bitcoin/${APP_BITCOIN_NETWORK}/readonly.macaroon:/lnd/data/chain/bitcoin/${APP_BITCOIN_NETWORK}/readonly.macaroon:ro"
```
nmfretz, line 29: Here you can just mount the entire LND data directory as read only. Like this:
`${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro`

What you have done makes perfect sense in the context of sharing as little file system access as possible for increased isolation. But in this specific case since the TLS cert is generated dynamically by LND there is an extreme edge-case where WoofBot starts before LND has created the cert. In this case, `${APP_LIGHTNING_NODE_DATA_DIR}/tls.cert` path won't exist when the container is starting so Docker will create an empty directory at that path owned by root. This could lead to LND running into issues when it's trying to create it's TLS cert.

nmfretz:
@woofbotapp one tiny change to make [here](https://github.com/getumbrel/umbrel-apps/pull/442#discussion_r1125996257) and we can go live!

woofbot-lnd/docker-compose.yml
woofbotapp: @nmfretz Indeed WoofBot needs only readonly access to LND.

Is there no other way? What if the docker will deliberately crash if LND_TLS_PATH is defined but the file was not found?
Or maybe even have a small separated service that will mount `/lnd` and then copy the tls.cert and readonly.macaroon to the APP_DATA_DIR folder?

woofbotapp: @nmfretz I see that ln-visualizer is doing something similar:

https://github.com/getumbrel/umbrel-apps/blob/master/ln-visualizer/docker-compose.yml#L24-L25

Maybe adding `user: 1000:1000` would solve the issue of the directory won't be created as root and will be mounted when LND starts?

nmfretz: > What if the docker will deliberately crash if LND_TLS_PATH is defined but the file was not found?

Ah, good point. Yes, let's try `user 1000:1000` in the web container and see if the permissions allow your app to write properly.

woofbot-lnd/umbrel-app.yml
```diff
@@ -0,0 +1,33 @@
...
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+releaseNotes: >
+  - Notification when a lightning channel is opened
+  - Notification when a lightning channel is closed
+  - Notification on payment forwarding
```
nmfretz, line 31: for these points to show up on separate lines, they need to have another newline between them:

```suggestion
  - Notification when a lightning channel is opened
  
  - Notification when a lightning channel is closed
  
  - Notification on payment forwarding
```

woofbot-lnd/docker-compose.yml
nmfretz: I just tested with `user: "1000:1000"` and everything works as expected.

woofbot-lnd/umbrel-app.yml
woofbotapp: fixed

woofbot-lnd/docker-compose.yml
woofbotapp: Great, I updated the PR to have `user: 1000:1000`.
I'm a little bit confused whether the pre-start hook is still necessary. The relevant discussion about it can be found here:
https://github.com/getumbrel/umbrel-apps/issues/346
https://github.com/getumbrel/umbrel-apps/pull/353#pullrequestreview-1236142421

It doesn't hurt to have this script create and chown the `data/server-logs` folder, so I'll leave it as-is for now.

I'm also trying now to uninstall and reinstall the app, and will let you know if it works.

nmfretz: Ya, I read through these the other day and let's leave the pre-start hook for now because the next version of Umbrel that includes @nevets963's fix for this issue hasn't been released yet.

woofbotapp: Everything works smoothly. I think we are good to go.

woofbotapp:
@nmfretz For the meanwhile, let's keep it as two apps.

woofbot-lnd/docker-compose.yml
nmfretz: Excellent!

nmfretz:
> For the meanwhile, let's keep it as two apps.

Okay, sounds good!

Two questions related to logos and app names before going live. This is ultimately up to you as the developer of WoofBot, so please let me know if you don't like any of the suggestions below and we will go with your original logos and names.

**1) Logos**:

Here's three variants for WoofBot and WoofBot without Lightning. I am a big fan of the third variant (bottom). What do you think?

Screenshot_2023-03-08_at_11 56 08_AM (1)

** 2) Names**:

We were wondering if you'd consider naming the apps
- `WoofBot`
- `WoofBot (Lightning)` or `WoofBot + Lightning`

No pressure on any of these. Please do what you think is best for your apps.

woofbotapp:
My feeling is that most Umbrel users do run LND. A small minority may be running C-Lightning, and a tiny number of users are "onchain-only" and don't run any lightning stuff. But maybe I'm wrong (Do you collect any statistics? It would be interesting to make a Twitter poll).
So it's important to emphasize that the "Bitcoin" version is only for the specific users who do not run any lightning - i.e. name it "WoofBot (without Lightning)" or "WoofBot (onchain only)", and show the lightning icon with a "🚫" on it.

For the majority of users, I like the second row (white background) - but below the logo I would put only the purple lightning logo, nearby the LND logo, and name it "WoofBot (LND)". This will leave room for having a "WoofBot (C-Lightning)" in the future, if there will be demand for it.

The third row looks pretty but it's less clear what the difference between the apps is. I don't think people automatically associate lightning with the color purple (unlike the orange color that is associated with Bitcoin). Also, the lightning version has all the onchain related functionality as well.

The first row is similar to the original logo, but I think putting the orange dog on top of an orange background was a mistake in the first place.

It really won't matter in the future once "optional permissions" will be implemented.

Many thanks!

woofbotapp:
I updated the apps names (in umbrel-app.yml files). Let's go 🚀

woofbot-lnd/hooks/pre-start
nevets963: This pre-start hook is not needed because this is a new app and therefore the `server-logs` directory will exist from day 1 and will have the correct ownership. In the case of #346, the hook was needed to address the case of an app update (not app submission) not creating the directory.

woofbotapp: @nevets963 Removed the hook.

nmfretz:
This is now live! Fantastic work @woofbotapp!

> Gallery screenshots can be taken from the original WoofBot app (although it's about time I update both of them, because the dashboard UI has changed).

Let us know when you have new gallery images or if you would like us to help in any way.

PR #443, opened 2023-03-02
Update rotki 1.27.1

Xosten:
Hi there,

Minor update for rotki, the full changelog is available here: https://github.com/rotki/rotki/releases/tag/v1.27.1

I don't have access to my testing environment to check if it breaks a previous install. It shouldn't as there are no disclaimers on the release page.

### DIFF
--- a/rotki/docker-compose.yml
+++ b/rotki/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 80
 
   web:
-    image: rotki/rotki:v1.27.0@sha256:f1c814e3636f4a48544aca9e500ff0e54cf18e9b0e5a2eb23c309c376efdf3fb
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @Xosten! Really appreciate you keeping these apps up to date. I've reviewed the changelog and tested on amd64 and arm64 🚀 

I made a change to the release notes section. I'm making an effort to keep release notes concise for minor updates and then link to the full changelog so users can see lower level details if they want. Do you think this is an okay idea or would you personally rather see more details in the release notes section?

Xosten:
> Thanks @Xosten! Really appreciate you keeping these apps up to date. I've reviewed the changelog and tested on amd64 and arm64 🚀
> 
> I made a change to the release notes section. I'm making an effort to keep release notes concise for minor updates and then link to the full changelog so users can see lower level details if they want. Do you think this is an okay idea or would you personally rather see more details in the release notes section?

Thanks Nathan. I totally agree with you regarding the release notes. We don't need to provide a load of information for minor releases, vs major releases we could provide some more information.

PR #444, opened 2023-03-02
Ghostfolio: Remove environment variable BASE_CURRENCY

dtslvr:
`USD` is the default value of `BASE_CURRENCY`, therefore no need to set it explicitly.

FYI @BeauAgst.


nmfretz:
Thanks for this @dtslvr! 

@BeauAgst should we update the server image to the most recent ghostfolio release (1.241.0) and role out a new update with the `BASE_CURRENCY` env var removed?

jfongch:
@dtslvr @nmfretz @BeauAgst I created a PR with the 1.241.0 image #445

BeauAgst:
Sounds like a good idea! I don't actually use the app myself so wouldn't be best placed to offer guidance.

@dtslvr I added `BASE_CURRENCY` into the docker-compose so that it was easily modifiable for people installing the app. Currently they have to jump into the files via ssh and are probably using vim to do so. It's up to you whether you think it's actually necessary or not!

nmfretz:
hmmmm fair point @BeauAgst. Even though Umbrel's goal is to remove the need to ssh into a device, I do see [here](https://github.com/ghostfolio/ghostfolio#supported-environment-variables) that the value of `BASE_CURRENCY` cannot be  changed via the UI later, so maybe keeping it in the docker-compose.yml is beneficial for now.

That being said, `BASE_CURRENCY` is just used as the currency when buying/selling/tracking individual **cryptocurrencies** (e.g., in activities tab). You can still view the value of your portfolio and the individual holdings in whatever currency you want by changing it in the UI. Is that the correct understanding @dtslvr?

dtslvr:
You have informed yourself very well, @nmfretz! 🙂 

As you point out, you can control the account currency in the UI. With the latest version you can additionally choose a (custom) currency at the activity level. Therefore, the concept `BASE_CURRENCY` will disappear sooner or later.

nmfretz:
> As you point out, you can control the account currency in the UI. With the latest version you can additionally choose a (custom) currency at the activity level. Therefore, the concept BASE_CURRENCY will disappear sooner or later.

I tested this out in 1.241.0 and it works great! PR https://github.com/getumbrel/umbrel-apps/pull/445 has been merged. Closing this PR.

Thanks again @dtslvr!

PR #447, opened 2023-03-05
Update RoboSats to v0.4.0-alpha

Reckless-Satoshi:
Update RoboSats to latest v0.4.0-alpha

Tested only on `amd64`


nmfretz:
Thanks @Reckless-Satoshi, I will review tomorrow!

PR #448, opened 2023-03-06
Update esphome to 2023.2.4

mateosilguero:
update esphome version. fix platformio version error

### DIFF
--- a/esphome/docker-compose.yml
+++ b/esphome/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 6052
 
   server:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this PR to update ESPHome @mateosilguero! I added some release notes for this version, and have tested on amd64 and arm64 to check that the app does not break. 

Unfortunately I don't have access to any ESP8266/ESP32 devices to run some basic in-app checks. Did you end up testing this on your Umbrel?

mateosilguero:
Hi @nmfretz !
Yes, it works, but only over wifi apparently 🤔
On the previous version it throws an error  because of platformio version, but now im was able to upload a yml

nmfretz:
Muchas gracias, @mateosilguero! This update is now live.

PR #449, opened 2023-03-06
App Submission: OctoPrint

mateosilguero:
# App Submission

### App name
OctoPrint

### 256x256 SVG icon
https://octoprint.org/assets/img/logo.png

### Gallery images

Captura desde 2023-03-11 17-10-56

### I have tested my app on:

### DIFF
--- a/octoprint/docker-compose.yml
+++ b/octoprint/docker-compose.yml
@@ -9,6 +9,7 @@ services:
 
   web:
     image: octoprint/octoprint:1.9.2@sha256:80ed8b8a1be8713506e7e29fc5b2ec55241242e28a5d8a474a25d9f2aab0843f
+    # privileged: true
     volumes:
       - ${APP_DATA_DIR}/data:/octoprint
       - /dev:/dev
--- a/octoprint/docker-compose.yml
+++ b/octoprint/docker-compose.yml
@@ -11,4 +11,6 @@ services:
     image: octoprint/octoprint:1.8.7
     volumes:
       - ${APP_DATA_DIR}/data:/data
+    devices:
+      - "/dev/ttyACM0:/dev/ttyACM0"
     restart: on-failure
--- a/octoprint/docker-compose.yml
+++ b/octoprint/docker-compose.yml
@@ -1,0 +1,14 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: octoprint_server_1
+      APP_PORT: 5000
+      PROXY_AUTH_ADD: "false"
+
+  server:
+    image: octoprint/octoprint:1.8.7@sha256:f4a649c27917ad0624d16779f36b3a7e60626eddefccc51ec367a52a4288492c
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+    restart: on-failure
--- a/octoprint/docker-compose.yml
+++ b/octoprint/docker-compose.yml
@@ -3,12 +3,12 @@ version: "3.7"
 services:
   app_proxy:
     environment:
-      APP_HOST: octoprint_server_1
+      APP_HOST: octoprint_web_1
       APP_PORT: 5000
       PROXY_AUTH_ADD: "false"
 
-  server:
-    image: octoprint/octoprint:1.8.7@sha256:f4a649c27917ad0624d16779f36b3a7e60626eddefccc51ec367a52a4288492c
+  web:
+    image: octoprint/octoprint:1.8.7
     volumes:
       - ${APP_DATA_DIR}/data:/data
     restart: on-failure
--- a/octoprint/docker-compose.yml
+++ b/octoprint/docker-compose.yml
@@ -8,9 +8,8 @@ services:
       PROXY_AUTH_ADD: "false"
 
   web:
-    image: octoprint/octoprint:1.8.7@sha256:4cccc197fda1346798a66ecbab45383c8daaa193630b6c3ad40947940cbdbfce
-   volumes:
-     - ${APP_DATA_DIR}/data:/octoprint
-    devices:
-      - "/dev/ttyACM0:/dev/ttyACM0"
+    image: octoprint/octoprint:1.9.2@sha256:41d23f39d66bdcdac895938f1662f13ca1cb6ac910d2254d8123f2cc27229ccc
+    volumes:
+      - ${APP_DATA_DIR}/data:/octoprint
+      - /dev:/dev
     restart: on-failure
--- a/octoprint/docker-compose.yml
+++ b/octoprint/docker-compose.yml
@@ -9,7 +9,7 @@ services:
 
   web:
     image: octoprint/octoprint:1.9.2@sha256:80ed8b8a1be8713506e7e29fc5b2ec55241242e28a5d8a474a25d9f2aab0843f
-    # privileged: true
+    privileged: true
     volumes:
       - ${APP_DATA_DIR}/data:/octoprint
       - /dev:/dev
--- a/octoprint/docker-compose.yml
+++ b/octoprint/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   web:
-    image: octoprint/octoprint:1.9.2@sha256:41d23f39d66bdcdac895938f1662f13ca1cb6ac910d2254d8123f2cc27229ccc
+    image: octoprint/octoprint:1.9.2@sha256:80ed8b8a1be8713506e7e29fc5b2ec55241242e28a5d8a474a25d9f2aab0843f
     volumes:
       - ${APP_DATA_DIR}/data:/octoprint
       - /dev:/dev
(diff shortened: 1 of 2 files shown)


mateosilguero:
Sorry for bother you @nmfretz, but now im trying to add this app to the store. It doesnt work on local node. Could you help me to understand what im doing wrong ?

mateosilguero:
> Sorry for bother you @nmfretz, but now im trying to add this app to the store. It doesnt work on local node. Could you help me to understand what im doing wrong ?

It is now !

nmfretz:
Great job getting it to work on your local node @mateosilguero! I will review this soon.

ohenrik:
I'm really looking forward to getting octoprint on Umbrel! Thank you for adding this @mateosilguero!

octoprint/docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+      APP_PORT: 5000
+      PROXY_AUTH_ADD: "false"
+
+  web:
+    image: octoprint/octoprint:1.8.7
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${UMBREL_ROOT}/octoprint:/octoprint
+    devices:
+      - "/dev/ttyACM0:/dev/ttyACM0"
```
nmfretz, line 16: I think this could cause us some issues. 

If there is no printer connected to `/dev/ttyACM0` on the Umbrel device then when the container is booting up during install/restart it will error out and not start up because it won't be able to map `dev/ttyACM0` from the container to a non-existent `/dev/ttyACM0` file on the Umbrel device.

We could potentially get around this by just mapping `- "/dev:/dev"` but I don't think we should give an app access to the entire `/dev` folder for security reasons.

Have you tested whether you can connect a printer with these two line commented out? Unfortunately I do not have access to a 3d printer to check.

octoprint/umbrel-app.yml
```diff
@@ -0,0 +1,28 @@
+manifestVersion: 1
+id: octoprint
+category: Networking
```
nmfretz, line 3: We might need a whole new category now with OctoPrint! Could also fit under the current category "Productivity"

octoprint/umbrel-app.yml
```diff
@@ -0,0 +1,28 @@
...
+  OctoPrint provides a snappy web interface for controlling consumer 3D printers.
+releaseNotes: >-
+  Improvements
+   Add support for Python 3.11 by bumping up the version of a dependency.
+  Bug fixes
+    Do not send None to Locale.parse if no match can be made between the request's accepted languages and the installed languages. Fixes a 500 server error on the API in case of Babel being at of version of 2.12 or higher.
+developer: OctoPrint
+website: https://octoprint.org/
+repo: https://github.com/OctoPrint/OctoPrint
+support: https://github.com/OctoPrint/OctoPrint/issues
```
nmfretz, line 17: I found the community support page here: https://community.octoprint.org/

```suggestion
support: https://community.octoprint.org/
```

octoprint/umbrel-app.yml
```diff
@@ -0,0 +1,28 @@
...
+port: 5000
+submitter: OctoPrint
```
nmfretz, line 19: This gets to be you! Woooop! You can use your github handle like this, or choose something different:

```suggestion
submitter: mateosilguero
```

octoprint/umbrel-app.yml
```diff
@@ -0,0 +1,28 @@
...
+name: OctoPrint
+version: "1.8.7"
+tagline: The snappy web interface for your 3D printer.
+description: >-
+  OctoPrint provides a snappy web interface for controlling consumer 3D printers.
+releaseNotes: >-
+  Improvements
+   Add support for Python 3.11 by bumping up the version of a dependency.
+  Bug fixes
+    Do not send None to Locale.parse if no match can be made between the request's accepted languages and the installed languages. Fixes a 500 server error on the API in case of Babel being at of version of 2.12 or higher.
```
nmfretz, line 13: For the initial app release we can either remove `releaseNotes` entirely, or else just say:
` - Initial release for the Umbrel App Store`

octoprint/umbrel-app.yml
```diff
@@ -0,0 +1,28 @@
...
+developer: OctoPrint
+website: https://octoprint.org/
+repo: https://github.com/OctoPrint/OctoPrint
+support: https://github.com/OctoPrint/OctoPrint/issues
+port: 5000
```
nmfretz, line 18: I just checked and this port is already taken by [SatSale](https://github.com/getumbrel/umbrel-apps/blob/master/satsale/umbrel-app.yml). You can choose any free port for this (5003 is available for example). Do a global search of the umbrel-apps repo for `port: <your-port-choice>` and see if it is already taken in an apps `umbrel-app.yml`.

You don't have to change anything else for this to work. It is not the same thing as `APP_PORT` for the `app_proxy` service.

nmfretz (commented):
Thanks again for this submission @mateosilguero! See below for my review.

Once we get close to release, I can help get gallery images made.

octoprint/docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+    environment:
+      APP_HOST: octoprint_web_1
+      APP_PORT: 5000
+      PROXY_AUTH_ADD: "false"
+
+  web:
+    image: octoprint/octoprint:1.8.7
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${UMBREL_ROOT}/octoprint:/octoprint
```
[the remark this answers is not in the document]
mateosilguero: Great !

octoprint/docker-compose.yml
mateosilguero: Yes, youre right. If the printer is not connected, the image will not boot. I'm trying to write the dir if is doesnt exists. Let me try some things.

octoprint/umbrel-app.yml
mateosilguero: done

octoprint/umbrel-app.yml
mateosilguero: done !

octoprint/docker-compose.yml
mateosilguero: My workaround doesnt work :( i was inspired by this https://github.com/getumbrel/umbrel-apps/pull/43/files#diff-de62ede3db7fbfc30c6ec34ff877c60a6afad419a257528af36e1bc311d81a97

Do you find a solution for esphome ?

mateosilguero: Captura desde 2023-03-24 20-13-23

When the printer is connected, it works really well, and once the image is booted, you could disconnect and connect it at any time. The problem is on installation and restart

mateosilguero: @nmfretz Is there a possible solution to handle this ?

nmfretz:
Hey @mateosilguero, stay tuned we might have a solution for USB access soon.

nmfretz:
@mateosilguero, check out the recent Home Assistant update for an example of how to set up Octoprint with USB access:
https://github.com/getumbrel/umbrel-apps/blob/master/home-assistant/docker-compose.yml

mateosilguero:
Thanks ! I gonna update my code in a few days.

nmfretz:
Hey @mateosilguero, if I update the compose file would you be okay helping me to test this? Unfortunately I don't have a 3D-printer to check that everything within the app is working correctly.

mateosilguero:
Hi ! Sorry for my delay. Let me check this out, will commit new changes today, maybe including the new version (1.9)

nmfretz:
No worries at all, I didn't mean to put any immediate pressure! I can do review and testing related to Umbrel, but will just need confirmation of printing.

mateosilguero:
It works now :) Thank you for your help

nmfretz:
> It works now :) Thank you for your help

Awesome, glad to hear @mateosilguero. Can you confirm that everything worked for you without `privileged: true` being set for the container. See where I added this line commented out for you to test: https://github.com/getumbrel/umbrel-apps/pull/449/commits/1837b8c78cc79246bd0196d9801578d10738d549. Could you try:
- Starting the container with the printer USB plugged in.
- Starting the container with no USB plugged in --> app should still start
  - Then plug in USB and see if it is detected
- Check that printing functionality works as expected. It would be great to avoid granting extended privileges to the octoprint container, but we need to make sure that all functionality remains.

I have made some small changes to the compose and manifest files to get this finalized for the app store. @surajmahraj will work on the icon and gallery images and then we are ready!

mateosilguero:
Well, without privileged flag, starts ok and it shows the usb port on the list, but idk why the printer refuses the connection. With the privileged flag, it works as expected :)

nmfretz:
> With the privileged flag, it works as expected :)

Makes sense, that is what I was expecting. Thanks for testing this @mateosilguero!

nmfretz:
Octoprint is going live to the app store! Congratulations and thanks again @mateosilguero.

Here are the gallery assets ⬇

Gallery images:





Icon:
icon

mateosilguero:
Awesome news !

Thanks a lot for your help @nmfretz :)

Pogopooc:
I cannot install any plugins. In the dialog it says that the plugin has been installed, but after a restart it is unfortunately not listed under the installed plugins. What am I doing wrong?

TheLankyScottishNerd:
Is there anything specific required to enable the use of USB Webcams? Webcams don't seem to work or show up within the app even after specifying the ffmpeg path.

PR #450, opened 2023-03-08
Update specter-desktop app to v2.0.0

aphex3k:


aphex3k:
I also ran into this issue: https://github.com/cryptoadvance/specter-desktop/issues/2296

specter-desktop/docker-compose.yml
```diff
@@ -16,6 +16,7 @@ services:
     command:
       - --host=0.0.0.0
       - --specter-data-folder=/data
+      - --devstatus-threshold=beta
```
aphex3k, line 19: This change will prevent existing SWAN plugin users on Umbrel to not see the plugin after the upgrade.

jaonoctus:
I wouldn't upgrade it just yet because there is still some issues with the new version.

nmfretz:
Thanks for this @aphex3k.

> We probably want new artwork with this version as the user interface changed significantly?!?

Sounds great. I can help replace these with current screenshots

I am following https://github.com/cryptoadvance/specter-desktop/issues/2296, https://github.com/cryptoadvance/specter-desktop/issues/1963, and an issue that @jaonoctus opened here: https://github.com/cryptoadvance/specter-desktop/issues/2301

What are your thoughts on delaying vs upgrading now?

jaonoctus:
@moneymanolis just tested with my segwit and taproot wallets (`regtest` and `testnet`), works just fine 👍🏻

BitcoinMechanic:
Were you connected via Spectrum or direct to bitcoin core?

moneymanolis:
@aphex3k is there a way to specify in the docker-compose yml that the hostname used by Specter is localhost and not umbrel.local? 

This would solve https://github.com/cryptoadvance/specter-desktop/issues/1963.

aphex3k:
> What are your thoughts on delaying vs upgrading now?

I would err on the side of holding the update for now.

moneymanolis:
FYI with [this PR](https://github.com/cryptoadvance/specter-desktop/pull/2309) merged now we will bring back the Swan Plugin to prod in the next release but it will not be available for the umbrel.local hostname (until the backend can handle this hostname as redirect URI). We added a specific hint for Umbrel users on how they can use the plugin:

grafik

It's the best I could do for now. @aphex3k

moneymanolis:
> I believe this would clash with the fundamentals of umbrel's innerworkings and the underlying docker architecture.

You know this better, was just an idea. In general, two different hostnames can resolve to 127.0.0.1 so I thought there maybe a way for Specter on umbrel to use localhost instead of umbrel.local while both exist in the etc/hostnames.

nmfretz:
> I would err on the side of holding the update for now.

Sounds good @aphex3k, I will keep my eye on this.

PR #452, opened 2023-03-10
Update WoofBot to v0.7.1

woofbotapp:
update both woofbot and woofbot-lnd.
They use the same docker image, with different permissions (configured in the .yml files).

Tested on my Raspberry Pi.


nmfretz:
Thanks @woofbotapp. Reviewed and tested on both amd64 and arm64. Going live now!

PR #459, opened 2023-03-16
Nutstash wallet

gandlafbtc:
Nutstash is a cashu wallet for the browser.

### DIFF
--- a/nutstash-wallet/docker-compose.yml
+++ b/nutstash-wallet/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
       
   web:
-    image: ghcr.io/gandlafbtc/nutstash-wallet:0.1.6@sha256:1fc859c57c15918c19774c95325c67e12a10191dc3e2f09c184241dffe478776
+    image: ghcr.io/gandlafbtc/nutstash-wallet:0.1.8@sha256:83f4c041b89de570bbc3df41983668e6858d257562e50a82a521899c2975c614
     restart: on-failure
     stop_grace_period: 1m
     
--- a/nutstash-wallet/docker-compose.yml
+++ b/nutstash-wallet/docker-compose.yml
@@ -1,0 +1,15 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: nutstash-wallet_web_1
+      APP_PORT: 4173
+      PROXY_AUTH_ADD: "false"
+      
+  web:
+    image: ghcr.io/gandlafbtc/nutstash-wallet:0.1.6@sha256:1fc859c57c15918c19774c95325c67e12a10191dc3e2f09c184241dffe478776
+    restart: on-failure
+    stop_grace_period: 1m
+    init: true
+    
--- a/nutstash-wallet/docker-compose.yml
+++ b/nutstash-wallet/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
       
   web:
-    image: ghcr.io/gandlafbtc/nutstash-wallet:0.1.8@sha256:83f4c041b89de570bbc3df41983668e6858d257562e50a82a521899c2975c614
+    image: ghcr.io/gandlafbtc/nutstash-wallet:0.1.10-a@sha256:70a32529ccda47a4673e28a2dd343ca532a73253332a74c0c0c61f2cbceae1cf
     restart: on-failure
     stop_grace_period: 1m
     
--- a/nutstash-wallet/docker-compose.yml
+++ b/nutstash-wallet/docker-compose.yml
@@ -11,5 +11,4 @@ services:
     image: ghcr.io/gandlafbtc/nutstash-wallet:0.1.6@sha256:1fc859c57c15918c19774c95325c67e12a10191dc3e2f09c184241dffe478776
     restart: on-failure
     stop_grace_period: 1m
-    init: true
     
--- a/nutstash-wallet/docker-compose.yml
+++ b/nutstash-wallet/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
       
   web:
-    image: ghcr.io/gandlafbtc/nutstash-wallet:0.1.10@sha256:2d33851ef8bf7008a4f0daef7905d055580976213987e7a09c67145250c84220
+    image: ghcr.io/gandlafbtc/nutstash-wallet:0.1.10@sha256:8a585c4a6545e68e9a1c8fcf2a0a15ef401ad040accf925234451fe1b838d180
     restart: on-failure
     stop_grace_period: 1m
     
--- a/nutstash-wallet/docker-compose.yml
+++ b/nutstash-wallet/docker-compose.yml
@@ -4,11 +4,11 @@ services:
   app_proxy:
     environment:
       APP_HOST: nutstash-wallet_web_1
-      APP_PORT: 4173
+      APP_PORT: 3000
       PROXY_AUTH_ADD: "false"
       
   web:
-    image: ghcr.io/gandlafbtc/nutstash-wallet:0.1.10-a@sha256:70a32529ccda47a4673e28a2dd343ca532a73253332a74c0c0c61f2cbceae1cf
+    image: ghcr.io/gandlafbtc/nutstash-wallet:0.1.10@sha256:2d33851ef8bf7008a4f0daef7905d055580976213987e7a09c67145250c84220
     restart: on-failure
     stop_grace_period: 1m
     
--- a/nutstash-wallet/umbrel-app.yml
+++ b/nutstash-wallet/umbrel-app.yml
@@ -15,7 +15,7 @@ description: >-
 releaseNotes: >-
   - Initial release on the umbrel app store
   
-developer: Gandlafbtc
+developer: gandlafbtc
 website: https://nutstash.app/
 dependencies: []
 repo: https://github.com/gandlafbtc/nutstash-wallet
--- a/nutstash-wallet/umbrel-app.yml
+++ b/nutstash-wallet/umbrel-app.yml
@@ -1,0 +1,31 @@
+manifestVersion: 1
+id: nutstash-wallet
+category: Finance
+name: Nutstash Wallet
+version: "0.1.6"
+tagline: Nutstash is a Cashu wallet for your browser.
+description: >-
+  This app is in early development and most likely buggy. It does handle real sats though, so be sure to use only amounts that are small enough to lose.
+  
+  What is Cashu?
+
+  "Cashu is a free and open-source Chaumian ecash system built for Bitcoin. Cashu offers near-perfect privacy for users of custodial Bitcoin applications. Nobody needs to knows who you are, how much funds you have, and with whom you transact with." - cashu.space
+  
+releaseNotes: >-
+  - Initial release on the umbrel app store
+  
+developer: Gandlafbtc
+website: https://nutstash.app/
+dependencies: 
+repo: https://github.com/gandlafbtc/nutstash-wallet
+support: https://t.me/+DRy-rSe5wQtlYTk1
+port: 4173
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+submitter: gandlafbtc
+submission: https://github.com/getumbrel/umbrel-apps/pull/455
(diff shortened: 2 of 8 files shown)


nutstash-wallet/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: nutstash-wallet_web_1
+      APP_PORT: 4173
+      PROXY_AUTH_ADD: "false"
+      
+  web:
+    image: ghcr.io/gandlafbtc/nutstash-wallet:0.1.6@sha256:1fc859c57c15918c19774c95325c67e12a10191dc3e2f09c184241dffe478776
```
nmfretz, line 11: perfect. thanks for the multi-arch digest here

nutstash-wallet/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+    restart: on-failure
+    stop_grace_period: 1m
+    init: true
```
nmfretz, line 14: Are these two lines needed? 

`stop_grace_period` is used to set the time that a container tries to gracefully shut down and complete any ongoing tasks. The default value is 10s if `stop_grace_period` is not included. Once this time has expired, the container will be forcefully terminated. Since nutstash is a browser wallet I don't think we need to wait for anything to happen before shutting down, correct?

`init: true` is used to run an initialization process in the container before starting the main process. It doesn't look like we have any entrypoint scripts or anything to run, so we can likely remove this.

```suggestion
```

nutstash-wallet/umbrel-app.yml
```diff
@@ -0,0 +1,31 @@
...
+  This app is in early development and most likely buggy. It does handle real sats though, so be sure to use only amounts that are small enough to lose.
+  
+  What is Cashu?
+
+  "Cashu is a free and open-source Chaumian ecash system built for Bitcoin. Cashu offers near-perfect privacy for users of custodial Bitcoin applications. Nobody needs to knows who you are, how much funds you have, and with whom you transact with." - cashu.space
+  
+releaseNotes: >-
+  - Initial release on the umbrel app store
+  
+developer: Gandlafbtc
```
nmfretz, line 17: do you want this to be all lowercase like you have in `submitter`?

```suggestion
developer: gandlafbtc
```

nutstash-wallet/umbrel-app.yml
```diff
@@ -0,0 +1,31 @@
...
+website: https://nutstash.app/
+dependencies:
```
nmfretz, line 19: This needs to be an array.

```suggestion
dependencies: []
```

nutstash-wallet/umbrel-app.yml
```diff
@@ -0,0 +1,31 @@
...
+name: Nutstash Wallet
+version: "0.1.6"
+tagline: Nutstash is a Cashu wallet for your browser.
+description: >-
+  This app is in early development and most likely buggy. It does handle real sats though, so be sure to use only amounts that are small enough to lose.
+  
+  What is Cashu?
+
+  "Cashu is a free and open-source Chaumian ecash system built for Bitcoin. Cashu offers near-perfect privacy for users of custodial Bitcoin applications. Nobody needs to knows who you are, how much funds you have, and with whom you transact with." - cashu.space
+
```
nmfretz, line 13: Here's what this will look like in the app store:

image

If you want a blank line above `What is Cashu?` then you can add a second newline:

```suggestion
description: >-
  This app is in early development and most likely buggy. It does handle real sats though, so be sure to use only amounts that are small enough to lose.

    
  What is Cashu?

  "Cashu is a free and open-source Chaumian ecash system built for Bitcoin. Cashu offers near-perfect privacy for users of custodial Bitcoin applications. Nobody needs to knows who you are, how much funds you have, and with whom you transact with." - cashu.space
  
```

nmfretz (commented):
This is looking great. Thank you for this submission @gandlafbtc! I have left some minor suggestions below.

nmfretz:
Here's what we have put together for the icon and gallery images. Are you comfortable with the images and wording or would you like to see any changes? @surajmahraj tweaked the app theme for these images to work nicely with the nuts 🌰

Nutstash

gandlafbtc:
That is an interesting Issue. I have never tried over http, but, all the calls out to the mint are https anyway, so I don't think this would make a difference? 

The app runs almost entirely locally, so the only time http is used would be on the initial page load.

gandlafbtc:
> Here's what we have put together for the icon and gallery images. Are you comfortable with the images and wording or would you like to see any changes? @surajmahraj tweaked the app theme for these images to work nicely with the nuts chestnut
> 

The gallery is awesome! 2 minor things for the descriptions: 

1. On the first pic, it would be nice to have cashus main value prop. something like "pay lightning invoices privately with cashu ecash" or something like that

2. on the second pic: "mint __multiple__ tokens..." multiple can be replaced with 'cashu', that would make more sense. 

Thanks for this awesome artwork it looks damn cute <3 


Also maybe something to consider: 

1. The software is still alpha, maybe give the users some kind of warning. I have a message inside the app, but you never know.
2. Let people know that the mints are custodians, and that cashu ecash is not sats

I'm not sure how these things are displayed in the umbrel appstore, since i'm not a user, but maybe we can figure out together what makes more sense

nmfretz:
>The gallery is awesome! 2 minor things for the descriptions:...

Thanks for the edits! Here they are:






>I'm not sure how these things are displayed in the umbrel appstore, since i'm not a user, but maybe we can figure out together what makes more sense. I guess since we have a description text, it might be enough to have it there.

Here's how your app will be displayed in the app store when a user goes to install it. The `description` text shows up under "about this app", so it will be very visible to the user. I tweaked the wording a bit to direct the user to additional info. What do you think of this? Or would you like it to explicitly mention custodial mints and ecash vs satoshis? Also, I think it would be okay to put the early-development warning below the "what is cashu" description.

Screenshot 2023-04-03 at 2 12 09 PM

nmfretz:
> That is an interesting Issue. I have never tried over http, but, all the calls out to the mint are https anyway, so I don't think this would make a difference?
>
>The app runs almost entirely locally, so the only time http is used would be on the initial page load.

Ya, I think it's running into issues when trying to perform some crypto operation. For example, if I try to send tokens (same mint) from a https://wallet.nutstash.app/ wallet to a local umbrel wallet here's what I see (throw-away wallets):

https://user-images.githubusercontent.com/85373263/229632575-e22d963a-973b-48b1-bbe7-8a65a17c902e.mp4

gandlafbtc:
Thanks for the video, that cleared at least something up. We've made an update to the cashu protocol, it changes the format of the token. the old client can't read those yet, that's why it fails there.  I created a container for the latest version and updated the pr.

As for the crypto-lib issues, could you try what they explain here? https://github.com/damienbod/angular-auth-oidc-client/issues/1320#issuecomment-1476338328

If it works after that, then we would at least know if that it is indeed the issue. If it is, I don't know how easy it would be to fix that. how do other projects solve that issue? I would guess a lot of apps rely on crypto libs

Thanks!

gandlafbtc:
BTW @surajmahraj , if you still have config for the theme used in the pictures, you can send it to me or open a pr in https://github.com/gandlafbtc/nutstash-wallet , so we can let the users have a "nutty" theme

nmfretz:
> ...We've made an update to the cashu protocol, it changes the format of the token. the old client can't read those yet, that's why it fails there....

Thanks. I can confirm that this fixed the invalid token issue.

> As for the crypto-lib issues, could you try what they explain here?

Ah damn. Ya, this is what I feared. `window.crypto.subtle` doesn't exist over http so there is no `window.crypto.subtle.digest`. 

The developer of the Snort app had a similar issue. Check out this suggestion from @lukechilds re a webcrypto polyfill: https://github.com/getumbrel/umbrel-apps/pull/411#issuecomment-1447235072. I don't think Snort ended up going this route though... instead they require key manager (like alby) over http to handle event signing.

gandlafbtc:
Yes, I was thinking that we might need to go the shim/pollyfill route. Thank you for the research and testing, I will take a look into it.

nmfretz:
Ping me whenever you need help testing. Really looking forward to self-hosting this.

nmfretz:
> BTW @surajmahraj , if you still have config for the theme used in the pictures, you can send it to me or open a pr in https://github.com/gandlafbtc/nutstash-wallet , so we can let the users have a "nutty" theme

Briiiiiing on the nutty theme. Suraj ended up just using an overlay when designing the images and playing with the blend modes. Here's the relevant HEX codes: 443A2A (text), 664625 (primary dark), C9A163 (primary), E6E6E6 (off-white)

gandlafbtc:
https://www.npmjs.com/package/webcrypto-shim maybe this is something we can try

gandlafbtc:
https://github.com/PeculiarVentures/webcrypto-liner or this

gandlafbtc:
> https://github.com/PeculiarVentures/webcrypto-liner or this

I could make it work with this, check the branch https://github.com/gandlafbtc/nutstash-wallet/tree/webcrypto-shim

gandlafbtc:
I updated the PR! note, I changed a bunch of stuff:
* The app changed a bit (some new features, some new bugfixes) this docker image is the alpha release of the next version v0.1.10
* I switched to a multistage docker image build
* i added webcrypto shim for the docker images. I tested it in a http env, and it worked!

nmfretz:
You Legend!!! Minting, sending, receiving all work great!

**One thing to confirm before this goes live:** 
I noticed that for the newest image, the UI is served on port 3000, so I had to change `APP_PORT: 4173` in the `docker-compose.yml` to `APP_PORT: 3000` to get it up and running on Umbrel. I just want to confirm that port 3000 is what you intend to use, and not something from your development environment that made its way into the alpha image build. If so, the only thing that needs to change is `APP_PORT` in the `docker-compose.yml`. The `port` variable in `umbrel-app.yml` can remain unchanged.

gandlafbtc:
Oh. Yeah sorry about that. Since I switched to a multi stage build, the image now runs the app via node, where the default `port` is `3000`. It used to start the app in preview mode, where the default was `4173`. For me it doesn't matter which port we use :) . I will push another update, where I use minified dependencies for the shims, that are included in the image and not loaded via CDN. 

Will update the PR shortly!

nmfretz:
Looks like this newest image is back to using 4173. Do you want to keep it this way and change back the `APP_PORT` to 4173?

gandlafbtc:
i did an oopsie. I built from the wrong branch. let me fix that

gandlafbtc:
done. Fun fact: the image is 125MB smaller thanks to multistage build!

One issue: since switching to node, the container doesn't seem to exit when I kill the app (with ctrl+C), might this be a problem?

nmfretz:
Looks good! Everything is working. Shall we go live?

> done. Fun fact: the image is 125MB smaller thanks to multistage build!

Nice, that's some decent savings.

> One issue: since switching to node, the container doesn't seem to exit when I kill the app (with ctrl+C), might this be a problem?

This doesn't appear to cause any problems for us. All methods of stopping and starting the container, as well as restarting, updating, etc the app on Umbrel work just fine.

gandlafbtc:
Let's do it!

nmfretz:
It's live! Here it is on our public app store preview: https://apps.umbrel.com/app/nutstash-wallet.

Thanks a tonne for all your work getting this ready for self-hosting on Umbrel. Very stoked to add this to my own device. And thanks @aphex3k for your help on the initial PR as well.

nmfretz:
FYI, if you're ever looking to include a badge on your site that shows nutstash is available on Umbrel, our public app store site generates badges for all apps dynamically: https://twitter.com/umbrel/status/1633112340034998274?s=20

https://apps.umbrel.com/api/app/nutstash-wallet/badge-dark.svg
https://apps.umbrel.com/api/app/nutstash-wallet/badge-light.svg
https://apps.umbrel.com/badge-dark.svg
https://apps.umbrel.com/badge-light.svg

gandlafbtc:
LFG! Thank you everyone for the help.

I will have to push an update soon, there is a bug that snuck in: https://github.com/gandlafbtc/nutstash-wallet/pull/79
How easy is it to push updates to umbrel users? should I collect changes, into releases, or could I publish a single bugfix?

@nmfretz

nmfretz:
Ah sorry @gandlafbtc, I completely missed this. 

We can just publish a single bugfix.  Nothing is too small. User's will see that an update is available from the app store when they log in to their device:

image

For an update, open a PR to the same repo you submitted the app to (this repo), with the following changes:
1) Update the docker image in the `docker-compose.yml` to the new multi-arch tagged image
2) Update the `version` in the `umbrel-app.yml`. This is what triggers our software to show the user that an update is available. If you are keeping the version number the same you can just do something like `version: "0.1.10-hotfix"`
3) Update the `releaseNotes` section with anything you think is relevant. Can be as simple as `minor bug fixes` if you want.

gandlafbtc:
no problem @nmfretz . i have some changes ready, and am testing the feature, "sync tokens with server" (exclusive for umbrel/selfhosted) 😎. I think i will release everything together in the next version!

PR #461, opened 2023-03-18
Updated Passky Server to v8.1.4

zigazajc007:


nmfretz:
Thanks for this @zigazajc007! I have updated the `version` in the `umbrel-app.yml` so that an available update will be shown to user's in the dashboard UI. I also added some generic release notes.

Tested on amd64 and arm64. Going live now.

PR #462, opened 2023-03-20
Update Plex to latest version

0x346e3730:
Title says it all


nmfretz:
Leaving this here for future reference. I struggled a bit to find release notes for plex media server. This is the best resource I could find:
https://www.videohelp.com/software/Plex/version-history

PR #470, opened 2023-03-27
Update Torq to v0.19.2

maxwedwards:
Small update to Torq

If LND goes down for any reason Torq wasn't reconnecting so we thought it was probably worth the release

Tested on real raspberry pi


nmfretz:
Thanks for this quick fix @maxwedwards. Tested on both amd64 and arm64. Going live.

PR #485, opened 2023-03-31
Update Snort to v0.1.6

v0l:
Bumping version v0.1.6


nmfretz:
Thanks @v0l. Tested on amd64 and arm64. Going live now.

PR #491, opened 2023-04-07
update WoofBot to v0.8.2

woofbotapp:
update both woofbot and woofbot-lnd -
they use the same image, with different
permissions.

Tested on my Raspberry Pi.


nmfretz:
Thanks @woofbotapp. Nice lightning invoice addition!

Tested both apps on arm64 and amd64. Going live.

PR #493, opened 2023-04-10
Update LND to v0.16.2-beta

kroese:
Update LND to v0.16.2-beta

### DIFF
--- a/lightning/docker-compose.yml
+++ b/lightning/docker-compose.yml
@@ -42,7 +42,7 @@ services:
 
   lnd:
     hostname: "${DEVICE_DOMAIN_NAME}" # Needed so LND can generate a valid cert
-    image: lightninglabs/lnd:v0.16.1-beta@sha256:d2044a48fbe35b2941a5fc19328b1314bbe2cf68b628754ada9990b7e6dc1e3b
+    image: lightninglabs/lnd:v0.16.2-beta@sha256:795d29d94f845bd070832b097ff75cf0f46cf6daf5e55bf62dc1e4e406e0027d
     command: "${APP_LIGHTNING_COMMAND}"
     user: 1000:1000
     restart: unless-stopped
--- a/lightning/docker-compose.yml
+++ b/lightning/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3006
 
   app:
-    image: getumbrel/umbrel-lightning:v1.1.1@sha256:43e22766c38d19bdedc54e8eb9f4ce621d9e27157eae20b20a20fcf72b52405f
+    image: getumbrel/umbrel-lightning:v1.1.2@sha256:3ab819f6335abebb160d2c7b4bd979ad706e25208d219beb9f82d4c1dadd3eff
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for submitting this @kroese. Ya, it's a bit weird: sometimes when we update just LND and not the UI we include relevant LND release notes.

Sorry for not responding sooner when I started making changes. I was taking a look through the release notes and saw that there is a [breaking change](https://github.com/lightningnetwork/lnd/blob/master/docs/release-notes/release-notes-0.16.0.md#:~:text=Note%20that%20this%20pathfinding%20change%20introduces%20a%20breaking%20change%20in%20lnd.conf%20apriori%20parameters%20under%20the%20routing%20section%2C%20see%20sample%2Dlnd.conf%20for%20an%20updated%20configuration.%20The%20behavior%20of%20lncli%20setmccfg/getmccfg%20is%20altered%20as%20well.) related to `lnd.conf` that will cause LND to crash on update. I'm just updating the relevant logic in [the app itself](https://github.com/getumbrel/umbrel-lightning) to handle this and then I will update the app image in this PR

nmfretz:
It's locked and loaded! Thanks for the reminder.

nmfretz:
Thanks for bringing that to our attention! I looked into it:

We persist everything in a bind mount with `${EXPORTS_APP_DIR}/data/lnd:/data/.lnd`. We run as non-root and set `$HOME=/data ` so nothing will be written to /root. This means that there is no issue with data persistence, but it has the side-effect you noted where a new anonymous volume defined in [LND's Dockerfile]( I've confirmed that
https://github.com/lightningnetwork/lnd/blob/158e93f11440c4be1a284ff476799f734a5e3594/Dockerfile#L33) is created every time the container is brought up and down, and the old anonymous volumes are not deleted.

We'll look into cleaning up unused anonymous volumes in our app start/stop/restart/etc logic.

nmfretz:
Also, I'm now just waiting on an imminent (today?) LND 0.16.1 release before updating the app.

nmfretz:
Thanks for your patience on this @kroese, and thanks for continually updating the LND image as new ones were released. I pushed commits to your patch-1 branch, so please pull to avoid conflicts down the road.

- I bumped the app image, which fixes https://github.com/getumbrel/umbrel/issues/1594, https://github.com/getumbrel/umbrel/issues/1344, and https://github.com/getumbrel/umbrel-dashboard/issues/299
- I also updated the release notes.

Going live ⚡️

nmfretz:
Hey, sorry I hadn't pushed to https://github.com/getumbrel/umbrel-lightning yet. It should now be viewable.

nmfretz:
Ya, I get your point here. This is the main drawback with our current development structure. The reasons we are operating this way right now are:
- It helps us with clear project management and issue tracking internally within the team
- It allows us to build features in a way that we can retain excitement around the release

Our longer-term plan after we have a stable release of our OS is to do less and less work in private repos and move the work public branches which will allow for feedback on changes before they are committed to master.

Very open to more feedback on this.

PR #494, opened 2023-04-10
Bump Elements to 22.1.1

tiero:
This also picks a more recent block for assume valid.

### DIFF
--- a/elements/docker-compose.yml
+++ b/elements/docker-compose.yml
@@ -16,8 +16,9 @@ services:
     volumes:
       - ${APP_DATA_DIR}/data:/home/elements/.elements
     command:
-      - -listen=1
+      - -listen=0
       - -txindex=1
+      - -trim_headers=1
       # Current Elements Core requires >8GB of RAM to verify confidential proofs during IBD
       # We skip historical block verification for now, as we don't have a way to set the memory limit
       - -assumevalid=d5c8278b7c1c926eea46776445aaa94906a0e2a753fa007bda1841139cfb6543
--- a/elements/docker-compose.yml
+++ b/elements/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8080
 
   node:
-    image: ghcr.io/vulpemventures/elements:0.21.0.2@sha256:60f082b13b366743ef3cd113cb20397b586fd7032c0eb9c0a2a242ebc528f537
+    image: ghcr.io/vulpemventures/elements:22.1.1@sha256:7cb6c86df3a9332dd950e1774b2b34e33084b81fde08f2655334adc2285d90ac
     restart: on-failure
     stop_grace_period: 1m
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this update @tiero!

Can you please also update the following in the `umbrel-app.yml`:
- `version`: update to `22.1.1`. This line is what triggers the app store in Umbrel to show users that an update is available.
- Add `releaseNotes`. See [here](https://github.com/getumbrel/umbrel-apps/blob/master/lightning/umbrel-app.yml) for example. The `umbrel-app.yml` is a little finicky. You need to make sure that any text in the `releaseNotes` section is indented as shown in the example. If you want something to appear on a newline, then add a blank line. If you want something to have a blank line between it then add 2 newlines.

tiero:
> What's your thoughts on this? Have you tested successfully?

I don't have a RPi, I tested on an online machine.

In terms of performance, I guess we should forward to Blockstream devs (as I am only making the docker packaging here). Do you have clue on how to make it run on RPi? Any extra options to activate in elements.conf? @psgreco

psgreco:
to activate the memory reduction, you have to add `trim_headers=1` to the config. The drawback of adding this is that the node can't be used to feed blocks to other nodes, so it's almost like having `listen=0`

tiero:
> trim_headers=1

Thanks! That was it provably! Do you suggets to have also `listen=0` on?

tiero:
> actually I just quickly did this. Let me know if you think it looks okay

Amazing! Looks perfect, thx

nmfretz:
Thanks @psgreco and @tiero I will test again.

@tiero Are you okay with the implications of this drawback:
> The drawback of adding this is that the node can't be used to feed blocks to other nodes, so it's almost like having listen=0

tiero:
>  implications of this drawback
I think it's a good compromise for the time being for users to have a Liquid wallet to be used from their node

psgreco:
We're hoping that this limitation is just temporary, but yeah, it should help in many situations. TBF, the theory is that the node should be able to provide blocks, but not headers, the problem is that there is no such flag in the protocol.

@nmfretz I wanna see numbers!

nmfretz:
Darn, I'm still seeing relatively high RAM usage with this new config. Here's an aws instance that is currently running (100% synced) and it's using 6GB:

```
CONTAINER ID   NAME              CPU %     MEM USAGE / LIMIT    MEM %     NET I/O           BLOCK I/O         PIDS
52b9824c0e28   elements_node_1   4.91%     5.797GiB / 15.5GiB   37.40%    27.8GB / 1.99GB   9.63GB / 62.2GB   15
```

Since this behaviour isn't something new that's being introduced, I feel okay pushing this update and then we can continue to look into lowering RAM usage for future updates. I will just add a note below the main app description that outlines RAM requirements.

Thanks again for this update @tiero!

PR #499, opened 2023-04-13
Update torq to v0.20.3

kroese:
This should fix the huge CPU usage Im seeing by the current container.


maxwedwards:
Thanks for the PR @kroese, I created one too just with slightly different release notes.

nmfretz:
Thanks @kroese. Merged https://github.com/getumbrel/umbrel-apps/pull/500.

PR #500, opened 2023-04-14
Update Torq v0.20.3

maxwedwards:
Fixes high CPU usage and adds peers page

Tested on a real RaspberryPi

I realise there is already a PR  for this (#499) but we went for simpler release notes and a warning that db migrations might take a while on first start-up as they clean up a lot of old logs.

### DIFF
--- a/torq/docker-compose.yml
+++ b/torq/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 7028
 
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this @maxwedwards. And thanks @kroese for your PR as well, much appreciated.

Tested on arm64 and amd64. As a datapoint, database migration on my RaspberryPi took about 5 minutes.

PR #502, opened 2023-04-14
update WoofBot to 0.9.0

woofbotapp:
Tested on my Raspberry Pi


nmfretz:
Thanks @woofbotapp! Tested on arm64 and amd64. I like the new /about command.

PR #505, opened 2023-04-21
Update: Snort

v0l:
Moving to 0.1.7


nmfretz:
Thanks @v0l. I added some white space to the release notes so they appear nicely in the app store UI. Please pull down changes to your local repo to avoid conflicts.

Tested on amd64 and arm64. Going live.

PR #509, opened 2023-04-25
Update Urbit to v2.2

yapishu:
- Updated Urbit to official v2.2 binary: https://github.com/urbit/vere/releases

PR #512, opened 2023-04-28
Revert "Update Urbit to v2.2"

yapishu:
Reverts getumbrel/umbrel-apps#509

It turns out the 2.2 binary is not actually released yet and this breaks the app :( I'm not sure if there is a preferred way to submit this as a PR but happy to oblige.

### DIFF
--- a/urbit/docker-compose.yml
+++ b/urbit/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   manager:
-    image: mopfelwinrux/urbit-umbrel:v2.2@sha256:936b83f6047bb70889a2c154b1fe8b915702affa050fb0ef7b533cceb16c7093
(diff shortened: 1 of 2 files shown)


yapishu:
@nmfretz sorry to add to the hassle, but 2.2 was for real released today -- should I just close this and create a new PR with it?

Btw, the broken image will still launch the management UI successfully, the problem is that the urbit binary isn't present, so when you try to boot a ship in the app it will fail

nmfretz:
No problem at all @yapishu.
> should I just close this and create a new PR with it?

Opening a new PR would be great, thank you. In terms of version number for the `version` field in the `umbrel-app.yml`, you will need to choose something other than `v2.2` so that users' Umbrels are triggered to show an available update in the App Store. It's completely up to you what you decide to choose, but san idea could be to do something like `v2.2-hotfix`

PR #514, opened 2023-05-01
Update transmission to 4.0.3

route:
Dunno why it's v3, I just bumped to v4 and everything works just fine.

### DIFF
--- a/transmission/docker-compose.yml
+++ b/transmission/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 9091
 
   server:
-    image: linuxserver/transmission:version-3.00-r5@sha256:f0b885f211dd6fde7b64f0d6b9e5931929cfc2331fb83d6f3171e412f22b3f87
+    image: linuxserver/transmission:4.0.3@sha256:fdd97348ad51844411a6d97e9393df8c29d13581b67d2be1fb3f0919debe2bbd
     environment:
       - PUID=1000
       - PGID=1000
--- a/transmission/docker-compose.yml
+++ b/transmission/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 9091
 
   server:
-    image: linuxserver/transmission:4.0.3@sha256:fdd97348ad51844411a6d97e9393df8c29d13581b67d2be1fb3f0919debe2bbd
+    image: linuxserver/transmission:4.0.3@sha256:be6b4fd40e4baf685633a9ccc7974fe350115fb4b2afdf244ebc0ccdf9a33cb7
     environment:
       - PUID=1000
       - PGID=1000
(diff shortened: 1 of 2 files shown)


transmission/docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       APP_PORT: 9091
 
   server:
-    image: linuxserver/transmission:version-3.00-r5@sha256:f0b885f211dd6fde7b64f0d6b9e5931929cfc2331fb83d6f3171e412f22b3f87
+    image: linuxserver/transmission:4.0.3@sha256:fdd97348ad51844411a6d97e9393df8c29d13581b67d2be1fb3f0919debe2bbd
```
nmfretz, line 10: We need to include the sha256 digest for the mult-arch Docker image, so that the app works for Umbrel users on both arm64 and x86(amd64) devices. 

In this case, it looks like the correct digest for 4.0.3 is `sha256:be6b4fd40e4baf685633a9ccc7974fe350115fb4b2afdf244ebc0ccdf9a33cb7`

You can get this by running the command `docker pull <image>` and seeing the digest in the output. For example:

```
$ docker pull linuxserver/transmission:4.0.3
4.0.3: Pulling from linuxserver/transmission
bc9794e5ef26: Pull complete
d2e3055e0657: Pull complete
601b52927c1d: Pull complete
6dbe19f465c4: Pull complete
29ff433cf9b3: Pull complete
49f060fe6fcf: Pull complete
Digest: sha256:be6b4fd40e4baf685633a9ccc7974fe350115fb4b2afdf244ebc0ccdf9a33cb7
Status: Downloaded newer image for linuxserver/transmission:4.0.3
docker.io/linuxserver/transmission:4.0.3
```

transmission/umbrel-app.yml
nmfretz, line 1: Since this will be the first time this app is updated, we can add a `releaseNotes` section to this file. See here for example: https://github.com/getumbrel/umbrel-apps/blob/master/pi-hole/umbrel-app.yml

I quickly checked https://github.com/transmission/transmission/releases and there doesn't seem to be anything major that will break the app in Umbrel.

This can be as simple as:
```
releaseNotes: >-
  This release updates Transmission from version 3.00 to 4.0.3/
  
  
  Full release notes can be found at https://github.com/transmission/transmission/releases
```

nmfretz (commented):
Thanks for taking on this update @route, much appreciated!

> Dunno why it's v3, I just bumped to v4 and everything works just fine.

We rely on the community and app developers to help keep apps up-to-date, so thank you for doing this.

I have left a review below.

transmission/docker-compose.yml
route: Yep, looks like I got the wrong one somehow. Fixed

route:
Done!

transmission/umbrel-app.yml
route: Added

nmfretz:
Thanks again for this update @route. I made a tiny tweak to my previously suggested release notes.

I have tested on amd64 and arm64 and everything looks good. Going live now!

PR #515, opened 2023-05-02
App Submission: Joplin Server

danvergara:
# App Submission

### App name
Joplin Server

### 256x256 SVG icon

[https://void.cat/NF1f95ZbQAy9eQy5YrCouC](https://void.cat/NF1f95ZbQAy9eQy5YrCouC)

### Gallery images

[https://void.cat/HGrQV4qJfN8K4u5EJyrS7J](https://void.cat/HGrQV4qJfN8K4u5EJyrS7J)
[https://void.cat/qxFRwcZ6Z7BKqNEGvGgHA](https://void.cat/qxFRwcZ6Z7BKqNEGvGgHA)
[https://void.cat/Us5fS4hWCEqwnLS2PGTXcH](https://void.cat/Us5fS4hWCEqwnLS2PGTXcH)

### I have tested my app on:

I had to get the current host public IP to set the  `APP_BASE_URL` env var as a public URL. My solution is a little bit hacky  and I think it can be done using Nginx, but I don't know how to do it, since we don't have a fixed domain for every installation.

### DIFF
--- a/joplin-server/docker-compose.yml
+++ b/joplin-server/docker-compose.yml
@@ -14,6 +14,7 @@ services:
       - "22300"
     environment:
       - APP_PORT=22300
+      - APP_BASE_URL=
       - DB_CLIENT=pg
       - POSTGRES_PASSWORD=postgres
       - POSTGRES_USER=postgres
--- a/joplin-server/docker-compose.yml
+++ b/joplin-server/docker-compose.yml
@@ -10,8 +10,8 @@ services:
     image: joplin/server:2.10.10-beta
     restart: on-failure
     stop_grace_period: 1m
-    ports:
-      - "22300:22300"
+    expose:
+      - "22300"
     environment:
       - APP_PORT=22300
       - DB_CLIENT=pg
--- a/joplin-server/docker-compose.yml
+++ b/joplin-server/docker-compose.yml
@@ -3,18 +3,16 @@ version: "3.7"
 services:
   app_proxy:
     environment:
-      APP_HOST: joplin_web_1
+      APP_HOST: joplin-server_web_1
       APP_PORT: 22300
 
   web:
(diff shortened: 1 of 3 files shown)


danvergara:
Thanks for the support @nmfretz . Let me test this out over the weekend ot make sure it works as expected.

danvergara:
@nmfretz I think this is ready for review, but I have a little concern with my current approach.

danvergara:
I experimented with nginx but was pointless, adding a public URL/IP is mandatory for some odd reason that should be out of the scope of Joplin's configuration.

@nmfretz I think this is ready for review. If you have comments or questions or suggestions, let me know and I'll be more than happy to address them.

danvergara:
This is not working when the request is coming through the Tor network. Is there a way to tell when this happens to properly set the APP_BASE_URL  variable with the value of $TOR_PROXY_IP dynamically?

nmfretz:
Thanks for your hard work in submitting Joplin @danvergara. I just started reviewing, but noticed that the [joplin/server Docker images](https://hub.docker.com/r/joplin/server/tags) are **only** available for the amd64 architecture. This means that the app will break for users who run Umbrel on an arm device like a Raspberry Pi. We cannot accept any app that doesn't work on both arm64 and amd64.

I found [this closed issue](https://github.com/laurent22/joplin/issues/5932) on the Joplin GitHub that requested an arm64 image, but it didn't get any traction. I think the best bet is to be vocal on Joplin's GitHub about them building and pushing multi-arch images to Docker Hub.

nmfretz:
@danvergara I am closing this issue for now since there is no multi-arch docker image. I have added my voice to this Joplin issue here https://github.com/laurent22/joplin/issues/5932. Let's keep an eye on this and reopen if a multi-arch image gets built.

PR #516, opened 2023-05-02
Update Pi-Hole to latest

0x346e3730:
Everything is in the title. I also linked the changelog rather than copy/pasting it because it has a lot of links, it's easier this way I think

### DIFF
--- a/pi-hole/docker-compose.yml
+++ b/pi-hole/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   server:
-    image: pihole/pihole:2023.03.1@sha256:56c7c10facf93eaedebabec4b7ac68b148aa2dbeed4a28c92d53e0daba5f87b3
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this update @0x346e3730. I agree with your release notes approach for pi-hole. I just added a note as to what versions of the components the release includes.

I also corrected the Docker image digest so that it refers to the multi-arch digest, which we need to reference so that the app will run for both arm64 and x86 users. In the future if you want to grab the multi-arch image you can run  `docker pull <image>` and use the digest displayed in the output.

For example:
```
$ docker pull pihole/pihole:2023.03.1
2023.03.1: Pulling from pihole/pihole
Digest: sha256:a74dde4800f54d3c0b0839babbac9f2cc7e4b8239ab4a5bc2c25c7328ec1c019
Status: Downloaded newer image for pihole/pihole:2023.03.1
```

Going live 🥧

0x346e3730:
Didn't know about that multi-arch digest, took the one from https://hub.docker.com/layers/pihole/pihole/2023.03.1/images/sha256-56c7c10facf93eaedebabec4b7ac68b148aa2dbeed4a28c92d53e0daba5f87b3?context=explore and indeed it's linked to a specific arch. Thank you for letting me know 👍  And thank's for merging !

PR #519, opened 2023-05-04
Update-BTCPay-Server-to-1.9.2

JimSB:
Updating to 1.9.2

### DIFF
--- a/btcpay-server/docker-compose.yml
+++ b/btcpay-server/docker-compose.yml
@@ -31,7 +31,7 @@ services:
       NBXPLORER_BTCHASTXINDEX: 1
 
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this, @JimSB! I have reviewed and tested on amd64/arm64.

Since there were a number of updates between 1.7.12 and 1.9.2, I decided to modify the release notes section so that it references some of the major new features in a general way and then links to the full release notes. What do you think of this approach?

JimSB:
Sounds good to me, https://github.com/btcpayserver/btcpayserver/releases being a consitent place to see patch notes
Unless maybe @Kukks things we should include anything else! 
ty!

PR #520, opened 2023-05-04
update WoofBot to 0.9.1

woofbotapp:
Tested on my Raspberry Pi.

Updated the [lightning package](https://www.npmjs.com/package/lightning) dependency to fix a bug with the `/about` command. Also fixed a few typos.


nmfretz:
Thanks @woofbotapp. Checked both apps on arm64/amd64. Going live 🐶

PR #521, opened 2023-05-05
Update Urbit to v2.3

yapishu:
- Updated Urbit to official v2.3 binary: https://github.com/urbit/vere/releases


nmfretz:
Excellent, thanks for this. Tested on amd64 and arm64. Going live now.

PR #522, opened 2023-05-05
Update LNbits to 0.10.5

nmfretz:
This PR addresses two concerns:
1. Critical bug: https://twitter.com/lnbits/status/1654416778804580352?s=20
2. Newer versions of LNbits have extensions [removed by default](https://github.com/lnbits/lnbits/releases/tag/0.10) for better scalability and decreased attack vectors. Users now need to be at least Admin-level to manage extensions.

(1) Is fixed by bumping LNbits to version 0.10.5
(2) Needs to be fixed by ~~setting new and existing users as Admins~~ giving users access to an admin or super user url. Currently, if a user were to update to the newest LNbits version, any extensions they had previously isntalled would disappear and there would be no way to manage extensions without manually updating `LNBITS_ADMIN_USERS` in the `docker-compose.yml` to include a relevant user id, which would reset on future app update.

Pinging @arcbtc and @lukechilds for troubleshooting help to get this out quickly.

### DIFF
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -32,5 +32,4 @@ services:
       LNBITS_DEFAULT_WALLET_NAME: "LNbits wallet"
       LNBITS_DISABLED_EXTENSIONS: "amilk"
       LNBITS_ADMIN_EXTENSIONS: "ngrok"
-      LNBITS_ADMIN_USERS: ""
       LNBITS_ADMIN_UI: "true"
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   web:
-    image: lnbitsdocker/lnbits-legend:0.10.5@sha256:602dd4a490605ffeb33a1223cb824b605c062112867085c0146b9847d935b662
+    image: lnbitsdocker/lnbits-legend:0.10.6@sha256:a11aaa6d2b211db4c1ec2cafb7c35057ef45bcf4b98fb3bbf23c5635a0304aff
     init: true
     restart: on-failure
     stop_grace_period: 1m
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   web:
-    image: lnbitsdocker/lnbits-legend:0.9.7@sha256:1fccd9611a23b9d9259ca4adff860e3e2d441ed42261d192141e76b261cd3b74
+    image: lnbitsdocker/lnbits-legend:0.10.5@sha256:602dd4a490605ffeb33a1223cb824b605c062112867085c0146b9847d935b662
     init: true
     restart: on-failure
     stop_grace_period: 1m
@@ -32,3 +32,5 @@ services:
       LNBITS_DEFAULT_WALLET_NAME: "LNbits wallet"
       LNBITS_DISABLED_EXTENSIONS: "amilk"
       LNBITS_ADMIN_EXTENSIONS: "ngrok"
+      LNBITS_ADMIN_USERS: ""
+      LNBITS_ADMIN_UI: "true"
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -33,3 +33,4 @@ services:
       LNBITS_DISABLED_EXTENSIONS: "amilk"
       LNBITS_ADMIN_EXTENSIONS: "ngrok"
       LNBITS_ADMIN_UI: "true"
+      SUPER_USER: "$APP_PASSWORD"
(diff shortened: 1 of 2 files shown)


nmfretz:
For our reference. @arcbtc on twitter (Thanks Ben!):
> [@dnilabs](https://twitter.com/dnilabs) made a useful file that's created if env ADMIN_UI is set to true, it will put it in a text doc on route, so you easily pass link to your users.
From that link users can install extensions, and all their previous entries will be there.

nmfretz:
Okay, I've played around a bit. @arcbtc do you see anything below that I have misunderstood?:
- I found the relevant commit that saves the super user id at `/app/.super_user` (inside the Docker container in our case): https://github.com/lnbits/lnbits/commit/667de4d1f092f3a3d5ce0a4fad5108da3db2d58f. Additionally, the entire url is logged in the container.
- `/app/.super_user` is created when we include `LNBITS_ADMIN_UI: "true"` in the `docker-compose.yml`.
- For an Umbrel accessible at umbrel.local, the super user account is accessible at `umbrel.local:<APP_PORT>/wallet?usr=<user-id-from-.super_user-file>`.
- By default, this url will include a wallet named `LNbits wallet`.
- The super user can manage extensions, and also has access to the full admin dashboard where additional admin users can be set,  settings can be tweaked that override settings from the env vars we set in `docker-compose.yml`, etc.
- The default behaviour is that any extensions installed by a super user or admin then becomes available to other users, who can then enable those extensions for themselves.

@lukechilds - From what I can tell, adding `LNBITS_ADMIN_UI: "true"` to our compose file is a good solution for us. New and existing users can just visit the super user url to install extensions and/or make existing wallets into admins. We just need to convey the process clearly to users and give them a simple way to access the correct super user url. 

I'm trying to think of the best way to do this. Should we create a hook to boot up lnbits, grab the id saved in `/app/.super_user` and then display a url to the user in the app store, similar to what we do for default passwords and usernames?

lnbits/docker-compose.yml
nmfretz, line 1: Do we want to add or change the volume bind mounts to make sure we persist `.super_user`, which is located inside the Docker container at `/app/.super_user`? Perhaps we should just be mounting the entire `/app` directory.

```
# ls -a /app
.   .dockerignore  .flake8  .prettierignore  Dockerfile  flake.lock  lnbits  poetry.lock     tools
..  .env.example   .github  .super_user      data	 flake.nix   nix     pyproject.toml
```

nmfretz: Looks like `LNBITS_DATA_FOLDER: "/data"` should ideally be set to `/app/data` since there is a default `data` folder in /app. See comment above.

Not sure this is worth the migration headache though

lukechilds: If we don't persist it, it gets written out with the same contents each time right? If that's the case then we don't need to persist. It's not trivial because mounting the file directly is problematic because Docker will create an empty dir on first start if it doesn't exist, and mounting `/app` means app is an empty dir on fresh install unless we check the entire lnbits source into the app repo.

lukechilds:
>@lukechilds - From what I can tell, adding LNBITS_ADMIN_UI: "true" to our compose file is a good solution for us. New and existing users can just visit the super user url to install extensions and/or make existing wallets into admins. We just need to convey the process clearly to users and give them a simple way to access the correct super user url.
>
>I'm trying to think of the best way to do this. Should we create a hook to boot up lnbits, grab the id saved in /app/.super_user (should bind this) and then display a url to the user in the app store, similar to what we do for default passwords and usernames?

Sounds good, so new users just become admin when they setup the first account right? It's only existing users that we need to explain things have changed and how to access the admin?

Also for existing users does it upgrade the first created account to admin? Or there is a totally new account that never existed before that is now admin?

lnbits/docker-compose.yml
nmfretz: ah yes, I've confirmed that with our current setup `.super_user` gets written out with the same contents on app stop/start/restart/update. Thanks for this.

nmfretz:
> Sounds good, so new users just become admin when they setup the first account right? It's only existing users that we need to explain things have changed and how to access the admin?

Unfortunately no. The way things appear to work with the current version of the app on Umbrel is that no users are admin. This does not pose any issues because in this version, non-admins can manage extensions. 

In the newer lnbits version; however, extensions are now removed by default and can only be installed by an admin or super user account. The way that  `LNBITS_ADMIN_UI=true` solves this issue is that it creates a super user account and spits out that account id to `/app/.super_user`.  Both new and existing users can then access that account if we provide them with the url. From this account they can install extensions and those extensions will be available to be enabled in all other wallets/accounts. A user can also upgrade their existing wallets/accounts to admin status from the super user page.

As far as I can tell, there is no way to make the first account an admin apriori. You would have to create an account, and then add that account to the env var `LNBITS_ADMIN_USERS`, which will not work for our use case.

nmfretz:
A huge thank you to the LNbits team for [merging a new feature](https://github.com/lnbits/lnbits/pull/1688) on priority to allow the creation of a super user account from a valid hex string supplied as an environment variable. Thank you @motorina0, @dni, @callebtc, @arcbtc, and any others who helped get this done.

PR #523, opened 2023-05-06
Update Specter Desktop to v2.0.1

usernameisJim:


JimSB:
I did add release notes in case we wanted to include

Also we may want to add this line from time of release notes somewhere:

> "Please create a full backup before migrating or any major internal changes like switching to an electrum based installation. You can easily create a backup in Settings --> Backup Specter (zip file)"

nmfretz:
Thanks @JimSB I will take a look at this soon! 
Pinging @aphex3k for general thoughts on this update if you have any? We had previously decided to hold off on the v2.0.0 update: https://github.com/getumbrel/umbrel-apps/pull/450

nmfretz:
Tested on amd64 and arm64. As-is, there are no issues for existing users who update, because the previous connection to the local Bitcoin Core instance is maintained. This was set up automatically for users via the environment variables passed to the `web` container:
https://github.com/getumbrel/umbrel-apps/blob/7f3335f400d1ce1a8689c1b63a29afd4f27ac73a/specter-desktop/docker-compose.yml#L19-L25

However, the initial set up of Specter Desktop has changed with the introduction of the option to connect to Bitcoin Core **or** a public Electrum Server:
image

The connection to Bitcoin Core is not made by default with the current `docker-compose` setup, and so a user needs to manually retrieve the Bitcoin Node IP, rpcpassword, and rpcuser to set up their local instance of Bitcoin Core. This is not ideal for UX. We need to see if there is an environment variable we can pass to specter that sets up a Bitcoin Core connection by default.

nmfretz:
@moneymanolis in 2.0.1 is there still a way to connect to a Bitcoin Core node by passing in environment variables? Previous versions prior to the introduction of the Electrum Server option allowed us to just pass in these:
https://github.com/getumbrel/umbrel-apps/blob/7f3335f400d1ce1a8689c1b63a29afd4f27ac73a/specter-desktop/docker-compose.yml#L19-L25

Now we would need some way of telling specter to connect to Bitcoin Core not an Electrum Server. The purpose of this is for easy initial set-up on Umbrel.

moneymanolis:
@nmfretz should still work like this. But, we haven't tested it yet.

nmfretz:
@moneymanolis It doesn't work unfortunately, but I may be missing a variable that tells Specter to connect to Bitcoin Core specifically instead of an Electrum server. When the app is installed I am met with this screen, where I am not connected to Bitcoin Core:

image

If I then go to add a Bitcoin Core configuration I see this form with nothing pre-populated:
image

moneymanolis:
Bringing @k9ert in, he is more familiar with the environment variables usage.

PR #525, opened 2023-05-08
Update Tailscale to v1.40.1

JimSB:
dockerhub had 1.40 and 1.40.0 
I used 1.40.0 becasue didn't see a 1.40 in [latest releases](https://github.com/tailscale/tailscale/releases)


nmfretz:
Updated to 1.40.1 and tested on amd64 and arm64. Thanks @JimSB!

PR #528, opened 2023-05-11
Update RoboSats to v0.5.0-alpha

Reckless-Satoshi:
Update RoboSats to latest v0.5.0-alpha. 

This update is a must as the RoboSats coordinators have deprecated the auth method used in v0.4.0-alpha.

Untested, but exactly replicated the steps to create the v0.4.0-alpha PR https://github.com/getumbrel/umbrel-apps/pull/447

### DIFF
--- a/robosats/docker-compose.yml
+++ b/robosats/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 12596
 
   web:
-    image: recksato/robosats-client:v0.4.0-alpha@sha256:9f9a759be3a4faf1305ab8ec17d0041fc430b604767b1b4838015d365080433d
+    image: recksato/robosats-client:v0.5.0-alpha@sha256:b240e5909df5db571c815900506a55ba8ac7dd8741ba8e7dab0dced43bb0c71b
     restart: on-failure
     stop_grace_period: 1m
     init: true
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @Reckless-Satoshi. I pushed some very small formatting changes to the release notes. Please pull down the changes so I don't put you out of sync.

I tested on amd64 and arm64. There's one small item we should address before shipping this. For people who update the app, the v0.5.0 update warning modal still shows up every time you open robosats even though the client app has been updated to v0.5.0 (shown below). This does not happen for people who have never installed robosats before. It seems to be a client-side caching issue, because once I hard refresh the browser page the message never shows up again.

image

Did you want to handle this somehow in robosats, or do you think we should just add a warning to the release notes that instructs users to hard refresh `http://umbrel.local:12596/` if the warning still pops up after update?

Reckless-Satoshi:
Hey @nmfretz thank you for testing and fixing this PR!

> Did you want to handle this somehow in robosats, or do you think we should just add a warning to the release notes that instructs users to hard refresh `http://umbrel.local:12596/` if the warning still pops up after update?

Indeed, it seems the browser is caching the last client main.js . The only way as of now to tackle this issue is to clear browser cache by instructing the user to do `Ctrl+Shift+R` (or to click the third option in the pop up, will have the same effect) if the pop up shows up. 

By next release of the robosats client app we could improve this pop up to detect if this client is running on `:12596` and be explicit about the update process: 1) update the Umbrel app, and 2) if that has been done, then "Click to force refresh".

nmfretz:
Sounds good to me! I have updated the release notes with instructions.

Going live 🤖

PR #530, opened 2023-05-13
Update Urbit to v2.5

yapishu:
- Updated Urbit to official v2.5 binary: https://github.com/urbit/vere/releases


nmfretz:
Thanks very much, @yapishu. Tested on arm64 and amd64. Merging now.

PR #532, opened 2023-05-14
Update Ghostfolio to 1.268.0

robertclarkson:
Yahoo Finance changed something and this older version of ghostfolio is no longer working correctly. You are unable to add new assets on the portfolio screen. 

My discussions with the Ghostfolio developer can confirm this here: https://ghostfolio.slack.com/archives/C02DMFGS0B1/p1684051691077229

Here is an update to a near new version of Ghostfolio recommended by the Ghostfolio developer

### DIFF
--- a/ghostfolio/docker-compose.yml
+++ b/ghostfolio/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3334
 
   server:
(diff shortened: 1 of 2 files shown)


robertclarkson:
Will push a new PR with most recent ghostfolio

PR #533, opened 2023-05-14
Update app: Snort v0.1.8

v0l:


nmfretz:
Thanks @v0l. Tested on amd64 and arm64 and everything seems good. I made some very small changes to the whitespace in the `releaseNotes`. Going live ⚡️

PR #534, opened 2023-05-15
Update Ghostfolio to 1.269.0

robertclarkson:
Yahoo Finance changed something and this older version of ghostfolio is no longer working correctly. You are unable to add new assets on the portfolio screen.

My discussions with the Ghostfolio developer can confirm this here: https://ghostfolio.slack.com/archives/C02DMFGS0B1/p1684051691077229

Here is an update to the newest version of Ghostfolio recommended by the Ghostfolio developer

### DIFF
--- a/ghostfolio/docker-compose.yml
+++ b/ghostfolio/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3334
 
   server:
-    image: ghostfolio/ghostfolio:1.269.0@sha256:b3bdc2bfab3c5917cae124c6aee6a2411d6e17f1fdea1ef3845d434b4eb612d1
+    image: ghostfolio/ghostfolio:1.269.0@sha256:096ea10a7e9c8c471701da78d3aeb8434879e465b36b3564d6bb424bc47c2b3d
     restart: on-failure
     environment:
       NODE_ENV: production
--- a/ghostfolio/docker-compose.yml
+++ b/ghostfolio/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3334
 
   server:
-    image: ghostfolio/ghostfolio:1.241.0@sha256:0fd0b0bd722420bcc8fb92076f4bcfd8e1dcfb9b8b136367539e82184e992e1c
+    image: ghostfolio/ghostfolio:1.269.0@sha256:b3bdc2bfab3c5917cae124c6aee6a2411d6e17f1fdea1ef3845d434b4eb612d1
     restart: on-failure
     environment:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this @robertclarkson! I will review soon. I can't access the slack messages, but I see that @dtslvr approves with a thumbs up!

dtslvr:
Maybe @robertclarkson or @nmfretz can update the title of this issue to `1.269.0`? 🙂

ghostfolio/docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3334
 
   server:
-    image: ghostfolio/ghostfolio:1.241.0@sha256:0fd0b0bd722420bcc8fb92076f4bcfd8e1dcfb9b8b136367539e82184e992e1c
+    image: ghostfolio/ghostfolio:1.269.0@sha256:096ea10a7e9c8c471701da78d3aeb8434879e465b36b3564d6bb424bc47c2b3d
```
nmfretz, line 10: I updated this with the sha256 digest for the multi-arch Docker image. It needs to be the multi-arch digest so that the app can run on Umbrel's installed on both arm64 and x86 devices. You can get the multi-arch digest by running:
`docker pull ghostfolio/ghostfolio:1.269.0`, replacing `1.269.0` with whichever tag you are grabbing and the output of that command will include the multi-arch digest.

ghostfolio/umbrel-app.yml
```diff
@@ -31,19 +31,17 @@ path: ""
...
-  - Considered the user's language in the Stripe checkout
+  Version 1.269.0 release notes:
 
-  - Upgraded the Stripe dependencies
+  - Added FINANCIAL_MODELING_PREP as a new data source type
 
-  - Upgraded twitter-api-v2 from version 1.10.3 to 1.14.2
+  - Improved the market price on the first buy date in the chart of the position detail dialog
   
+  - Restructured the admin control panel with a new settings tab
```
nmfretz, line 45: In order to have each of these show up on their own separate lines in the UI, I added a newline between each.

nmfretz (commented):
Thanks again for doing this @robertclarkson! I made some changes and pushed to the app store instead of going back-and-forth with you to make sure users could add new securities as soon as possible.

I have included comments below for your reference in case you end up submitting more app updates (please do!).

Going live now 👻

robertclarkson:
Ah nice one @nmfretz  I don't have a development version of umbrel installed so i was unable to test what I'd submitted. 

Thanks for the info, i did wonder why the hash of the old version did not match with the one listed in docker hub!

PR #537, opened 2023-05-18
Update Urbit to v2.6

yapishu:
- Updated Urbit to official v2.6 binary: https://github.com/urbit/vere/releases


nmfretz:
Tested on arm64/amd64. Thanks for the regular updates @yapishu.

PR #547, opened 2023-05-23
Updating Nitter to Latest

JimSB:
Changed tag from[ ceramicwhite](https://hub.docker.com/r/ceramicwhite/nitter) from 8 months ago in dockerhub to https://hub.docker.com/r/zedeus/nitter/tags

I don't know the implications of this 

I did not change the redis tag

Without update Nitter is currently broken with "rate instance is limited" error message

this looks like the official tag/image above from developer Zedeus who is also helpful in their Matrix channel (confirmed previous version is broken right now): https://matrix.to/#/#nitter:matrix.org

Web UI for Matrix and link: https://app.element.io/#/room/#nitter:snopyta.org/$yuy1_I3Mn9D9q02RMtn7FikVIQIKE818cyowKYHxGSQ

Zedeus shared with me there's no currently versioning setup so updated the dates and didn't add any release notes...

### DIFF
--- a/nitter/docker-compose.yml
+++ b/nitter/docker-compose.yml
@@ -1,4 +1,4 @@
-version: "3.7"
+version: "2023.05.20"
 
 services:
   app_proxy:
@@ -9,7 +9,7 @@ services:
   web:
     # Official Nim image used as base image for nitter doesn't offer Arm variant so using a custom image w/ patch for umbrel
     # https://github.com/ceramicwhite/umbrel-patch-ci/blob/master/nitter/umbrel.patch
-    image: ceramicwhite/nitter:build-20220908@sha256:cfb1227aca5014804c47d10d7ff77f24416a3fd571549eaeee9850f915507447
+    image: zedeus/nitter:latest@sha256:2172275e119f39e5e2e4d554bf1bd70feae70ac872818ede2c1d566e601678ae
     restart: on-failure
     stop_grace_period: 1m
     init: true
--- a/nitter/umbrel-app.yml
+++ b/nitter/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: nitter
 category: Social
 name: Nitter
-version: "20220908"
+version: "2023.05.20"
 tagline: Browse Twitter without tracking or ads
 description: >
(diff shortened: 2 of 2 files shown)


nmfretz:
Thanks @JimSB. 

Unfortunately, the [official image from zedeus](https://hub.docker.com/r/zedeus/nitter/tags) is not multi-arch and only works for amd64, so we cannot use it out of the box.

@ceramicwhite - you did fantastic work packaging this app for Umbrel with some custom changes to the image build. Would you have any interest in updating your image to a version of nitter that no longer has rate limiting issues?

ceramicwhite:
So when this was initially added the nim-alpine didn't build for Arm64 so zedeus wasn't offering a build for the arch.
The other problem that still stands is that adding an envs from the conf to the compose won't work and Umbrel wouldn't allow submissions including the .conf & entrypoint.sh ( Is this still the case?) so I added them at build.

I see zedeus has fixed his arm64 dockerfile and is now pushing it to docker hub but they aren't creating a manifest list that include amd64 & arm64, which isn't problem we can just create our own manifest list with the official images if we could add the conf and entrypoint to the app repo.

Anyways in the meantime I've updated my build:
`ceramicwhite/nitter:build-38985af@sha256:b81f61f1527000d27daf5f222ea27122fee7bb489900587e732e252a5654c002`
https://github.com/ceramicwhite/nitter/tree/master/.github/workflows/dockerfile

I tested it works until you search a user but error is thrown from the umbrel app-proxy not nitter:
`There was an error connecting to Nitter. Error code: ECONNRESET` @nmfretz 
https://github.com/ceramicwhite/umbrel-apps/tree/nitter

nmfretz:
Thanks @ceramicwhite, much appreciated. We are still trying hard to stay away from including .conf and entrypoint.sh in the repo due to the complication it adds downstream for us when needing to make changes to these files without clobbering an existing users config.

> https://github.com/ceramicwhite/umbrel-apps/tree/nitter
> lmk if you want me to pr this branch instead

Yes, can you please PR that branch. I will close this one.

> There was an error connecting to Nitter. Error code: ECONNRESET

Ah, thanks for this. I will look into it.

PR #549, opened 2023-05-26
Update Torq to v0.22.4

laitang2000:
Torq can now we used with CLN! It is still an experimental version

Bug fixes and improvements

Tested on real Rasberry PI

### DIFF
--- a/torq/docker-compose.yml
+++ b/torq/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 7028
 
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this @laitang2000! I have tested on amd64 and arm64.

Would the LN Capital team be okay including a note in the release notes that Torq is now closed source? Our next Umbrel update will include the ability to mark apps as having Public vs Private source code, but for now Torq will still show in the app store as having Public source code, so it would be good to give the user an indication of this change:

image

laitang2000:
Hi @nmfretz, I've added the note you requested.

PR #551, opened 2023-05-26
Update WoofBot to 0.9.6

woofbotapp:
Tested on my Raspberry Pi.

This version does not have new features, but contains important performance improvements.


nmfretz:
Sorry for the delay on this @woofbotapp. Tested on arm64 and amd64. Going live 🐕

PR #606, opened 2023-05-30
App Submission - Homebridge

nmfretz:
https://homebridge.io/
https://github.com/homebridge/homebridge

Bringing HomeKit support where there is none. Homebridge allows you to integrate with smart home devices that do not natively support HomeKit. There are over 2000 Homebridge plugins supporting thousands of different smart accessories.


nmfretz:
Tested on amd64 and arm64. Going live to the app store 🏠🌉

PR #607, opened 2023-05-30
App Submission - Overseerr

nmfretz:
https://overseerr.dev/
https://github.com/sct/overseerr

Overseerr is a request management and media discovery tool built to work with your existing Plex ecosystem.


nmfretz:
Tested on arm64 and amd64. Going live 🎬

PR #609, opened 2023-05-30
App Submission - LibreSpeed

nmfretz:
https://librespeed.org/
https://github.com/librespeed/speedtest

Free and open source speedtest. Measure internet speeds between your devices and your Umbrel.


nmfretz:
Tested on amd64 and arm64. Going live 🏁

PR #610, opened 2023-05-30
App Submission - Bazarr

nmfretz:
https://www.bazarr.media/
https://github.com/morpheus65535/bazarr

Bazarr is a companion application to Sonarr and Radarr that manages and downloads subtitles based on your requirements. You define your preferences by TV show or movie and Bazarr takes care of everything for you.


nmfretz:
Tested on amd64 and arm64. Going live 🎬

PR #614, opened 2023-05-30
Update Urbit to v2.7

yapishu:
- Updated Urbit to official v2.7 binary: https://github.com/urbit/vere/releases


nmfretz:
Thanks @yapishu! I have tested on arm64 and amd64. Going live.

I made a small change to the whitespace in the release notes so that they show up like this:

image

yapishu:
Apreciado!

PR #615, opened 2023-05-30
Add explicit hostnames to Ghostfolio

YellowWords:
Fixes https://github.com/getumbrel/umbrel/issues/1630

### DIFF
--- a/ghostfolio/docker-compose.yml
+++ b/ghostfolio/docker-compose.yml
@@ -8,6 +8,7 @@ services:
 
   server:
     image: ghostfolio/ghostfolio:1.269.0@sha256:096ea10a7e9c8c471701da78d3aeb8434879e465b36b3564d6bb424bc47c2b3d
+    hostname: ghostfolio_server_1
     restart: on-failure
     environment:
       NODE_ENV: production
@@ -25,6 +26,7 @@ services:
   
   postgres:
     image: postgres:14-bullseye@sha256:c2a30d08a6f9e6c365595fd086c9e0436064c52425f15f72379ecf0807bac518
+    hostname: ghostfolio_postgres_1
     restart: on-failure
     stop_grace_period: 1m
     user: "1000:1000"
@@ -38,6 +40,7 @@ services:
 
   redis:
     image: redis:6.2-bullseye@sha256:5cdb2ac6f780e1d250787f4a887c1ed1166e3389757189ea913059409acc6f7c
+    hostname: ghostfolio_redis_1
     restart: on-failure
     user: "1000:1000"
     command: >


nmfretz:
Thanks for this PR @YellowWords. Great work troubleshooting your issue! 

I believe this hostname issue is happening because you are using a version of Docker Compose that is not compatible with the way we have hardcoded some service names in our `docker-compose.yml` files for apps. Compose V1 uses `_` when generating container names, while Compose V2 using hyphens `-`:
https://docs.docker.com/compose/migrate/#:~:text=Service%20container%20names&text=In%20Compose%20V2%2C%20a%20hyphen,network%20via%20consistent%2C%20predictable%20hostnames

I think your idea of setting explicit hostnames is a great solution that will allow apps to work on modern compose versions. Before we go down this route though, which will require making changes to all apps in this repo, could you please help me understand your specific case a little better?
1. Are you on a custom linux install?
2. Did you have an existing Docker Compose version installed prior to installing Umbrel or did you upgrade Docker Compose at some point?
3. Do you have other docker containers unrelated to Umbrel on your machine?
4. Can you please give me the output of these commands:
```
docker --version
docker-compose --version
```
Running these commands on Ubuntu 22.04.2 LTS with docker installation being handled by Umbrel install script I get:
```
ubuntu@umbrel-dev-test:~$ docker --version
Docker version 24.0.2, build cb74dfc
ubuntu@umbrel-dev-test:~$ docker-compose --version
docker-compose version 1.29.2, build unknown
```

5. Is your issue isolated to ghostfolio, or do are there other apps that you are experiencing issues with? If your Compose version is naming services with dashes instead of underscores I would assume that some other apps are stuck in `starting...` on your dashboard or have containers that cannot communicate with each other. Most of our apps sit behind an App Proxy container that uses container names with underscores to properly proxy traffic to the app's UI.

YellowWords:
Hello @nmfretz - That's very helpful context to some issues I've been having!

My Umbrel install is pretty custom, I run it on a old PC running Linux Mint 20.3 that I converted into a homelab. Besides Umbrel the homelab also runs [Tipi](https://github.com/meienberger/runtipi). I have 20 services running in Tipi and 4 in Umbrel. The ones on Umbrel are besides Ghostfolio: Bitcoin, Electrs and Mempool.

At current I have these versions:
```
 ~ % docker --version
Docker version 24.0.2, build cb74dfc
~ % docker-compose --version
Docker Compose version v2.18.1
~ % which docker-compose 
/usr/local/bin/docker-compose
~ % cat /etc/issue
Linux Mint 20.3 Una \n \l

~ % which docker-compose    
/usr/local/bin/docker-compose
```

For the most part running Tipi and Umbrel side by side worked pretty well up to this point - I ran into a couple of minor issues installing it on the same system as Tipi, like i.e. default port conflicts, but that was pretty minor stuff I could solve with some small edits of the compose files of both projects.

I did have one issue that was a bit thornier though which I actually [reported earlier](https://github.com/getumbrel/umbrel/issues/1512). I didn't comment at that time what the resolution was, but I eventually tracked it to docker-compose on the homelab. I noted that my docker-compose was installed [standalone](https://docs.docker.com/compose/install/standalone/), and version 1.x. Tipi OTOH uses the [plugin installation](https://docs.docker.com/compose/install/linux/) of docker-compose, so you invoke it like `docker compose $command` instead.

I thought that it was a 1.x installation was just a mistake on my part when installing, so I updated it to compose V2 at the time, which fixed the segfault issue I experienced earlier and didn't think much more of it. Up to this point everything worked well with compose V2 installed so my mind didn't go to it as a cause of the issue I'd been having.

I can try to downgrade the docker-compose standalone to 1.29.2 as the Umbrel installation script uses again. However I think that may cause some arcane issues where you're running `docker-compose` with a higher version on the same system. I'll do some searching and experimentation and get back to you.

YellowWords:
I did some quick testing and confirmed that If I switched to compose V1 Ghostfolio works out of the box. And I can uninstall apps now.

Regardless, my issues aren't related to Ghostfolio config as I thought initially and it's not your responsibility to get it to work on my bespoke system so I'll close the PR. Thanks anyway!

nmfretz:
@YellowWords I'm glad the context was helpful, and that downgrading compose got your Umbrel installation back up and running. Hopefully it doesn't cause new issues for you with Tipi.

> The ones on Umbrel are besides Ghostfolio: Bitcoin, Electrs and Mempool.

Ah, this makes sense as to why you weren't seeing issues with apps other than Ghosfolio. All three of those apps are special cases where we use the actual IP address of the containers instead of relying on the Docker bridge and Compose to resolve container names. See Bitcoin Node app proxy for example, where `$APP_BITCOIN_IP` is set to 10.21.22.2:
https://github.com/getumbrel/umbrel-apps/blob/b6c5dc6d8f10a31a8a4d5dad5ef680cf7f899d56/bitcoin/docker-compose.yml#L3-L7

Again, thanks for this PR. This Compose version issue is now on our radar and we will be considering addressing it down the line.

PR #618, opened 2023-06-02
Update ThunderHub to latest v0.13.19

aphex3k:


nmfretz:
Thanks for this update @aphex3k. Tested on amd64 and arm64. Going live ⚡️.

PR #619, opened 2023-06-04
feat(trilium-notes): bump version to 59.4

jgmontoya:
Updates trilium notes from version 55.1 to version 59.4 (currently latest stable)

### DIFF
--- a/trilium-notes/docker-compose.yml
+++ b/trilium-notes/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api/*"
 
   server:
-    image: zadam/trilium:0.55.1@sha256:9bf48a49359cd567bb8e9ce87c1d310713ba98ac8d517d75b07b8de907525077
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this PR to update Trilium Notes @jgmontoya! I made a few changes, so please pull to your local repo to avoid any conflicts:

- Indented and tweaked release notes: the yaml format is finicky about indentation and if there are any errors, the app won't show up in the app store UI. I indented the `releaseNotes` section to fix this https://github.com/getumbrel/umbrel-apps/pull/619/commits/f0b90da5d32ed9855d1de262424eee89dea97ed6

- Added the image digest for the multi-arch Docker image so that Umbrel users on both arm64 and x86 devices can use the app https://github.com/getumbrel/umbrel-apps/pull/619/commits/34eebe201020e800d70484d6508bcb798b98625d
  - You can get the multi-arch digest by running: `docker pull zadam/trilium:0.59.4`, replacing 0.59.4 with whichever tag you are grabbing and the output of that command will include the multi-arch digest if it is available.

I have tested on arm64 and amd64. Going live 📝.

jgmontoya:
Amazing @nmfretz, thanks!

PR #622, opened 2023-06-06
Update Audiobookshelf to 2.2.22

ceramicwhite:


nmfretz:
Thanks very much for this update @ceramicwhite! Tested on amd64 and arm64. Going live 🎧

PR #623, opened 2023-06-06
Update Whoogle to 0.8.2

ceramicwhite:


nmfretz:
Thanks for this update as well @ceramicwhite. Tested on amd64 and arm64 and it is working well. Going live 🔍

nmfretz:
I was playing around with Whoogle a little more and noticed that we don't bind mount a directory to persist the config, so it gets nuked on container restart. For next update, we could consider mounting:
```
volumes:
      - ${APP_DATA_DIR}/data/config:/config
```

PR #621, opened 2023-06-06
Update LND to v0.16.3-beta

kroese:
Update LND to v0.16.3-beta

PR #625, opened 2023-06-07
Update lnd app to v0.16.3-beta

aphex3k:
Update lnd app to v0.16.3-beta

The release notes in the prior manifest mentioned upgrading from a specific version, which I is not necessarily true for all upgrade paths...


aphex3k:
Sorry, just realized I duplicated #621.

nmfretz:
> The release notes in the prior manifest mentioned upgrading from a specific version, which I is not necessarily true for all upgrade paths...

Great point @aphex3k. I will make sure not to include this language in the future.

PR #627, opened 2023-06-07
Update Bitcoin Node to Core v25.0

nmfretz:
This PR updates the bitcoind image to the most recent 25.0 version


nmfretz:
Tested on amd64 and arm64. Also tested integration with other bitcoin and lightning apps.

PR #628, opened 2023-06-08
Update Nitter to zedeus/nitter 38985af

ceramicwhite:


ceramicwhite:
@nmfretz Can use official images going forward. Also the app proxy error is gone.

nmfretz:
Perfect, thank you for this

ceramicwhite:
> Perfect, thank you for this. I will play around with persisting the nitter.conf and not nuking existing users config when updating to this new image

Oh, I removed the volume because I noticed preferences were being saved as cookies. There is no need to mount a directory anymore.

nmfretz:
Sweet, okay tested on arm64 and amd64. Merging now. Thanks again for this fix.

I have kept the `./data/nitter` directory in this repo (even though we no longer bind mount to this directory), in case we want to persist state later down the line. By keeping the directory, we won't have to do any funny pre-start hook shenanigans to make sure that existing users pull down these directories.

PR #630, opened 2023-06-08
Update Urbit to v2.9

yapishu:
- Updated Urbit to official v2.9 binary: https://github.com/urbit/vere/releases


nmfretz:
Thanks for this update @yapishu. Tested on arm64 and x86. Merging now!

PR #631, opened 2023-06-09
Update WoofBot to 0.9.7

woofbotapp:
Just updating dependencies which broke a little with LND v0.16.3-beta.

Tested on my Raspberry-Pi.


nmfretz:
Thanks @woofbotapp. Tested on amd64 and arm64. /about works for me!

PR #637, opened 2023-06-16
Update Snort v0.1.11

v0l:


nmfretz:
Sorry for the delay @v0l.  It looks like `git.v0l.io/kieran/snort:v0.1.10` is only built for arm64, unlike previous tags that were multi-arch and so would work for both arm and x86 Umbrel users. Please ping me when the multi-arch image is up and I'll get on this right away to not delay you any further.

nmfretz:
@v0l pinging again in case you missed this.

v0l:
Hey, yea sorry forgot to come back to this, i guess Gitea doesn't support multi-arch builds because i certainly am building them, will push images to docker hub instead

nmfretz:
> i guess Gitea doesn't support multi-arch builds because i certainly am building them, will push images to docker hub instead

Damn that's too bad about Gitea. Thanks for pushing to docker hub. I changed the digest to the multi-arch one and tested on amd64 and x86. Working great!

PR #638, opened 2023-06-17
Update Immich to v1.61.0

nmfretz:
https://github.com/immich-app/immich/releases/tag/v1.61.0


nmfretz:
Tested on arm64 and x86. Going live 📸

rpjcbs:
Thanks Nathan

PR #640, opened 2023-06-19
Update mempool app description

hunicus:
Add mining dashboard, add gallery image, add language note, and remove TV view.

This PR adds a reference to a fourth image, which is added in https://github.com/getumbrel/umbrel-apps-gallery/pull/40, so that needs to be merged first.


nmfretz:
Looks great @hunicus! Merging.

Issue #635, opened 2023-06-14
[Update Request] n8n - 15 releases behind


itsdoublearon:
Can n8n please be updated. It's a few releases behind now.

Thanks

Xosten:
> Can n8n please be updated. It's a few releases behind now.
> 
> Thanks

Hey @itsdoublearon, this will be fixed with #653 😉

nmfretz:
Thanks @itsdoublearon. Closed by https://github.com/getumbrel/umbrel-apps/pull/653

PR #653, opened 2023-06-27
Up N8N version to 0.233.1

benborges:


nmfretz:
Thanks @benborges. Based on the n8n releases here it looks like the latest stable release is [0.233.1](https://github.com/n8n-io/n8n/releases/tag/n8n%400.233.1): https://github.com/n8n-io/n8n/releases

There's a ton of different tags available on docker hub, but the latest stable one is what we should use.

If you run `docker pull n8nio/n8n:0.233.1` the output will show you the correct sha256 digest to use for the multi-arch image.

benborges:
Oh alright, I was indeed confused by the tag to use, it's now 0.233.1

Xosten:
Closes #635 when merged

nmfretz:
Thanks @benborges! I see that n8n just came out with another update (0.234.1), so I bumped the image and updated the release notes.

Going live.

PR #655, opened 2023-06-28
chore(trilium-notes): update to v0.60.4

jgmontoya:
updates trilium notes from v0.59.4 to v0.60.4 (currently the latest stable)


nmfretz:
Thanks a bunch for this update @jgmontoya! really appreciate it. I have tested on amd64 and arm64. Going live now 📝

FYI, I added some whitespace to the release notes section so that they were formatted for the UI. To have a section of text show up on its own line we need to add a blank newline above it in the .yml. To have a blank newline space between two sections of text there needs to be two blank newline in the .yml.

image

PR #656, opened 2023-06-29
Fix no access to download directories in bazaar

gbaygon:
Fixes an issue where bazaar tries to read the media file and returns an error:

FileNotFoundError: [Errno 2] No such file or directory: \'/downloads/movies/[movie_file_name]

thus never downloading nor finding any subtitles for Radaar movies or Sonar shows

### DIFF
--- a/bazarr/docker-compose.yml
+++ b/bazarr/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api/*"
 
   server:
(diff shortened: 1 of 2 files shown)


nmfretz:
oh wow, great catch @gbaygon! Thanks a ton for this. I'm quite embarrassed that I did that. I had this set up properly when originally testing and then did not push my commit with a bind to the downloads folder.

I took this opportunity to bump bazarr to the latest 1.2.2 and added a release notes section. Tested on amd64 and arm64.

PR #659, opened 2023-07-03
Update Immich to 1.65.0

usernameisJim:
Getting Immich update PR going

added tag for 
immich-server
Looks like one for microservices is the same

immich-machine-learning
immich-web
&
immich-proxy

didn't change the new typesense
redis
or 
postgres

### DIFF
--- a/immich/docker-compose.yml
+++ b/immich/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api/*,/search/*"
 
   server:
-    image: altran1502/immich-server:v1.65.0@sha256:92b21221e62129505e24adc8f42ecf867b5b00155668a9a15bf4039e7ef75e65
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for keeping on top of Immich updates @usernameisJim! By the time I got to this, 1.66.1 had been released so I bumped the image versions again and updated the release notes.

Also, the docker-compose.yml and env vars are constantly changing due to Immich being under such active development. As a result, I updated the redis / typesense / postgress images, and made other necessary changes to bring our compose file in line with this one: https://github.com/immich-app/immich/blob/main/docker/docker-compose.yml. 
- One issue I ran into that we need to remember for future updates: The redis service needs a bind mount to persist data despite what the example docker-compose.yml shows on Immich's GitHub:
```
volumes:
      - ${APP_DATA_DIR}/data/redis:/data
```

Tested on amd64 and arm64. Everything works for me on update and fresh install.

swarupbhc:
Hi @nmfretz v1.68.0 has been released, can you please update the app once again? The current version is not compatible with the latest ios client.

PR #662, opened 2023-07-04
chore: update ghostfolio to 1.286.0

ulm0:
Update ghostfolio to its latest release


nmfretz:
Thanks for updating Ghostfolio @ulm0! A newer version was released before I could get to reviewing this, so I bumped the image to 1.287.0 and updated the release notes. I tested on amd64 and arm64. Going live to the app store now.

PR #667, opened 2023-07-05
Update Urbit to v2.11

yapishu:
- Updated Urbit to official v2.11 binary: https://github.com/urbit/vere/releases


nmfretz:
Thanks @yapishu. Tested on arm64 and x86 (amd64). Merged.

PR #669, opened 2023-07-07
Update rotki to v1.29.0

Xosten:
Tested on ARM64 VM (Ubuntu Multipass) ✅

### DIFF
--- a/rotki/docker-compose.yml
+++ b/rotki/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 80
 
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks for this @xosten! Glad you're using Multipass!

Tested on arm64 and x86 (amd64). Going live 🐦

Xosten:
Always a pleasure @nmfretz! 

I don't know if Multipass behaves differently than a classical VM, so I rather mention it case of ^^

PR #670, opened 2023-07-08
Upgrade Node-Red-standalone to latest version (3.0.4)

hieronymousch:
Upgrade Node-Red-Standalone to 3.0.4.
Tested on Custom Umbrel install on Linux

### DIFF
--- a/node-red-standalone/docker-compose.yml
+++ b/node-red-standalone/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       PROXY_AUTH_WHITELIST: "/public/*"
   
   web:
-    image: nodered/node-red:2.2.2-12@sha256:7b8e58892801f01af48acfb18c21b845a6f35029e7654ca6e19ba86bbe810d04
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for bringing us up-to-date with Node Red (standalone) @hieronymousch! 

I added release notes and also changed the image digest so that it refers to the multi-arch image for Node Red. That way, the app will work for both arm and x86 Umbrel users.

For future updates you can just run `docker pull <the-image-you-want>` and it will spit out the correct multi-arch digest if available. For example here is what I get for 3.0.2-14:

```
$ sudo docker pull nodered/node-red:3.0.2-14
3.0.2-14: Pulling from nodered/node-red
...
Digest: sha256:5041fe85e55705d91594980330fc447d86d4f138723befe1628c686bace01de8
Status: Downloaded newer image for nodered/node-red:3.0.2-14
docker.io/nodered/node-red:3.0.2-14
```

PR #671, opened 2023-07-11
Update mempool description+tagline (v2.5, mid-2023)

hunicus:
Simplify, simplify, simplify.

Tagline on main repo also updated with https://github.com/getumbrel/umbrel/pull/1665.


nmfretz:
nice, nice nice. I like the simplification. I pushed a fix for the yaml indentation in the description section, which was resulting in a yaml parsing error that meant mempool would not show up in the app store. I also added whitespace so the description shows up like this in the UI:

image

PR #676, opened 2023-07-11
App Submission: Seafile

IMPranshu:
# App Submission

### SeaFile
...

### 256x256 SVG icon
https://svgshare.com/i/vV3.svg
...

### Gallery images
Screenshot from 2022-11-14 14-31-56
Screenshot from 2022-11-14 14-32-06
Screenshot from 2022-11-14 14-32-19
Screenshot from 2022-11-14 14-32-36
Screenshot from 2022-11-14 14-32-44

...

### I have tested my app on:

### DIFF
--- a/seafile/umbrel-app.yml
+++ b/seafile/umbrel-app.yml
@@ -1,6 +1,6 @@
 manifestVersion: 1
 id: seafile
-category: Files
+category: files
 name: SeaFile
 version: "10.0.1"
 tagline: High performance file syncing and sharing
--- a/seafile/umbrel-app.yml
+++ b/seafile/umbrel-app.yml
@@ -17,6 +17,12 @@ description: >
   🛡️ Built-in File Encryption
 
   A library can be encrypted by a password choosen by you. Files are encrypted before syncing to the server. Even the system admin can't view the files.
+
+
+  🛠️ SET-UP:
+
+  Seafile is configured to work out-of-the-box when installed on Umbrel. However, if you want to be able to upload files from a domain name that is not umbrel.local,
+  open the Seafile app, go to System Admin > Settings and change "FILE_SERVER_ROOT" to /seafhttp.
 releaseNotes: ""
 developer: Seafile
 website: https://www.seafile.com/
--- a/seafile/umbrel-app.yml
+++ b/seafile/umbrel-app.yml
@@ -99,4 +99,4 @@ deterministicPassword: false
 defaultUsername: "me@example.com"
 defaultPassword: "asecret"
 submitter: Pranshu Agrawal
-submission:
+submission: https://github.com/getumbrel/umbrel-apps/pull/281
--- a/seafile/umbrel-app.yml
+++ b/seafile/umbrel-app.yml
@@ -1,0 +1,102 @@
+manifestVersion: 1
+id: seafile
+category: Files
+name: SeaFile
+version: "9.0.9"
+tagline: High performance file syncing and sharing
+description: >-
+  Seafile is an open source cloud storage system with privacy protection and teamwork features. Collections of files are called libraries. Each library can be synced separately. A library can also be encrypted with a user chosen password. Seafile also allows users to create groups and easily sharing files into groups.
+
+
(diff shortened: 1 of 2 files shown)


nmfretz:
Awesome @IMPranshu, this has been requested in the past and is quite popular. I will review soon!

nmfretz:
Hey @IMPranshu, it looks like this might still be a work-in-progress so I have converted this PR to a Draft while you are working on it and testing. 

Here's a few things to check that I noticed at first glance:
- The Umbrel `app_proxy` container is missing from the `docker-compose.yml`. Was this intentional?
- seafile is running as root (may not be necessary).
- The images are quite old (~1yr).
- Can you please check if we can remove the custom `seafile-net` network and custom `container_name`'s, so that we simplify the yml and let Docker Compose automatically handle this? I think everything should still work correctly.
- The data directories for the volume bind mounts are missing from this repo.

Let me know when you have this working nicely and I'll review and test!

swarupbhc:
Hi, @nmfretz please review this.

seafile/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
+manifestVersion: 1
+id: seafile
+category: Files
```
ParthJadhav, line 3: Hey @IMPranshu , due to a recent change the category values should be lowercase. 

```suggestion
category: files
```

ParthJadhav (changes requested):
Tested on Umbrel-dev server. Works fine. I didn't test further because I didn't know where to create an account to test it.

there are couple of changes which I've mentioned.

Thanks for the contribution ❤️

johnpc:
This is awesome! 

However it does require that you set these manually to the host-ip:port after installation, otherwise all uploads fail: 


Additionally, the username/password of `me@example.com` and `asecret` are not conventional for umbrel. I think they should be changed (possibly `umbrel` and `moneyprintergobrrr` instead).

Finally, it is frustrating for me that the drive folder is nested under `app-data`. This means that if the app is deleted, all files uploaded via seafile are destroyed. I would prefer if it pointed to the data volume like ` - ${UMBREL_ROOT}/data/storage/seafile-data:/shared` for example.

Seafile is an excellent program and it certainly belongs in the umbrel app store. I hope this PR gets merged

nmfretz:
Alrighty, I went through this today and got Seafile working https://github.com/getumbrel/umbrel-apps/pull/676/commits/dfb9b366a5e640c2f61a26388f8324f738057954. This is pretty much ready to go! @IMPranshu @smolgrrr @johnpc may I please get one of you to test as well?

@johnpc:

> Is it possible to set those values to the correct ones by default using environment variables?

Yep, I set env vars so that uploading works out-of-the-box after installing as long as you upload from the UI using your main hostname... so `umbrel.local` for most people.

image



Users who want to customize this can do so in the UI and it will persist.


> Additionally, the username/password of me@example.com and asecret are not conventional for umbrel. I think they should be changed (possibly umbrel and moneyprintergobrrr instead).

Good call. I have changed the username to `umbrel` and the default password to a unique, deterministic one as we do for other apps. These values will be available to users in the app store when they install the app.

> Finally, it is frustrating for me that the drive folder is nested under app-data. This means that if the app is deleted, all files uploaded via seafile are destroyed. I would prefer if it pointed to the data volume like  - ${UMBREL_ROOT}/data/storage/seafile-data:/shared for example.

Ya, I hear you. The reason we do this is because most users would expect deleting an app to also delete all data associated with it (especially if they are doing so to free up room on their device). Let me chat with the team and get back to you on this.

IMPranshu:
Tested this on my Pi, working 💪

nmfretz:
Thanks @IMPranshu!

I added some additional instructions to the app description. I'll merge this once we have the gallery assets ready.

IMPranshu:
😎

nmfretz:
Gallery assets are now ready! Going live to the app store 🎉

https://github.com/getumbrel/umbrel-apps-gallery/tree/master/seafile
image

PR #688, opened 2023-07-11
Update btc-rpc-explorer app to 3.4.0

pointbiz:
There's been a new release of this app by the author:
https://github.com/janoside/btc-rpc-explorer/releases/tag/v3.4.0

### DIFF
--- a/btc-rpc-explorer/docker-compose.yml
+++ b/btc-rpc-explorer/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8080
 
   web:
-    image: getumbrel/btc-rpc-explorer:v3.3.0@sha256:cfd14f8e722cfbf1ad106ba224569c8babe685422461a641abc210e13913c636
+    image: getumbrel/btc-rpc-explorer:v3.4.0
     restart: on-failure
     stop_grace_period: 1m
     environment:
--- a/btc-rpc-explorer/docker-compose.yml
+++ b/btc-rpc-explorer/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8080
 
   web:
-    image: getumbrel/btc-rpc-explorer:v3.4.0
+    image: getumbrel/btc-rpc-explorer:v3.4.0@sha256:e85a1fe80919d308b1f80de2dc7174e7b61ec79384d695304fbf259b67b53594
     restart: on-failure
     stop_grace_period: 1m
     environment:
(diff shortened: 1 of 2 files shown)


nmfretz:
I have built the new 3.4.0 image, included the digest in the compose file, and added release notes. Tested on amd64 and arm64 running Bitcoin Core 25.0.

Thanks again for flagging this @pointbiz and for updating the outdated website link. Going live to the app store.

pointbiz:
Amazing, thank you. I'm a contributor to btc-rpc-explorer and a user of Umbrel. So, thanks to you for your work on Umbrel. 

I was clueless how to include the digest, is that related to docker?

nmfretz:
>  I'm a contributor to btc-rpc-explorer

Oh nice! I'm a big fan of btc-rpc-explorer. Such a fun tool.

> I was clueless how to include the digest, is that related to docker?

Yep, the digest is a unique SHA256 hash that is specific to a certain Docker image. What it means for us is that we lock in the exact image that Umbrel users will get when installing the app. Umbrel can be run on both x86 (e.g., amd64) and arm64 (i.e. raspberry pi) devices, so it is important that we  include a digest that refers to a multi-architecture image that can be run on both devices. If we accidentally include a digest that refers only to the x86 architecture, then the app will break for raspberry pi users.

I pushed a multi-architecture image to docker hub for v3.4.0. Here's an example of how to get the correct digest. Pretend the 3.4.0 image already existed, you would:
- Go and look at the [image's repository on docker hub](https://hub.docker.com/r/getumbrel/btc-rpc-explorer/tags). You'll see that the tag v3.4.0 has been built for both arm and amd64 architectures. Digests are listed for the individual architectures, but we need the digest that refers to the multi-arch image, which can't be seen on Docker Hub.
image
- To get the multi-arch digest, run `docker pull getumbrel/btc-rpc-explorer:v3.4.0` in a terminal. This will pull down the image and spit out the digest in the output:
```
$ sudo docker pull getumbrel/btc-rpc-explorer:v3.4.0
v3.4.0: Pulling from getumbrel/btc-rpc-explorer
Digest: sha256:e85a1fe80919d308b1f80de2dc7174e7b61ec79384d695304fbf259b67b53594
Status: Image is up to date for getumbrel/btc-rpc-explorer:v3.4.0
docker.io/getumbrel/btc-rpc-explorer:v3.4.0
```

If you are ever interested in submitting app update PRs for other apps, you can use this method to get the correct digest. Just find the image you need and run `docker pull [image-name]:[tag]`.

PR #689, opened 2023-07-12
Home Assistant - Enable USB device access

lukechilds:
This PR enables Home Assistant to have access to USB devices connected to the host machine, while not requiring the USB to be connected prior to starting the Docker container.

### DIFF
--- a/home-assistant/docker-compose.yml
+++ b/home-assistant/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   server:
-    image: homeassistant/home-assistant:2023.6.3@sha256:1dd89ec1ce8de5ea869a489d5699bd91f77058ffaa6f5484b3e81323128abd2c
+    image: homeassistant/home-assistant:2023.7.1@sha256:a86ff5d05ce46520c53d67c8da55aba310de9b9b4ca8eead1ae0b5ab1c068f97
     network_mode: host
     # UI at default port 8123
     privileged: true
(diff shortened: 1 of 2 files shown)


arguser:
Hello, after app_proxy has been removed I couldn't access home-assistant anymore via the new 8123 port. I'm running umbrel on linux.

nmfretz:
Hey @arguser, sorry you're running into issues with this. The Home Assistant container now runs in host network mode, meaning that it shares your linux machine's networking namespace. This means that if any other service is using port 8123 on your device, Home Assistant will clash with it and not be accessible.

Can you please check the following:
1. Are you by chance running an outdated version of the [Tallycoin Connect app](https://apps.umbrel.com/app/tallycoin-connect)? Previous versions of this app used port 8123 so if it is not up-to-date and you run the new Home Assistant update there will be a port clash and neither app will start. If this is the case, then update Tallycoin Connect and restart your Umbrel.
2. Are you running other services besides Umbrel on your linux machine that are using port 8123 already? This would also result in a port clash.
3. If non of the above, can you please post your troubleshooting logs here. [Instructions](https://community.umbrel.com/t/official-umbrel-troubleshooting-guide-and-faq/9873#:~:text=Generating%20Troubleshooting%20Logs).

arguser:
@nmfretz  I've managed to sort my issue by

```
services:
  server:
    image: homeassistant/home-assistant:2023.7.1@sha256:a86ff5d05ce46520c53d67c8da55aba310de9b9b4ca8eead1ae0b5ab1c068f97
    # UI at default port 8123
    ports:
      - "8123:8123"
    network_mode: bridge
    privileged: true
    volumes:
      - ${APP_DATA_DIR}/data:/config
      - /dev:/dev
```

nmfretz:
Glad you got it sorted. Although in this set up you won't get auto discovery of devices because the container won't be able to open the necessary ports on the host machine or have bluetooth access. You're still mapping to port 8123 on your machine so it looks like the issue wasn't related to a port clash for that specific port.

If you switch it back to host network mode at some point, I'd love to see the logs to figure out what's going on.

PR #693, opened 2023-07-15
Update Node-Red-Standalone to 3.0.2-18

oren-z0:
NodeJS 14 is deprecated. This version uses NodeJS 18 instead.

Tested on Raspberry Pi.

### DIFF
--- a/node-red-standalone/docker-compose.yml
+++ b/node-red-standalone/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       PROXY_AUTH_WHITELIST: "/public/*"
   
   web:
-    image: nodered/node-red:3.0.2-14@sha256:5041fe85e55705d91594980330fc447d86d4f138723befe1628c686bace01de8
+    image: nodered/node-red:3.0.2-18@sha256:e401cd81a43bca7db51fe8dac2c9cfc9b4ffb922ba424241280f7392e13b06a0
     restart: on-failure
     stop_grace_period: 1m
     volumes:
--- a/node-red-standalone/umbrel-app.yml
+++ b/node-red-standalone/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: node-red-standalone
 category: automation
 name: "Node-RED"
-version: "3.0.2-14"
+version: "3.0.2-18"
 tagline: Wire together the Internet of Things
 description: >-
   Node-RED is a visual programming tool for wiring together hardware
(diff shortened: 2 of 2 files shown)


IMPranshu:
Appreciate the PR @oren-z0. Node 14 might be deprecated but the docker images built are still functioning and supported. Updating it might be not necessary at this point unless you have faced any issues while running the app which is specific to node version 14 requiring this update. Do share a screenshot of the issue faced, will help better understand.


Screenshot 2023-07-16 at 8 50 26 PM

Both Images are being maintained similarly.

Screenshot 2023-07-16 at 8 50 08 PM

oren-z0:
I wanted to use [@nostr-dev-kit/ndk](https://github.com/nostr-dev-kit/ndk) which depends on [@noble/hashes](https://github.com/paulmillr/noble-hashes), which requires the engine to be >= 16:

https://github.com/paulmillr/noble-hashes/blob/54c3197fc53bbfa4521b6883eed0e7bd1430beab/package.json#L44

nmfretz:
Oh, good call @oren-z0. Looks like 16.x is recommended by Node-Red [https://nodered.org/docs/faq/node-versions](https://nodered.org/docs/faq/node-versions) anyways. You okay going with [16](https://hub.docker.com/layers/nodered/node-red/3.0.2-16/images/sha256-e2cd6b89e18361593276e3ef5282b403e0879040c96d3099c15e8f0200a36339?context=explore) or would you rather see the bump all the way to 18?

Also, how are you finding ndk? I'm currently using nostr-tools and curious to here about experience with other tools.

oren-z0:
Thanks @nmfretz ,
I don't see a reason not to upgrade directly to Node.JS 18 - it's stable, and Node.JS 16 will become deprecated too this year: https://endoflife.date/nodejs . 

NDK is cool, does a lot of work for you, like all the multi-step procedure for creating a "zap" (a single `event.zap(...)` call will do all the zap-request, http call to the lnurlp etc. and give you the invoice).
But it has this annoying feature that automatically adds `e` and `p` tags referencing events that are mentioned in your event content (before signing an event, it will parse it's content).
https://github.com/nostr-dev-kit/ndk/issues/53
Sometimes you don't want to add `p` and `e` tags to events that you mention, because it will make your event appear as a reply to the mentioned events in some clients - and may also break the nip-57 and create a zap request with more than one `p` tag.
 I ended up not using the Node-RED functionally directly, I just created the code in one .js file and a .sh script to execute it, and use Node-RED to re-run the .sh script if it crashes.
But it was useful to have a docker environment with Node.JS installed and a /data folder to put the files and install dependencies.

Check out this bot: http://damus.io/npub1qm05ghrt2xuc970rxpsuchs504ah4rllxsgevdyjc0v5uha77m6qujz78x

nmfretz:
> I don't see a reason not to upgrade directly to Node.JS 18 - it's stable, and Node.JS 16 will become deprecated too this year.

Sound good to me @oren-z0, thanks for keeping on top of this. I have tested on arm64 and amd64 and we are good to go live. FYI, I tweaked the release notes and pushed to your fork, just so you're aware for potential future merge conflicts.

>Check out this bot: http://damus.io/npub1qm05ghrt2xuc970rxpsuchs504ah4rllxsgevdyjc0v5uha77m6qujz78x

I just tested this. You are the king of bots!

> But it has this annoying feature that automatically adds e and p tags referencing events that are mentioned ...

Ah I see, ya thanks for the breakdown. It's such early days with these tools that it's often hard to use them out-of-the-box without modifying for your specific use case. I have found similar quirks when using nostr-tools.

PR #701, opened 2023-07-17
Remove Transmission dependencies 

awholegnuworld:
There are three different torrent clients available in the main app store. At least two of these are supported by the *arr apps. I found it strange that Transmission is listed as a dependency for the *arr apps when qBittorrent works as well. I’ve removed the requirement from the config file to allow users to manually chose in app what client to use.

### DIFF
--- a/lidarr/umbrel-app.yml
+++ b/lidarr/umbrel-app.yml
@@ -14,6 +14,7 @@ description: >-
   It can also be configured to automatically upgrade the quality of existing files in the library when a better quality format becomes available.
 developer: Lidarr
 website: https://lidarr.audio/
+dependencies: []
 repo: https://github.com/Lidarr/Lidarr
 support: https://github.com/Lidarr/Lidarr/issues
 port: 8686
--- a/lidarr/umbrel-app.yml
+++ b/lidarr/umbrel-app.yml
@@ -14,8 +14,6 @@ description: >-
   It can also be configured to automatically upgrade the quality of existing files in the library when a better quality format becomes available.
 developer: Lidarr
 website: https://lidarr.audio/
-dependencies:
-  - transmission
 repo: https://github.com/Lidarr/Lidarr
 support: https://github.com/Lidarr/Lidarr/issues
 port: 8686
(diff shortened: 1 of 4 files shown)


awholegnuworld:
Just fixed some accidental formatting errors in the .yml files. I forgot to add brackets. Also had forgot to add Bazaar and Sonarr.

nmfretz:
Thanks for this @awholegnuworld. You're right that now with multiple compatible torrent client options, it looks a bit weird that we have transmission listed as a dependency for these apps. Originally this was done when Transmission was the only compatible client on Umbrel... what we did at the time was create a [media-app-configurator](https://github.com/getumbrel/media-app-configurator) that takes care of configuration of the download client for some of the *arr apps. You'll see this as the `mac` container in the docker-compose.yml files for these apps. We are trying to make it as easy as possible for someone to download an app and just have it work, without having to muddle through configuration.

Let me chat with the Team about the best way forward with this one. We have "optional dependencies" on our roadmap, so this should be something we are thinking about. As it stands right now I believe you can change the download client manually in the apps, but you would still have to download transmission since it's a dependency, which would be pretty clunky experience.

I'll get back to you.

awholegnuworld:
Ok, thanks! The optional dependencies are a great idea.

awholegnuworld:
On the local side, could I uninstall Transmission after installing another client? Or does Umbrel check dependencies at app start or something like that?

nmfretz:
> could I uninstall Transmission after installing another client?

I don't want to officially recommend this in case some weird bugginess happens, but juuuuuust hypothetically of course (😉) you could:
- Install Transmission
- Install the *arr apps you want
- Remove Transmission via script. (`sudo ~/umbrel/scripts/app uninstall transmission`)
- Restart Umbrel for good measure
- Nothing should error on app start as dependencies are checked at app install and uninstall

nmfretz:
Also, I spoke with the Team and at this point we are going to keep the Transmission dependency until we, at least, have optional dependencies implemented. 

I will close this PR for now, but really want to thank you for opening it and bringing this to our attention. Down the line we will allow for easy configuration of multiple download clients. Hope to see more PRs from you @awholegnuworld!

PR #709, opened 2023-07-26
App Submission: Blocky

awholegnuworld:
This isn't working in Multipass or my Umbrel, but it could be because I have Pi-Hole installed on my main Umbrel. It complains about port 53 being taken even on a clean Multipass install.


blocky/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: "2.1"
+services:
+  blocky:
```
IMPranshu, line 3:
```suggestion
  app_proxy:
    environment:
      APP_HOST: blocky_web_1
      APP_PORT: 4000
  web:
```

blocky/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+    image: spx01/blocky
+    container_name: blocky
+    restart: unless-stopped
+    # Optional the instance hostname for logging purpose
+    hostname: blocky-hostname
+    ports:
```
IMPranshu, line 9:
```suggestion
```

IMPranshu (changes requested):

IMPranshu:
I have left some comments do try them and see if the error occurs again.

awholegnuworld:
Now would be a great time to say that I'm awful at Docker. 😄 I don't know if there is an arm64 image for the front end of this app, but the repository says there is. Ideas?

PR #716, opened 2023-08-04
Update Torq to v1.0.11

ohenrik:
This update fixes a bug when creating tags. In the last release, new tags would overwrite an existing one.


nmfretz:
Thanks again @ohenrik. Tested on arm64 and x86(amd64).

PR #718, opened 2023-08-07
Update lnbits to 0.10.9

chloehjung15:
Lnbits extension is few versions behind now.
This is to update lnbits to the latest version (0.10.9)

### DIFF
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   web:
-    image: lnbitsdocker/lnbits-legend:0.10.6@sha256:a11aaa6d2b211db4c1ec2cafb7c35057ef45bcf4b98fb3bbf23c5635a0304aff
+    image: lnbitsdocker/lnbits-legend:0.10.9@sha256:ec4d6055e1de1b340c66bb6b473635ade1e95b57a69e0cb6effe21af310385d6
     init: true
     restart: on-failure
     stop_grace_period: 1m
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks very much @chloehjung15! Really appreciate you keeping us up-to-date. I updated the release notes a bit and tested on arm64 and x86(amd64). Going live ⚡️.

### umbrel-app.yml tip:
I updated the whitespace in the `umbrel-app.yml` so the release notes show nicely in the UI. It's a bit finicky, but to get multiple text-blocks to show up on separate lines you need a single newline between each block. And to get a blank line to show between text-blocks you need to add **2** newlines.

You can see the difference between this:

```yaml
releaseNotes:  >-
  🔧 The new Security panel

  The Security panel has six components:

  - Server logs
  - IP blocker 
  - Rate limiter
  - Security notifications
  - Killswitch
  - Watchdog 

  
  📄 Full Release Notes and detailed information is available at: https://github.com/lnbits/lnbits/releases
  ```

image

and this one where additional whitespace has been added my adding newlines

```yaml
releaseNotes:  >-
  In this release, we introduce a whole suite of new security features to harden your LNbits instance. Our goal is to make these features
  accessible and keep every instance as safe as possible to protect user funds from possible security breaches.


  You will find a new Security panel in your Manage Server AdminUI that you can access as the super user. All features are opt-in and are also
  accessible via the LNbits API.


  The Security panel has six components:

  - Server logs 📝

  - IP blocker 🚫

  - Rate limiter 🐌

  - Security notifications 🔔

  - Killswitch 🪓
  
  - Watchdog 🐕 (coming soon)

  
  📄 Full Release Notes and detailed information is available at: https://github.com/lnbits/lnbits/releases
  ```

image

chloehjung15:
@nmfretz
Right I didn't realise I needed a new line in between to get an actual new line. Thanks for fixing it for me and merging

PR #724, opened 2023-08-13
update pi hole v2023.05.2

f321x:
https://github.com/pi-hole/docker-pi-hole/releases/tag/2023.05.2

Not tested

### DIFF
--- a/pi-hole/docker-compose.yml
+++ b/pi-hole/docker-compose.yml
@@ -14,6 +14,7 @@ services:
     environment:
       - VIRTUAL_HOST=${APP_DOMAIN}
       - WEBPASSWORD=${APP_PASSWORD}
+      - WEB_PORT=8082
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this @f321x! 

I have made a change to the the docker-compose.yml, which I had earmarked to do for the next pi-hole update so I am piggy-backing off your PR.
The pi-hole container now runs in host network mode, which allows users to easily configure their pi-hole as a DHCP server. This has been a heavily requested feature.

For those curious, the `DNSMASQ_LISTENING=all` environment variable sets pi-hole's Interface listening behaviour to "Listen on all interfaces, permit all origins", which is safe for Umbrel's use-case as a typical at-home setup where your Pi-hole is located within your local network (and you have not forwarded port 53 in your router).

Tested on arm64 and x86(amd64). Going live.

PR #733, opened 2023-08-19
Update Vaultwarden to v1.29.1

ankit5902:
Update Vaultwarden from v1.25.0 to v1.29.1.

Full release notes can be found at: https://github.com/dani-garcia/vaultwarden/releases

### DIFF
--- a/vaultwarden/docker-compose.yml
+++ b/vaultwarden/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   server:
-    image: vaultwarden/server:1.29.1@sha256:89ffbe952338f42265dbff0033d0fa8a88458a52674a05f5ed918875e403dccd
+    image: vaultwarden/server:1.29.1@sha256:c2849f8189e4d425a9d80db0380566cc38577b289b9e6c88330e190e19af5a30
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/vaultwarden/docker-compose.yml
+++ b/vaultwarden/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   server:
-    image: vaultwarden/server:1.25.0@sha256:f3ebede27f1cf5e78373c3c4a429cf1fdd8d6b13528a2b9ca4fb3cb7cc681ba9
+    image: vaultwarden/server:1.29.1@sha256:89ffbe952338f42265dbff0033d0fa8a88458a52674a05f5ed918875e403dccd
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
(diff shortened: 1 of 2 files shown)


vaultwarden/umbrel-app.yml
```diff
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: vaultwarden
 category: files
 name: Vaultwarden
-version: "1.25.0"
+version: "1.29.1"
```
nmfretz, line 5: Looks like this will be the first update we've done of Vaultwarden, so we can now add a `releaseNotes` section to the `umbrel-app.yml`.

It can be a bit tricky trying to decide what to include in the release notes; the amount of detail really depends on the situation. Sometimes the release notes are just direct copies of the the official release notes, like this Torq example submitted by Torq themselves:
https://github.com/lncapital/torq/releases/tag/v1.0.1
image

and other times we simplify the release notes if we are updating several versions at once, like this Plex example:
image

For this update where we are moving forward several updates (1.25.0 all the way to 1.29.1), so it might make sense to just do a simplification like this:

```yaml
releaseNotes: >-
  This release updates Vaultwarden from v1.25.0 to v1.29.1. It includes many bug fixes and performance improvements, as well as the following new features:
  
  - Support for Forward Email
  
  - WebSocket notifications now work via the default HTTP port
  
  - Support for Argon2 key derivation

  - and more!


  The full release notes are available at https://github.com/dani-garcia/vaultwarden/releases
```

What do you think?

vaultwarden/docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   server:
-    image: vaultwarden/server:1.25.0@sha256:f3ebede27f1cf5e78373c3c4a429cf1fdd8d6b13528a2b9ca4fb3cb7cc681ba9
+    image: vaultwarden/server:1.29.1@sha256:89ffbe952338f42265dbff0033d0fa8a88458a52674a05f5ed918875e403dccd
```
nmfretz, line 11: I think this may be the wrong digest for the multi-arch digest. I get:

```
$ sudo docker pull vaultwarden/server:1.29.1
...
Digest: sha256:c2849f8189e4d425a9d80db0380566cc38577b289b9e6c88330e190e19af5a30
```

nmfretz (commented):
Thanks @ankit5902. Looks like this is the first Vaultwarden update! See below for suggestions.

vaultwarden/umbrel-app.yml
ankit5902: Appreciate your suggestions! I think your proposal to simplify the release notes makes sense. Considering the multiple versions being covered, this will help keep the release notes concise while highlighting the key changes

vaultwarden/docker-compose.yml
ankit5902: Please allow me a bit of time to cross-reference and confirm the correct digest. I'll provide an update once I have more information.

Thank you for your patience and vigilance in maintaining the integrity of the processes.

ankit5902: You're absolutely right, the correct sha256 digest should be indeed `c2849f8189e4d425a9d80db0380566cc38577b289b9e6c88330e190e19af5a30`. I realized that I mistakenly provided the `SHA256` for the image "`ghcr.io/dani-garcia/vaultwarden:1.29.1`" instead of the correct one, which should have been for "`vaultwarden/server:1.29.1`".

Thank you for bringing this to my attention, and I appreciate your understanding. I'll promptly rectify this mistake.

ankit5902:
The issues have been resolved, and the code is now ready for review.

PR #734, opened 2023-08-19
Update rotki 1.30.0

Xosten:
✅  Tested on Ubuntu Multipass on MacOS


nmfretz:
Thanks very much for this update @Xosten! I tested on arm64 and x86(amd64). Database migration and asset update worked well for me. Going live.

Xosten:
Ey! Thank @nmfretz ❤️ !

PR #735, opened 2023-08-19
Update Immich to v1.74.0

ankit5902:
Updated image to v1.74.0 for

`immich-server` (the same `immich-server` tag is used under `server` and `microservices`)
`immich-machine-learning`
`immich-web`
`immich-proxy`

Did not change `typesense:0.24.1`, `redis:6.2-alpine` or `postgres:14-alpine`

### DIFF
--- a/immich/docker-compose.yml
+++ b/immich/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api/*,/search/*"
 
   server:
-    image: ghcr.io/immich-app/immich-server:v1.73.0@sha256:9ffecf7b8e1d65de4a82e1c47a5d1297a353703b4f511dc4b08fbb2a822da576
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for staying so on top of Immich updates @ankit5902! Really appreciate this. I made a small tweak to the release notes since it looks like the mobile app doesn't need to be on latest v1.74.0 to be fully compatible with this update. 

I also double checked the recent changes to the example compose file that Immich provides, but we want to retain debug messages so I did not add `GLOG_minloglevel=1`:
[GLOG_minloglevel=1](https://github.com/immich-app/immich/commit/66490d5db416856c75412945fe2d77f366aa20e3#diff-423deb13b7c401b1a7f41ee91c77f722e11d2f317d6a66b546524e8a04cc8b03)

Tested on arm64 and amd64. Going live 📸

PR #741, opened 2023-08-23
Update Tailscale to v1.48.1

ankit5902:
This update fixes the `nftables` interaction between `tailscale` and `ufw` which resulted in blocking subnet routed traffic. Also, it now determine correct CPU architecture on Synology platforms during `tailscale update`


nmfretz:
Excellent, thank you @ankit5902! I have tested on arm64 and x86(amd64). Going live.

FYI, I ended up simplifying the release notes for this one, since there wasn't anything big.

PR #742, opened 2023-08-23
Update Gitea to v1.20.3

ankit5902:
I've updated the below images:
- `Gitea` image from `v1.19.1` to `v1.20.3` (latest)
- `MariaDB` image from `v10.11.2` to `v11.0.3` (latest)

I've also updated the `releaseNotes` in `umbrel-app.yml` file.

### DIFF
--- a/gitea/umbrel-app.yml
+++ b/gitea/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1.1
 id: gitea
 category: developer
 name: Gitea
-version: "1.19.1"
+version: "1.20.3"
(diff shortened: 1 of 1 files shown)


nmfretz:
Thanks @ankit5902, excellent update PR 🙏.

I took a look through the mariadb release notes and tested gitea pretty thoroughly to make sure there were no database migration issues. Going live 🍵

For database versions with long-term support (e.g., 5 years for each mariadb release), we don't have to update the database to the newest version each time we update the main application. Then every once in a while, like you've done here, we can update the database and test thoroughly.

PR #745, opened 2023-08-24
Update Invidious to 2023.8.7-3450896

ceramicwhite:


nmfretz:
Excellent, thanks for this @ceramicwhite. I just attempted a quick test and noticed that a new env var is now required for Invidious to work: `hmac_key`. https://docs.invidious.io/installation/#set-up-invidious_1:~:text=You%20must%20set%20a%20random%20generated%20value%20for%20the%20parameter%20hmac_key%3A!%20On%20Linux%20you%20can%20generate%20it%20using%20the%20command%20pwgen%2020%201.

Looks like it is supposed to be a random string. I'll look into this, but potentially we can just use `$APP_SEED`:
```
$APP_SEED - Unique 256 bit long hex string (128 bits of entropy) deterministically derived from user's Umbrel seed and your app's ID
```

nmfretz:
Tested on amd64 and arm64 with $APP_SEED as hmac_key and all works well. Thanks again @ceramicwhite for the updated image! Going live

PR #746, opened 2023-08-24
Update File Browser to v2.24.2

ankit5902:
Update File Browser image from `v2.22.4` to `v2.24.2`


nmfretz:
Thanks for this one as well @ankit5902! I really appreciate all these updates. I made some adjustments to the release notes so that the "new features" list only highlights features, and not bug fixes or performance improvements.

PR #749, opened 2023-08-24
Update Syncthing to  v1.23.7

ankit5902:
Update the `Syncthing` image from `v1.23.2` to `v1.23.7`


nmfretz:
Great Syncthing update @ankit5902! I have tested on arm64 and x86(amd64). Going live ⏱️

FYI, I changed syncthing's developer release notes slightly so that they are more understandable for the average user.

PR #750, opened 2023-08-25
Update Uptime Kuma to v1.23.0

ankit5902:
Updated image of Uptime Kuma from v1.21.2 to v1.23.0


nmfretz:
Wonderful @ankit5902. No changes from me! Tested on arm64 and x86(amd64).

PR #751, opened 2023-08-25
Update rotki 1.30.1

Xosten:
✅ Tested on Ubuntu Multipass on ARM


nmfretz:
Perfect, thanks for this @Xosten! Everything looks great. I have tested on arm64 and x86(amd64). Going live to the app store 🪶

PR #752, opened 2023-08-26
Update Jellyfin to v10.8.10

ankit5902:
Update Jellyfin image from `v10.8.9` to `v10.8.10`, which fixes a critical bug along with other bug fixes and improvements.

⚠️ CRITICAL SECURITY ADVISORY: [GHSA-9p5f-5x8v-x65m](https://github.com/jellyfin/jellyfin/security/advisories/GHSA-9p5f-5x8v-x65m) and [GHSA-89hp-h43h-r5pq](https://github.com/jellyfin/jellyfin-web/security/advisories/GHSA-89hp-h43h-r5pq) can be combined to allow remote code execution for any authenticated Jellyfin user including non-admin users. While the particular execution mechanism of the former dates to the 10.8.0 release, the latter was present for all Jellyfin releases before this point. It is thus absolutely critical for all Jellyfin administrators, regardless of version, to upgrade to this version if they allow any untrusted users and/or expose their instance to the Internet.

### DIFF
--- a/jellyfin/docker-compose.yml
+++ b/jellyfin/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   server:
-    image: linuxserver/jellyfin:10.8.9@sha256:af281e1b23198076a2ce848d710e8c7ee6f96134d18c61e9c23128e3419d9e86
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this jellyfin update @ankit5902. And thanks again for all your hard work updating many many apps last week, I really appreciate it 🤝.

Tested on arm64 and x86(amd64). Going live 🎥

PR #757, opened 2023-08-29
Update: Snort v0.1.13

v0l:


nmfretz:
Thanks @v0l! Nailed the multi-arch digest 😉.

Tested on arm64 and x86(amd64). This release is fantastic. Going live.

PR #759, opened 2023-08-30
Update docker-compose.yml to reflect umbrel-lnd.conf update

smolgrrr:
The latest version of the Lightning Node app does not include an lnd.conf file by default. Instead, it has a file named umbrel-lnd.conf located at:

`/lnd/umbrel-lnd.conf`

Updating RTL to reflect this


nmfretz:
Great sleuthing @smolgrrr! I looked into it and `CONFIG_PATH` is an optional config parameter that allows a user to view, but not edit, their local lnd.conf (see below). Changing this env var does not impact anything else in RTL. Tested on arm64 and x86(amd64). Going live ⚡️.

image

nmfretz:
FYI, I bumped the `version` and updated the `releaseNotes` in the manifest so that users will see an update.

PR #765, opened 2023-09-09
Upgrade to Syncthing 1.24.0

jon-ryan:


nmfretz:
Thanks so much for your regular Syncthing updates @jon-ryan! I really appreciate it. The PR is perfect.

I have tested on arm64 and x86(amd64). Going live 🔁

PR #768, opened 2023-09-13
Update Torq to v1.1.1

ohenrik:
New version of Torq is ready :)


nmfretz:
Excellent, thanks @ohenrik! And thanks very much for keeping the umbrel-app.yaml up-to-date.

Tested on arm64 and x86(amd64). Going live ⚡️

PR #774, opened 2023-09-18
Add Portainer

nmfretz:
Portainer on Umbrel has been configured for power users aiming to run custom applications on umbrelOS. Portainer is not interfacing directly with the primary host's Docker instance active on Umbrel. Instead, it communicates with a Docker-in-Docker (DinD) service.

Follow these best practices to avoid issues:

  1. Data persistence: Make sure to only used named Docker volumes for your stacks and containers. Data in bind-mounted volumes will be lost when the Portainer app is restarted or updated.

  2. Port management: Watch out for potential port conflicts between your custom containers and umbrelOS' service containers, apps you have installed from the Umbrel App Store or Community App Stores, and any apps you go to install in the future.

  3. Container restart policy: Set your containers to "unless-stopped" or "always" restart policies. This will allow your containers to restart automatically when the Portainer app is restarted or updated.

  4. Web access to containers: Access your custom containers in your browser at umbrel.local:<mapped-port>. For example, for a container with a web UI running on port 4545, navigate to umbrel.local:4545 to access it.

### DIFF
--- a/portainer/default-password
+++ b/portainer/default-password
@@ -1,0 +1,1 @@
+changeme
--- a/portainer/docker-compose.yml
+++ b/portainer/docker-compose.yml
@@ -28,7 +28,7 @@ services:
 
   portainer:
     image: portainer/portainer-ce:2.19.0
-    command: --admin-password-file=/default-password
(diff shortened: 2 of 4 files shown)


nmfretz:

PR #780, opened 2023-09-24
Bump Oak Node to v0.3.8

ok300:
Tested on:

- amd64 (VM)
- arm64 (raspi)

### DIFF
--- a/oak-node/docker-compose.yml
+++ b/oak-node/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8100
 
   web:
-    image: oak-node.net/oak:v0.3.8@sha256:a529320a398eeb11511fcb56a7f98199c435ecd227bc11f6e395b5fe29b5b067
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this update @ok300!

The works well for me on amd64, but I get the following error on arm64(raspi 4):

`/oak: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ./oak)`

Would you be able to check the arm64 image again? Your base image is debian:stable-slim correct?

ok300:
Good catch!

I saw that
```
docker pull debian:stable-slim
```
indeed pulled a new version of the image on my raspi, which means I've used an older `debian:stable-slim` when I built.

I've built the raspi image again and it starts when I run it.

Can you please check if this one is ok?

nmfretz:
Just tested and it is working well! Going live now 🌳. Thanks again @ok300.

PR #782, opened 2023-09-25
Update umbrel-app.yml: Add submitter & submission fields

ParthJadhav:


nmfretz:
Thanks for adding these fields to the app manifest template @ParthJadhav! 

For those making their way to this PR in the future:
As of the date of this comment, apps on Umbrel **do not** break if the umbrel-app.yml is missing the `submitter` and `submission` fields, but these are really nice to have for documentation.

PR #783, opened 2023-09-25
Update: Snort v0.1.19

v0l:


ParthJadhav:
Hey @v0l thank you so much for the update. Although there are few things which needs to be addressed:

- The copy to clipboard button isn't working since the connection is over `http` and not `https` ([Ref](https://stackoverflow.com/a/51823007)). This can be solved by doing something like [this](https://github.com/getumbrel/umbrel-nostr-relay/blob/e39aa894a533778ba0ae9fe9e2bbc646741cbbb7/ui/src/components/utility/CopyText.js#L6-L22). OR the full key can be shown so the user can copy it manually.
- DMs don't work out of the box. @nmfretz might be able to explain more about it.

v0l:
Hi, Thanks for notifying me of this. I pushed a fix now for the copy code, also i notices Nos2x doesnt work at all i had to use Alby to login with the extension.

But after logging in DM's appeared to work, so im not sure what the issue might be there

ParthJadhav:
> Hi, Thanks for notifying me of this. I pushed a fix now for the copy code, also i notices Nos2x doesnt work at all i had to use Alby to login with the extension.

Yeah that's right, when I create a new account normally it doesn't work. 

For this feature to work you should use an extension like Alby to manage your nostr keys and then use "Login with Extension (NIP-07)".

--- 

Also did you update the image for the pushed code?

v0l:
I didnt push a build yet, you mentioned that DM's are not working?

ParthJadhav:
> I didnt push a build yet, you mentioned that DM's are not working?

Yes, I thought you pushed the build for the `copy to clipboard` change. But no worries.

nmfretz:
Excellent, thanks for this update to Snort @v0l. Been using snort as my main client for a while now and love these updates.

I pushed the following minor changes, so please pull them down to avoid potential future conflicts:
- included v0.1.19 tag in image for clarity [64c968c](https://github.com/getumbrel/umbrel-apps/pull/783/commits/64c968cbc30e5321ab2867bc9b5e6daa034b9365)
- tweaked whitespace in release notes [1e9f49f](https://github.com/getumbrel/umbrel-apps/pull/783/commits/1e9f49f008789003c0a0d01a05f039ed9e69e19f)

@ParthJadhav great review and testing, and good find on the copy-to-clipboard functionality. This is ultimately an issue on our end, not Snort's, so thanks for accommodating us @v0l.   


I have tested on arm64 and x86(amd64). Going live. See below for an explanation of Parth's DM finding.

### ----------------------

I just took a look, and the reason @ParthJadhav wasn't able to DM **after creating an account through the UI** is because the Web Crypto API (well... essentially all of it) cannot be run over insecure contexts. The `window.crypto.subtle` property isn't available so anything that requires private key access with window.crypto.subtle.importkey will result in the type error seen when attempting to DM for instance:
```
nip4.ts:44 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'importKey')
    at
```

The solution for this is already in place in Snort, which is for users to manage keys and sign in with an Extension, which is more secure and enables signing with private keys for necessary actions. 

If you attempt to sign in with a private key over http, Snort will show this error to the user:
image

And signing in with a public key does not give you access to posting or DMs, so not having access to  `window.crypto.subtle` doesn't matter.

Where the issue arises is for the specific case where:
- user creates account through the UI, over http
- they are then automatically logged in
- The option to view and show DMs is shown in the UI, which is not the case when signing in with a pubkey
- a user attempts to DM and just sees an infinite load-spinner beside the text-input. The only error shown is in the console.

If the user logs out, then they won't get into this situation again because the UI will not allow them to log in with a private key. It's obviously up to you on if/how you want to handle this case (e.g., leave as-is, show account creation warning over http, etc).

PR #785, opened 2023-09-28
App Submission: Mutiny Wallet

benthecarman:
# App Submission

### App name

Mutiny Wallet

### 256x256 SVG icon

https://raw.githubusercontent.com/MutinyWallet/mutiny-web/master/public/mutiny_logo_mask.svg

### Gallery images
Screenshot from 2023-09-27 22-46-55

image

image

### I have tested my app on:
- [x] [Umbrel dev environment](https://github.com/getumbrel/umbrel-dev)

### DIFF
--- a/mutiny-wallet/docker-compose.yml
+++ b/mutiny-wallet/docker-compose.yml
@@ -1,0 +1,51 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: mutiny-wallet_nginx_1
+      APP_PORT: 14499
+
+  proxy:
+    image: ghcr.io/mutinywallet/ln-websocket-proxy:v0.3.1@sha256:b5cd29c10acbc4182b2be7644043ad565407d62e4046c19e6e4a6450886d847e
+    restart: on-failure
+    stop_signal: SIGINT
+
+  postgres:
+    image: postgres:15@sha256:546445ad21cb5893c0997080b831ee45945e798c4359270e16413c8bd93575db
+    restart: on-failure
+    stop_signal: SIGINT
+    volumes:
+      - ${APP_DATA_DIR}/data/postgresql:/var/lib/postgresql/data
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: docker
+      POSTGRES_DB: vss
+
+  vss:
+    image: ghcr.io/mutinywallet/vss-rs:v0.1.0@sha256:f64b08c3ec38f0091187b49bf509ab75903d76300a95428c5e14f71cbc053761
+    restart: on-failure
+    stop_signal: SIGINT
+    environment:
+      DATABASE_URL: postgres://postgres:docker@postgres/vss
+      SELF_HOST: "true"
+    links:
+      - postgres
+    depends_on:
+      - postgres
+
+  web:
+    image: ghcr.io/mutinywallet/mutiny-web:build-93@sha256:da1d0d75cb7562ea3a2babd392c0f7ca8ac5ad179d0d5e8fdebc0d2fefd13602
+    restart: on-failure
+
+  nginx:
+    image: nginx:1.21-alpine@sha256:686aac2769fd6e7bab67663fd38750c135b72d993d0bb0a942ab02ef647fc9c3
+    init: true
+    restart: on-failure
+    volumes:
+      - ${APP_DATA_DIR}/config/nginx.conf:/etc/nginx/nginx.conf:ro
+      - ${APP_DATA_DIR}/data/public:/var/www/app/public:ro
+    depends_on:
+      - web
+      - proxy
+      - vss
--- a/mutiny-wallet/umbrel-app.yml
+++ b/mutiny-wallet/umbrel-app.yml
@@ -1,0 +1,32 @@
+manifestVersion: 1
+id: mutiny-wallet
+category: bitcoin
+name: Mutiny Wallet
+version: "0.4.20"
+tagline: Mutiny is a self-custodial lightning wallet that runs in the browser.
+description: >-
+  Mutiny is a self-custodial lightning wallet that runs in the browser. It is a
+  not a traditional lightning node that is always online, but rather a wallet
+  that runs in your browser.
+
+  Mutiny comes with full functionality for sending and receiving on-chain and lightning
+  payments. It also supports many social features like many Nostr integrations like
+  Nostr Wallet Connect, Nostr Contacts, and a Zap Feed.
+
+  WARNING: This is an beta software, don't put too much money in.
+releaseNotes: ""
+developer: Mutiny Wallet
+website: https://mutinywallet.com
+repo: https://github.com/mutiny-wallet/mutiny-web
+support: https://matrix.to/#/#mutiny-community:lightninghackers.com
+port: 14499
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
+path: ""
+dependencies: []
+defaultUsername: ""
+defaultPassword: ""
+submitter: MutinyWallet
+submission: https://github.com/getumbrel/umbrel-apps/pull/785
--- a/mutiny-wallet/umbrel-app.yml
+++ b/mutiny-wallet/umbrel-app.yml
@@ -3,14 +3,14 @@ id: mutiny-wallet
 category: bitcoin
 name: Mutiny Wallet
 version: "0.4.20"
-tagline: Mutiny is a self-custodial lightning wallet that runs in the browser.
+tagline: Mutiny is a self-custodial lightning wallet that runs in the web.
 description: >-
-  Mutiny is a self-custodial lightning wallet that runs in the browser. It is a
+  Mutiny is a self-custodial lightning wallet that runs in the web. It is a
   not a traditional lightning node that is always online, but rather a wallet
-  that runs in your browser.
+  that runs in your browser of the device of you're accessing it from.
 
   Mutiny comes with full functionality for sending and receiving on-chain and lightning
-  payments. It also supports many social features like many Nostr integrations like
+  payments. It also supports many social features like Nostr integrations like
   Nostr Wallet Connect, Nostr Contacts, and a Zap Feed.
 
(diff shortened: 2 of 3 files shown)


mutiny-wallet/docker-compose.yml
```diff
@@ -0,0 +1,62 @@
...
+
+  nginx:
+    image: nginx:alpine
+    container_name: nginx
+    restart: always
+    stop_grace_period: 1m
+    ports:
+      - "4499:80"
+    volumes:
+      - ./nginx.conf:/etc/nginx/nginx.conf
```
benthecarman, line 58: do i need to put this in the data folder?

benthecarman:
Trying to test in the dev environment, when I try to install I get this error

```
ERROR: for mutiny-wallet_nginx_1  Cannot start service nginx: driver failed programming external connectivity on endpoint mutiny-wallet_nginx_1 (99a64fb287bd9a2a4221912d8628b2aac2298ddce4516fb72533c8c195e8b76d): Bind for 0.0.0.0:14499 failed: port is already allocated

ERROR: for nginx  Cannot start service nginx: driver failed programming external connectivity on endpoint mutiny-wallet_nginx_1 (99a64fb287bd9a2a4221912d8628b2aac2298ddce4516fb72533c8c195e8b76d): Bind for 0.0.0.0:14499 failed: port is already allocated
```

however if I run `sudo docker ps` I can see it running

benthecarman:
Ready for review!

mutiny-wallet/umbrel-app.yml
```diff
@@ -0,0 +1,33 @@
...
+  Mutiny is a self-custodial lightning wallet that runs in the browser. It is a
+  not a traditional lightning node that is always online, but rather a wallet
+  that runs in your browser.
+
+  Mutiny comes with full functionality for sending and receiving on-chain and lightning
+  payments. It also supports many social features like many Nostr integrations like
+  Nostr Wallet Connect, Nostr Contacts, and a Zap Feed.
+
+  WARNING: This is an beta software, don't put too much money in.
+releaseNotes: >
```
ParthJadhav, line 17: Since this a initial release you can ignore this field.

```
releaseNotes: ""
```

ParthJadhav:
https://github.com/getumbrel/umbrel-apps/pull/787
Ran workflow from above PR, Let me know if something isn't right.

Here is the result:
---
Checklist:

### `docker-compose.yml`
- 🛑  Image ghcr.io/mutinywallet/mutiny-web:v0.4.20@sha256:16e4730bf0049453a74bc41cd622c4d37e631fc2d305b0cb7188a323c7c301eb does not support both linux/amd64 and linux/arm architectures
- ✅ Versioned Images 
- ✅ No proxy auth errors 

### `umbrel-app.yml`
- 🛑  <code>dependencies</code>,  : required fields missing in umbrel-app.yml
- ✅ All optional fields present 
- ✅ Category value is correct

ParthJadhav:
`dependencies` is required because currently there's a bug on the frontend which causes the app page to not render if dependencies are missing.

Will be fixed by: https://github.com/getumbrel/umbrel/pull/1705

benthecarman:
@ParthJadhav fixed!

ParthJadhav:
Hey @benthecarman that looks good. Thanks for the changes.

When I tested it locally it worked properly but when I tested it on a cloud instance where I had multiple other apps. It showing me `snort - nostr` interface instead of `mutiny-wallet`.  

I even deleted `snort - nostr` and restarted Umbrel. Still the problem persists. 

---

beside from that, When I go to receive money there's a copy to clipboard button won't working since the connection is over `http` and not `https` ([Ref](https://stackoverflow.com/a/51823007)). This can be solved by doing something like [this](https://github.com/getumbrel/umbrel-nostr-relay/blob/e39aa894a533778ba0ae9fe9e2bbc646741cbbb7/ui/src/components/utility/CopyText.js#L6-L22). OR the full key can be shown so the user can copy it manually.

ok300:
> When I tested it locally it worked properly but when I tested it on a cloud instance where I had multiple other apps. It showing me `snort - nostr` interface instead of `mutiny-wallet`.

Not 100% sure, but this might be due to an `APP_PORT` conflict. Snort is also using `APP_PORT: 80`.

@benthecarman [you have](https://github.com/getumbrel/umbrel-apps/pull/785/files#diff-196514e71f9b2d3ac03914a4870b91e243942b6764fec973adc3f9956dec7144R7) `APP_PORT: 80` in your `docker-compose.yml`. It might work if you choose one that's not used by anyone else (see [query for used ports](https://github.com/search?q=repo%3Agetumbrel%2Fumbrel-apps+app_port+language%3AYAML&type=code&l=YAML)).

Had a similar issue months ago and that's what fixed it for me.

@ParthJadhav what do you think? Could this be it?

ParthJadhav:
> > When I tested it locally it worked properly but when I tested it on a cloud instance where I had multiple other apps. It showing me `snort - nostr` interface instead of `mutiny-wallet`.
> 
> Not 100% sure, but this might be due to an `APP_PORT` conflict. Snort is also using `APP_PORT: 80`.
> 
> @benthecarman [you have](https://github.com/getumbrel/umbrel-apps/pull/785/files#diff-196514e71f9b2d3ac03914a4870b91e243942b6764fec973adc3f9956dec7144R7) `APP_PORT: 80` in your `docker-compose.yml`. It might work if you choose one that's not used by anyone else (see [query for used ports](https://github.com/search?q=repo%3Agetumbrel%2Fumbrel-apps+app_port+language%3AYAML&type=code&l=YAML)).
> 
> Had a similar issue months ago and that's what fixed it for me.
> 
> @ParthJadhav what do you think? Could this be it?

The app ports seems to be fine. 

`APP_PORT: 80` in `APP_PROXY` points to the port inside the container.  `port: 14499` in `umbrel-app.yml` is the one which is exposed for users to use.

http://umbrel-os.local:14499 (from your browser) -> http://localhost:80 (Inside the container)

benthecarman:
@ParthJadhav that sounds like a port issue, are you sure `APP_PORT: 80` isn't the problem?

benthecarman:
Tried something that might work

ParthJadhav (approved):
Amazing, That works now. Only thing remaining is the PR to be merged.

TonyGiorgio:
Recommending the disability of security features is the wrong approach. Further, it's not just the copy button, there's other browser APIs that depend on HTTPS like camera. And your suggestion to use deprecated APIs and methods is not good. 

This should only be ran in an HTTPS context, either umbrel should provide this or users should not use HTTPS workarounds but use this with tailscale explicitly.

ParthJadhav:
Hey @TonyGiorgio,

These are some thread regarding the security concerns:

https://github.com/getumbrel/umbrel/blob/master/SECURITY.md
https://github.com/getumbrel/umbrel/issues/985#issuecomment-920790411

It's you're right that it's not optimal to disable the security features but it's the team's first priority and hopefully it would be resolved soon. 

Until then it would be amazing if we could have the features proposed in the above PR.

TonyGiorgio:
> Hey @TonyGiorgio,
> 
> These are some thread regarding the security concerns:
> 
> https://github.com/getumbrel/umbrel/blob/master/SECURITY.md
> https://github.com/getumbrel/umbrel/issues/985#issuecomment-920790411
> 
> It's you're right that it's not optimal to disable the security features but it's the team's first priority and hopefully it would be resolved soon. 
> 
> Until then it would be amazing if we could have the features proposed in the above PR.

It's not going to work for us. Https is a requirement. It's far more than just a copy button not working. 

What options do we have in having umbrel users installing certs and informing them this is a requirement?

nmfretz:
Thanks for submitting Mutiny Wallet @benthecarman! Excited for this to be self-hosted.

>It's not going to work for us. Https is a requirement. It's far more than just a copy button not working.
>
>What options do we have in having umbrel users installing certs and informing them this is a requirement?

Understood @TonyGiorgio. Let us discuss internally and get back to you.

TonyGiorgio:
> Thanks for submitting Mutiny Wallet @benthecarman! Excited for this to be self-hosted.
> 
> > It's not going to work for us. Https is a requirement. It's far more than just a copy button not working.
> > What options do we have in having umbrel users installing certs and informing them this is a requirement?
> 
> Understood @TonyGiorgio. Let us discuss internally and get back to you.

I believe I've seen people manually set up their own DNS or tailscale on their umbrel's before. Even if it's something we have to instruct users in the umbrel description, that's an option we can work with.

nmfretz:
@TonyGiorgio @benthecarman, sorry for the delay.

We discussed this internally. The https-via-tailscale option may be too finicky for users to set up since it will require running tailscale commands in docker and also configuring the main umbrel nginx which then won't persist on update.

One interim option while umbrelOS doesn't support https out-of-the-box could be to serve an SSL cert as part of the mutiny web server. This would involve leg-work on your end in order to implement and walk users through the UX of accepting the cert if http is detected. This option might break the auth proxy so that may need to be disabled.

Another possible option with different drawbacks is to instruct users to set up a cloudflare tunnel. There is now a cloudflare app on Umbrel which walks users through setting up a tunnel and public hostnames: https://apps.umbrel.com/app/cloudflared
- This would require users to configure a hostname to the mutiny nginx container directly (bypassing umbrel auth).
- The drawback here is that the tunnel would be public and anyone could access the mutiny web server since there is no in-built auth.
- We could include the cloudflare app as a dependency in the Mutiny app's umbrel-app.yml so that users are required to install it before installing Mutiny (although this doesn't stop them from not using it afterwards).

Obviously we can just wait to add Mutiny Wallet to the app store until we have https support in umbrelOS, but I wanted to give you some options to consider.

TonyGiorgio:
Thanks for that information. 

We've been very happy with our start9 integration and I think we'll leave it at that for now. That's more effort than I think we (and what I expect out of your type of users) to be able to setup and I don't want to support it.

 @benthecarman can you close this? We've gone through a few iterations of docker versioning and setup since then.

nmfretz:
Thanks for your efforts on this anyways @benthecarman and @TonyGiorgio, we really appreciate it.

We'll reach out when we are providing https out-of-the-box. In the meantime, best of luck with continued success for Mutiny Wallet!

PR #790, opened 2023-10-02
feat: update element to v1.11.45

ParthJadhav:
!! This update doesn't work with current Synapse Version !!

Tested on:
- [x] x86_64 -> Digital Ocean droplet
- [x] Arm64 -> Multipass Instance on Apple M2

### DIFF
--- a/element/docker-compose.yml
+++ b/element/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 80
 
   web:
-    image: getumbrel/element-web:v1.8.4@sha256:16b2776278a8a9d8a2397ccd75484f8728fd3223cb3329bf572b50b8f950636c
+    image: vectorim/element-web:v1.11.45@sha256:bfbe1745c57a19579bb4bf1e8760c1f11413cd6d75b036e80787671e1911f838    
     # NGINX parent container requires root
     # user: "1000:1000"
     restart: on-failure
--- a/element/umbrel-app.yml
+++ b/element/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: element
 category: social
 name: Element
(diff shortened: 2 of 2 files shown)


nmfretz:
Awesome, thanks for staging this Parth. We can roll this out at the same time as a compatible Synapse update and include the necessary warning in each app's release notes to update to the latest version for both apps.

Please feel free to take a crack at adding release notes to this PR. This update fast forwards us through a lot of releases, with a large number of changes so there a couple ways we could approach the release notes without making them overwhelming. 
- [Trilium Notes example](https://github.com/getumbrel/umbrel-apps/blob/6d4ae9c4a2af59e6ad9a9d8707771a6a16683f7b/trilium-notes/umbrel-app.yml#L63-L84) where we include the full release notes for the latest version and then reference the release notes for past versions.
- [Electrs example](https://github.com/getumbrel/umbrel-apps/blob/6d4ae9c4a2af59e6ad9a9d8707771a6a16683f7b/electrs/umbrel-app.yml#L32-L35) where we keep it quite generic, but include info about any major change that is relevant for users
- [Elements Core example](https://github.com/getumbrel/umbrel-apps/blob/6d4ae9c4a2af59e6ad9a9d8707771a6a16683f7b/elements/umbrel-app.yml#L15-L31) where we list some select features/improvements that are relevant to users.

Don't get tied down to the examples I've provided if you think of something that makes more sense!

ParthJadhav:
Awesome, Thanks for the resources @nmfretz ! I'll take a look and make changes.

nmfretz:
Closing in favour of:
- Element: https://github.com/getumbrel/umbrel-apps/pull/1228
- Synapse: https://github.com/getumbrel/umbrel-apps/pull/1357

PR #797, opened 2023-10-04
lightning-terminal: update to `v0.11.0-alpha`

ViktorTigerstrom:
In this PR we bump Litd to `v0.11.0-alpha`.

See the release notes here: https://github.com/lightninglabs/lightning-terminal/releases/tag/v0.11.0-alpha

There's one change that I think might be important for Umbrel that I've noted in the release notes:
"One important change to note for users running LiT through docker, is that due to the new status server feature, the startup process of LiT won’t error and end if a sub-server or sub-system fails to start. So any platforms that rely on docker to automatically restart the LiT container if the startup process ends due to the sub-servers or sub-system not having started yet, will now not restart the container as the start process won't exit."

I don't know if this affects your startup process of the container, so just thought I'd make you aware of this in case that's something that will affect you.

Happy to address any potential feedback! Thanks!

### DIFF
--- a/lightning-terminal/docker-compose.yml
+++ b/lightning-terminal/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3004
 
   web:
-    image: lightninglabs/lightning-terminal:v0.10.5-alpha@sha256:cc305b855e55d040a10cb8d0b374e03092ac85e30a531036d6970c36511dfef4
+    image: lightninglabs/lightning-terminal:v0.11.0-alpha@sha256:2c533b7598c536c567eef56250408b96b20d3eaa5f905ff09f017504cf2cdb52
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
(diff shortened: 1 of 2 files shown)


ParthJadhav (approved):
- Tested Multi-Arch Image
- Tested updating the app

nmfretz:
Thanks @ViktorTigerstrom! I have tested with the current LND version on Umbrel (v0.16.4-beta), which is above the [minimum version required](https://github.com/lightninglabs/lightning-terminal/releases/tag/v0.11.0-alpha#:~:text=NOTE%20that%20the%20minimum%20version%20of%20lnd%20that%20can%20be%20used%20in%20%2D%2Dlnd%2Dmode%3Dremote%20is%20v0.16.0%2Dbeta). Going live.

FYI, I made a slight change to the release notes so that the individual paragraphs are separated by a blank line for increased readability:

image

> I don't know if this affects your startup process of the container, so just thought I'd make you aware of this in case that's something that will affect you.

Ah, thanks for letting us know. Really appreciate it.

PR #725, opened 2023-08-14
Update Prowlarr to v1.7.4

jimbrend:
It looks like we used the nightly build before 
I used what looked like the [latest nightly build](nightly-version-1.7.4.3768) for [linux server](https://hub.docker.com/r/linuxserver/prowlarr/tags?page=2&name=1.7.4): 
image

Not sure when next version is out of pre-release does look like 1.8.3 may be ready soon FYI: 
https://github.com/linuxserver/docker-prowlarr/releases

Didn't change media-app-configurator

PR #804, opened 2023-10-14
update Radarr to 5.0.3

ParthJadhav:
Tested on:
- [x] x86_64 -> Digital Ocean droplet
- [x] Arm64 -> Multipass Instance on Apple M2

### DIFF
--- a/radarr/umbrel-app.yml
+++ b/radarr/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1.1
 id: radarr
 category: media
 name: Radarr
-version: "4.1.0.6175"
+version: "5.0.3.8127"
 tagline: Your movie collection manager
 description: >-
   Radarr is a movie collection manager for Usenet and BitTorrent users. It can monitor multiple RSS feeds for new movies and will interface with clients and indexers to grab, sort, and rename them. It can also be configured to automatically upgrade the quality of existing files in the library when a better quality format becomes available. Note that only one type of a given movie is supported. If you want both an 4k version and 1080p version of a given movie you will need multiple instances.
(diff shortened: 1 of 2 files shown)


radarr/umbrel-app.yml
```diff
@@ -18,10 +18,14 @@ gallery:
   - 2.jpg
   - 3.jpg
 path: ""
+releaseNotes: >-
+  Not sure where to find the releaseNotes for the update,
+  releases page on Github seem to be including not much in it.
+
```
nmfretz, line 24: Ya, it looks like our only option is to mash something together from the changes shown in the Github releases. I would suggest that we just say something very simple about this update taking Radarr from 4.1.0.6175 all the way to v5.0.3.8127 and then mention any major changes that change the functionality of the app. For instance it looks like as of v5, UI authentication is mandatory. We can then link to the releases page.

nmfretz (commented):
Thanks for this, Parth. 

I just tested and noticed that there is a similar issue to the [one encountered for Prowlarr](https://github.com/getumbrel/umbrel-apps/pull/725#issuecomment-1756314999), where the mac container does not set up radarr with transmission for new installs.

Also, I see that UI authentication is now mandatory. Have you checked that things still work correctly with this new requirement?

image

ParthJadhav:
> Thanks for this, Parth.
> 
> I just tested and noticed that there is a similar issue to the [one encountered for Prowlarr](https://github.com/getumbrel/umbrel-apps/pull/725#issuecomment-1756314999), where the mac container does not set up radarr with transmission for new installs.
> 
> Also, I see that UI authentication is now mandatory. Have you checked that things still work correctly with this new requirement?
> 
> image

I tested it and it still works. Even with the login page; Transmission is configured properly.

radarr/umbrel-app.yml
ParthJadhav: Okays, I updated the release notes !

nmfretz:
Perfect, thanks Parth! Going live 🎬

PR #803, opened 2023-10-14
update plex to 1.32.6

ParthJadhav:
Tested on:
- [x] x86_64 -> Digital Ocean droplet
- [x] Arm64 -> Multipass Instance on Apple M2

PR #807, opened 2023-10-17
update file browser to v2.25.0

ParthJadhav:
Tested on:
- [x]  x86_64 -> Digital Ocean droplet
- [x]  Arm64 -> Multipass Instance on Apple M2

### DIFF
--- a/file-browser/docker-compose.yml
+++ b/file-browser/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 80
 
   server:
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks @ParthJadhav! Great work. Going live now.

FYI, I removed the intro sentence from the release notes like we did for the recent [plex](https://github.com/getumbrel/umbrel-apps/pull/803#issuecomment-1765677094) update since we aren't skipping any versions between 2.24.2 and 2.25.0, and we are including the full release notes not just the "the following new features".

PR #808, opened 2023-10-17
update Immich to v1.82.0

ParthJadhav:
Tested on:
- [x]  x86_64 -> Digital Ocean droplet
- [x]  Arm64 -> Multipass Instance on Apple M2

Tested both fresh install & version update and verified the functionality by uploading images, deleting images etc.

Very cool update, Adds Trash feature & Websocket support so we don't have to refresh to see updates.

### DIFF
--- a/immich/docker-compose.yml
+++ b/immich/docker-compose.yml
@@ -25,7 +25,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api/*,/search/*"
 
   server:
-    image: ghcr.io/immich-app/immich-server:v1.82.0@sha256:b8368f3a58940f383fd8c4f91389100797ab9df1a66f7cff71d9413ffe1a29d2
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @ParthJadhav, your update works great! I ended up bumping to version 1.82.1 because it came out before I found time to review this, and it includes a bug fix for the newly added websocket support. 

I also updated the security-warning wording in the release notes (https://github.com/getumbrel/umbrel-apps/pull/808/commits/b10b9af0593d97916097ee93675bbad7037184cf) because it is actually referring to two different security flaws than for the last update. And I also added the EXIF metadata warning (https://github.com/getumbrel/umbrel-apps/pull/808/commits/9768a5de7eb4a1174f34d16165902d04cf79e304) that immich noted in their release notes since it seems pretty important considering Immich is aimed at privacy-focused self-hosters.

PR #810, opened 2023-10-18
App submission: Sabnzbd

ParthJadhav:
# App Submission

### 256x256 SVG icon
icon

...

### Gallery images

...

### I have tested my app on:
- [x] [Umbrel dev environment](https://github.com/getumbrel/umbrel-dev)
- [x] Both `x86_64` & `Arm64 architecture.

### DIFF
--- a/sabnzbd/docker-compose.yml
+++ b/sabnzbd/docker-compose.yml
@@ -1,0 +1,19 @@
+version: '3.7'
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: sabnzbd_web_1
+      APP_PORT: 8080
+
+  web:
+    image: lscr.io/linuxserver/sabnzbd:4.1.0@sha256:84a54b2bd29198bc76e253c8592ac72996737d79f845c1db56ab053739cb61bc
+    restart: unless-stopped
+    stop_grace_period: 1m
+    volumes:
+      - ${APP_DATA_DIR}/data/config:/config
+      - ${UMBREL_ROOT}/data/storage/downloads:/downloads
+    environment:
+      - PUID=1000
+      - PGID=1000
+      - TZ=Etc/UTC
--- a/sabnzbd/docker-compose.yml
+++ b/sabnzbd/docker-compose.yml
@@ -12,7 +12,8 @@ services:
     stop_grace_period: 1m
     volumes:
       - ${APP_DATA_DIR}/data/config:/config
-      - ${UMBREL_ROOT}/data/storage/downloads:/downloads
+      - ${UMBREL_ROOT}/data/storage/downloads:/config/Downloads
+      - ${UMBREL_ROOT}/data/storage/downloads/incomplete:/config/incomplete-downloads
     environment:
       - PUID=1000
       - PGID=1000
--- a/sabnzbd/docker-compose.yml
+++ b/sabnzbd/docker-compose.yml
@@ -5,6 +5,7 @@ services:
     environment:
       APP_HOST: sabnzbd_web_1
       APP_PORT: 8080
+      PROXY_AUTH_WHITELIST: "/api/*"
 
   web:
     image: lscr.io/linuxserver/sabnzbd:4.1.0@sha256:84a54b2bd29198bc76e253c8592ac72996737d79f845c1db56ab053739cb61bc
--- a/sabnzbd/docker-compose.yml
+++ b/sabnzbd/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     environment:
       APP_HOST: sabnzbd_web_1
       APP_PORT: 8080
-      PROXY_AUTH_WHITELIST: "/api/*"
+      PROXY_AUTH_WHITELIST: "/api*"
 
   web:
     image: lscr.io/linuxserver/sabnzbd:4.1.0@sha256:84a54b2bd29198bc76e253c8592ac72996737d79f845c1db56ab053739cb61bc
--- a/sabnzbd/docker-compose.yml
+++ b/sabnzbd/docker-compose.yml
@@ -14,7 +14,6 @@ services:
     volumes:
       - ${APP_DATA_DIR}/data/config:/config
       - ${UMBREL_ROOT}/data/storage/downloads:/config/Downloads
-      - ${UMBREL_ROOT}/data/storage/downloads/incomplete:/config/incomplete-downloads
     environment:
       - PUID=1000
       - PGID=1000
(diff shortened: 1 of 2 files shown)


sabnzbd/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+      APP_HOST: sabnzbd_web_1
+      APP_PORT: 8080
+
+  web:
+    image: lscr.io/linuxserver/sabnzbd:4.1.0@sha256:84a54b2bd29198bc76e253c8592ac72996737d79f845c1db56ab053739cb61bc
+    restart: unless-stopped
+    stop_grace_period: 1m
+    volumes:
+      - ${APP_DATA_DIR}/data/config:/config
+      - ${UMBREL_ROOT}/data/storage/downloads:/downloads
```
nmfretz, line 15: Nice, this looks good. The container is running as 1000:1000 so any directories created in `${UMBREL_ROOT}/data/storage/downloads` will have the proper permissions 👌.

### Some thoughts on initial set up

One thing I noticed when checking out the app is that the default downloads folders are set up like this:

image

So in the container downloads are stored at `/config/Downloads/`. We bind `${APP_DATA_DIR}/data/config:/config`, so by default a user would be storing downloads on their Umbrel at `~/umbrel/app-data/sabnzbd/config/Downloads` instead of where we want them to be accessible at `${UMBREL_ROOT}/data/storage/downloads`. This means they wouldn't be able to access them via an app like File Explorer and would have to ssh in to their Umbrel to access them.

Unfortunately it doesn't look like we can pass environment variables to override this (I could be wrong). That leaves us a couple options. I'm not sure what's best and would love your thoughts:

1. We include instructions in the app description to change the download folders to `/downloads/complete` and `/downloads/incomplete` in the UI:

image

2. We add script logic to edit the config so that by default `/downloads` inside the container is used by default

3. We change the bind mounts in such a way that we bind `/config/Downloads` in the container to the main downloads directory on the host.

nmfretz (commented):
Thanks @ParthJadhav, solid gallery assets you wizard!

I've left some comments below. We're very close, I think we just need to address a couple things so that users can easily set up sabnzbd.

sabnzbd/docker-compose.yml
ParthJadhav: Hey @nmfretz , I just quickly went over it and discovered that this is actually an Issue.

https://github.com/linuxserver/docker-sabnzbd/issues/71#issuecomment-681632799

When the Sabnzbd wizard runs it seems to change / modify the `sabnzbd.ini` file. So I don't think we can create a PostInstal script to change the paths. 

I haven't looked at option 3 yet, I'll look into it.

sabnzbd/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
+version: '3.7'
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: sabnzbd_web_1
+      APP_PORT: 8080
```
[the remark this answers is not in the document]
ParthJadhav: Right, I'll take a look soon.

sabnzbd/docker-compose.yml
ParthJadhav: Hey @nmfretz, I fixed this issue by the 3rd option you mentioned. 

> We change the bind mounts in such a way that we bind /config/Downloads in the container to the main downloads directory on the host.

sabnzbd/docker-compose.yml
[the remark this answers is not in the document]
ParthJadhav: Regarding this issue, I enabled the access to `/api/*` route. But I still wasn't able to add `Sabnzbd` as download client in other apps. 

It seems there are couple of steps before being able to do that.

1. We need to add `sabnzbd_web_1` to `host_whitelist` in `settings -> special  -> values -> host_whitelist` OR We can use the `container-ID` as Host. It is shown right after the `Sabnzbd` wizard is done along with 2 other URLs. Reference: https://sabnzbd.org/wiki/extra/hostname-check.html 

2. We need to setup the category `Indexer Categories / Groups` in `settings -> categories`. It should be equal to what you're accessing from the respective `*arr` app. For eg: if I'm adding `Sabnzbd` as download client in `Sonarr` then I have to go to `settings -> categories` find `Indexer Categories / Groups` for `tv` and add `tv` to that field.

Should I write an article to communicate this to the User or is there any other way to do it?

sabnzbd/docker-compose.yml
nmfretz: Nice. Ya, I like this option as well as I think it is the one with the least overhead for us moving forward. We don't have to worry about the structure of the sabnzbd.ini file changing in the future like we would with option 2. And I like it better than Option 1 because we reduce the amount of set up that a user has to do manually.

> Hey @nmfretz , I just quickly went over it and discovered that this is actually an Issue.

Oh ya, good find. The issue actually links to the README where they say this is the desired behaviour:
https://github.com/linuxserver/docker-sabnzbd#application-setup

We still have to make a change to your current bind mounts:
```yaml
volumes:
      - ${APP_DATA_DIR}/data/config:/config
      - ${UMBREL_ROOT}/data/storage/downloads:/config/Downloads
      - ${UMBREL_ROOT}/data/storage/downloads/incomplete:/config/incomplete-downloads
```

The default location that the set-up wizard uses for incomplete downloads is `/config/Downloads/incomplete` instead of `/config/incomplete-downloads`. That being said, I think we should just bind `/config/Downloads` and we will get both complete and incomplete. So we would do:

```yaml
volumes:
      - ${APP_DATA_DIR}/data/config:/config
      - ${UMBREL_ROOT}/data/storage/downloads:/config/Downloads
```

image

ParthJadhav: Ah right, My bad on that one. I've removed the incomplete downloads mount and only included the downloads dir.

sabnzbd/docker-compose.yml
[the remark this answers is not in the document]
ParthJadhav: Thanks for the detailed explanation @nmfretz  !! I've made sure to fix those issues in the latest commits.

nmfretz:
Thanks for the changes Parth. I have tested on arm64 and x86(amd64) and this is now working well. We're just making some changes to the gallery assets so that they are differentiated a bit from autobrr and then we'll go live!

nmfretz:
Gallery assets are done. Going live. Thanks again @ParthJadhav!

ernesto75:
Hello @ParthJadhav i want to thank you for your amazing work! I use Sabnzbd on Umbrel a lot. Can you or @nmfretz  just add ffmpeg in the container? Will be so useful for running in scripts to automate the format conversion of videos. Thanks a lot again!

PR #811, opened 2023-10-19
lightning-terminal: update to `v0.12.0-alpha`

ViktorTigerstrom:
In this PR we bump Litd to `v0.12.0-alpha` which adds support for Taproot Assets on the Bitcoin mainnet!

See the release notes here: https://github.com/lightninglabs/lightning-terminal/releases/tag/v0.12.0-alpha

There's one extra important release note for Umbrel users that I'll include until we have added some additional backup procedure to mitigate the current backup procedure for Umbrel Taproot Assets users:

## Important note for Umbrel/Lightning Terminal users
**DO NOT UNDER ANY CIRCUMSTANCE** uninstall (or re-install) the "Lightning Terminal" app without first making a manual backup of all local tapd data, if you are using Taproot Assets as part of the "Lightning Terminal" app with Umbrel. Uninstalling Umbrel apps deletes application data. This Taproot Assets application data encumbers Taproot Assets AND bitcoin funds. Receiving and sending tapd assets updates the daemon's funds-custody material. Merely having the lnd seed phrase is NOT enough to restore assets minted or received. WITHOUT BACKUP BEFORE DELETION, FUNDS ARE DESTROYED.

IMPORTANT NOTE: To avoid loss of funds, it's imperative that you read the [Operational Safety Guidelines](https://github.com/lightninglabs/taproot-assets/blob/main/docs/safety.md) before using tapd on mainnet!

Happy to address any feedback!

### DIFF
--- a/lightning-terminal/docker-compose.yml
+++ b/lightning-terminal/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3004
 
   web:
-    image: lightninglabs/lightning-terminal:v0.11.0-alpha@sha256:2c533b7598c536c567eef56250408b96b20d3eaa5f905ff09f017504cf2cdb52
+    image: lightninglabs/lightning-terminal:v0.12.0-alpha@sha256:11eab3e46437844e7e227cb32880a087bca7637685914cab07b758eab8211fea
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks for this update as usual @ViktorTigerstrom! I have tested on arm64 and x86(amd64). Going live now.

What are your thoughts on adding the new backup warning to the `description` in the umbrel-app.yml? That way it will show up front and centre in the app store description when users go to install or check out Lightning Terminal. We can add this to the description anytime without updating the app.

ViktorTigerstrom:
Thanks @nmfretz! Sorry for the late reply to your question. That sounds like a great idea! Would you like me to create a new PR with the warning text added?

PR #814, opened 2023-10-24
Fix to use RPC port

usernameisJim:
Fixes issue when running Bitcoin in testnet, Core Lightning will also now automatically open in testnet successfully, thank you!
Worked and opened app at 97% Bitcoin testnet synced and opening at 100% showing: 
image

Not sure if denoting that it is indeed in testnet?

something app proxy can handle?

### DIFF
--- a/core-lightning/docker-compose.yml
+++ b/core-lightning/docker-compose.yml
@@ -64,7 +64,7 @@ services:
       - --bitcoin-rpcconnect=${APP_BITCOIN_NODE_IP}
       - --bitcoin-rpcuser=${APP_BITCOIN_RPC_USER}
       - --bitcoin-rpcpassword=${APP_BITCOIN_RPC_PASS}
-      - --bitcoin-rpcport=8332
+      - --bitcoin-rpcport=${$APP_BITCOIN_RPC_PORT} 
       - --lightning-dir=${CORE_LIGHTNING_PATH}
       - --proxy=${TOR_PROXY_IP}:${TOR_PROXY_PORT}
       - --bind-addr=${APP_CORE_LIGHTNING_DAEMON_IP}:9735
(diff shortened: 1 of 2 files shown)


nmfretz:
Awesome, thanks @usernameisJim! I will test soon.

core-lightning/docker-compose.yml
```diff
@@ -64,6 +64,7 @@ services:
       - --bitcoin-rpcconnect=${APP_BITCOIN_NODE_IP}
       - --bitcoin-rpcuser=${APP_BITCOIN_RPC_USER}
       - --bitcoin-rpcpassword=${APP_BITCOIN_RPC_PASS}
+      - --bitcoin-rpcport=8332
```
nmfretz, line 67: We can use the environment variable `${$APP_BITCOIN_RPC_PORT}` here instead of `8332` because we export it from the [Bitcoin app](https://github.com/getumbrel/umbrel-apps/blob/a1579886cd0c366b81f6312ef5eba43c8b0180b5/bitcoin/exports.sh#L7). This way any changes we make to the Bitcoin app will automatically get carried over here.

nmfretz:
@usernameisJim in order for users to see an update on their Umbrel we'll need to change the `version` in the umbrel-app.yml. Shall we go with something like "v23.08-hotfix-1"?

Then we can also add a note to the bottom of the existing release notes that says something like:

"Note: This update fixes a bug that was resulting in the app not working for networks other than mainnet"

core-lightning/docker-compose.yml
usernameisJim: fixing

usernameisJim:
I believe I correctly added 
`- --bitcoin-rpcport=${APP_BITCOIN_RPC_PORT}`
to line 67

Updated version umbrel-app.yml to
`23.08-hotfix-1`

and I think I added spacing correctly for Note, 

Will defer to you if that looks good now?  Ty for your input!

nmfretz:
Thanks again @usernameisJim! I have tested and this is working well. Going live ⚡️

PR #815, opened 2023-10-25
Circuitbreaker: Upgrade to v0.5.0

carlaKC:
This PR upgrades circuitbreaker to v0.5.0, which contains an upgrade to start storing more granular HTLC forwarding data. 

The final section of the release notes has a call to action to help out with ongoing research for protocol work - happy to remove it if it's not appropriate!

### DIFF
--- a/circuitbreaker/docker-compose.yml
+++ b/circuitbreaker/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 9235
       
   web:
-    image: ghcr.io/lightningequipment/circuitbreaker:v0.4.3@sha256:b8e0505da638ff47d69f0c3af8140d82dde970dadadf5949565426c4f3c9ded9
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks very much for this update to Circuit Breaker @carlaKC. It is totally fine to put a call to action in the release notes. Hopefully you get some people reaching out to you because of it! Tested on arm64 and x86(amd64). Going live.

-------------------

FYI, I made the following changes:

I changed the release notes by adding additional blanklines so they look like this in the UI:

image

Instead of like this:

image

In the current version of umbrelOS we don't render markdown in the app store for security reasons, so I changed the hyperlink markdown notation so that `[]` and `()` don't show up around the text and urls.

I also tweaked the docker image reference so that we don't get a manifest error when installing/upgrading the app https://github.com/getumbrel/umbrel-apps/pull/815/commits/6a9b230ba0e38559ce540f95430116b16ca3b478

carlaKC:
Thanks for the fixups and the quick review 🚀

PR #818, opened 2023-11-03
Update Torq to v1.4.1

ohenrik:
New Torq version with a lot of bug fixes.

  List of changes:

  -  You can now duplicate views
  
  -  Add payment request to payments page
  
  -  Closed channels sometimes reappeared as active when delayed gossip was processed
  
  -  Pagination indicated to many pages when filtering
  
  -  You can now choose refresh interval for each list view
  
  -  Fix turnover layout and calculation
  
  -  Channel filter "Active" is now true when both sides of the channel are active (used to be local side only)
  
  -  Improvements on channel inspect page:
  
  -  Show channel open and close dates on charts
  
  -  Added basic channel info
  
  -  Indicators if data is loading
  
  Note: This release contains database migrations and cannot be rolled back to a previous version.

### DIFF
--- a/torq/docker-compose.yml
+++ b/torq/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 7028
 
   web:
-    image: lncapital/torq:1.2.1@sha256:68fdcd6739d2828fd68beec07b9a1f0bd0949736ecc793e8e144cff0be5fba47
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks very much @ohenrik. Tested on arm64 and x86(amd64). Going live.

PR #820, opened 2023-11-07
Update Syncthing to 1.26.0

jon-ryan:


nmfretz:
Perfect, thank you for this @jon-ryan. Tested on arm64 and x86(amd64). Going live.

PR #823, opened 2023-11-10
Update Immich to v1.85.0

nmfretz:


nmfretz:
Tested on arm64 and x86(amd64) with current mobile app. New features work.

PR #824, opened 2023-11-10
Update Plex to v1.32.7.7621

nmfretz:


nmfretz:
Tested on arm64 and x86(amd64) 🍿

PR #825, opened 2023-11-10
Update Pi-hole to v2023.10.0

nmfretz:


nmfretz:
Minor bug-fix release. Tested on arm64 and x86(amd64).

PR #827, opened 2023-11-11
Update qbittorrent to 4.6.0

nmfretz:


nmfretz:
Tested on arm64 and amd64.

PR #828, opened 2023-11-11
Update alby-nostr-wallet-connect to 0.3.0

nmfretz:


nmfretz:
Tested on arm64 and amd64. Tested connection to amethyst and zap.

PR #830, opened 2023-11-11
Update autobrr to v1.33.0

nmfretz:


nmfretz:
Tested on arm64/amd64.

PR #831, opened 2023-11-11
Update Bazarr to v1.3.1

nmfretz:


nmfretz:

PR #842, opened 2023-11-11
Update tailscale to v1.52.0

smolgrrr:
Release notes here https://github.com/tailscale/tailscale/releases

Update works pretty seamlessly which is nice. I updated on my promox umbrel instance and it didn't even drop out.
Did a completely fresh install on Pi instance, and added to tailnet- no issues!
Changed the formatting for release notes so newlines worked.

* [x]   x86_64 -> UmbrelOS on proxmox Debian instance (update)

### DIFF
--- a/tailscale/umbrel-app.yml
+++ b/tailscale/umbrel-app.yml
@@ -29,6 +29,7 @@ deterministicPassword: false
 torOnly: false
 releaseNotes: |
   This release updates Tailscale from v1.48.1 to v1.52.0 which includes: 
+  
   v1.50.0
   - tailscale ping now sends an ICMP Ping code of 0.
   - UPnP falls back to a permanent lease if a limited lease fails, some servers only support permanent.
--- a/tailscale/umbrel-app.yml
+++ b/tailscale/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: tailscale
 category: networking
 name: Tailscale
-version: "v1.48.1"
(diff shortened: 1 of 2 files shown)


smolgrrr:
Please advise if this checksum check needs to be updated.
https://github.com/getumbrel/umbrel-apps/blob/master/tailscale/exports.sh

tailscale/umbrel-app.yml
```diff
@@ -27,9 +27,29 @@ gallery:
 path: ""
 deterministicPassword: false
 torOnly: false
-releaseNotes: >-
-  This release updates Tailscale from v1.46.1 to v1.48.1.
+releaseNotes: |
+  This release updates Tailscale from v1.48.1 to v1.52.0 which includes:
```
nmfretz, line 31: Dude, brilliant use of  the literal `|` instead of the folded `>-` to preserve formatting. This is way better for the description and release notes.

nmfretz:
> Please advise if this checksum check needs to be updated.
https://github.com/getumbrel/umbrel-apps/blob/master/tailscale/exports.sh

Thanks for the thoroughness here @smolgrrr. We don't need to change these checksums. These are checksums for specific versions of the umbrel `app` script (https://github.com/getumbrel/umbrel/blob/master/scripts/app), not for specific versions of Tailscale. We essentially check if a user has certain `app` scripts and patch them if so.

nmfretz:
Great work on this @smolgrrr. Going live to the app store.

PR #846, opened 2023-11-12
Update uptime-kuma to 1.23.3

smolgrrr:
Release notes: https://github.com/louislam/uptime-kuma/releases

Only weird issue I noticed is that it takes a weirdly long time to install on the Pi. Wasn't able to resolve it, if you dont mind, maybe check if this happens on your Pi too.

* [x]    x86_64 -> UmbrelOS on proxmox Debian instance (update and fresh install)
* [x]    Arm64 -> Pi 4 8GB (fresh install)

### DIFF
--- a/uptime-kuma/docker-compose.yml
+++ b/uptime-kuma/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3001
 
   server:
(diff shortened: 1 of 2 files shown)


nmfretz:
Wonderful. Install was super quick for me both on my RPi, so it may have been a RAM related issue for you if you've got a lot of stuff on the Pi for testing currently. Going live.

PR #852, opened 2023-11-15
Update trilium-notes to 0.61.13

smolgrrr:
Release notes: https://github.com/zadam/trilium/releases

Note the main dev does say this: 'It's possible to upgrade to 0.61 only from 0.60.X. If you're on an older version, download and run 0.60.4 first. Only after that upgrade to 0.61.'

I had no issues updating (it was 0.60.4 before this), but should we be aware of users that may be on an older version because they can't manually go between versions?


smolgrrr:
As this is an upgrade skipping a major release, it will need to be changed to be non-breaking

PR #858, opened 2023-11-20
Update Torq to v1.4.2

ohenrik:
New Update fixing some important things:

Improve CLN's funds performance
Fix CLN payments import
Fix DST problem in profit chart
Update of packages i.e. Open-Telemetry, gRPC, cobra, ...
Reduce logging


ohenrik:
Adding 1.4.3 soon. Discovered bug that should be included.

PR #860, opened 2023-11-21
Torq v1.4.3

ohenrik:


ohenrik:
@nmfretz I would really appreciate a swift merge of this update. It contains a fix to a bug that prevents users from signing up on Umbrel after the trial.

nmfretz:
Tested on arm64 and x86(amd64). Both app update and fresh install work well. Thanks for the update @ohenrik!

ohenrik:
Awesome! Thank you!

Issue #865, opened 2023-11-22
Immich 1.88.2


felipe2102:
Can someone update Immich to version 1.88.2?

johnpc:
This is needed for the iOS mobile app to allow logging in to Immich

PR #867, opened 2023-11-26
Update Immich to 1.88.2

johnpc:
For compatibility with mobile clients, Immich must be updated. See https://github.com/getumbrel/umbrel-apps/issues/865

1.88.0 of immich includes a breaking change, including breaking login functionality for mobile clients.

See documentation on update path here: https://github.com/immich-app/immich/discussions/5086

This PR performs the update as documented.

### DIFF
--- a/immich/docker-compose.yml
+++ b/immich/docker-compose.yml
@@ -19,7 +19,7 @@ services:
   app_proxy:
     environment:
       APP_HOST: immich_server_1
-      APP_PORT: 8080
+      APP_PORT: 3001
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for getting on this @johnpc! I have pushed some commits to get our app proxy working with the new Immich changes, to get it working on both arm64 and x86 devices, and to make it so that people currently running Immich will get an update notification.

I have tested on arm64 and x86(amd64). Going live!

PR #871, opened 2023-11-30
Update esphome to 2023.11.6

smolgrrr:
Release notes: https://github.com/esphome/esphome/releases


nmfretz:
Looks great. I added a warning to the release notes which we can include in all updates for esphome going forward. I took a look through their release notes for the past while and it looks like they often introduce breaking changes that will affect configurations for individual devices.

PR #884, opened 2023-12-03
Update node-red-standalone to 3.1.1

smolgrrr:
Release notes: https://github.com/node-red/node-red/releases/tag/3.1.0


nmfretz:
Nice work. We switched to the node 18 image last update (https://github.com/getumbrel/umbrel-apps/pull/693#issuecomment-1639655954), but I see that node 20.x is also recommended by node-red so I agree with your change of image going forward https://nodered.org/docs/faq/node-versions:

image

PR #888, opened 2023-12-04
Update prowlarr to 1.10.5

smolgrrr:
Release notes: https://github.com/Prowlarr/Prowlarr/releases

This is a major update and will require users to set-up a login.


nmfretz:
I updated the media-app-configurator (mac container) to work with newer versions of Prowlarr that now require a `collections` field to be passed when setting up transmission via the api. Going live 🐅.

PR #893, opened 2023-12-05
Update qbittorrent to 4.6.2

smolgrrr:
Release notes: https://www.qbittorrent.org/news#mon-nov-27th-2023---qbittorrent-v4.6.2-release

Starting from v4.6.1 the default password is no longer adminadmin, so you cannot log in and the moment without pulling the new password from the .config directory directly. Will need to add that to umbrel-app.yml

### DIFF
--- a/qbittorrent/docker-compose.yml
+++ b/qbittorrent/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   server:
-    image: hotio/qbittorrent:release-4.6.0@sha256:78ac9dfad4e1bbb7233e11e09781c929e692d92d5de8f499c0f4477eee266622
+    image: hotio/qbittorrent:release-4.6.2@sha256:9fec00002f576ea20faea2934b4cce9e5aec06a64ffe46ba077bca5644be1b0a
     environment:
       - PUID=1000
       - PGID=1000
--- a/qbittorrent/umbrel-app.yml
+++ b/qbittorrent/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: qbittorrent
 category: networking
 name: qBittorrent
-version: "4.6.0"
+version: "4.6.2"
 tagline: Free and reliable P2P Bittorrent client
 description: >-
   qBittorrent is an open-source software alternative to µTorrent. It's designed to meet the needs of most users while using as little CPU and memory as possible.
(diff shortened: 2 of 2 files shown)


nmfretz:
>Starting from v4.6.1 the default password is no longer adminadmin, so you cannot log in at the moment without pulling the new password from the .config directory directly. Will need to add that to umbrel-app.yml

Ah, damn that's actually a giant pain. And it looks like it's actually even worse than that. Instead of a new password being written to the config directory on fresh install / update, a temporary password is actually output to the container logs. You need to read the qbittorrent container logs to get the temporary password, use this to login, and then remember to change your password in qBittorrent's settings to something custom. If you don't do this, then the temporary password will change on app restart and you'll have to go through the same process again.

Ideally we want the account configuration process to happen via qBittorrent itself (like through its UI), so that we aren't taking on the overhead of having to make sure that any custom scripts we write to modify a config file still work with each app update. But I think our only real option here is to write a custom script.

What do you think of this approach for a pre-start script:
- check to see if a qBittorrent.conf exists AND has a line starting with `WebUI\Password_PBKDF2`
- If both are true, then a user has already set a custom password. W do nothing so we don't overwrite a users custom password.
- Otherwise, we need to set a default password for the user.
- We start up the qBittorrent container and wait until it generates a default qBittorrent.conf file
- Once generated we then write out a default password to file and restart the container.

The default password can be a deterministic password generated from a users seed like we do for other apps like RTL so that it will be different for everyone. We can show it to the user in the app store immediately after clicking "Install".

nmfretz:
Finally came back to this.

I have made two main changes:

1. Changed the downloads bind path to the default downloads directory that qBittorrent comes installed with: https://github.com/getumbrel/umbrel-apps/pull/893/commits/79abd9834df561777b2c1cf4d036811a88e073b6

    This makes it so that without any additional configuration, qBittorrent will save downloads to the main Umbrel downloads folder. This closes https://github.com/getumbrel/umbrel/issues/1765.

2. Added a post start script to handle the change where qBittorrent prints a temporary password to container logs instead of allowing a default password to be set on install: https://github.com/getumbrel/umbrel-apps/pull/893/commits/6d42737486a3522c12ad8570900e156942ed7051

    We check if a password has already been set by the user. If it has, we do nothing. If it has not been set, we set it to `adminadmin`.

    I added a note to the app description recommending that users change the default password

PR #894, opened 2023-12-05
Update specter-desktop to 2.0.2

smolgrrr:
Release notes: https://github.com/cryptoadvance/specter-desktop/releases?page=1

### DIFF
--- a/specter-desktop/docker-compose.yml
+++ b/specter-desktop/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 25441
 
   web:
-    image: lncm/specter-desktop:v1.14.5@sha256:beea686b08e1ed47d293fd39c125dfd4be7629f60b4591474e6bcce7dcd0466f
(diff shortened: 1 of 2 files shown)


nmfretz:
@smolgrrr with this newest update does Specter connect automatically to Bitcoin Core? Take a look at this previous PR where we made the decision to stay on v1.14.5 https://github.com/getumbrel/umbrel-apps/pull/523. Perhaps this has now been resolved.

nmfretz:
Updated the app description and tweaked release notes. Tested on arm64 and x86(amd64) as well. Thanks @smolgrrr.

With the updated UI we should consider updating the gallery images (done outside of this PR).

For our records:
We have decided to push this update of Specter Desktop even though environment variables can no longer be used to connect a local Bitcoin Core connection via docker compose: https://github.com/cryptoadvance/specter-desktop/issues/2342

Users who installed the app prior to this update (i.e., v1.14.5 and below) will remain connected to Bitcoin Core running on their Umbrel without issue. New installs will go through Specter's in-app set-up.

@k9ert - pinging you here again just to keep this on your radar for if/when the team has time to look into the break in this functionality.

PR #896, opened 2023-12-05
Update README.md

smolgrrr:
Removing references to umbrel-dev in App Submission process


nmfretz:
Good call @smolgrrr! Looks good to me. We could consider adding info for using multipass to virtualize a local Umbrel instance in a separate PR if we want.

```bash
multipass launch --name umbrel-dev --disk 10G
multipass exec umbrel-dev -- sh -c 'curl -L https://umbrel.sh | bash'
multipass shell umbrel-dev
```

PR #907, opened 2023-12-15
Update invoice ninja to 5.7.62

smolgrrr:
Release notes: https://github.com/invoiceninja/invoiceninja/releases

**I'm having two slight issues**

- Invoice ninja requires app_key now: I've hardcoded it for the moment, and not sure on the best way to do with it so it persists for users in future updates. See my idea for exports.sh (doesn't work)

- After installing you will get nginx errors for a min or so, not a big issue, but a lil confusing.

### DIFF
--- a/invoice-ninja/config/nginx.conf
+++ b/invoice-ninja/config/nginx.conf
@@ -2,6 +2,8 @@ server {
     listen 80 default_server;
     server_name _;
 
+    server_tokens off;
+
     client_max_body_size 100M;
 
     root /var/www/app/public/;
@@ -13,6 +15,10 @@ server {
 
     location = /favicon.ico { access_log off; log_not_found off; }
     location = /robots.txt  { access_log off; log_not_found off; }
+
+    location ~* /storage/.*\.php$ {
+        return 503;
+    }
 
     location ~ \.php$ {
         fastcgi_split_path_info ^(.+\.php)(/.+)$;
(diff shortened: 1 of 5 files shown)


smolgrrr:
Retested with new APP_KEY set-up

Working great @nmfretz

nmfretz:
> Retested with new APP_KEY set-up

Awesome, thanks. I've bumped to latest 5.8.15 and retested. Going live.

For our records, we are generating the laravel APP_KEY with a pre-start script:

```bash
LARAVEL_APP_KEY_FILE_PATH="${APP_DATA_DIR}/laravel-app-key.txt"


# We only generate and save the Laravel APP_KEY if it doesn't exist already
if [[ ! -f "${LARAVEL_APP_KEY_FILE_PATH}" ]]; then
	#  generates a laravel APP_KEY using the invoice-ninja container --> php artisan key:generate --show
	APP_KEY=$("${UMBREL_ROOT}/scripts/app" compose "${APP_ID}" run --rm app php artisan key:generate --show)
	# we remove the color codes and newlines from the output which cause laravel to fail to parse the APP_KEY
	echo "${APP_KEY}" | sed 's/\x1b\[[0-9;]*m//g' | tr -d '\r\n' > "${LARAVEL_APP_KEY_FILE_PATH}"
fi
```

- check for the existence of a file that we write out the APP_KEY value to
- if the file doesn't exist then compose up the invoice ninja container and run `php artisan key:generate --show`. We then write this value out to a persistent file.
- The exports.sh file exports a variable called APP_KEY that consists of the contents of the persistent file.

I did it this way so that we're completely sure that the APP_KEY has been generated properly via `php artisan key:generate --show`. The key is only generated once and we do not overwrite it afterwards, so that it persists across restarts and updates. If a user were to back up the invoice ninja data directory then this key will also be backed up, which would be a benefit if Invoice Ninja ever uses this to encrypt the data and a user accidentally deletes invoice ninja but has a backup.

PR #910, opened 2023-12-16
Update bazarr to 1.4.0

smolgrrr:
Release notes: https://github.com/morpheus65535/bazarr/releases

Only test I did for data persistence was adding an indexer- I noticed that it didn't stick across the update, but I may not have set it up properly. Another than that all worked well.


nmfretz:
Excellent, looks perfect @smolgrrr. 
> I noticed that it didn't stick across the update, but I may not have set it up properl

All my providers persisted across the update without issue, so may have just been a hiccup when testing.

PR #916, opened 2023-12-21
Update Thunderhub  v0.13.23 --> v0.13.29

davisv7:
Update performed manually locally. Data persisted after update.


nmfretz:
Thanks @davisv7! I added the multi-arch digests for the docker image so that the update will work for both arm64 and x86 users. I also updated the release notes.

Tested on arm64 and x86(amd64). Going live 🌩️

PR #918, opened 2023-12-22
Update BTCPay Server to v1.12.3

nmfretz:


nmfretz:

PR #921, opened 2023-12-27
Update LNbits to v0.11.3

nmfretz:


nmfretz:
Test on arm64 and x86(amd64).

PR #926, opened 2024-01-02
Update octoprint to 1.9.3

smolgrrr:
Release notes: https://github.com/OctoPrint/OctoPrint/releases/tag/1.9.3

This is a requested update. Part of the request is that they've been having trouble connecting via USB- `privileged: true` is specified, and `- /dev:/dev` is correct, maybe this needs to be booted in host mode also?

### DIFF
--- a/octoprint/docker-compose.yml
+++ b/octoprint/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   web:
-    image: octoprint/octoprint:1.9.2@sha256:80ed8b8a1be8713506e7e29fc5b2ec55241242e28a5d8a474a25d9f2aab0843f
+    image: octoprint/octoprint:1.9.3@sha256:32e81a9d8c73511855ad5b9d7212ef61e0ba7608c47ce215ec2c6402c52423ee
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @smolgrrr! 

> Part of the request is that they've been having trouble connecting via USB

Host network mode will make it so that the octoprint container shares the host's networking namespace (would then need to remove the app proxy), but unfortunately it wouldn't do anything for giving the container access to USB devices.

Is this a Raspberry Pi user having issues? My suspicion is that this may be due to a power consumption issue with Raspberry Pi where the printer is unable to connect when an SSD is already being used by the Raspberry Pi.

@mateosilguero, do you by chance have a Raspberry Pi that you can test printer connection on?

Let's get this update out now and we can always push an update, if necessary, once someone is able to test printer connection on a Raspberry Pi 4.

mateosilguero:
Hi ! I updated to the new version, and works as expected. (raspi 4)
I have a printer, so let me know if you need help for testing.

nmfretz:
> Hi ! I updated to the new version, and works as expected. (raspi 4)

Amazing, thanks a ton for confirming @mateosilguero!

PR #927, opened 2024-01-04
Update firefly to 6.1.6

smolgrrr:
Release notes: https://github.com/firefly-iii/firefly-iii/releases

### DIFF
--- a/firefly-iii/docker-compose.yml
+++ b/firefly-iii/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api/*"
 
   server:
-    image: fireflyiii/core:version-6.1.4@sha256:b9905d30934be51453bbb956be88c9ff9db35edce6f3779f8fae1ba87e820a69
+    image: fireflyiii/core:version-6.1.6@sha256:da0121d50fdf40443eef0f70f983711a45508ffff6ff09ef698dfd06966927e5
     # This image runs under the user/group www-data:www-data
     # user: "1000:1000"
     restart: on-failure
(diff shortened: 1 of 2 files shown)


smolgrrr:
Moving this to draft, and new data import feature is not working

nmfretz:
@smolgrrr I got the importer working. Can you give this a test as well please?

The `FIREFLY_III_URL` gets set to the internal compose service name (http://server:8080), but then we need to include a `VANITY_URL` as well so that the importer makes the correct requests (http://${DEVICE_DOMAIN_NAME}:30009):
https://raw.githubusercontent.com/firefly-iii/data-importer/main/.env.example#:~:text=%23%20Imagine%20Firefly%20III,of%20the%20URL.

It's kind of a pain that the importer is completely separate and running on a different port, but there's not much we can do about it. We'll need to instruct users in the app description to visit http://umbrel.local/30010 to get to the importer which isn't great UX.  I suppose we could do some fancy proxying and instruct users to visit /importer from the firefly app, but then we'd be making it quite non-standard, which would make it less easy for users to troubleshoot issues by searching the internet since those instructions and troubleshooting steps wouldn't match out set up.

smolgrrr:
Ah yes this is a really good approach- I was thinking it was gonna need to be a seperate app lol.
Both are working great for me.

Added a little something about it to the description.

nmfretz:
Actually @smolgrrr let's do a separate app like you were thinking. At first I was thinking this needed to be in the same compose file because the importer depends on the server already being up and running before starting, but I think we could get around this with a pre-start script that checks that firefly is running before starting the importer. And we should just be able to use `http://firefly-iii_server_1:8080` for the FIREFLY_III_URL.

This way there is an obvious way users to get to the importer app.

I will close this PR in favour of https://github.com/getumbrel/umbrel-apps/pull/943/files and add the importer as a separate app.

PR #941, opened 2024-01-17
Update Invidious to 2024.1.10-1c0b420

ceramicwhite:


nmfretz:
Perfect, thanks for this update @ceramicwhite! Tested on arm64 and x86(amd64).

PR #945, opened 2024-01-23
Update umbrel-app.yml

RomainRouphael:
description update


nmfretz:
Thanks @RomainRouphael! I added in the newlines so that the yml gets rendered properly in the app store. It will look like this:

image

RomainRouphael:
That's perfect, thanks Nathan!

PR #948, opened 2024-01-24
App Submission: Firefly III Data Importer

nmfretz:
Repo: https://github.com/firefly-iii/data-importer
Website: https://docs.firefly-iii.org/explanation/data-importer/about/introduction/

I have added this as a separate app instead of including the data importer as a service within the Firefly III compose file. This way there is an obvious way for users to get to the importer app (by installing it and having the app visible on their dashboard). If the importer was a service within the Firefly III compose file, then it would be invisibly running on port 30010 and we would need to provide instructions in the app description on how to access it (e.g., navigate to http://umbrel.local/30010) which isn't great UX.

Tested in production on arm64 and x86.

### DIFF
--- a/firefly-iii-importer/docker-compose.yml
+++ b/firefly-iii-importer/docker-compose.yml
@@ -1,0 +1,15 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: firefly-importer_importer_1
+      APP_PORT: 8080
+
+  importer:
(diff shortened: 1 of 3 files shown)


johnpc:
Any interest in getting the plaid connector set up for us folks in the US? 

https://gitlab.com/GeorgeHahn/firefly-plaid-connector

nmfretz:
> Any interest in getting the plaid connector set up for us folks in the US?

@johnpc thanks for the suggestion! I don't have the capacity at the moment to add it, but I will keep it on my radar. In the meantime please feel free to take a crack at it. Briefly reading the description on gitlab it looks like there would be no easy way for a user to set up the connector through the UI and they'd have to manually add info the their config FYI.

PR #956, opened 2024-02-01
Update Torq to v1.5.1

ohenrik:
Release notes:

- CLN v23.11+ is required (pagination)
- Integration with Kraken
- New workflow triggers: Invoice, Forward, Payment, Transaction (on-chain)
- New workflow filters: Mempool, Exchange (Kraken) balance
- New workflow action: Payment Attempt (experimental: tries to find a route to pay an exchange invoice)
- New variables in workflow actions: API client, notification
- Replace intercom with chatwoot
- Small changes to rebalancer + new metrics and logs
- Timelock Delta in seconds on list screens
- Update of packages

Notes:

- This release contains a large set of database updates. So the migration might take some time depending on the hardware and stored data. Please create a backup before updating, there is no automatic rollback process other then a database restore from before the migration.
- This release (re)imports CLN data so it might take some time to get all data fetched from your CLN node. (No impact to LND nodes)

### DIFF
--- a/torq/docker-compose.yml
+++ b/torq/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 7028
 
   web:
-    image: lncapital/torq:1.4.3@sha256:3aec9ef341588f34de72167a5a69021a4680e2ee62ff78c85edd3b1f75f8913c
+    image: lncapital/torq:1.5.1@sha256:f705b5a1b1da79668264dd9197b7c2b0020cfe6d9f0fdb946672d2e763ba176b
     user: "1000:1000"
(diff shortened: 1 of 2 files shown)


smolgrrr:
Tested on:

Working great for me- good to go @nmfretz. Would be cool if we could find a way to automatically import CLN without making it a dependency in a future update.

ohenrik:
> Agreed. @ohenrik we will be implementing optional dependencies at some point in the future, which will allow us to set up Torq with LND and/or CLN depending on what the user has installed and chooses.

That's great! Looking forward to it!

PR #961, opened 2024-02-06
Update Helipad to v0.1.11

ericpp:
Updates the Helipad app from `v0.1.10.5` to `v0.1.11`.


nmfretz:
Excellent, thanks for this update @ericpp! Tested on x86(amd64) and arm64. Going live.

PR #964, opened 2024-02-07
Update ln markers version

vafanassieff:
New version of LNM with a small fix, the image sha is pointing to the arm64 version.

### DIFF
--- a/lnmarkets/docker-compose.yml
+++ b/lnmarkets/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: $APP_LNMARKETS_PORT
   
   lnmarkets:
-    image: ghcr.io/ln-markets/umbrel:v2.0.0@sha256:47b836cd23f7aa9f31aa4ae99ba66f74ffc513388016e31a00f6cb0767144bae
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @vafanassieff! Sorry for the delay in reviewing. I tried pushing a couple small changes but I do not have access to push to your fork of umbrel-apps.

Can you please make the following changes (or else give me access whichever is easier and more comfortable for you):

In the docker-compose.yml:
- Change the sha to the multi-arch digest, which is `391798e0a18ec8135abf94037f879fa83e8394fa3a8881fec3aa934338e978d2`.

In the umbrel-app.yml:
- change the version to `"2.0.1"`. Doing this is what tells a user's Umbrel that there is an update of ln markets available.
- update the `releaseNotes`. Something simple like `This is a minor bugfix release.` is totally fine.

If you end up giving me access to your fork of the apps repo I can make any other changes needed when I test and I won't hold you up any further

vafanassieff:
@nmfretz Thanks for your feedback, I have updated the PR accordingly.

nmfretz:
Wonderful, thanks again @vafanassieff! Going live ⚡️

PR #970, opened 2024-02-14
Update immich to v1.94.1

smolgrrr:

* [x]   x86_64 -> UmbrelOS on proxmox Debian instance (install and update)


nmfretz:
Thanks @smolgrrr, going live.

PR #972, opened 2024-02-14
Update ghostfolio to 2.51.0

smolgrrr:
https://github.com/ghostfolio/ghostfolio/releases

* [x]   x86_64 -> UmbrelOS on proxmox Debian instance (install and update)


nmfretz:
Thanks @smolgrrr! Tested on arm64 and x86 as well. Going live 👻

PR #974, opened 2024-02-16
Update FreshRSS to v1.23.1

nmfretz:
Bumps FreshRSS to version 1.23.1 and also whitelists the api route in the proxy so that other RSS feed apps like NetNewsWire can connect.

Tested on arm64 and x86(amd64). Tested adding FreshRSS to NetNewsWire.


nmfretz:
wrong branch

PR #978, opened 2024-02-22
Update btcpay to 1.12.5

smolgrrr:
https://github.com/btcpayserver/btcpayserver/releases

* [ ]   x86_64 -> UmbrelOS on proxmox Debian instance
* [x]   Arm64 -> Pi 4 8GB (install and update)

@nmfretz do you mind checking btcpay on a linux instance for me please. should be fine, but im having issues with my linux instance at the moment so i couldn't check.


nmfretz:
Awesome @smolgrrr. I tested on x86_64 and it is working well. Going live.

PR #984, opened 2024-02-25
pi-hole: fixed the systemd service check in the pre-install hook

markshep:
The old code was checking two systemd units - one called "service" and the other the actual one it was supposed to be checking.  If you want to restrict the check to units of type "service" you need to append ".service" to the unit name, _not_ add it as a separate argument.


nmfretz:
Thanks for catching this @markshep! In fact, we don't even need to append `.service` since systemctl will default to this, but I like the specificity.

I will update Pi-hole today so that existing installs will get the update.

PR #993, opened 2024-03-05
Update immich to v1.97.0

mateosilguero:

Arm64 -> Pi 4 8GB (update)

Captura de pantalla 2024-03-05 a la(s) 09 43 24


smolgrrr:
Thanks for the submission @mateosilguero 

Looks like we'll need to include some migration logic as the postgres version update will break apps that are being updated, not freshly installed.

mateosilguero:
I saw that an sql script is needed on 1.95, but i didnt know how to do it from the container. Anyways my instance seems to works fine without this

cawis68:
Hi, is there any update on this? The entire Immich ecosystem is broken on Umbrel now, as due to the old version the mobile apps can't connect to any Immich instance running on any Umbrel due to the version mismatch. An update would be highly appreciated so we could continue using this amazing app with the amazing Umbrel platform. Thank you in advance!

ddavidel:
I need to know it updating it from v1.94.1 will break it. Can't use the mobile app because of different versions.

nmfretz:
Thanks @mateosilguero and @smolgrrr.  I have bumped to lated v1.98.2 and added a post-update script to make sure that postgres migration works. Going live 📸. Thanks for your patience everyone!

PR #994, opened 2024-03-05
Update Audiobookshelf to v2.8.0

nmfretz:


nmfretz:
Tested on x86 and arm64

PR #995, opened 2024-03-06
Update Autobrr to v1.39.1

nmfretz:


nmfretz:
Tested on arm64 and x86

PR #997, opened 2024-03-06
Update Chatpad-AI to latest commit f45cd5

nmfretz:


nmfretz:
Tested on arm64 and x86.

PR #998, opened 2024-03-06
Update Code Server to v4.22.0

nmfretz:


nmfretz:
Tested on arm64 and x86

PR #999, opened 2024-03-06
Update ESPHome to v2024.2.2

nmfretz:


nmfretz:
Tested on arm64 and x86

PR #1000, opened 2024-03-06
Update File Browser to v2.27.0

nmfretz:


nmfretz:
Tested on arm64 and x86

PR #1001, opened 2024-03-06
Update Firefly-III to v6.1.10

nmfretz:


nmfretz:
Tested on arm64 and x86

PR #1003, opened 2024-03-06
Update Firefox to v122.0

nmfretz:


nmfretz:
Tested on x86 and arm64

PR #1004, opened 2024-03-06
Update Gitea to v1.21.7

nmfretz:


nmfretz:
Tested on x86 and arm64

PR #1005, opened 2024-03-06
Update Invoice Ninja to v5.8.33

nmfretz:


nmfretz:
Tested on x86 and arm64

PR #1006, opened 2024-03-06
Update Jackett to v0.21.1946

nmfretz:


nmfretz:
To look into tomorrow:
image

nmfretz:
The security risk warning shown above does not apply to Umbrel users because we have packaged Jackett with our app proxy that provides in-built authentication.

If users want to remove the warning, they can set a password from Jackett's UI:

image

Tested on arm64 and x86.

PR #1007, opened 2024-03-06
Update Lidarr to 2.1.7

nmfretz:


nmfretz:
Tested on x86 and arm64 and confirmed that it still works with the media-app-configurator

PR #1008, opened 2024-03-06
Update LNbits to v0.12.2

nmfretz:


smolgrrr:

Working great. FYI i think we can get rid of the deterministic password now with their new login system.

PR #1014, opened 2024-03-10
Update frigate to 0.13.2

smolgrrr:
Release notes: https://github.com/blakeblackshear/frigate/releases/tag/v0.13.2

I had to add a /config directory to fix some migration issues with the update, so it would be good if someone with a camera can test that everything's working fine. @arguser if you don't mind thanks.
I was getting some weird errors, but idk if that's just because I don't have a camera set-up.
image


arguser:
@smolgrrr those errors happen because the default configuration points to a non-existent camera. We can't leave the configuration empty because Frigate would not start, so that configuration works as a placeholder until you do a proper setup. 

Should be safe to advance with this changes.

frigate/docker-compose.yml
```diff
@@ -10,12 +10,13 @@ services:
...
     privileged: true # this may not be necessary for all setups
     restart: unless-stopped
-    image: ghcr.io/blakeblackshear/frigate:0.12.1@sha256:bb7f7e76a13eccef0b12704e5851cc774a12af1f12df387d6a70a796a3e938c3
+    image: ghcr.io/blakeblackshear/frigate:0.12.1@sha256:2906991ccad85035b176941f9dedfd35088ff710c39d45ef1baa9a49f2b16734
     shm_size: "128mb" # update for your cameras based on calculation above
     devices:
      - /dev:/dev
     volumes:
       - /etc/localtime:/etc/localtime:ro
+      - ${APP_DATA_DIR}/data/config:/config
```
arguser, line 19: Adding this line wouldn't render unnecessary to have  ` - ${APP_DATA_DIR}/data/config.yml:/config/config.yml`?

nmfretz:
Thanks @smolgrrr and @arguser.

> I had to add a /config directory to fix some migration issues with the update, so it would be good if someone with a camera can test that everything's working fine.

Thanks for that. 0.13.0 introduced a [breaking change](https://arc.net/l/quote/qaqrvtsb) that now requires the entire `/config` directory to be bind mounted. To make the bind mounts a little less confusing, I've removed the direct config.yml bind mount and moved the default config.yml file into the new config directory.

Also, since this update requires a new config directory on the host, we need to create it with proper permissions for existing installs. New installs will have this directory created automatically, but existing installs will not due to the way we [whitelist files](https://github.com/getumbrel/umbrel/blob/49d37581c8233f808fc5a79c2cd38810c6cd1b0b/packages/umbreld/source/modules/apps/legacy-compat/app-script#L594-L595) for app updates. 
I added a pre-start script that creates the new config directory on the host for current installs with the proper permissions. Without doing this, Docker will create this directory with root permissions. The script also handles the migration of a users existing config.yml to this new directory. Nothing needs to happen for new installs.

frigate/docker-compose.yml
nmfretz: I've removed the direct config.yml bind mount and added a pre-start script to migrate users to this new set up.

PR #1017, opened 2024-03-12
updated databag version

balzack:
new databag version changelog:
- a smaller container size
- support of dark mode
- support of additional languages
- some guidance text on login screen for account creation


nmfretz:
Thanks @balzack! I have changed the `version` in the umbrel-app.yml, which makes it so that existing installs will receive a notification in the umbrel UI that a databag update is available.

I also added release notes.

balzack:
Thank you!

PR #1018, opened 2024-03-13
export core-lightning datadir

jackstar12:
Applications that want to connect to core-lightning via grpc need access to the datadirectory in order to authorize.

The `APP_CORE_LIGHTNING_DATA_DIR` variable was used internally in the app service, but that can be replaced with `CORE_LIGHTNING_PATH`.

### DIFF
--- a/core-lightning/docker-compose.yml
+++ b/core-lightning/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     restart: on-failure
     volumes:
       - ${APP_DATA_DIR}/data/app:${APP_CONFIG_DIR}
-      - ${APP_DATA_DIR}/data/lightningd:${APP_CORE_LIGHTNING_DATA_DIR}
+      - ${APP_CORE_LIGHTNING_DATA_DIR}:${CORE_LIGHTNING_PATH}
       - ${APP_CORE_LIGHTNING_REST_CERT_DIR}:${APP_REST_CERT_VOLUME_DIR}
     environment:
       SINGLE_SIGN_ON: "true"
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this @jackstar12. This change looks good to me.

One consideration here is that if a different app uses the new `APP_CORE_LIGHTNING_DATA_DIR` variable it will only work if a user has Core Lightning on a version where this variable has been introduced. So any app you add this to may need to have that disclaimer in case a user is on a previous version of Core Lightning. @ShahanaFarooqui would you / the team be willing to update the Core Lightning app on Umbrel? Once an update goes out, anyone who updates the app will have access to this new environment variable.


> Is the change from APP_DATA_DIR to EXPORTS_APP_DIR backwards compatible?

These are actually two different variables:
- `EXPORTS_APP_DIR` is the directory where the exports.sh file of an app resides. CLN for example is at `/home/umbrel/umbrel/app-data/core-lightning`
- `APP_DATA_DIR` is one directory level deeper at `EXPORTS_APP_DIR/data`, which equates to this for the CLN app: `/home/umbrel/umbrel/app-data/core-lightning/data`

The way you have implemented it in the exports.sh file is totally fine:
`APP_CORE_LIGHTNING_DATA_DIR="${EXPORTS_APP_DIR}/data/lightningd"`

You could have also done:
`APP_CORE_LIGHTNING_DATA_DIR="${APP_DATA_DIR}/lightningd"`

ShahanaFarooqui:
@nmfretz Hey Nate, I would prefer to wait till next month if possible. As cln-application only has 3 issues fixed and cln's latest release doesn't have arm image. Core lightning's next release is due in May and by that time, hopefully I will have some more enhancements for cln-application also.

nmfretz:
Yep, that's totally fine @ShahanaFarooqui! I just wanted to check in case we could get out an update to users so that @jackstar12 and @kilrau would have the new `APP_CORE_LIGHTNING_DATA_DIR` active for most users.

I will merge this PR for now. New installs of CLN will have the env var available, but existing installs will not get this env var until we issue an app update.

jackstar12:
> @jackstar12 @kilrau if it is critical for your timeline that we issue a CLN update earlier then we can just bump the version in CLN's umbrel-app.yml to something like `23.08-hotfix-2` with applicable release notes.

We won't need the feature until the next CLN release, so its fine.

PR #1019, opened 2024-03-13
upgrade to boltz-client in rtl

jackstar12:
This pr replaces `boltz-lnd` with its successor [boltz-client](https://github.com/BoltzExchange/boltz-client).
Its completely backwards compatible and the migration is well tested.

### DIFF
--- a/ride-the-lightning/docker-compose.yml
+++ b/ride-the-lightning/docker-compose.yml
@@ -35,7 +35,7 @@ services:
         BOLTZ_MACAROON_PATH: "/boltz/.boltz-lnd/macaroons"
 
   boltz:
-    image: boltz/boltz-client:2.0.0-compat@sha256:3a29501de1576569e797d0f199f6c96c523aa44dd3b3610ab7356506ca4ef397
+    image: boltz/boltz-client:2.0.2-compat@sha256:dc7643381306d29a035a5422b5c217eccf0e08678c4b863c0c62e2dcb6bf6ed6
     user: "1000:1000"
     restart: "on-failure"
     stop_grace_period: "1m"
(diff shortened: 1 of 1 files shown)


kilrau:
Could you give this a look? @nmfretz 

We tested the migration from boltz-lnd to boltz-client in a smaller beta [with a couple of external umbrel users](https://twitter.com/Boltzhq/status/1758533595310793102). All worked as expected. Also note, that merging this doesn't change anything feature-wise for the user but is a major requirement for us to not block our work @ RTL to enable new Boltz Client like e.g. [CLN support](https://github.com/Ride-The-Lightning/RTL/pull/1366) and more.

kilrau:
Friendly bump @nmfretz

nmfretz:
Hi @jackstar12 and @kilrau, sorry for the delay.

It looks like the `2.0.x-compat` docker images are not multi-architecture (right now they are only amd64), so right now an update will break the app for Raspberry Pi users.

Can you please push a multi-arch image? In the meantime I'll test this on armd64.

nmfretz:
Also, one other consideration:

In order for current RTL users to be notified of an app update we need to update the `version` in the umbrel-app.yml. The current version in the yml file is `0.14.1-beta`. What would you like to change this to? It can be something as simple as just doing `0.14.1-beta-boltzclient` or something.

jackstar12:
@nmfretz I've pushed and updated to the multi-version platform of the image. Your proposed `0.14.1-beta-boltzclient` sounds like a good, we can go with that.

kilrau:
With regards to release notes: while most users won't notice the change, boltz-client uses [taproot swaps ](https://blog.boltz.exchange/p/introducing-taproot-swaps-putting)by default now (boltz-lnd didn't support taproot swaps). Might be worth to mention.

nmfretz:
Perfect, thanks @jackstar12 and @kilrau. I have tested on both arm64 and x86. Going live now.

It looks like I do not have permission to push changes to `BoltzExchange/rtl-boltz-client` so I will bump the version and add release notes to the manifest in a separate commit

PR #1020, opened 2024-03-14
add morphos server

danvergara:
# App Submission

### App name
Morphos Server

### 256x256 SVG icon

[Morphos icon](https://svgshare.com/s/14x5)

### Gallery images

morphos
select_options_morphos
modal_morphos

### I have tested my app on:

### DIFF
--- a/morphos/docker-compose.yml
+++ b/morphos/docker-compose.yml
@@ -7,10 +7,6 @@ services:
       APP_PORT: 8080
 
   web:
-    image: ghcr.io/danvergara/morphos-server:0.2.0
+    image: ghcr.io/danvergara/morphos-server:0.3.0@sha256:b09a886b6ae1c2c3ab3911d0b3513b1a8c9953e80edeb942d33814700ebadfc7
     restart: on-failure
     stop_grace_period: 1m
-    ports:
-      - 8080:8080
-    volumes:
-      - ${APP_DATA_DIR}/tmp:/tmp
--- a/morphos/docker-compose.yml
+++ b/morphos/docker-compose.yml
@@ -1,0 +1,16 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: morphos_web_1
+      APP_PORT: 8080
+
+  web:
+    image: ghcr.io/danvergara/morphos-server:0.2.0
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - 8080:8080
+    volumes:
+      - ${APP_DATA_DIR}/tmp:/tmp
--- a/morphos/umbrel-app.yml
+++ b/morphos/umbrel-app.yml
@@ -1,19 +1,30 @@
 manifestVersion: 1
 id: morphos
 category: files
-name: Morphos Server
-version: "0.1.1"
-tagline: Self-hosted file converter server
+name: Morphos server
+version: "0.3.0"
+tagline: Self-hosted file converter
 description: >-
   Today we are forced to rely on third party services to convert files to other formats. This is a serious threat to our privacy, if we use such services to convert files with  highly sensitive personal data. It can be used against us, sooner or later.
 
+
   Morphos server aims to solve the problem mentioned above, by providing a self-hosted server to convert files privately. The project provides an user-friendly web UI.
 
-  For now, Morphos only supports images. Documents will be added soon.
-releaseNotes: >-
-  PDF support is finally here! Easily convert to images to PDF files.  Otherwise is also possible.
+
+  Features:
+
+  - Convert images to other image formats
+
+  - Convert PDF to images and vice versa
+
+  - Convert DOCX to PDF and vice versa
+
+  - Convert XLSX to CSV and vice versa
+releaseNotes: ""
 developer: Dan Vergara
 website: https://github.com/danvergara/morphos
+submitter: danvergara
+submission: https://github.com/getumbrel/umbrel-apps/pull/1020
 dependencies: []
 repo: https://github.com/danvergara/morphos
 support: https://github.com/danvergara/morphos/issues
--- a/morphos/umbrel-app.yml
+++ b/morphos/umbrel-app.yml
@@ -1,0 +1,27 @@
+manifestVersion: 1
+id: morphos
+category: files
+name: Morphos Server
+version: "0.1.1"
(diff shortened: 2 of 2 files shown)


morphos/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: morphos_web_1
+      APP_PORT: 8080
+
+  web:
+    image: ghcr.io/danvergara/morphos-server:0.2.0
```
nmfretz, line 10: We need to add the multi-arch image digest here to make sure all installs and app restarts result in the same image being downloaded.

Since you are using ghcr.io and have a multi-arch build for your image you can find the right digest listed under the tag. Here's the digest for your latest 0.3.0 for example:

image

So for 0.3.0 you would use:
```
ghcr.io/danvergara/morphos-server:0.3.0@sha256:b09a886b6ae1c2c3ab3911d0b3513b1a8c9953e80edeb942d33814700ebadfc7
```

morphos/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - 8080:8080
```
nmfretz, line 14: We don't need to expose a port on the host here. Instead, the app_proxy container + docker's internal dns handles this. You can delete the entire ports directive.

morphos/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+    volumes:
+      - ${APP_DATA_DIR}/tmp:/tmp
```
nmfretz, line 16: Do you need to persist the `tmp` directory on the host machine between app restarts? If not you can just remove this entire directive.

If you do need the files in `tmp` to persist across app restarts then leave this as-is, but we'll beed to commit this directory to the app repo so that it gets installed with correct permissions on a users Umbrel. If we don't do this then Docker will create the directories with root permissions. What you can do is create a `data` directory in the morphos app directory, and then inside the `data` directory create a `tmp` directory with a `.gitkeep` file in it to make sure it is committed even though it is an empty directory.

Take a look at Gitea as an example: https://github.com/getumbrel/umbrel-apps/tree/master/gitea

morphos/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
+manifestVersion: 1
+id: morphos
+category: files
+name: Morphos Server
+version: "0.1.1"
```
nmfretz, line 5: This can be updated to the current version of the app

morphos/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
...
+tagline: Self-hosted file converter server
```
nmfretz, line 6: Up to you here, but feel free to remove "server" from the tag so that the average user just understands the app as a self-hosted file converter

morphos/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
...
+description: >-
+  Today we are forced to rely on third party services to convert files to other formats. This is a serious threat to our privacy, if we use such services to convert files with  highly sensitive personal data. It can be used against us, sooner or later.
+
+  Morphos server aims to solve the problem mentioned above, by providing a self-hosted server to convert files privately. The project provides an user-friendly web UI.
+
+  For now, Morphos only supports images. Documents will be added soon.
```
nmfretz, line 12: Rendering the yaml nicely is a bit tricky. To get a blank line for whitespace between paragraphs you need to add **2** blank lines in the yaml.

Otherwise it looks like this:

image

morphos/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
...
+releaseNotes: >-
+  PDF support is finally here! Easily convert to images to PDF files.  Otherwise is also possible.
```
nmfretz, line 14: For the initial release on umbrelOS, we can leave the release notes blank.

```suggestion
releaseNotes: ""
```

morphos/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
...
+developer: Dan Vergara
```
nmfretz, line 15: The other two pieces of info to add here are:

```yaml
submitter: your-name-or-handle-here
submission: https://github.com/getumbrel/umbrel-apps/pull/1020

nmfretz (commented):
Thanks for this converter app @danvergara! It looks great. I have left a review below with some packaging items to address. Once these are done I'll do a final test before sending this to the app store!

nmfretz:
Also @danvergara do you have any ideas for text that you'd like to see on each of the gallery images?

morphos/docker-compose.yml
danvergara: I just replaced the old one with the one you suggest.
Thanks!

morphos/docker-compose.yml
danvergara: Gotcha!
It's gone!

morphos/umbrel-app.yml
danvergara: I just rephrased the half of the description added a new list of feature, and of course, extra lines for better rendering.

morphos/umbrel-app.yml
danvergara: Gone, too!

morphos/umbrel-app.yml
danvergara: Nice! I just added those missing fields.
Really appreciated!

danvergara:
Thanks @nmfretz for taking the time to review this PR!

I just addressed all your comments. Thanks!

Regarding the text for the gallery I'd like: Convert your files privately.

nmfretz:
Excellent, thanks @danvergara this is working great. We'll get the gallery images and icon ready and then launch on the app store! Give us just a few days to do this due to our current workload.

danvergara:
No rush!

Looking forward to the launch on the app store!!

nmfretz:
New gallery assets are ready @danvergara! Thanks again for the app submission, it is now live.

image

danvergara:
Sorry, I got busy over the past week.

I was able to download and use Morphos in my home lab running Umbrel. I was great!!!

Thank you so much @nmfretz!!

PR #1021, opened 2024-03-15
App Submission: Saifa (lnd Invoice manager)

patranakamo:
App Submission 
...
app name (Saifa)
...

256x256 SVG icon
https://svgur.com/i/1452.svg

description: 
    Saifa is designed to enhance the interoperability and efficiency of applications interacting with the Lightning Network Daemon (LND).
  Our project simplifies the integration process, allowing applications to seamlessly connect to LND through an intuitive webhook system. 
  With Saifa, users can effortlessly create new invoices using a straightforward REST API, cancel invoices, and receive real-time updates on invoice status.
  This system is crafted to streamline operations and facilitate a smoother transaction experience on the Lightning Network,
  making it an essential tool for developers looking to optimize their applications for better performance and user satisfaction.

...

Gallery images

I have tested my app on:
 [Umbrel OS on a Raspberry Pi 4](https://github.com/getumbrel/umbrel-os)

### DIFF
--- a/saifa/docker-compose.yml
+++ b/saifa/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       APP_HOST: saifa_frontend_1
       APP_PORT: 80
       PROXY_AUTH_ADD: "false"
-      
+
   server:
     image: patranakamo/saifa-be:v1.3.3@sha256:95321f638a5198c66ec45e72adfe41dd7abbdc3ab611a88bdaa12976ed2b0242
     restart: on-failure
@@ -23,7 +23,7 @@ services:
       APP_PASSWORD: $APP_PASSWORD
       BITCOIN_NETWORK: $APP_BITCOIN_NETWORK
       LN_SERVER_URL: "https://$LND_IP:$LND_REST_PORT"
-      REDIS_HOST: redis
+      REDIS_HOST: saifa_redis_1
 
   frontend:
     image: patranakamo/saifa-fe:v1.3.2@sha256:06ab9f0246e77a63a102111d00e2da7dfe1a94ed21032e56dbdc63dded9717bc
@@ -36,8 +36,6 @@ services:
   redis:
     restart: on-failure
     image: redis:latest@sha256:f44e91787b843612a3878d8d8fae227b9db63f471214314f5b2e73576ca633d0
-#    ports:
-#      - "6381:6379"
     volumes:
       - ${APP_DATA_DIR}/redis_data:/data/redis_data
     environment:
--- a/saifa/docker-compose.yml
+++ b/saifa/docker-compose.yml
@@ -5,7 +5,7 @@ services:
 #    environment:
 #      APP_HOST: saifa_web_1
 #      APP_PORT: 9989
-  server:
+  backend:
     image: patranakamo/saifa-be:v1.0.0
     restart: on-failure
     stop_grace_period: 1m
@@ -23,7 +23,7 @@ services:
       MACAROON_PATH: "/lnd/data/chain/bitcoin/$APP_BITCOIN_NETWORK/invoice.macaroon"
       REDIS_HOST: redis
 
-  web:
+  frontend:
     image: patranakamo/saifa-fe:v1.0.0
     ports:
       - "9989:80"
--- a/saifa/docker-compose.yml
+++ b/saifa/docker-compose.yml
@@ -32,8 +32,8 @@ services:
       - backend
 
   redis:
-    restart: on-failure
     image: redis:7.2.4@sha256:f14f42fc7e824b93c0e2fe3cdf42f68197ee0311c3d2e0235be37480b2e208e6
+    restart: on-failure
     volumes:
       - ${APP_DATA_DIR}/redis_data:/data/redis_data
     environment:
--- a/saifa/docker-compose.yml
+++ b/saifa/docker-compose.yml
@@ -1,12 +1,11 @@
 version: "3.7"
 
 services:
-  app_proxy:
-    environment:
-      APP_HOST: saifa_frontend_1
-      APP_PORT: 9989
-      PROXY_AUTH_ADD: "false"
-
+#  app_proxy:
+#    environment:
+#      APP_HOST: saifa_frontend_1
+#      APP_PORT: 80
+#      PROXY_AUTH_ADD: "false"
   server:
     image: patranakamo/saifa-be:v1.3.3@sha256:95321f638a5198c66ec45e72adfe41dd7abbdc3ab611a88bdaa12976ed2b0242
     restart: on-failure
@@ -28,6 +27,8 @@ services:
   frontend:
     image: patranakamo/saifa-fe:v1.3.2@sha256:06ab9f0246e77a63a102111d00e2da7dfe1a94ed21032e56dbdc63dded9717bc
     restart: on-failure
+    ports:
+      - "9989:80"
     depends_on:
       - server
 
--- a/saifa/docker-compose.yml
+++ b/saifa/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   backend:
-    image: patranakamo/saifa-be:v1.0.3
+    image: patranakamo/saifa-be:v1.1.0
     restart: on-failure
     stop_grace_period: 1m
     ports:
--- a/saifa/docker-compose.yml
+++ b/saifa/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   backend:
-    image: patranakamo/saifa-be:v1.2.2
+    image: patranakamo/saifa-be:v1.2.3
     restart: on-failure
     stop_grace_period: 1m
     ports:
@@ -19,7 +19,7 @@ services:
       REDIS_HOST: redis
 
   frontend:
-    image: patranakamo/saifa-fe:v1.1.0
+    image: patranakamo/saifa-fe:v1.2.0
     ports:
       - "9989:80"
     depends_on:
--- a/saifa/docker-compose.yml
+++ b/saifa/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   backend:
-    image: patranakamo/saifa-be:v1.0.0
+    image: patranakamo/saifa-be:v1.0.1
     restart: on-failure
     stop_grace_period: 1m
     ports:
--- a/saifa/docker-compose.yml
+++ b/saifa/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   backend:
-    image: patranakamo/saifa-be:v1.2.0
+    image: patranakamo/saifa-be:v1.2.1
     restart: on-failure
     stop_grace_period: 1m
     ports:
--- a/saifa/docker-compose.yml
+++ b/saifa/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   backend:
-    image: patranakamo/saifa-be:v1.2.4
+    image: patranakamo/saifa-be:v1.3.0
     restart: on-failure
     stop_grace_period: 1m
     ports:
@@ -19,7 +19,7 @@ services:
       REDIS_HOST: redis
 
   frontend:
-    image: patranakamo/saifa-fe:v1.2.1
+    image: patranakamo/saifa-fe:v1.3.0
     ports:
       - "9989:80"
     depends_on:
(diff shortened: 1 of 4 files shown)


patranakamo:
Hi nmfretz great to hear that.
Thank you for you comment.

saifa/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
+version: "3.7"
+
+services:
+  backend:
+    image: patranakamo/saifa-be:v1.3.3
```
nmfretz, line 5: For each image in this compose file, we need to add the multi-arch digest to make sure that we are locking in a specific image that has been tested for this release. With just the tag, a new image can be pushed to docker hub with the same tag and then a new image would be downloaded by the user just by restarting the app (and of course by fresh installs).

Your image is already multi-arch which is fantastic:
image

So all you need to do is get the digest by running:
```
docker pull patranakamo/saifa-be:v1.3.3
```

and grabbing the digest from the output. In this case I get:
```
Digest: sha256:a04bbaba219b35d7bb59426665ae813ad2d219cb1b2e0eceed1c79170ede6228
```

So in this example you would do:
```
image: patranakamo/saifa-be:v1.3.3@sha256:a04bbaba219b35d7bb59426665ae813ad2d219cb1b2e0eceed1c79170ede6228
```

saifa/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    restart: on-failure
```
nmfretz, line 6: Perfect. You'll want to add this same `restart: on failure` to the frontend and redis services in this compose file as well.

saifa/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    stop_grace_period: 1m
+    ports:
+      - "9988:80"
```
nmfretz, line 9: Do you need this port bound to the host? This would be necessary if external applications are trying to send requests directly to this backend container.

Or is it just your frontend container that is connecting to the backend? If so, then there is no need to bind the port to the host, since your frontend container can resolve the ip of the backend container using Docker's inbuilt dns and communicate on container port 80 directly.

saifa/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    volumes:
+      - ${APP_DATA_DIR}/data/app:/data
```
nmfretz, line 11: Can you please add this directory to this repo so that it is created for users when installing the app. If we don't do this then Docker will create the directory when the container is spun up, which has the potential to cause permissions issues.

See the Lightning Node app as an example:
https://github.com/getumbrel/umbrel-apps/tree/master/lightning

You'll want to create a `data` directory and then a `app` directory inside of that with a `.gitkeep` file to make sure the empty directory  is committed.

saifa/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    depends_on:
+      - redis
+    environment:
+      APP_PASSWORD: $APP_PASSWORD
+      BITCOIN_NETWORK: $APP_BITCOIN_NETWORK
+      LN_SERVER_URL: "https://$LND_IP:$LND_REST_PORT"
+      REDIS_HOST: redis
+
+  frontend:
+    image: patranakamo/saifa-fe:v1.3.2
```
nmfretz, line 22: add multi-arch digest here

saifa/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    ports:
+      - "9989:80"
```
nmfretz, line 24: The entire `ports` section can be removed. Instead you can add the app_proxy to this compose file as explained here: <add-link> and point it to port 80 of the frontend container.

saifa/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    depends_on:
+      - backend
+
+  redis:
+    image: redis:latest
```
nmfretz, line 29: add specific version tag and multi-arch digest here

saifa/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+#    ports:
+#      - "6381:6379"
+    volumes:
+      - ${APP_DATA_DIR}/redis_data:/data
```
nmfretz, line 33: You'll probably want to change the bind mount on the host to be:
`${APP_DATA_DIR}/data/redis_data`
so that it lives in the same parent directory as your backend container's app data.

You'll also need to commit the `redis_data` directory to this repo as explained here: <link>

saifa/umbrel-app.yml
```diff
@@ -0,0 +1,36 @@
+manifestVersion: 1
+id: saifa
+category: finance,development tool
+name: Saifa - invoice lightning network manager
+version: "1.0.0"
```
nmfretz, line 5: You can update this to your current version

saifa/umbrel-app.yml
```diff
@@ -0,0 +1,36 @@
+manifestVersion: 1
+id: saifa
+category: finance,development tool
```
nmfretz, line 3: We currently only support one category (may change in the future). I'd suggest `bitcoin` for now so that it will show up with related apps.

saifa/umbrel-app.yml
```diff
@@ -0,0 +1,36 @@
...
+  making it an essential tool for developers looking to optimize their applications for better performance and user satisfaction.
+releaseNotes: >-
+  - Base Project Launch: The foundational framework of our project has been successfully deployed, establishing the core functionalities and setting the stage for future enhancements.
+  - Webhook System Integration: We've introduced a robust webhook system designed to facilitate seamless updates on invoice statuses. This feature ensures that applications remain synchronized, providing timely notifications to users and other connected applications.
+  - User Interface for Invoice Management: A user-friendly interface has been implemented, allowing users to effortlessly monitor the status of their invoices. This intuitive UI simplifies the management process, ensuring a smooth and efficient user experience.
+  - Enhanced Invoice Operations: Users now have the ability to add and cancel invoices directly from the simple UI. This enhancement streamlines the invoice management process, making it more accessible and convenient for our users.
```
nmfretz, line 17: For the initial release we can just leave these blank, since no one will get an app update notification.

You can change to:

`releaseNotes: ""`

saifa/umbrel-app.yml
```diff
@@ -0,0 +1,36 @@
...
+developer: patra nakamo
+website: https://github.com/patranakamo
+email: patranakamo@gmail.com
+dependencies:
+  - bitcoin
+  - lnd
```
nmfretz, line 23: the id for lnd is actually `lightning` for this.

saifa/umbrel-app.yml
```diff
@@ -0,0 +1,36 @@
...
+category: finance,development tool
+name: Saifa - invoice lightning network manager
+version: "1.0.0"
+tagline: Simple, webhook for connect with invoice and invoice lightning network manager
+description: >-
+    Saifa is designed to enhance the interoperability and efficiency of applications interacting with the Lightning Network Daemon (LND).
+  Our project simplifies the integration process, allowing applications to seamlessly connect to LND through an intuitive webhook system. 
+  With Saifa, users can effortlessly create new invoices using a straightforward REST API, cancel invoices, and receive real-time updates on invoice status.
+  This system is crafted to streamline operations and facilitate a smoother transaction experience on the Lightning Network,
+  making it an essential tool for developers looking to optimize their applications for better performance and user satisfaction.
```
nmfretz, line 12: Rendering yaml in the UI makes it a bit finicky to create whitespace. here's what you can do to make sure that the app description looks good in the app store:

If you want a paragraph to show up on a separate line **without** a blank line above it, then add a single blank line to the yaml

e.g.
```yaml
description: >-
  I am the first block.
  
  I will show up on my own line but directly under the first block.
  ```
  
  The above will render like this:
```
I am the first block.
I will show up on my own line but directly under the first block.
```

If you want a paragraph to show up on a separate line **with** a blank line above it, then add **two** blank lines to the yaml

e.g.
```yaml
description: >-
  I am the first block.
  

  I will show up on my own line with a blank line above.
  ```
  
  The above will render like this:
```
I am the first block.

I will show up on my own line with a blank line above.
```

**Review by nmfretz — COMMENTED** [review]
Hi @patranakamo, thanks for bearing with me. And thanks again for this submission!

I've reviewed and left some packaging suggestions below. Once these are addressed I'll test out the app.

**Comment by patranakamo** [comment]
HI @nmfretz 
Thank you so much for your detailed review. I've made the edits based on your comments. However, I’m encountering an issue with adding app_proxy. 
Whenever I try to add it, I receive the following error:

ERROR: The Compose file '/home/umbrel/umbrel/scripts/support/docker-compose.app_proxy.yml' is invalid because:
services.app_proxy.ports contains an invalid type, it should be a number, or an object

The system runs normally when I remove and reinstall. (I push it in docker-composer.yml as comment in case you want to see it.)
Could you suggest any solutions for this issue? Your help would be greatly appreciated.

Thanks again.

**Comment by nmfretz** [comment]
> Could you suggest any solutions for this issue? Your help would be greatly appreciated.

No worries at all, I have pushed some commits that resolve a couple issues. The app_proxy is using Docker's DNS to resolve the IP of the frontend container and proxy to port 80 running in the frontend container.

The app now installs fine, but the `frontend` container has these errors and won't start up:

```
$ sudo docker logs -f saifa_frontend_1
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: /etc/nginx/conf.d/default.conf differs from the packaged version
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2024/05/13 12:13:19 [emerg] 1#1: host not found in upstream "backend" in /etc/nginx/conf.d/default.conf:12
```

patranakamo:
I pushed an update after you helped me fix the frontend port issue. I tested it on my Umbrel-on-Raspberry, and it can be installed normally, with the frontend running smoothly. It should work on your Umbrel too.

I just wanted to let you know that I appreciate your help in pushing some fixes. I didn't expect this level of assistance from anyone.

nmfretz:
> I just wanted to let you know that I appreciate your help in pushing some fixes. I didn't expect this level of assistance from anyone.

You're most welcome, glad to help!

I have tested and made a couple changes to make sure it runs for both arm64 and x86 users and to show users their default app password. We are pretty much ready to go.

Can you please let me know the following:

1) Do you want your app version to still be listed as 1.0.0 (https://github.com/getumbrel/umbrel-apps/pull/1021#discussion_r1596125239), or do you want to change it to 1.3.3 like in your Docker images?

2) Just to confirm: are you binding the backend port (9988:80) on the host so that you can make requests to the backend from outside the app (e.g., via REST API)? If so, then we leave this as-is. If not and only the frontend should talk to the backend then we can remove this port binding. https://github.com/getumbrel/umbrel-apps/pull/1021#discussion_r1596119735

3) Are you okay with me changing the app name from `Saifa - Invoice lightning network manager` to just `Saifa` so that it shows up nicely on the homescreen when someone installs it? I moved `Invoice lightning network manager` to the tagline.

4) For your 3 gallery images, what would you like the text to say on each image?

patranakamo:
HI again @nmfretz

Let me answer your questions:

1 ) Let's keep the app version as 1.0.0, even if it's not the same as the Docker images. The Docker images include the front-end as well, so sometimes they need different versions.

2 ) Yes, I still need to keep that port open. It has a feature for an external API that can call this backend directly (with a token) to create or get invoices without opening the front-end UI.

3 ) Sure, that sounds good.

4 ) For the gallery images, please use the following text:
First picture: "Add New Invoice Page."
Second picture: "Invoice List - Users can see all invoices, both active and inactive, on this page."
Third picture: "Webhook Routing Config Page - Shows how users can configure to receive real-time invoice data sent to other apps."
Last picture: "Token Config Page - Shows how users can call the Saifa API without logging in through the UI."

nmfretz:
Thanks @patranakamo. We'll get the gallery assets ready and then launch on the app store!

nmfretz:
@patranakamo the gallery assets are ready: https://github.com/getumbrel/umbrel-apps-gallery/tree/master/saifa

Going live to the app store now. Congratulations! And thanks again for all your hard work bringing Saifa to Umbrel.

image

patranakamo:
Wow, the app looks great on the App Store! 
Thank you for your help and suggestions. 
I will try to continue contribute to this community.

PR #1027, opened 2024-03-28
Update Bitcoin Knots to v26.1

Retropex:
Update Bitcoin Knots to [v26.1](https://github.com/bitcoinknots/bitcoin/releases/tag/v26.1.knots20240325)

**note:** I could not test this update because it is no longer possible to install umbrel on linux and I no longer have my umbrel VMs available.

### DIFF
--- a/bitcoin-knots/docker-compose.yml
+++ b/bitcoin-knots/docker-compose.yml
@@ -41,7 +41,7 @@ services:
         ipv4_address: $APP_BITCOIN_KNOTS_IP
 
   bitcoind:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @Retropex! I added release notes, which you can now update in future releases.

Tested on arm64 and x86. 

> note: I could not test this update because it is no longer possible to install umbrel on linux and I no longer have my umbrel VMs available.

We'll be updating instructions for testing apps in 1.0 very soon. It's very simple to spin up a local VM with Multipass for testing on 1.0 now. With Multipass installed you can run the following from the root directory of https://github.com/getumbrel/umbrel:

```
npm run vm:provision
```

A locally running umbrelOS instance will then be available at umbrel-dev.local.

PR #1028, opened 2024-03-29
[UPDATE] Uptime Kuma 1.23.11

shubhankartrivedi:


uptime-kuma/docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3001
 
   server:
-    image: louislam/uptime-kuma:1.23.7@sha256:658b77607d872913cc1217e43e2288d35dd712c3d943c7c21fab0dfbd471b8ad
+    image: louislam/uptime-kuma:1.23.11@sha256:658b77607d872913cc1217e43e2288d35dd712c3d943c7c21fab0dfbd471b8ad
```
nmfretz, line 10: One thing to note when updating the image:
The image digest takes precedence over the tag. So in this case the image with digest `sha256:658b77607d872913cc1217e43e2288d35dd712c3d943c7c21fab0dfbd471b8ad` will be downloaded no matter what the tag (1.23.11) has been changed to. Even though you updated the tag, Docker would still pull the image associated with the old digest which was 1.23.7

nmfretz (commented):
Thanks for this update @shubhankartrivedi and sorry for the delay!

I have bumped this to the latest version 1.23.12 that came out a few days ago and left a comment below that you can follow the next time you go to update an app.

Thanks again for staying on top of this!

PR #1030, opened 2024-04-01
Update Invidious to 2024.3.31-08390ac

johnpc:
Fixes a bug that prevented videos from loading (see https://github.com/iv-org/invidious/issues/4551)

Essentially, YouTube made a service update that broke Invidious, and this build contains the workaround.

PR #1031, opened 2024-04-01
Update invidious image version

RenanDelfanti:
Workaround used to fetch streaming URLs (see https://github.com/iv-org/invidious/issues/4498)


johnpc:
looks like a duplicate of https://github.com/getumbrel/umbrel-apps/pull/1030

PR #1032, opened 2024-04-01
Update docker-compose.yml

racompton:
Typo in APP_HOST name causes librespeed-app_proxy-1 docker container to not find the docker host.   Changed underscores to dashes:
librespeed-server-1

### DIFF
--- a/librespeed/docker-compose.yml
+++ b/librespeed/docker-compose.yml
@@ -3,7 +3,7 @@ version: "3.7"
 services:
   app_proxy:
     environment:
-      APP_HOST: librespeed_server_1
+      APP_HOST: librespeed-server-1
       APP_PORT: 80
 
   server:


racompton:
librespeed-app_proxy-1 was failing to connect to librespeed-server-1 because it was looking for librespeed_server_1

highghlow:
The issue with this is that umbrel still uses docker compose v1, which uses underscores. The onlybway to update to docker compose v2 would be to update all the apps at the same time (including umbrel itself), or come up with a new way of handling app_proxy.

nmfretz:
Hey @racompton-

@highghlow is correct that this is a docker compose version issue. It will be related to your custom set-up using a Docker Compose version that was not the default that umbrelOS installs. umbrelOS 1.x includes logic that prevents this issue from happening anymore, but here is some additional background:

compose v1 names containers with underscores:
`<project-name>_<service-name>_1`

while compose v2 names containers with dashes:
`<project-name>-<service-name>-1`

What's happening for you is that you likely already had Docker Compose v2 installed on your machine when you did a custom install of umbrelOS or else you updated later, does that sound right? If so, then this breaks Umbrel in places where containers are referenced via name or where Docker DNS references containers. If you are running a custom linux install with umbrelOS 0.5.4 then the solution here is to downgrade to compose v1. Alternatively, umbrelOS 1.x comes out for custom linux users this month with Compose v2 and includes the following patch: https://github.com/getumbrel/umbrel/blob/ace3d9d2e528c798fab0b75b8000b16f30d2aa3b/packages/umbreld/source/modules/apps/app.ts#L93-L109

nmfretz:
Closing this PR @racompton, but feel free to tag me here still if you need additional info.

PR #1034, opened 2024-04-05
Update Ordinals to  v0.17.1

nmfretz:
- update ord from 0.15.0 to 0.17.1
- built 0.17.1 image using the [official Dockerfile in the ord repo](https://github.com/ordinals/ord/blob/cb4de8dedf8485d265e6cd0edab7c2c894685190/Dockerfile). Had previously been using a custom Dockerfile because the ord repo did not include one in the past. The official Dockerfile does not include an entrypoint with the ord command, so I have added it to the command in the compose file.
- moved appropriate cli flags to environment variables. Users will now see accurate config values when running `ord settings` and can now run ord commands without passing data-dir/config/rpc options. For example `ord wallet create`.
- added --index-runes
- modified the pre-start hook to nuke index.db for users updating so that reindexing for runes occurs automatically.

### DIFF
--- a/ordinals/hooks/pre-start
+++ b/ordinals/hooks/pre-start
@@ -24,6 +24,11 @@ if [[ -f "${APP_DATA_DIR}/index.redb" ]]; then
   set_correct_permissions "${APP_DATA_DIR}"
 fi
 
+# re-index the ordinals data for runes
+if [[ ! -f "${APP_DATA_DIR}/INDEXED_0-17-1" ]]; then
+	rm -f "${ORDINALS_DATA_DIR}/index.redb"
+	touch "${APP_DATA_DIR}/INDEXED_0-17-1"
+fi
 
 
 
--- a/ordinals/hooks/pre-start
+++ b/ordinals/hooks/pre-start
@@ -17,6 +17,7 @@ set_correct_permissions() {
 	fi
 }
 
+# handle generating the correct ord data dir for pre-0.15.0 versions
 if [[ -f "${APP_DATA_DIR}/index.redb" ]]; then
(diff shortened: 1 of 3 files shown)


nmfretz:
In progress. Untested.

ordinals/hooks/pre-start
```diff
@@ -24,6 +24,11 @@ if [[ -f "${APP_DATA_DIR}/index.redb" ]]; then
   set_correct_permissions "${APP_DATA_DIR}"
 fi
 
+# re-index the ordinals data for runes
+if [[ ! -f "${APP_DATA_DIR}/INDEXED_0-17-1" ]]; then
```
mrv777, line 28: Nice idea here

nmfretz: Thanks @mrv777,  it's a little hacky but will work for now. I've renamed the file to `AT_LEAST_0-17-1` since future installs will have this written to file.

mrv777: Another idea I had but haven't had time today to think it through for pros/cons would be a VERSION file that just has the version number as text in it.  The script can then just check the contents of the file to see what it should do, and update it at the end.

I know quite a few people excited for this to be updated though as they prefer the official umbrel apps and your other fix to simplify the ord commands will be greatly appreciated.

Great work and thank you!

nmfretz:
Tested on arm64 and x86.

ordinals/hooks/pre-start
nmfretz: That's a good idea if this starts getting out of hand because it will keep the data directory cleaner and less confusing for someone who views it and sees a bunch of strangely named files

PR #1037, opened 2024-04-09
[FIX] [BUG] Update lnbits docker-compose.yml (#1033)

Ghooosstt:
Add `PROXY_TRUST_UPSTREAM: "true"` in **lnbits/docker-compose.yml** to make lnbits work in clearnet using nginx reverse proxy with SSL certificate.

### DIFF
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -6,6 +6,7 @@ services:
       APP_HOST: lnbits_web_1
       APP_PORT: 3007
       PROXY_AUTH_ADD: "false"
+      PROXY_TRUST_UPSTREAM: "true"
   
   web:
     image: lnbits/lnbits:0.12.2@sha256:ce11b1127ab9a6314ed5a57792dcd938ee189c2f549888f6cdda21a181a12fc9


EpiicDream:
I have tested the update on umbrelOS 1.0.4 on my Raspberry Pi 5.

nmfretz:
Hey @Ghooosstt nice work finding the `PROXY_TRUST_UPSTREAM` environment variable for the app_proxy and troubleshooting 👌.

We've looked into this exact issue before https://github.com/getumbrel/umbrel-apps/pull/102#issuecomment-1207184997, but have decided that we cannot have lnbits blindly trusting the upstream as a default setting for the app, as it makes it trivial to spoof certain potentially important HTTP headers. `PROXY_TRUST_UPSTREAM="true"` is only safe when running Umbrel's app proxy behind a trusted proxy that sanitizes header values, not when exposing the Umbrel app proxy directly which is the common use case.

### Solution

While we can't have this set as default for everyone, there is a way that you can set this variable and have it persist across app updates:

You can create a `.env.app_proxy` config file in the lnbits directory with `PROXY_TRUST_UPSTREAM=true`. This file will persist across app updates so you only need to set it once.

For lnbits the file should live at `~/umbrel/app-data/lnbits/.env.app_proxy`. As of umbrelOS 1.1 you can access the terminal from the Settings page > Advanced settings > Terminal > umbrelOS and run your commands there to add the .env.app_proxy file. After saving the file you'll need to restart lnbits by right-clicking on the app and restarting

I'm going to close this PR, but feel free to tag me to continue the conversation.

nmfretz:
@Ghooosstt, yes that is exactly right!

PR #1041, opened 2024-04-11
fix: ensure ord data dir exists before touch

mrv777:
I could be missing something, but on a new install when the pre-start file runs the data directory doesn't seem created yet so the touch command fails and doesn't create that version check file

### DIFF
--- a/ordinals/hooks/pre-start
+++ b/ordinals/hooks/pre-start
@@ -5,6 +5,9 @@ APP_DATA_DIR="$(readlink -f $(dirname "${BASH_SOURCE[0]}")/../data)"
 ORDINALS_DATA_DIR="${APP_DATA_DIR}/ord"
 DESIRED_OWNER="1000:1000"
 
+# Ensure the data directory exists
+mkdir -p "${ORDINALS_DATA_DIR}"
+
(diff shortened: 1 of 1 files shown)


mayankchhabra:
Thanks for the PR, @mrv777! The `data/ord` directory is [committed in the repo](https://github.com/getumbrel/umbrel-apps/tree/master/ordinals/data/ord), so we don't need to create it in the `pre-start` hook. I wasn't able to reproduce the touch command failing to create the version check file on a fresh install of Ordinals. Can you double-check, or help me with the steps to reproduce it?

mrv777:
Hmm, I think you could be right as maybe this only effects a custom app store.  The new command:
`umbreld client apps.install.mutate --appId ordinals`
doesn't seem to give the output anymore of the install so can't confirm, but I'm guessing that is the issue on the custom app store and that its fine for the main repo maybe.

mayankchhabra:
Makes sense, yeah. If you commit `data/ord` directory in https://github.com/mrv777/umbrel-community-app-store/tree/master/mrv777-ordinals, it should be working fine without `mkdir -p`. Closing the issue for now! :)

PR #1042, opened 2024-04-12
Add FlareSolverr to Prowlarr

Julienpeps:
For some indexers, it is required to proxy Prowlarr's traffic through FlareSolverr, as shown on the screenshot below.

image

My solution was to add FlareSolverr to Prowlarr's docker-compose, and configure the proxy in Prowlarr as shown below.

image

From my understanding, it would also be possible to use "getumbrel/media-app-configurator" to configure Prowlarr, however, I have no clue how to do that. Any help would be appreciated.

### DIFF
--- a/prowlarr/docker-compose.yml
+++ b/prowlarr/docker-compose.yml
@@ -36,3 +36,7 @@ services:
       SONARR_CONFIG_XML: "${APP_PROWLARR_SONARR_CONFIG_XML}"
       READARR_URL: "http://readarr_server_1:8787"
       READARR_CONFIG_XML: "${APP_PROWLARR_READARR_CONFIG_XML}"
+
+  proxy:
+    image: flaresolverr/flaresolverr:v3.3.17@sha256:5f5661db1e69a6f80ac24d47d9fa5580f6f741ee5ec967818396ae0dacecd7ea
+    restart: on-failure


hutchYy:
up!

nmfretz:
Great idea adding FlareSolverr to the compose file @Julienpeps.  

> From my understanding, it would also be possible to use "getumbrel/media-app-configurator" to configure Prowlarr, however, I have no clue how to do that. Any help would be appreciated.

Yep that's correct, we should be able to autoconfigure FlareSolverr as an indexer proxy when Prowlarr is installed. I need to update the media-app-configurator to include some other configuration items for other apps, so I can do this at the same time.

Do you know if there is any reason why a user might not want Flaresolverr set up by default?

Julienpeps:
Users can choose whether they want to use FlareSolverr for certain indexer. There is therefore no impact on users who don't use it, except an extra container of course.

After setting up the proxy as shown in my first comment, enabling it for a specific indexer only requires tagging it as shown below:
image

nmfretz:
> Users can choose whether they want to use FlareSolverr for certain indexer. There is therefore no impact on users who don't use it, except an extra container of course.

Excellent, thanks for running me through that. I'm going to focus on getting some app submissions out the door and then will come back to this and update the media-app-configurator.

nmfretz:
working on the configurator now.

nmfretz:
Thanks again for this @Julienpeps. I have merged this into a working branch that will be merged into master with additional servarr app updates.

PR #1045, opened 2024-04-12
Update Ordinals to 0.18.1

mayankchhabra:
- update ord from 0.17.1 to 0.18.1
- modified the pre-start hook to nuke index.db for all bitcoin networks, so that reindexing for runes occurs automatically

### DIFF
--- a/ordinals/hooks/pre-start
+++ b/ordinals/hooks/pre-start
@@ -28,9 +28,12 @@ fi
 # re-index the ordinals data for runes
 # for installs post 0.18.1 this file will be harmlessly created 
 if [[ ! -f "${APP_DATA_DIR}/AT_LEAST_0-18-1" ]]; then
-    if [[ -f "${ORDINALS_DATA_DIR}/index.redb" ]]; then
-        rm -f "${ORDINALS_DATA_DIR}/index.redb"
-    fi
+    # delete index db for all bitcoin networks
+    for file in "${ORDINALS_DATA_DIR}/index.redb" "${ORDINALS_DATA_DIR}/testnet3/index.redb" "${ORDINALS_DATA_DIR}/regtest/index.redb" "${ORDINALS_DATA_DIR}/signet/index.redb"; do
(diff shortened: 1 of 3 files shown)


ordinals/hooks/pre-start
```diff
@@ -28,9 +28,12 @@ fi
 # re-index the ordinals data for runes
 # for installs post 0.18.1 this file will be harmlessly created 
 if [[ ! -f "${APP_DATA_DIR}/AT_LEAST_0-18-1" ]]; then
-    if [[ -f "${ORDINALS_DATA_DIR}/index.redb" ]]; then
-        rm -f "${ORDINALS_DATA_DIR}/index.redb"
-    fi
+    # delete index db for all bitcoin networks
+    for file in "${ORDINALS_DATA_DIR}/index.redb" "${ORDINALS_DATA_DIR}/testnet3/index.redb" "${ORDINALS_DATA_DIR}/regtest/index.redb" "${ORDINALS_DATA_DIR}/signet/index.redb"; do
```
mrv777, line 32: Just an FYI, I don't think this app will currently work for other networks since it doesn't update the bitcoin data DIR and cookie file location when its a different chain:
https://github.com/mrv777/umbrel-community-app-store/blob/master/mrv777-ordinals/hooks/pre-start#L12

I could be wrong since I haven't had a chance to confirm.  I also don't think many people that run umbrel use it outside mainnet :) but just wanted to mention it while I saw this commit

mayankchhabra: Oh yeah, good point. Fair enough, I guess we can keep the deletion logic in place for now and make changes later for the app to be able to work with the correct Bitcoin network.

mrv777: Just documenting that I was wrong here :) and the directories get updated correctly by ord then:
https://github.com/ordinals/ord/blob/e8084bab3a77cef5225cc5edcd9f6db307560d62/src/chain.rs#L80

Issue #991, opened 2024-03-04
[App Request] — Technitium DNS Server


wont-stream:
Would be nice to have something a little more professional than Pi-hole

highghlow:
I have started working on this

highghlow:
Here you go

nmfretz:
Technitium DNS is now available in the app store thanks to the work of @highghlow! Closed by https://github.com/getumbrel/umbrel-apps/commit/d7b4e80fc905c3f9ac2f5914c5b877e26e550950

wont-stream:
Thank you ❤️

PR #1046, opened 2024-04-12
Add technitium dns

unhighghlow:
App request: #991 

# App Submission

### App name
Technitium DNS Server

### 256x256 SVG icon

If I don't forget, I will make this into an svg

https://styles.redditmedia.com/t5_3z4eb4/styles/communityIcon_snskdmw5lzh61.png

### Gallery images

landing-original

screenshot-edit-zone

screenshot-donut-charts

ScreenShot1

### I have tested my app on:
- [x] umbrelOS on Linux VM (NOTE: a full test cannot be done, as the admin password is provided by the $APP_PASSWORD, which cannot be viewed: getumbrel/umbrel#1790)

### DIFF
--- a/technitium-dns/umbrel-app.yml
+++ b/technitium-dns/umbrel-app.yml
@@ -8,7 +8,7 @@ description: >-
   Technitium DNS Server is an open source authoritative, as well as recursive, DNS server that can be used for self hosting a DNS server for privacy & security.
   It works out-of-the-box with no or minimal configuration.
 releaseNotes: ""
-developer: Shreyas Zare
+developer: Technitium
 website: https://technitium.com/dns/
 dependencies: []
 repo: https://github.com/TechnitiumSoftware/DnsServer


technitium-dns/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+      # - "53443:53443/tcp" #DNS web console (HTTPS)
+      - "53:53/udp" #DNS service
+      - "53:53/tcp" #DNS service
+      # - "853:853/udp" #DNS-over-QUIC service
+      # - "853:853/tcp" #DNS-over-TLS service
+      # - "443:443/udp" #DNS-over-HTTPS service (HTTP/3)
+      # - "443:443/tcp" #DNS-over-HTTPS service (HTTP/1.1, HTTP/2)
+      # - "80:80/tcp" #DNS-over-HTTP service (use with reverse proxy or certbot certificate renewal)
+      # - "8053:8053/tcp" #DNS-over-HTTP service (use with reverse proxy)
+      # - "67:67/udp" #DHCP service
```
nmfretz, line 22: What do you think about adding other ports? Obviously we don't have inbuilt https on umbrelOS quite yet, but users may want to use technitium dns as a DHCP server for example?

Actually, thinking about it I wonder if we need to run technitium dns in host network mode for now to get the full capabilities of the app (like we do with pi-hole in order to get the DHCP server functionality working).

What do you think?

technitium-dns/umbrel-app.yml
```diff
@@ -0,0 +1,25 @@
...
+category: networking
+name: Technitium DNS server
+version: "12.1"
+tagline: Block ads & malware at DNS level for your entire network!
+description: >-
+  Technitium DNS Server is an open source authoritative as well as recursive DNS server that can be used for self hosting a DNS server for privacy & security. It works out-of-the-box with no or minimal configuration and provides a user friendly web console accessible using any modern web browser.
+
+developer: Shreyas Zare
+website: https://technitium.com/dns/
+dependencies:
```
nmfretz, line 12: We just need to close this with an empty array:
```suggestion
dependencies: []
```

technitium-dns/umbrel-app.yml
```diff
@@ -0,0 +1,25 @@
...
+repo: https://github.com/TechnitiumSoftware/DnsServer
+support: https://github.com/TechnitiumSoftware/DnsServer/discussions
+port: 5380
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
+path: ""
+defaultUsername: ""
+defaultPassword: ""
```
nmfretz, line 22: > (NOTE: a full test cannot be done, as the admin password is provided by the $APP_PASSWORD, which cannot be viewed

This is actually where we can set the default credentials so that username and password are shown in the app store when the app is installed, and when a user right clicks on the app icon from the homescreen and selects "Show default credentials".


As an example, say the username is `admin` and we want to show `${APP_PASSWORD}`, then you would do:
```suggestion
defaultUsername: admin
deterministicPassword: true
```

nmfretz (commented):
Thanks for this submission @highghlow! I've left a review with some minor suggestions below, as well as a question on ports/network-mode.

We'll start getting the gallery assets prepared.

technitium-dns/docker-compose.yml
highghlow: OK, since it's done with network_mode: "host" in pi-hole, I'll replicate that here

nmfretz: Sounds good, I think that is the right call in order to make Technitium DNS useful.

FYI, I have made the following additional changes so that it can work with host networking:
- removed the app proxy: https://github.com/getumbrel/umbrel-apps/pull/1046/commits/ddf54df6ff67efdd81948cfcfce3971ce813f14a
- removed the sysctls port range: https://github.com/getumbrel/umbrel-apps/pull/1046/commits/d69edd429b2285250aed1f035fc1e00b89bcea5b

nmfretz:
@highghlow I've made some additional changes to get Technitium DNS working with host networking, and have tweaked the packaging slightly.

The gallery assets are here: https://github.com/getumbrel/umbrel-apps-gallery/tree/master/technitium-dns

Technitium DNS Server is going live to the app store. Thanks again for keeping your eye out for app requests 🙏

Issue #1035, opened 2024-04-05
Add an app for paperless-ngx


maxime1992:
https://docs.paperless-ngx.com is in my top 3 of apps to self host.

Compared to some apps that have a non open source equivalent at least, I don't think I've ever used an app that comes anywhere near paperless-ngx.

It'd be fantastic if you could add support for it :pray:!

Thanks

highghlow:
I have started working on this request

highghlow:
It's done

nmfretz:
Paperless-ngx is now available in the app store thanks to @highghlow! Closed by https://github.com/getumbrel/umbrel-apps/commit/3cb4c100c1a428629680d53d2e479c1e04e73d8d.

PR #1047, opened 2024-04-12
Add paperless

highghlow:
App request: #1035 

# App Submission

### App name
Paperless

### 256x256 SVG icon

Again, that will be a pain to make into an svg

paperless

### Gallery images

proxy-image3
proxy-image2
proxy-image

### I have tested my app on:
- [x] umbrelOS on Linux VM


pavlojs (changes requested):
I think we need to stick to specified version with sha256 provided

`image: ghcr.io/paperless-ngx/paperless-ngx:2.7.2@sha256:703c990a790dfd4d25fb56df3afec27b13cb0926a3818bf265edac9c71311647`

paperless/docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+      APP_PORT: 8000
+
+  broker:
+    image: docker.io/library/redis:7
+    restart: unless-stopped
+    volumes:
+      - ${APP_DATA_DIR}/redisdata:/data
+
+  webserver:
+    image: ghcr.io/paperless-ngx/paperless-ngx:latest
```
pavlojs, line 16: `image: ghcr.io/paperless-ngx/paperless-ngx:2.7.2@sha256:703c990a790dfd4d25fb56df3afec27b13cb0926a3818bf265edac9c71311647`

highghlow: Yeah, sorry, I forgot to pin the version, thanks for pointing that out. Fixed it!

nmfretz:
Thanks for another submission of a requested app @highghlow! 

I've gone through and made the following changes to get it working. Can you please take a look through and give it a quick test? Everything works for me:

- removed some accidentally committed files from the paperless data dir
- committed all the data subdirectories to ensure proper ownership
- added postgres as the database (recommended by Paperless)
- added an exports.sh to ensure that the `consume` and `export` directories in the main Umbrel data directory are created with the correct ownership
- Added environment variables to create a super user and included default credentials in the app manifest. Without a superuser, you can't log in to Paperless after install unless you run a super user creation command inside the container, which we don't want users to have to do.
- Fixed the Tika digest. It was pointing to a non-existent manifest so was resulting in a failed install.

We'll get gallery assets made and then send Paperless-ngx to the app store!

highghlow:
> * removed some accidentally committed files from the paperless data dir

I think I added them to resolve some username+password issue...

> * Added environment variables to create a super user and included default credentials in the app manifest. Without a superuser, you can't log in to Paperless after install unless you run a super user creation command inside the container, which we don't want users to have to do.

Yes, that's was the issue. Thanks for fixing it in a non-terrible way!

(how do you add emojis, anyway, hooray_emoji)

nmfretz:
Thanks again for submitting Paperless-ngx @highghlow! Gallery assets are here: https://github.com/getumbrel/umbrel-apps-gallery/tree/master/paperless

I have made a few commits to finalize the submission. Going live to the app store now.

Issue #1049, opened 2024-04-14
Update BTCPay Server


calebdame:
A new release came out 2 weeks ago with fixes to counter changes in the Shopify API.  Please update!

Thx.

highghlow:
Awaiting merge

PR #1052, opened 2024-04-15
[UPDATE] Btcpay server

highghlow:
Fix: #1049

Updated containers:
 - nicolasdoriev/nbxplorer:2.5.0 -> 2.5.2
 - btcpayserver/btcpayserver:1.12.5 -> 1.13.1
 - btcpaayserver:13.10 -> 13.13

I have partially tested the app on my Debian Umbrel, but because I don't have a bitcoin node (the synchronization takes Ages!) I can't fully verify that the app is working. It gets to starting the app, but because APP_LIGHTNING_NODE_DATA_DIR is null, it fails. The app should be working for people who have the lightning node. Test it on your machine. Just in case.

### DIFF
--- a/btcpay-server/docker-compose.yml
+++ b/btcpay-server/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   nbxplorer:
-    image: nicolasdorier/nbxplorer:2.5.0@sha256:bbff2b6703ed240da263e6ba01deacf2c6520d371d6a0202e87a9e8c7f97b158
+    image: nicolasdorier/nbxplorer:2.5.2@sha256:a85a697d1cb35fd3736ed3bd4690076141beb32a965569b003b209e0b6fd4631
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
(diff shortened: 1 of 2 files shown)


Radiokot:
I'm running it without the Lightning node, just comment mounting APP_LIGHTNING_NODE_DATA_DIR in the `docker-compose` file

highghlow:
> I'm running it without the Lightning node

Well, lightning is required for BTCPay. Even if the app itself supports running without lighnting, most people use it with lnd, so I'll keep that it.

Radiokot:
Sure, I did not suggest commenting it in the `umbrel-apps` repo – just on your local machine, to test that it works.
You can use BTCPay without your own Lightning node and still use Lightning checkout, for example by connecting it to Alby in the store settings.

nmfretz:
Thanks for this update @highghlow! Really appreciate all your work.

I have confirmed that there are no breaking changes, and have tested on arm64 and x86.
I also updated BTCPay's release notes to be a bit more user friendly since they are generally obscure and don't mean too much to the average user.

If you are ever interested in testing out the Bitcoin Node app you can install it, open the app, hit the 3 dots to get to Advanced Settings, change the network to `signet` and restart the node from within the Advanced Settings menu. This will download a 1GB blockchain on a test network. You could then install the Lightning Node and BTCPay server and they will also be on that test network.

Issue #1054, opened 2024-04-16
Request to update homeassistant


KieranSt97:
Please could I request that homeassistant is updated to at least version - 2024.3.2

This is to fix some integrations between my Umbrel HA and Unify protect detailed here - https://github.com/home-assistant/core/issues/112298

I'm not sure if I can contribute these changes but let me know as I am more than happy to!

Thanks

highghlow:
I have started working on the request

highghlow:
Awaiting merge

PR #1055, opened 2024-04-16
Updated Home Assistant

highghlow:
Fix: #1054 

I haven't tested it, but the update is pretty straight-forward, so I'm almost entirely sure it works.


nmfretz:
Thanks @highghlow! Great work using the multi-arch image digest 👌. I have updated the umbrel-app.yml with release notes so that users see relevant info when they go to update. Going live.

Issue #1053, opened 2024-04-16
Update Plex


marciasmedia:
Please update server plex : my version : 1.40.0 / now new version 1.40.1

Thanks

highghlow:
I have started working on this request

highghlow:
Awaiting merge

marciasmedia:
how long can you wait for a simple update??? Have mercy, people

highghlow:
> how long can you wait for a simple update??? Have mercy, people

The developers are busy porting v1.0 to Linux. Give them time!
Luke said the devs will review my PRs this week.

marciasmedia:
Ok I see

nmfretz:
Thanks @marciasmedia and @highghlow. Plex has now been updated to latest 1.40.2: https://github.com/getumbrel/umbrel-apps/pull/1056

PR #1056, opened 2024-04-16
Updated Plex

highghlow:
Fix: #1053 
1.40.0 -> 1.40.1

Haven't tested the update, sorry. It almost certainly works though

### DIFF
--- a/plex/docker-compose.yml
+++ b/plex/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   server:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for all your submission and update PR's @highghlow! Much appreciated. I'll be making my way through these later this week so thanks for your patience in bearing with me.

nmfretz:
Thanks @highghlow, I have bumped this to 1.40.2 which came out after this PR. I also updated the release notes and tested. Going live.

PR #1057, opened 2024-04-18
Update Ordinals to v0.18.2

nmfretz:
This PR updates ord to 0.18.2, which requires the following migration:
https://github.com/ordinals/ord/releases/tag/0.18.2
> Wallet databases are now stored in the /wallets subdirectory of the data dir. To use old wallet databases with 0.18.2, move <WALLET_NAME>.redb files into the /wallets subdirectory of the data dir. Currently, the only information stored in wallet databases are pending etchings.

I have updated the pre-start script to migrate all wallet database files for all bitcoin networks (mainnet, testnet, signet, regtest). I have assumed that all .redb files in the ord data dir that **are not** index.db are instead wallet database files.

I have also removed the `ORD_FIRST_INSCRIPTION_HEIGHT` env var. Ord now handles this automatically for each Bitcoin network, so the app will work well on test networks because we haven't hard coded the first inscription height for mainnet.

Tested on arm64 and x86

@smolgrrr @mrv777 can I please get some eyes and additional tests on this 🙏

### DIFF
--- a/ordinals/hooks/pre-start
+++ b/ordinals/hooks/pre-start
@@ -43,35 +43,42 @@ fi
 # for installs post 0.18.2, the AT_LEAST_0-18-2 file will be harmlessly created
 
 if [[ ! -f "${APP_DATA_DIR}/AT_LEAST_0-18-2" ]]; then
-    declare -a dirs=("" "/testnet3" "/regtest" "/signet")  # Array of bitcoin network subdirectories
+    echo "AT_LEAST_0-18-2 not found, proceeding with wallet database migration if necessary..."
+    # declare bitcoin network directories
+    declare -a dirs=("" "/testnet3" "/regtest" "/signet")
     for dir in "${dirs[@]}"; do
-        if [[ -d "${ORDINALS_DATA_DIR}${dir}" ]]; then
-            has_wallets=false
-            files_to_move=()  # Array to store files that need to be moved
+        dir_path="${ORDINALS_DATA_DIR}${dir}"
+        echo "Checking directory: ${dir_path}"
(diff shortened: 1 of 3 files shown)


ordinals/hooks/pre-start
```diff
@@ -37,7 +37,48 @@ if [[ ! -f "${APP_DATA_DIR}/AT_LEAST_0-18-1" ]]; then
     touch "${APP_DATA_DIR}/AT_LEAST_0-18-1"
 fi
 
+# migrate all wallet database files to ord/wallets, ord/testnet3/wallets, ord/regtest/wallets, and ord/signet/wallets for pre-0.18.2 versions: https://github.com/ordinals/ord/releases/tag/0.18.2
```
mrv777, line 40: Hmm, I'll have to double check but I didn't think ord stored the wallets in redb files.  I thought when you do wallet create it creates a folder with a wallet.dat file in the bitcoin directory.

So if you did `ord wallet --name test1 create` a new directory would be created in `umbrel/app-data/bitcoin/data/bitcoin` called `test` with the wallet.dat file in it.

I just woke up though, so maybe I'm misunderstanding this code and will look more at it in a bit

nmfretz: It's all good, this is for the ord wallet database files (which are redb) not the bitcoind wallet.dat files.

I tested with multiple wallet database files on different networks and the migration works well.

mrv777: Okay, I see now.  Yeah, looks good reading it over again.  I haven't looked into how I can test an app without being in the app store or a community app store, but I think the code looks good

nmfretz:
Going live.

PR #1058, opened 2024-04-18
update immich 1.101

mateosilguero:


nmfretz:
Thanks @mateosilguero! I've bumped the version to 1.103.0 which just came out a few minutes ago. Tested on arm64 and amd64 and working well. 

### A couple notes for our records:
Due to the way we have implemented the docker-compose.yml we are not impacted by the breaking change from 1.102.0.

I reverted a few items to bring us back in-line with the official immich compose file and the general format of other apps on umbrelOS. For commit https://github.com/getumbrel/umbrel-apps/pull/1058/commits/e3c83d55fb2bd20ae3e14ac75f12bcb199803c92, I removed the `container_name` properties from the services. umbrelOS 1.x makes sure that containers follow the `<project-name>_<service-name>_1` naming convention regardless of whether Compose v1 or v2 is used: https://github.com/getumbrel/umbrel/blob/ace3d9d2e528c798fab0b75b8000b16f30d2aa3b/packages/umbreld/source/modules/apps/app.ts#L93-L109

mateosilguero:
Thank you !

mateosilguero:
@nmfretz The immich logo has changed, could we updated at the store ?

nmfretz:
oh nice @mateosilguero, I'll get that updated. Thanks!

nmfretz:
@mateosilguero updated here https://github.com/getumbrel/umbrel-apps-gallery/commit/caaa722b4239ff7999d8aa6519b029d1465f22a1 🤝

PR #1061, opened 2024-04-21
Add issue templates

highghlow:
This is not an app addition

This PR adds github issue templates. They include:
 - App bug report
 - App request
 - App update request


nmfretz:
Hey @highghlow, with the umbrelOS apps rewrite approaching there is going to be significant changes to the apps framework and to the use of this repo, so let's avoid adding templates for now. Once the rewrite is in effect and the new framework is in place we can make sure there is a standardized way to submit bug reports, etc.

PR #1062, opened 2024-04-22
Fix bad gateway error in Seafile-app

ngutech21:
This fixes 2 issues:
- bad gateway error after starting the app: The seafile container has to use the container-name seafile-mysql to access the database
-  When connecting through one of the desktop seafile apps, there was an internal server error: This was caused by the additional proxy auth, that the apps could not handle so I deactivated it. Seafile already provides an authentification, so this seems redundant anyway.

I have also updated seafile to the latest patch version and tested it on umbrel home 1.1.

### DIFF
--- a/seafile/docker-compose.yml
+++ b/seafile/docker-compose.yml
@@ -5,28 +5,29 @@ services:
     environment:
       APP_HOST: seafile
       APP_PORT: 80
+      PROXY_AUTH_ADD: "false"
 
   db:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this PR @ngutech21, much appreciated. I'll be taking a look at this late this week / early next week.

ngutech21:
@nmfretz Did you find the time to review my PR?

PR #1063, opened 2024-04-24
updating databag container

balzack:
Normally my releases include more features, but this version of Go has resolved several security vulnerabilities.


nmfretz:
Excellent, thanks for this @balzack. Tested and going live.

PR #1067, opened 2024-04-26
Added Home Assistant media storage and updated Home Assistant

highghlow:
This commit adds a a new volume:
```
${UMBREL_ROOT}/data/storage/home-assistant-media:/config/media
```
as requested in getumbrel/umbrel#1800

This PR also updates hass for the change to apply. There is a conflict though, so I think you should apply umbrel-app.yml version change and docker-compose.yml version change (because this is required for clients to receive the update), but not apply the changelog (it didn't change between the repo version and the PR version)


highghlow:
It turns out the contributor can resolve conflicts themselves.

nmfretz:
Great idea, I had not realized this was a use case that people wanted. Thanks for keeping your eyes out for these types of requests.

> This commit adds a a new volume:
>```
>${UMBREL_ROOT}/data/storage/home-assistant-media:/config/media
>```

I think the default media path in Home Assistant is `/media` not `/config/media`:
https://www.home-assistant.io/more-info/local-media/setup-media

So what we want for this to work automatically is:
```
<path-to-persistent-media-storage>:/media
```

### But just a thought here on where the persistent storage on the host lives:

How would most Home Assistant users use the media folder in HA? @TowyTowy I'd love to hear your thoughts here as well.

With most apps on umbrelOS, we make it so that app data is stored within the app data directory. That way, if an app is deleted by a user, that app's data is deleted too and doesn't stick around and confuse users by continuing to take up disk space. If that's how we want the media folder on Home Assistant to function, then we want to bind it to the host like this:

```
${APP_DATA_DIR}/data/media:/media
```

Under this scenario, users would use the Home Assistant UI to view, add, and remove media. If the user then deletes Home Assistant, all media they have added through the UI will be deleted.

Alternatively, if users are wanting to have access to all media on their umbrel device regardless of whether it was added via Home Assistant or by some other app instead like one of the *arr apps, then we should probably bind this to the entire `downloads` folder like we do for Plex, Jellyfin, etc. In that case, we would want to bind:
```
${UMBREL_ROOT}/data/storage/downloads:/media
```

Under this scenario, HA would have full access to the downloads folder and all subfolders that get generated by other apps (movies, shows, incomplete, etc). Any media files added through the Home Assistant UI would persist if the user deletes Home Assistant.

highghlow:
I don't really know what the media folder is used for, but I moved it to $storage/downloads

TowyTowy:
I personally used it for a sound effect that I wanted to play in one of my automations. But I'm guessing there's a big variety of use cases for it.

KieranSt97:
Plus one to TowyTowys comment above. Is this able to be merged? Its frustrating having to reupload the media every time its restarted.

nmfretz:
> Plus one to TowyTowys comment above. Is this able to be merged? Its frustrating having to reupload the media every time its restarted.

Thanks for the input @KieranSt97! Can you let me know how you usually use the media folder? Do you add media through the Home Assistant UI? This will help me determine where to mount /media on the host.

KieranSt97:
Just to add to the above. In my case the media lives in /media/Chime.mp3 (for example) inside the container.

nmfretz:
Thanks for the responses @KieranSt97 and @TowyTowy.

Let's go with  binding /media in the container to the main umbrelOS storage/dowloads folder:
`${UMBREL_ROOT}/data/storage/downloads:/media`

We can always change this later. But this way any media added to umbrelOS from other media apps will be accessible through Home Assistant, so users can properly do things like play downloaded media with a vlc integration, or play audio through smart speakers, etc.

Thanks again for being on top of this @highghlow! I have bumped the image to the latest 2024.6 release, updated the app manifest, and added storage directory permissions.

KieranSt97:
Hi @nmfretz I know this has been merged now but how can I apply these changes? There's no update for home assistant showing and my compose file doesn't contain the new volume mount

nmfretz:
Darn, thanks for reporting @KieranSt97! Looks like I  wasn't careful at the last second and missed that we reverted the version in the manifest: https://github.com/getumbrel/umbrel-apps/pull/1067/commits/15bcb5326f1c82fe2375d5e88c3fd225718adfe7

I have fixed here https://github.com/getumbrel/umbrel-apps/commit/c781a5cdf95bc044c05355fc64e6f9664e6db1a8. You should now see an update available on your Umbrel. Thanks again!

PR #1077, opened 2024-05-06
Update Overseer from 1.32.5 to 1.33.2

rtrive:
Updates Overseer from `1.32.5` to `1.33.2`.

Tested on rpi5 with Umbrel OS 1.1


nmfretz:
Thanks a bunch for this update @rtrive! I added release notes to the manifest so users see them when they get the update notification. Going live.

PR #1079, opened 2024-05-08
App Submission: localai

unhighghlow:
noGPUpasthu4u

# App Submission

### App name
LocalAI

### 256x256 SVG icon

For some reason I was only able to find this as a jpeg

https://github.com/go-skynet/LocalAI/assets/2420543/0966aa2a-166e-4f99-a3e5-6c915fc997dd

### Gallery images
...

Screenshot 2024-05-08 at 08-44-05 LocalAI API - v2 14 0 (b58274b8a26a3d22605e3c484cf39c5dd9a5cf8e)

image

I can't get more screenshots because running this without gpu passthrough is really, **really** slow.

### I have tested my app on:
- [x] umbrelOS on Linux


nmfretz:
Thanks for getting this together @highghlow. I've been playing around with it this morning to package it so that it installs on a users machine with no/minimal models downloaded initially. The image and the models are quite large, so install time is very long.

But I've just realized embarrassingly late in my testing that the localai images are not multi-architecture. They are only available for x86 right now. This means that the install would fail for arm users (e.g., RPi 5) and leave users in a broken state. Let's close this for now until an official arm image is available. I will open an Issue in their repo asking about official arm support.

nmfretz:
Ohhhhh, hold up I see that arm support is on the horizon so let's keep this PR open!
https://github.com/mudler/LocalAI/issues/1249#issuecomment-2134721490

Once they are available I'm hoping we can use their "standard container images" which don't include pre-installed models:
https://localai.io/basics/container/#standard-container-images

nmfretz:
Looks like some images are now multi-arch. e.g.,:

image

al-lac:
@nmfretz can this be looked at again? Would be great to have this.

highghlow:
It seems like the localai docker image is still missing arm64 support, which is the main issue here

nmfretz:
> It seems like the localai docker image is still missing arm64 support, which is the main issue here

Ah that's too bad. Thanks for checking @highghlow. This multiarch image I noticed [here](https://github.com/getumbrel/umbrel-apps/pull/1079#issuecomment-2316498509) might have just been a trial they were doing that didn't pan out longterm for them.

al-lac:
Not sure where you are checking, but using the latest version (2.24.1) i was able to install it on my raspberry pi 5:
CleanShot 2024-12-09 at 13 26 30

See the container here: https://registry.hub.docker.com/layers/localai/localai/v2.24.1-aio-cpu/images/sha256-592397759ab3fd3fc58643fd1c9437ad556ee32b31e944ec6cc622bdb547e0d7?context=explore

Did not get a response yet for my query though, so really not performant, at least on a pi 5.

nmfretz:
@al-lac, ya that's where I was checking previously as well. So looks like at least the all-in-one cpu image is still multiarch.

I recall that the AIO image downloads a bunch of models on install, correct? How long did it take for the models to download, and was the UI available during that time?

al-lac:
Yeah exactly. So it downloads around **2.25GB** for the container, which results in about **6GB** uncompressed. The models it downloads automatically come up to around **14GB**. So really depends on the network speed. Not sure if the UI was available during the model download as I was away for a bit. 

But after downloading everything and even downloading a really small model i was never able to get a response or get anything to properly work. Also had to stop nearly every other application running and during its peak it consumed around **4,8GB** of Memory which is really a lot to handle for the **8GB** Raspberry Pi.

All of this could of course work better on the Umbrel Home or on other machines.

highghlow:
Boom, done. @nmfretz

al-lac:
Hey @unhighghlow,

Sorry that this took so long. Finally was able to test this again with the latest version, and it works really well on my Umbrel Home. I even managed to generate an image, but that is just really slow, took over 10 minutes. Chatting and tts works great though!

We will look into getting the assets ready and finally ship it! 

FYI @nmfretz

al-lac:
Hey @unhighghlow,

finally going live with this one! 🚀 🤖 

CleanShot 2025-04-17 at 21 00 50

Niclassslua:
image

<br><br>

The LocalAi icon is very low quality.

I would suggest using the one I uploaded. It's 900x900.

<br><br>

localai

PR #1084, opened 2024-05-10
Add WireGuard

sharknoon:
# App Submission

### App name
WireGuard

### 256x256 SVG icon
wireguard

### I have tested my app on:
- [x] umbrelOS on Linux VM

### DIFF
--- a/wireguard/docker-compose.yml
+++ b/wireguard/docker-compose.yml
@@ -1,0 +1,24 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: wireguard_app_1
+      APP_PORT: 51821
+  
+  app:
+    image: ghcr.io/wg-easy/wg-easy:12@sha256:daa94b77901bdef3cb10151416c095e64cf66a6539fa8f748c09812b60b97f49
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - "51820:51820/udp"
+    volumes:
+      - ${APP_DATA_DIR}/app/wireguard:/etc/wireguard
+    environment:
+      WG_HOST: ${DEVICE_DOMAIN_NAME}
+    cap_add:
+      - NET_ADMIN
+      - SYS_MODULE
+    sysctls:
+      - net.ipv4.ip_forward=1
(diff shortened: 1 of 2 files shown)


wireguard/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+      APP_PORT: 51821
+  
+  app:
+    image: ghcr.io/wg-easy/wg-easy:12@sha256:daa94b77901bdef3cb10151416c095e64cf66a6539fa8f748c09812b60b97f49
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - "51820:51820/udp"
+    volumes:
+      - ${APP_DATA_DIR}/data/wireguard:/etc/wireguard
```
nmfretz, line 16: Can you please add this host path with a `.gitkeep` as described here (https://github.com/getumbrel/umbrel-apps/pull/1083#discussion_r1609144592)?

wireguard/umbrel-app.yml
```diff
@@ -0,0 +1,50 @@
+manifestVersion: 1
+id: wireguard
+name: WireGuard
+tagline: Run your own free and unlimited WireGuard VPN
+category: networking
+version: "12"
+port: 38062
```
nmfretz, line 7: The default Wireguard UI port of 51821 is free right now. No other app uses it, so this can be changed to `51821` so it behaves as users would expect if they were looking at troubleshooting guides on the web.

nmfretz (commented):
Another fantastic submission @sharknoon! We'll get gallery images prepared.

I have left some minor suggestions below.

sharknoon:
@nmfretz Thank you for your suggestions, I have implemented them 😃

sharknoon:
Unfortunetly also for wireguard `user: 1000:1000` keeps the container from starting up

nmfretz:
> @nmfretz Thank you for your suggestions, I have implemented them 😃

Excellent, thanks @sharknoon.

Gallery assets are here: https://github.com/getumbrel/umbrel-apps-gallery/tree/master/wireguard

I have bumped wireguard to the latest version 13 and pushed a couple final commits. Going live now. Thanks again for all your hard work!

PR #1087, opened 2024-05-13
Update Vikunja to v0.22.1

nmfretz:
Updates Vikunja to v0.22.1 and adds specific db host name.


nmfretz:
Tested on arm64 and x86

PR #1093, opened 2024-05-16
[URGENT] Updated tailscale to v1.66.0

highghlow:
This updated fixes a security issue: TS-2024-005
Fixed: getumbrel/umbrel#1831

I'll try to figure out a way to automate updates.

### DIFF
--- a/tailscale/docker-compose.yml
+++ b/tailscale/docker-compose.yml
@@ -3,7 +3,7 @@ version: "3.7"
 services:
   web:
     network_mode: "host" # TODO: We can remove this later with some iptables magic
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @highghlow, testing now.

FYI, with the way Tailscale is currently packaged with docker compose on the app store I don't think we are vulnerable to this particular security issue because we don't pass in flags for exit nodes or app connectors. Still super important to keep this up-to-date so thanks a ton.

We also shouldn't be affected by [this breaking change](https://arc.net/l/quote/wexrihvx).

nmfretz:
Tested and working great. Thanks again @highghlow!

Issue #1085, opened 2024-05-11
[App Request] scanservjs


sund3RRR:
scanservjs is the SANE scanner nodejs web ui

Project link: https://sbs20.github.io/scanservjs/
Sources link: https://github.com/sbs20/scanservjs

highghlow:
If everything goes according to plan. This should work. I've started working on this request

nmfretz:
If others make there way to this app request, we would love some help testing this app submission 🙏
https://github.com/getumbrel/umbrel-apps/pull/1095#issuecomment-2144264517

PR #1095, opened 2024-05-17
App submission: scanservjs

unhighghlow:
Request: #1085 

This app will require the scanner to be connected by usb to the Umbrel. Less that ideal, but that's what usb ports on the Umbrel home are for... right?

Also, btw, can you give me the power to close issues in the `umbrel` repo? There're a lot of resolved issues where the authors are not responding.

# App Submission

### App name
scanservjs

### 256x256 SVG icon

This project doesn't have an icon. There're no google image results for `scanservjs`. You'll have to make an icon of your own.

### Gallery images

image

image

There's not much about `scanservjs` on the internet

### I have tested my app on:
I haven't tested this app. Both because I don't currently have access to my server and because I don't have a scanner. But it should hopefully work.

### DIFF
--- a/ordinals/docker-compose.yml
+++ b/ordinals/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   ord:
-    image: nmfretz/ord:v0.17.1@sha256:ee391d9152c0ac6f0f6909c19fc14e747af72abf9813d6a79b22d6f4a55ef29c
+    image: mayankchhabra/ord:0.18.1@sha256:9cc88552cca2b59ed6417790ec1b43f4dbfaf0c9ea3fdf79bf055dcd6dcb60b8
     # This needs to run as root
     # user: "1000:1000"
     restart: on-failure
--- a/ordinals/hooks/pre-start
+++ b/ordinals/hooks/pre-start
@@ -26,12 +26,15 @@ if [[ -f "${APP_DATA_DIR}/index.redb" ]]; then
 fi
 
 # re-index the ordinals data for runes
-# for installs post 0.17.1 this file will be harmlessly created 
-if [[ ! -f "${APP_DATA_DIR}/AT_LEAST_0-17-1" ]]; then
-    if [[ -f "${ORDINALS_DATA_DIR}/index.redb" ]]; then
-        rm -f "${ORDINALS_DATA_DIR}/index.redb"
-    fi
-    touch "${APP_DATA_DIR}/AT_LEAST_0-17-1"
+# for installs post 0.18.1 this file will be harmlessly created 
+if [[ ! -f "${APP_DATA_DIR}/AT_LEAST_0-18-1" ]]; then
+    # delete index db for all bitcoin networks
+    for file in "${ORDINALS_DATA_DIR}/index.redb" "${ORDINALS_DATA_DIR}/testnet3/index.redb" "${ORDINALS_DATA_DIR}/regtest/index.redb" "${ORDINALS_DATA_DIR}/signet/index.redb"; do
+        if [[ -f "${file}" ]]; then
+            rm -f "${file}"
+        fi
+    done
+    touch "${APP_DATA_DIR}/AT_LEAST_0-18-1"
 fi
 
 
--- a/ordinals/umbrel-app.yml
+++ b/ordinals/umbrel-app.yml
@@ -2,11 +2,11 @@ manifestVersion: 1
 id: ordinals
 category: bitcoin
 name: Ordinals
-version: "0.17.1"
+version: "0.18.1"
 tagline: Run your own index, block explorer, and command-line wallet for Ordinals
 description: >
   Run your own index, block explorer, and command-line wallet for Ordinals. The app automatically connects to your Bitcoin node on umbrelOS for trustless operation. Simply install the app and wait for Ordinals to index inscriptions and runes.
-  
+
 
   Use the block explorer to view Ordinal inscriptions. Search blocks, transactions, outputs, and individual satoshis.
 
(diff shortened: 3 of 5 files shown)


sund3RRR:
Probably it will not work, at least for hp scanners.
There are different targets in [Dockerfile](https://github.com/sbs20/scanservjs/blob/master/Dockerfile#L115), you can view build instructions [here](https://github.com/sbs20/scanservjs/blob/master/docs/50-development.md#using-docker-to-build)
Ideally, it is better to make a choice in umbrel store to install basic image or hp/airscan image, but as I understand, it is impossible.

And I think you also need to map usb device `/dev/bus/usb/`

highghlow:
> And I think you also need to map usb device `/dev/bus/usb/`

They said in the docker guide that `--priveleged` is used for this

sund3RRR:
It can work without privileged mode. I installed libsane-hpaio in the container and mapped usb device, it works at least for hp scanner.

highghlow:
> It can work without privileged mode. I installed libsane-hpaio in the container and mapped usb device, it works at least for hp scanner.

But we don't want every usb device to be mapped to scanservjs (And we don't which specific device to mount)

sund3RRR:
> > It can work without privileged mode. I installed libsane-hpaio in the container and mapped usb device, it works at least for hp scanner.
> 
> But we don't want every usb device to be mapped to scanservjs (And we don't which specific device to mount)

Hard question. I think grant container privileged rights is worse than mapping the whole usb bus. At least for ubmrel store application.

highghlow:
> Hard question. I think grant container privileged rights is worse than mapping the whole usb bus. At least for ubmrel store application.

Other apps (and umbrel itself) might want to use usb

nmfretz:
Thanks for submitting scanservejs @highghlow! And thanks for your input @sund3RRR, much appreciated.

We'll need some help from the community to properly test this app because both @highghlow and I don't have a scanner. This may require some back and forth here. @sund3RRR do you have a SANE compatible scanner?

As I understand it, the idea here is that for this packaging of scanservjs we will assume that a user will plug their scanner directly into their umbrelOS device.

> They said in the docker guide that --priveleged is used for this

The increased container privileges may indeed be enough in this case because we have also mapped `/var/run/dbus`; however, we need to test that this is actually the case.

I have committed some changes to get scanservjs to install correctly and have added what I assume would be the maximum permissions we might need to give this container:
- `privileged: true`
- mapping `/var/run/dbus`
- mapping `/dev`. (see my comment here for why we may need to do this https://github.com/getumbrel/umbrel-apps/pull/1044#discussion_r1606376243)

This should allow it to work, but our goal now should be to reduce permissions as much as possible without reducing the functionality of scanservjs. We should test the following:

1) Does this PR work as of https://github.com/getumbrel/umbrel-apps/pull/1095/commits/1bc54379cfe648f4ee0d6aef35a5a3bb7063ec5a with with `privileged: true`, `/var/run/dbus`. and `/dev`?

1) Remove the `/dev` bind mount. Can you boot the scanservjs container without a scanner plugged in, then plug in a scanner and have it be discovered? 

2) Keep the `/dev` bind mount but remove `privileged: true`. Does it still work?

3) remove both `/dev` mapping and `privileged: true`. Does it still work?

4) If it turns out that the `/dev` bind mount is required in the testing above, does reducing the mounts scope to `/dev/bus/usb` still work?

nmfretz:
@sund3RRR bumping this in case you are able to help test.

nmfretz:
@sharknoon do you have a scanner to test by chance?

sund3RRR:
I don't have an umbrel anymore because I moved my server on orange pi 5 machine.

nmfretz:
Ah no worries, thanks for the response @sund3RRR

sharknoon:
@nmfretz Unfortunately I don't have a scanner at the moment, mine broke just a couple of weeks ago... But I will buy a new one soon, mainly because of german bureaucracy 😅

nmfretz:
> But I will buy a new one soon, mainly because of german bureaucracy 😅

Ha! Sounds good @sharknoon.

nmfretz:
@al-lac do you by chance happen to have access to a scanner to test this app?

al-lac:
I have a HP All-In-One, could try if that one works. Is not on the officially supported list, but I can give it a try.

al-lac:
Hey @nmfretz, I was able to test this app yesterday with my HP Envy 4500. 

It worked flawlessly via airscan or the network. The USB was also mapped into the container, but it never showed up as an option. For me scanning via the network is anyways way more practical 👍 

Some stuff from the terminal:
```bash
# in the container
root@ddea0e312dc3:/usr/lib/scanservjs# ls /dev/usb/
lp0
root@ddea0e312dc3:/usr/lib/scanservjs# scanimage -L
device `escl:http://192.168.0.15:8080' is a HP ENVY 4500 platen scanner
device `airscan:e0:HP ENVY 4500' is a eSCL HP ENVY 4500 ip=192.168.0.15

# on the umbrel home
root@umbrel-home:~# lsusb
Bus 004 Device 001: ID 1d6b:0003 Linux Foundation 3.0 root hub
Bus 003 Device 003: ID 8087:0aaa Intel Corp. Bluetooth 9460/9560 Jefferson Peak (JfP)
Bus 003 Device 005: ID 03f0:c511 HP, Inc ENVY 4500 series
Bus 003 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Bus 002 Device 001: ID 1d6b:0003 Linux Foundation 3.0 root hub
Bus 001 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
```
CleanShot 2025-01-28 at 19 09 17

nmfretz:
Excellent, thanks @al-lac we can get gallery assets prepared!

> It worked flawlessly via airscan or the network. The USB was also mapped into the container, but it never showed up as an option.

Let's launch as-is then with these:

```yaml
    volumes:
      - ${APP_DATA_DIR}/data/scans:/var/lib/scanservjs/output
      - ${APP_DATA_DIR}/data/config:/etc/scanservjs
      - /var/run/dbus:/var/run/dbus
      # map the host's /dev to allow discovery of usb devices without needing to know apriori which devices are connected
      - /dev:/dev 
    privileged: true
 ```
 
 and revisit if we get feedback from users

nmfretz:
Gallery assets are ready. Going live!

Thanks again for your efforts on this @unhighghlow.

image

PR #1098, opened 2024-05-19
Remove lndboss from app store

niteshbalusu11:
I am not maintaining the app anymore, please remove it from the app store. I don't think an unmaintained app should be on it.


nmfretz:
Hey @niteshbalusu11 - sad to see LndBoss go as it was a great little BoS GUI, but we totally understand your reasoning.

We can't completely remove the app from this repo yet or else it will break current installs of LndBoss for users still on umbrelOS 0.5.x, but here is what I have done with the following commit: https://github.com/getumbrel/umbrel-apps/commit/7287da88a7ea9638019e18e5be8b32e71a7ccc81:
- **Remove the app from umbrelOS 1.x app store:** added `disabled: true` to the app manifest so that umbrelOS 1.x users do not see the app in the app store any more. No one on umbrelOS 1.x will be able to install the app through the official app store UI.
- **Show app as deprecated in umbrelOS 0.5.x app store:** added deprecation notice right at the top of the app description so users are aware that this app is no longer maintained and will not be receiving updates. 

We are working on an apps ecosystem rewrite which will make it so these workarounds don't need to be implemented.

Hope to see you you back here for future app submissions Nitesh!

1.x:
image

0.5.x:
image

niteshbalusu11:
Thank you.

PR #1109, opened 2024-05-28
Add missing `releaseNotes` keys and logo fix

sharknoon:
Hello,

I have found some small issues:

- Some apps are missing the key for the release notes in the `umbrel-app.yml`. 
- Adguard Home has the wrong [logo](https://github.com/getumbrel/umbrel-apps-gallery/blob/master/adguard-home/icon.svg). According to their [repo](https://github.com/AdguardTeam/AdGuardHome), the app should have this logo:
adguard
(This is a 256x256 optimized svg, ready to be used as app icon)

Kind regards

Sharknoon

### DIFF
--- a/agora/umbrel-app.yml
+++ b/agora/umbrel-app.yml
@@ -27,3 +27,4 @@ defaultUsername: umbrel
 torOnly: false
 submitter: Jonathan Zernik
 submission: https://github.com/getumbrel/umbrel/pull/1282
+releaseNotes: ""
--- a/bitfeed/umbrel-app.yml
+++ b/bitfeed/umbrel-app.yml
@@ -29,3 +29,4 @@ defaultUsername: ""
 defaultPassword: ""
 submitter: Mononaut
 submission: https://github.com/getumbrel/umbrel/pull/1202
+releaseNotes: ""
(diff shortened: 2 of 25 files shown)


nmfretz:
Thanks for this PR @sharknoon! The `releaseNotes` key is optional in all versions of umbrelOS, but I think it is a really good idea to include it as a placeholder in every manifest like you have done here. This way it reminds people to update the release notes when the first app update goes through.

I have let our designer @surajmahraj know about the Adguard Home logo 🤝

--------------------------
Also, for some context on recent commits where we added the `path` and `dependencies` keys to a couple apps. These keys are [optional in umbrelOS 1.x](https://github.com/getumbrel/umbrel/blob/5a042334e36e65128058f512c29e7af422f4be0c/packages/umbreld/source/modules/apps/schema.ts#L19-L50), but they actually cause problems when they are missing in umbrelOS 0.5.x so we added them in to maintain compatibility with older versions. What happens when they are missing is that the app's install screen in the app store will not render in umbrelOS 0.5.x.

sharknoon:
Thank you for clarifying. I have based my linter on the [schema](https://github.com/getumbrel/umbrel/blob/5a042334e36e65128058f512c29e7af422f4be0c/packages/umbreld/source/modules/apps/schema.ts#L19-L50) and was confused, that it seemd to be mandatory. Looking forward to the new App Framwork to detect such errors early 😀

Thank you for reviewing this PR 👍

PR #1112, opened 2024-05-29
Add TeamSpeak

sharknoon:
# App Submission

### App name
TeamSpeak

### 256x256 SVG icon
logo

### Gallery images
Bildschirmfoto am 2024-05-29 um 16 01 09
Bildschirmfoto am 2024-05-29 um 13 27 18
Bildschirmfoto am 2024-05-29 um 13 27 05
Bildschirmfoto am 2024-05-29 um 13 27 01
Bildschirmfoto am 2024-05-29 um 13 26 19
Bildschirmfoto am 2024-05-29 um 13 26 15
Bildschirmfoto am 2024-05-29 um 13 25 50
Bildschirmfoto am 2024-05-29 um 13 25 38

### I have tested my app on:
- [x] umbrelOS on Linux VM


nmfretz:
Thanks for yet another excellent app submission @sharknoon. I've just tested TeamSpeak (never used before) and it is great and is working well.

I modified the instructions slightly. Any changes you think we should make to them?:
image

We'll get gallery images made up.

sharknoon:
Thanks for reviewing @nmfretz. I am glad that you take a look at the english texts, as it is not my mother tongue 😀

sharknoon:
The new description looks great 👍

nmfretz:
Thanks again for submitting TeamSpeak @sharknoon! Gallery assets are here: https://github.com/getumbrel/umbrel-apps-gallery/tree/master/teamspeak

Going live 🎙️

image

sharknoon:
It was my pleasure 😀 Thank you @nmfretz for reviewing it 👍

jimbrend:
Wow this is awesome, nice one @sharknoon!
<3 teamspeak

sharknoon:
@jimbrend Glad to hear that 😀 I hope you enjoy TeamSpeak 🎉

sharknoon:
@pimushkin thank you for the catch! I haven't tested it on arm64 and totally missed this issue.

Unfortunetely teamspeak only provides binaries (no source code available). They also [have no plans](https://github.com/TeamSpeak-Systems/teamspeak-linux-docker-images/issues/6#issuecomment-410220789) to support it in the future which is very unfortunate.

Therefore we need to emulate it. I have looked around and found one [docker image](https://hub.docker.com/r/ertagh/teamspeak3-server) that is working on arm, but that does not run on x86...

Right now I am looking how to build my own image that can be deployed as a multi arch image.

sharknoon:
The problem is, that umbrel has no qemu installed and therefore we can't simply emulate teamspeak. I basically hope one of the two things to happen here:

- TeamSpeak adds a official arm64 docker image or at least an arm64 binary
- Umbrel supports apps, that don't have support for all platforms (amd64 and arm64), for example via emulation

Therefore unfortunately I am out of ideas on how to make TeamSpeak compatible with arm64 as well...

nmfretz:
Ah darn. This is my fault @sharknoon. I must have only checked the `joni1802/ts3-manager` image. Thinking about this...

sharknoon:
Same goes for me @nmfretz. I somehow assumed it has `arm64` support, without actually checking it 🫣

nmfretz:
@sharknoon I have temporarily removed TeamSpeak to prevent arm64 broken installs https://github.com/getumbrel/umbrel-apps/commit/8e1ee6c7fa53189cd39ac59cafad549aebe14f01 😢

TeamSpeak won't show up in the app store on umbrelOS, but user's who have installed it successfully on x86 devices will continue to be able to use the app. It will also still show up on https://apps.umbrel.com/app/teamspeak with a "temporarily disabled" notice. Let's re-enable it as soon as we either 1) come up with an arm64 solution, or 2) umbrelOS implements architecture-specific apps.

sharknoon:
Thank you @nmfretz for disabling it. I think we should support non arm apps as well. Either through qemu emulation (best solution in my opinion, but slow, better something slow than not having it), or by filtering the store to only show compatible apps (arm users will miss out especially on game servers).

PR #1115, opened 2024-05-31
Update ghostfolio to 2.83.0

Batgame:
Release notes: https://github.com/ghostfolio/ghostfolio/releases

Could not test it, but I don't see any breaking changes. 

Follows a repeated request for an update on the forum, as the current version is in some cases unusable. See here: 

https://community.umbrel.com/t/ghostfolio-needs-to-be-updated-due-to-yahoo-api-issue/16667

### DIFF
--- a/ghostfolio/docker-compose.yml
+++ b/ghostfolio/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3334
 
   server:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks very much for updating Ghostfolio @Batgame! 

I scanned the release notes for past versions for any additional changes. I've added `init: true` to the compose file to handle the change in v2.70.0 where Ghostfolio now relies on compose to set up Tini instead of handling it in the Dockerfile.
https://github.com/ghostfolio/ghostfolio/releases/tag/2.70.0

Tested and going live.

PR #1116, opened 2024-06-01
Updated to databag version supporting MFA

balzack:
- Added support for Multi-Factor Authentication for admin and users
-- integration tested with freeOTP(+) and [online test site](https://www.token2.com/site/page/token2-totp-toolset-sha256-version)


nmfretz:
Wonderful, thanks for the update @balzack. Going live.

PR #1118, opened 2024-06-02
Add ntfy

sharknoon:
# App Submission

### App name
ntfy

### 256x256 SVG icon
logo

### Gallery images
Bildschirmfoto am 2024-06-02 um 13 58 39
Bildschirmfoto am 2024-06-02 um 14 00 44
Bildschirmfoto am 2024-06-02 um 14 00 30
Bildschirmfoto am 2024-06-02 um 13 59 04
Bildschirmfoto am 2024-06-02 um 13 58 57
Bildschirmfoto am 2024-06-02 um 13 58 52

### I have tested my app on:
- [x] umbrelOS on Linux VM

### DIFF
--- a/ntfy/docker-compose.yml
+++ b/ntfy/docker-compose.yml
@@ -1,0 +1,33 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: ntfy_app_1
+      APP_PORT: 80
+      PROXY_AUTH_ADD: false
+      
+  init:
+    image: binwiederhier/ntfy:v2.11.0@sha256:4a7d0f0adc6d5d9fc36e64ab55ef676e76e124a2bdd50ce115b6d9c1c7430294
+    restart: 'no'
+    command: >
+      NTFY_PASSWORD=${APP_PASSWORD} ntfy user add --role=admin umbrel
+
+  app:
+    image: binwiederhier/ntfy:v2.11.0@sha256:4a7d0f0adc6d5d9fc36e64ab55ef676e76e124a2bdd50ce115b6d9c1c7430294
+    depends_on:
+      - init
+    user: "1000:1000"
+    restart: on-failure
+    stop_grace_period: 1m
+    command: serve
+    volumes:
+      - ${APP_DATA_DIR}/data/cache:/var/cache/ntfy
+      - ${APP_DATA_DIR}/data/data:/etc/ntfy
+      - ${APP_DATA_DIR}/data/lib:/var/lib/ntfy
+    environment:
+      NTFY_BASE_URL: http://${DEVICE_DOMAIN_NAME}:${APP_PROXY_PORT}
+      NTFY_CACHE_FILE: /var/cache/ntfy/cache.db
+      NTFY_ENABLE_LOGIN: true
+      NTFY_AUTH_FILE: /var/lib/ntfy/user.db
+      NTFY_AUTH_DEFAULT_ACCESS: deny-all
--- a/ntfy/docker-compose.yml
+++ b/ntfy/docker-compose.yml
@@ -6,22 +6,15 @@ services:
       APP_HOST: ntfy_app_1
       APP_PORT: 80
       PROXY_AUTH_ADD: false
-      
-  init:
-    image: binwiederhier/ntfy:v2.11.0@sha256:4a7d0f0adc6d5d9fc36e64ab55ef676e76e124a2bdd50ce115b6d9c1c7430294
-    restart: 'no'
-    command: >
-      NTFY_PASSWORD=${APP_PASSWORD} ntfy user add --role=admin umbrel
 
   app:
     image: binwiederhier/ntfy:v2.11.0@sha256:4a7d0f0adc6d5d9fc36e64ab55ef676e76e124a2bdd50ce115b6d9c1c7430294
-    depends_on:
-      - init
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
-    command: serve
+    entrypoint: [ "/bin/sh", "/entrypoint.sh" ]
     volumes:
+      - ${APP_DATA_DIR}/entrypoint.sh:/entrypoint.sh
       - ${APP_DATA_DIR}/data/cache:/var/cache/ntfy
       - ${APP_DATA_DIR}/data/data:/etc/ntfy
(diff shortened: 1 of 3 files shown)


nmfretz:
Thanks for submitting ntfy @sharknoon! I'll take a look at this tomorrow.

nmfretz:
This is a super fun app @sharknoon, I just played around with it. Excellent job packaging it; we'll start getting gallery images prepared.

I'm hoping to bounce a couple initial thought off you from testing:

### 1) Instant notifications for iOS users

Unfortunately it looks like instant notifications for iOS clients isn't as seemless as other platforms. Without additional configuration it can take hours for an iOS device to receive a notification:
https://docs.ntfy.sh/config/#ios-instant-notifications

To get around this, we can set `NTFY_UPSTREAM_BASE_URL` to `https://ntfy.sh` in the compose file. This forwards a poll request (not the actual notification) to the upstream ntfy.sh server, which sends the poll request to your iOS device so that it can download the notification from the self-hosted ntfy instance.

I'm a bit hesitant to force all users to use the centralized https://ntfy.sh server for poll requests, but the app becomes essentially unusable for iOS users without it. What are your thoughts here on including this by default vs providing instructions to add it manually for iOS users?

### 2) Exposing ntfy to the internet

For receiving notifications outside of a users local network, there are probably two easy ways to do this right now:
- One would be to install the [Tailscale app](https://apps.umbrel.com/app/tailscale) on your Umbrel and then just use the device's tailscale short-domain and port of the ntfy server in a client (e.g., `http://umbrel:13119` instead of `http://umbrel.local:13119`).

- The other way would be to install the [Cloudflare Tunnel app](https://apps.umbrel.com/app/cloudflared) on your Umbrel and create a tunnel to ntfy, and then use your custom domain in a client (e.g., `https://ntfy.mysupercooldomain.io`).

But either way there is a hiccup, which is that the environment variable ` NTFY_BASE_URL` must match what's in the client for two features:
1) instant notification on iOS 🤦‍♂️
2) features like sending attachments (since a download link is sent).

Of course, advanced users can go in and change the `NTFY_BASE_URL` in the compose file, but I'm wondering if we should try and make this "just work" for most users by detecting whether they have tailscale installed and if so, set `NTFY_BASE_URL` to their tailscale short-domain, and include instructions in the app description. I'm torn on this.

sharknoon:
Thank you @nmfretz for your comprehensive review 😀 Yes, this app is a bit tricky in many ways. I have spend a lot of hours to create the default user on startup and make it secure by default.

## 1) Instant notifications for iOS users
According to the [docs](https://docs.ntfy.sh/config/#config-options) the default value for `NTFY_UPSTREAM_BASE_URL` is already `https://ntfy.sh`, so we do not need to change that manually. I think we should provide the best possible expirience by default, but let the users know (in the description), that we are using this service, due to Apples restrictions.

This way, expert users can still change it if they want to.

## 2) Exposing ntfy to the internet
This is a tricky one. I don't think there is a good solution at the moment until Umbrel OS brings official support for domains. 
For now I also think checking for Tailscale is the right solution.

## Actions
- I will add a note in the description to explain the iOS instant notification situation and a short guide how to change the upstream base url.
- I will check for Tailscale in the `entrypoint.sh` and overwrite the `NTFY_BASE_URL` if Tailscale is detected.

## Questions
- How do I detect Tailscale? The `exports.sh` file does'nt export a env variable.

highghlow:
>     * How do I detect Tailscale? The `exports.sh` file does'nt export a env variable.

I guess you can try pinging of the tailscale's containers `tailscale_<name>_1`

nmfretz:
> I have spend a lot of hours to create the default user on startup and make it secure by default.

I noticed that with your custom entrypoint. Great work on that 👌 I was super impressed.

> According to the [docs](https://docs.ntfy.sh/config/#config-options) the default value for NTFY_UPSTREAM_BASE_URL is already https://ntfy.sh, so we do not need to change that manually.

Ah ya I saw that in that table as well, but it's actually a documentation mistake on their part. When testing I found that it actually isn't the case, and that the default is no upstream base url. We need to include `NTFY_UPSTREAM_BASE_URL: https://ntfy.sh` in the compose file ourselves for it to work on iOS devices (tested with my iPhone).

> I will add a note in the description to explain the iOS instant notification situation and a short guide how to change the upstream base url.

Perfect, thanks for doing that, it looks great 🙏.

> This is a tricky one. I don't think there is a good solution at the moment until Umbrel OS brings official support for domains. For now I also think checking for Tailscale is the right solution.

Ya, completely agree with you that there isn't a good solution until domains are supported. In fact, I'm kind of second guessing my suggestion to auto-detect Tailscale and change `NTFY_BASE_URL` because it may make it even more confusing for users.

Shall we just ship the app as-is for now and see how people do with it?

> How do I detect Tailscale? The exports.sh file doesn't export a env variable.

@sharknoon @highghlow here is our current script logic for apps. Anything in here will work on both pre 1.0 and post 1.0 umbrelOS: https://github.com/getumbrel/umbrel/blob/master/packages/umbreld/source/modules/apps/legacy-compat/app-script

So if we wanted to create a pre-start script that checks if Tailscale is installed we could do something like:

```bash
if "${UMBREL_ROOT}/scripts/app" ls-installed | grep --quiet 'tailscale'
do something

```


And we can use the script to pass anything to compose, so if the Tailscale container was up and running we could pass a command like this:

```bash
"${UMBREL_ROOT}/scripts/app" compose tailscale exec web <command>
```

But we'd probably have to include logic to wait until the Tailscale container is up and running so that we can query it for the short-domain name. For 99% of users the domain name would be `umbrel`, but if they have multiple Umbrels connected to their Tailscale then it may be different (e.g., `umbrel-1`).

nmfretz:
> I propose we leave the NTFY_BASE_URL as it is for now and revisit it as soon as domain support for Umbrel OS is out.

Sounds good! Okay, once the gallery assets are ready we will go live.

nmfretz:
We're going live with ntfy! Thanks again for the huge amount of work you put into this one @sharknoon.

Gallery assets are here: https://github.com/getumbrel/umbrel-apps-gallery/tree/master/ntfy

image

sharknoon:
You are welcome 😀 Thank for the corrections of the `umbrel-app.yml`, I totally forgot about downgrading the manifest version again after deciding to go with a custom entrypoint instead of hooks.

PR #1126, opened 2024-06-05
Update Penpot 2.0.1

JJeeff248:
Penpot has released v2 which has quite a lot of changes. https://github.com/penpot/penpot/releases/tag/2.0.0

I've updated the images in `docker-compose.yml` and I've updated the release notes + version in `umbrel-app.yml`. 

I tested the new container but I've been getting 502 errors which I can't resolve. The 502 error persists when reverting versions so I'm thinking I did something wrong when trying to test. The 502 occurs when a request to `http://umbrel.local:9001/api/rpc/command/get-profile` is made.

### DIFF
--- a/penpot/docker-compose.yml
+++ b/penpot/docker-compose.yml
@@ -1,4 +1,4 @@
-version: "3.8"
+version: "3.7"
 
 services:
   app_proxy:
--- a/penpot/docker-compose.yml
+++ b/penpot/docker-compose.yml
@@ -1,4 +1,4 @@
-version: "3.7"
+version: "3.8"
 
 services:
   app_proxy:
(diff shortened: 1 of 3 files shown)


nmfretz:
Thanks for this update @JJeeff248!  

I made a couple small changes, and added the `PENPOT_PUBLIC_URI` environment variable (https://github.com/getumbrel/umbrel-apps/pull/1126/commits/9c9a6084221502f69cbc59c2f031967e6fb32639). This fixes an issue in newer versions of Penpot where thumbnails need a custom URI to render thumbnails (if none is provided, then thumbnails are fetched in the browser from `http://localhost:9001/assets/...` which obviously won't work since Penpot won't be available at the localhost of the machine running the browser.

I have tested on arm64 and x86.

> I tested the new container but I've been getting 502 errors which I can't resolve. The 502 error persists when reverting versions so I'm thinking I did something wrong when trying to test. The 502 occurs when a request to http://umbrel.local:9001/api/rpc/command/get-profile is made.

Ah darn, ya that sounds like an accidental issue during testing. I can help you troubleshoot and get things back up and running. Were you testing on umbrelOS 0.5.x or 1.x?

JJeeff248:
> Thanks for this update @JJeeff248!

All good! Thank you for fixing and merging it :)

> Ah darn, ya that sounds like an accidental issue during testing. I can help you troubleshoot and get things back up and running. Were you testing on umbrelOS 0.5.x or 1.x?

I was testing on umbrelOS 1.x. It appears only penpot is affected. Any help would be amazing. Happy to move somewhere else if this isn't the right place for help.

PR #1127, opened 2024-06-05
update n8n from 1.18.1 to 1.44.1

joaovictor-local:
Tested in Umbrel OS 1.1.2 on Beelink EQ12 N100


nmfretz:
Perfect, thanks for this update @joaovictor-local. And thanks for catching the `N8N_SECURE_COOKIE` breaking change.

For my future reference, here is a nice live list of breaking changes for n8n:
https://github.com/n8n-io/n8n/blob/master/packages/cli/BREAKING-CHANGES.md

PR #1130, opened 2024-06-07
Update Immich to v1.105.1

nmfretz:
1.103.0 --> 1.105.1

Tested on x86 and arm64


mi875:
> thanks for this! would it be possible to bump it to at least v1.106? the latest ios update is not working with v1.105.x anymore…

I have the same problem on the Android app too.

PR #1141, opened 2024-06-11
Update prowlarr to 1.18.0

fabricionaweb:
Tested with 1.1.2 amd64


nmfretz:
Wonderful. Thanks for this update @fabricionaweb! And double thanks for confirming that the media-app-configurator is still working with this Prowlarr update 🙌

PR #1142, opened 2024-06-11
Update sonarr to 4.0.5

fabricionaweb:
Tested with 1.1.2 amd64

Sonarr v3 is no longer active, its recommended to update to v4.


nmfretz:
Thanks for updating Sonnar to v4 @fabricionaweb, and for upgrading the mac container at the same time!

I have [changed the `APP_URL`](https://github.com/getumbrel/umbrel-apps/pull/1142/commits/47af785f110270d06ada4b20e94b6c5bb1bf7cc8) env var to
```
APP_URL: "http://sonarr_server_1:8989"
``` 

So that it is using the sonarr container's port instead of prowlarr's. It's an easy one to miss because if you tested by updating the sonarr app (as opposed to testing a fresh install of the app) it would look like the mac container setup worked (even though it actually failed) because Transmission would have already been set up in Sonarr from the initial install.

PR #1143, opened 2024-06-11
Update radarr to to 5.6.0

fabricionaweb:
Tested with 1.1.2 amd64


nmfretz:
Wonderful, thanks for this update and for checking that the media-app-configurator is compatible 🙌.

Tested on fresh install and update. Going live.

PR #1153, opened 2024-06-12
Update sabnzbd to 4.3.2

fabricionaweb:
If a config file exists on first run, a merged will occour and values there will be used. We can use this to change download location and also add hosts whitelist.

I tested on 1.1.2 amd64 and it worked.

It would be cool to have the `mac` to add it to sonarr/radarr/prowlarr etc, maybe we can think on that later.

We could just use static value and create the config/sabnzbd.ini, but I thought the envs can change, so Im not sure. Feedback is appreciated.

### DIFF
--- a/sabnzbd/docker-compose.yml
+++ b/sabnzbd/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3.7'
+version: "3.7"
 
 services:
   app_proxy:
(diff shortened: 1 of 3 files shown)


nmfretz:
Thanks for this update PR @fabricionaweb! I will review this early next week. 

> It would be cool to have the mac to add it to sonarr/radarr/prowlarr etc, maybe we can think on that later.

Great idea. I will do that at the same time that I update the media-app-configurator for Prowlarr FlareSolverr:
https://github.com/getumbrel/umbrel-apps/pull/1042

nmfretz:
working on the configurator now.

nmfretz:
Thanks again for this @fabricionaweb. I have opened a PR here to update all Servarr apps and download clients https://github.com/getumbrel/umbrel-apps/pull/1194

I have added a post-install script that:
- restarts any Servarr apps (e.g., Radarr, Sonarr, etc.) after SABnzbd is installed so that the Servarr app runs the auto-configuration from our media-app-configurator. 
- The download directories are now set correctly at install without needing to modify them through the setup wizard
- The sabnzbd container name (`sabnzbd_web_1`) has been added to `host_whitelist` to allow connections coming in from Servarr apps.
- Categories are set up for Prowlarr (`prowlarr-fallback`), Radarr (`movies`), Sonarr (`shows`), Readarr (`books`), and Lidarr (`music`) so that the media-app-configurator can connect Servarr apps using these categories (Connecting to SABnzbd requires that a category set in a Servarr app is an exact match to a category set in SABnzbd).
- User's can make changes to the configurations after install and we won't overwrite their changes.

PR #1154, opened 2024-06-12
updated to latest databag container

balzack:
- added support for cloudflare turn service for audio and video calls as an alternative to self-hosting coturn
- added support for browser push notifications. existing session require log (out/in) to enable


nmfretz:
Thanks for this update @balzack! I have made a small edit to the release notes so that they render nicely in the app store.

image

Going live.

balzack:
Thanks for you help!

PR #1155, opened 2024-06-12
RTL v0.15.2 update on Core Lightning

ShahanaFarooqui:


ShahanaFarooqui:
- The latest RTL release requires nodejs >=v18.19.0. How can I confirm the nodejs version on Umbrel?

- RTL is now supporting block explorer with configurable BLOCK_EXPLORER_URL variable. It requires:
    1: mempool url and
    2: `Bitcoin network` value to append with block explorer url (eg. mempool.space/testnet).
	- Is there a way to configure the explorer url with local mempool url (if it is already installed on Umbrel) otherwise fallback to mempool.space url?
    - How can I capture the Bitcoin Network value?

nmfretz:
Thanks for this update @ShahanaFarooqui! I've responded to your other RTL-LND update here: https://github.com/getumbrel/umbrel-apps/pull/1156. Let's discuss there and then copy the changes here once we have finalized.

ShahanaFarooqui:
@nmfretz Hey, just pushed a commit with v0.15.2 update.

nmfretz:
Thanks @ShahanaFarooqui, responded here https://github.com/getumbrel/umbrel-apps/pull/1156#issuecomment-2191729658

nmfretz:
Hey @ShahanaFarooqui, please ping me whenever you are ready with the config fix for testing.

ShahanaFarooqui:
@nmfretz Hey, pushed the final update for core-lightning and core-lightning-rtl in this PR. Let me know if you would prefer separate PRs for them. Below are the highlights of the latest update:

- core-lightning needs to be available with clnrest options for RTL v0.15.0 and above. So I updated it to v24.05.

- Add clnrest host and port details for core-lightning and exposed the clnrest-port.

- Updated RTL to use clnrest NO c-lightning-REST anymore. Which required lnServerUrl to be updated with clnrest's url & MACAROON_PATH is replaced with RUNE_PATH.

- Also added default RTL-Config.json and entrypoint.sh for core-lightning-rtl as discussed in our last meeting.

But I found two strange behaviours in the setup:

- Due to a bug in RTL, LN_IMPLEMENTATION: CLN should not be working (https://github.com/Ride-The-Lightning/RTL/issues/1429). I am not sure how the old setup for CLN was working but I added lnImplementation: "CLN" for the temporary fix.

- I also had to create .commando-env file with rune value manually in the container even after mounting it with "${CORE_LIGHTNING_PATH}:/root/.lightning"

Can you please check if you are also unable to mount the `.commando-env` file?

nmfretz:
Hey @ShahanaFarooqui, thanks for the fixes and changes!

I have pushed a few commits, and I think we're nearly there:
- Removed the entrypoint for password migration https://github.com/getumbrel/umbrel-apps/pull/1155/commits/35a644c0297b579d811614c875cfd4aec4ec1027. This is only needed for the RTL-LND app because that app was being used by people prior to our current password creation scheme. RTL-CLN will work fine without this entrypoint.
- Added an `exports.sh` to export the `BLOCK_EXPLORER_URL` based on what Bitcoin network a user is running on. Like RTL-LND we can use a local mempool instance down the line https://github.com/getumbrel/umbrel-apps/pull/1156#issuecomment-2191729658.
- Changed the bind mount on the host for the Core Lightning data directory https://github.com/getumbrel/umbrel-apps/pull/1155/commits/cbc32b3b9cdba880829fe06c4e3d45579dab290c. The reason for the `.commando-env` file issues is that the bind mount on the host was set to `${CORE_LIGHTNING_PATH}` which is actually a variable for `/root/.lightning` that we use inside the container:
  > I also had to create .commando-env file with rune value manually in the container even after mounting it with "${CORE_LIGHTNING_PATH}:/root/.lightning"
- Added a hook that copies the default RTL-Config.json for legacy users who are updating their app. This file will be copied to a user's local data directory on app install, but not on app update, so the hook handles this.

Can you please take a look at 2 things:
1. Can you please check the pre-start hook to make sure I have added the default RTL-Config.json correctly.
2. When I visit "Node Config" in the UI, I no longer see "Core Lightning Config" at the top nav bar:

image

The logs for the RTL container don't show any irregularities, but I do see a rendering error in the console:

image

ShahanaFarooqui:
> 1. Can you please check the pre-start hook to make sure I have added the default RTL-Config.json correctly.

Looks perfect. Thanks.


> 2. When I visit "Node Config" in the UI, I no longer see "Core Lightning Config" at the top nav bar:

It should not be visible because we do not have CLN config file in Umbrel setup. All of our CLN config options are passed via [command-line](https://github.com/getumbrel/umbrel-apps/blob/2a69c7d83eb0708ac58ba9daf94b371165c6fa02/core-lightning/docker-compose.yml#L63). But I have opened another [issue](https://github.com/Ride-The-Lightning/RTL/issues/1424) on RTL to show CLN configuration via `listConfigs` rpc rather than config file itself. This will allow the UI to show all config options passed/over-ridden via cli, config, plugins etc.

> The logs for the RTL container don't show any irregularities, but I do see a rendering error in the console:

It is a known issue generated by graphs but it is not related to config file and does not impact anything significant.

nmfretz:
Wonderful, thanks very much @ShahanaFarooqui. I have tested on arm64 and amd64 with both fresh installs and app updates. Going live ⚡️

PR #1159, opened 2024-06-16
update lnd image to v0.18.0-beta

zapomatic:
Couldn't find any reference to anyone testing or merging this on the umbrel team. Is this update already in progress somewhere?


nmfretz:
Thanks for the LND update @zapomatic!  

I have checked the LND release notes for any breaking changes that need to be handled in our UI and there are none: https://github.com/lightningnetwork/lnd/releases

I have updated the app manifest (`umbrel-app.yml`) with the new version so that users receive an update notification on their umbrels. I also added release notes, including a warning that users running with "wtclient.active = true" cannot downgrade from this version.

Tested on arm64 and x86

PR #1161, opened 2024-06-18
App Submission: Samba

joaovictor-local:
I am already using it through a community store (https://github.com/joaovictor-local/umbrel-samba-app-store) but it requires me to change the password manually editing the files using SSH, before that I was using Samba directly installed on the host but it was deleted by a system update so I had to figure out how to eliminate this problem.

<s>I think it need at least to generate a random password for each installation, maybe a script to generate the password and insert it at the `/data`  files and the `umbrel-app.yml` </s>  :heavy_check_mark: our allow the user to sign in using his Umbrel password (I not sure if it's possible on Linux)

What do you guys think about the idea of this app? A web app to edit the config file and password would be cool too and I think I could code it. If we it would be possible to use the same password we would just need a better `index.html`.

Icon: https://github.com/getumbrel/umbrel-apps-gallery/pull/47


samba/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    container_name: samba_homepage_app_proxy_1
+
+  apache:
+    image: >-
+      httpd:2.4.59@sha256:10182d88d7fbc5161ae0f6f758cba7adc56d4aae2dc950e51d72c0cf68967cea
+    container_name: samba_homepage_1
+    restart: on-failure
+    volumes:
+      - ${APP_DATA_DIR}/data:/usr/local/apache2/htdocs
+      - ${APP_DATA_DIR}/data/httpd.conf:/usr/local/apache2/conf/httpd.conf
```
sharknoon, line 17: For just hosting a static web page, Apache httpd seems a bit heavy to me. Maybe consider alternatives like [Static Web Server](https://static-web-server.net/).
The image size of `httpd` is 220MB, while the image size of `joseluisq/static-web-server` is 15MB, therefore I think it it is really worth changeing it.

samba/umbrel-app.yml
```diff
@@ -0,0 +1,26 @@
...
+port: 9445
+description: >-
+  Samba is the standard Windows interoperability suite of programs for Linux and Unix.
+  Samba is Free Software licensed under the GNU General Public License, the Samba project is a member of the Software Freedom Conservancy.
+  Since 1992, Samba has provided secure, stable and fast file and print services for all clients using the SMB/CIFS protocol, such as all versions of DOS and Windows, OS/2, Linux and many others.
+  Samba is an important component to seamlessly integrate Linux/Unix Servers and Desktops into Active Directory environments. It can function both as a domain controller or as a regular domain member. 
+
+developer: Samba
+website: https://www.samba.org/
+submitter: João Victor
```
sharknoon, line 16: Add the permission to access the umbrelOS Downloads folder
```suggestion
submitter: João Victor
permissions:
  - STORAGE_DOWNLOADS
```

sharknoon (changes requested):
Hi @joaovictor-local,

thank you so much for your awesome app submission, I am really impressed! I also think this app is needed as umbrelOS currently has no app to export its storage via samba.

I have left a couple of suggestions below. If you have any questions, feel free to ask me anytime.

Note: I haven't tested the app yet, but I will do that as soon as you have updated this PR 😀

Thanks again for your awesome work, I can't wait to see your changes!

joaovictor-local:
> Hi @joaovictor-local,
> 
> thank you so much for your awesome app submission, I am really impressed! I also think this app is needed as umbrelOS currently has no app to export its storage via samba.
> 
> I have left a couple of suggestions below. If you have any questions, feel free to ask me anytime.
> 
> Note: I haven't tested the app yet, but I will do that as soon as you have updated this PR 😀
> 
> Thanks again for your awesome work, I can't wait to see your changes!

Hi @sharknoon!
I appreciate! Thank you for you review :smile:  I applied all your requests.

image

samba/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: samba_homepage_1
+      APP_PORT: 9445
+
+  homepage:
+    image: joseluisq/static-web-server:2@sha256:c8345c2c469ba249cbc986b6155dc275ce86bcd30e0b29755fda4deb501bbddd
+    container_name: samba_homepage_1
```
sharknoon, line 11: We can omit the container_name here
```suggestion
```

samba/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+      - SERVER_ROOT=/var/public
+      - SERVER_PORT=9445
+    volumes:
+      - ${APP_DATA_DIR}/data/public:/var/public
+      - ${APP_DATA_DIR}/data/smb.conf:/var/public/smb.conf
+
+  server:
+    image: >-
+      dockurr/samba:4.20.1@sha256:9a57887c0be1242ee2e1381653619a656820e1aee8e271885662d04c2c367be8
+    container_name: samba_server_1
```
sharknoon, line 22: We can also omit the container_name here
```suggestion
```

samba/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+    environment:
+      USER: umbrel
+      PASS: ${APP_PASSWORD}
```
sharknoon, line 25: To reduce the permissions, please change the environment variables to this
```suggestion
      PASS: ${APP_PASSWORD}
      UID: 1000
      GID: 1000
```
[Here](https://github.com/dockur/samba?tab=readme-ov-file#how-do-i-modify-the-permissions) is the explanation for that.

samba/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+    environment:
+      - SERVER_ROOT=/var/public
+      - SERVER_PORT=9445
+    volumes:
+      - ${APP_DATA_DIR}/data/public:/var/public
+      - ${APP_DATA_DIR}/data/smb.conf:/var/public/smb.conf
+
+  server:
+    image: >-
+      dockurr/samba:4.20.1@sha256:9a57887c0be1242ee2e1381653619a656820e1aee8e271885662d04c2c367be8
```
sharknoon, line 21: To reduce the permissions, set the user and group to 1000, also the image can be on one line
```suggestion
    image: dockurr/samba:4.20.1@sha256:9a57887c0be1242ee2e1381653619a656820e1aee8e271885662d04c2c367be8
    user: "1000:1000"
```

samba/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+version: '3.7'
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: samba_homepage_1
+      APP_PORT: 9445
```
sharknoon, line 7: I know it is confusing sometimes, but the `APP_PORT` does not need to be the same as the port in the `umbrel-app.yml`. Therefore I suggest changing this to the default value of the `static-web-server` to get rid of one environment variable
```suggestion
      APP_PORT: 80
```

samba/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+
+  homepage:
+    image: joseluisq/static-web-server:2@sha256:c8345c2c469ba249cbc986b6155dc275ce86bcd30e0b29755fda4deb501bbddd
+    container_name: samba_homepage_1
+    environment:
+      - SERVER_ROOT=/var/public
+      - SERVER_PORT=9445
```
sharknoon, line 14: This can be removed because of the default value of `80`
```suggestion
```

samba/data/smb.conf
```diff
@@ -0,0 +1,19 @@
...
+
+[Data]
+        path = /storage
+        comment = Shared
+        valid users = @smb
+        browseable = yes
+        writable = yes
+        read only = no
+        force user = root
+        force group = root
```
sharknoon, line 19: I am not an expert in samba, but with the reduced permissions to user 1000 and group 1000 we probably need to change it here too (or remove it completly)

samba/data/public/index.html
```diff
@@ -0,0 +1,352 @@
...
+        body.iframe .app-container {
+            display: none;
+        }
+
+        body.iframe .container.terminal-container {
+            border-radius: 0;
+            padding: 0;
+            margin: 0;
+        }
+    </style>
```
sharknoon, line 294: Some of the classes are not needed anymore, because we don't use them for example for the terminal. Please remove all non-needed classes

samba/data/public/index.html
```diff
@@ -0,0 +1,352 @@
...
+            <hr />
+        <h2>How to Connect to the Umbrel SMB Server</h1>
+            <p>Follow these steps to connect to the Umbrel SMB Server:</p>
+
+            <h3>For Windows Users</h3>
+            <ol>
+                <li>Open <strong>File Explorer</strong>.</li>
+                <li>In the address bar, type <code>\\umbrel.local</code> and press <strong>Enter</strong>.</li>
+                <li>When prompted, enter the username: <code>umbrel</code> and the password: <code>changeme</code>.</li>
+                <li>You should now see your <code>/storage</code> directory.</li>
```
sharknoon, line 317: The Password is the default generated one by umbrelOS. Please adjust it. Also the user doen't know that the folder internally is called `/storage`, maybe we could use something like `You should now see all of your umbrelOS data like downloads, movies, music, etc...`

samba/data/public/index.html
```diff
@@ -0,0 +1,352 @@
...
+            <h3>For macOS Users</h3>
+            <ol>
+                <li>Open <strong>Finder</strong>.</li>
+                <li>In the menu bar, click on <strong>Go</strong> and then select <strong>Connect to Server...</strong>.
+                </li>
+                <li>In the Server Address field, type <code>smb://umbrel.local</code> and click
+                    <strong>Connect</strong>.
+                </li>
+                <li>When prompted, enter the username: <code>umbrel</code> and the password: <code>changeme</code>.</li>
+                <li>You should now see your <code>/storage</code> directory.</li>
```
sharknoon, line 329: Same as above

samba/data/public/index.html
```diff
@@ -0,0 +1,352 @@
...
+            </ol>
+
+            <h3>For Linux Users</h3>
+            <ol>
+                <li>Open a file manager (such as <strong>Nautilus</strong> or <strong>Dolphin</strong>).</li>
+                <li>In the address bar, type <code>smb://umbrel.local</code> and press <strong>Enter</strong>.</li>
+                <li>When prompted, enter the username: <code>umbrel</code> and the password: <code>changeme</code>.</li>
+                <li>You should now see your <code>/storage</code> directory.</li>
```
sharknoon, line 337: Same as above

samba/data/public/index.html
```diff
@@ -0,0 +1,352 @@
...
+
+            <p>If you encounter any issues, make sure that the Umbrel SMB Server is running and that you have the
+                correct
+                permissions to access it. Additionally, ensure that your firewall or antivirus software is not blocking
+                the
+                connection.</p>
+
+            <p>For more detailed configuration information, <a href="smb.conf">click here to check your config
+                    file</a>.</p>
+            </p>
```
sharknoon, line 348: Maybe we can extend this section to tell users, how to change the config. They can change the config by opening the terminal (settings -> advanced settings -> terminal) and manually modifying the smb file (e.g. with `nano umbrel/app-data/samba/data/smb.conf`)

samba/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+version: '3.7'
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: samba_homepage_1
+      APP_PORT: 9445
+
+  homepage:
+    image: joseluisq/static-web-server:2@sha256:c8345c2c469ba249cbc986b6155dc275ce86bcd30e0b29755fda4deb501bbddd
```
sharknoon, line 10: Maybe try to reduce the permissions to a lower user to improve sandboxing of the container
```suggestion
    image: joseluisq/static-web-server:2@sha256:c8345c2c469ba249cbc986b6155dc275ce86bcd30e0b29755fda4deb501bbddd
    user: "1000:1000"
```

sharknoon (changes requested):
Wow, awesome work! You are really fast!

I have tested it on my umbrel and it worked flawlessly. I think I will keep this app on my personal umbrel :)

I have made some suggestions below. It can happen, that the user change (`user: "1000:1000"`) in the docker compose file will not work. Please test if it workd with less permissions and let me know.

Thank you again :)

samba/docker-compose.yml
sharknoon: Also, please change the tag of `2` to `2.32.0`
```suggestion
    image: joseluisq/static-web-server:2.32.0@sha256:c8345c2c469ba249cbc986b6155dc275ce86bcd30e0b29755fda4deb501bbddd
```

samba/docker-compose.yml
joaovictor-local: Got it! I had this impression when checking other apps, the external port that the user will access is defined only on umbrel-app.yml.

joaovictor-local:
> Wow, awesome work! You are really fast!
> 
> I have tested it on my umbrel and it worked flawlessly. I think I will keep this app on my personal umbrel :)
> 
> I have made some suggestions below. It can happen, that the user change (`user: "1000:1000"`) in the docker compose file will not work. Please test if it workd with less permissions and let me know.
> 
> Thank you again :)

I am excited to see it available in the store :smile:
It worked fine with less permission for the Static Web Server too :+1:

samba/data/public/index.html
```diff
@@ -0,0 +1,333 @@
...
+        </ol>
+
+        <p>If you encounter any issues, make sure that the Umbrel SMB Server is running and that you have the
+            correct
+            permissions to access it. Additionally, ensure that your firewall or antivirus software is not blocking
+            the
+            connection.</p>
+
+        <p>For more detailed configuration information, <a href="smb.conf">click here to check your config file</a>.
+            can change the config by opening the terminal (Settings -> Advanced Settings -> Terminal) and manually
```
sharknoon, line 326:
```suggestion
           You  can change the config by opening the terminal (Settings -> Advanced Settings -> Terminal) and manually
```

samba/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    user: "1000:1000"
+    environment:
+      - SERVER_ROOT=/var/public
+    volumes:
+      - ${APP_DATA_DIR}/data/public:/var/public
+      - ${APP_DATA_DIR}/data/smb.conf:/var/public/smb.conf
+
+  server:
+    image: dockurr/samba:4.20.1@sha256:9a57887c0be1242ee2e1381653619a656820e1aee8e271885662d04c2c367be8
+    restart: on-failure
```
sharknoon, line 21: The user key is missing
```suggestion
    restart: on-failure
    user: "1000:1000"
```

sharknoon (changes requested):
Really good work @joaovictor-local! I have found two small issues. After that, I think we are good to go 😀 It will take some time, until the app is in the app store, because this app gets custom made gallery images.

samba/docker-compose.yml
joaovictor-local: It only worked for the Static Web Server, when I tried to do the same on the samba server it didn't started.
```
samba_server_1    | Failed to create user umbrel
samba_server_1    | adduser: permission denied (are you root?)
samba_server_1    | Failed to create user umbrel
```
https://github.com/getumbrel/umbrel-apps/pull/1161#discussion_r1646095174 the environment variable seems to be the only way (samba/docker-compose.yml:25-25).

joaovictor-local: https://github.com/dockur/samba/blob/master/samba.sh
I just read this file, the script need root access to add the user to the samba group and if the user don´t exist it will be created. It seems to be a good idea to have force group and force group, since the script will override it with the umbrel user and group.

sharknoon (approved):
Awesome work, looks good to me 🎉

@nmfretz can you have a second look if everything looks okay from your side?

Thank you @joaovictor-local for submitting such a great and (in my opinion) much needed app

joaovictor-local:
@nmfretz @lukechilds @mayankchhabra @smolgrrr  :eyes:

nmfretz:
Hey @joaovictor-local, sorry for the delay on my end. I will look at this tomorrow!

nmfretz:
Wonderful addition @joaovictor-local! And fantastic review @sharknoon. I'm super impressed with both of you here 🤝

@joaovictor-local this is working really well. We'll start getting gallery assets together so this can go live!

In the meantime, there are a couple things you may want to do to make app updates easier for you:

### 1) Include the `public` assets (favicon/index.html/samba.svg) within the `homepage` Docker image itself.

Ideally, the best thing to do is have your frontend code be part of the Docker image so that frontend app updates are as easy as you rebuilding the image and changing the `image` in the compose file. 

The way umbrelOS currently works is that on app install, everything from the app's repo is copied over to a users umbrel data. But on app updates only a whitelist of files is copied over so that user data is not overwritten:
https://github.com/getumbrel/umbrel/blob/49d37581c8233f808fc5a79c2cd38810c6cd1b0b/packages/umbreld/source/modules/apps/legacy-compat/app-script#L594-L595

So right now, if you made changes to your UI (e.g., changed the instructions in the html), only new installs would receive the updates. Existing installs that update won't get the changes. You could do some hacky things like use a hook that runs on app update and changes certain files, but it's not ideal because it will add complex overhead for you.

Instead, what you can do is just build your own image from static-web-server that includes your public folder and then push it to DockerHub. 

Something as simple as this will likely work for a Dockerfile:
```Dockerfile
# Use the joseluisq/static-web-server as the base image
# You'll want to update this tag to the latest version each time you build
FROM joseluisq/static-web-server:2.32.0

# Copy your entire public directory to the /var/public directory which is set by the SERVER_ROOT env var
COPY public /var/public
```

You'll then need to make sure to build a multi-architecture image for amd64 and arm64. You'll run something like this:
```
docker buildx build --platform linux/amd64,linux/arm64 -t your-docker-hub-username/static-web-server-samba:version-tag .
```

### 2) Can the Samba container create it's own smb.conf and we still bind to it?

See if it's possible to let the Samba container create it's own smb.conf on initial startup, still have us bind it, but also allow us to set the things we need to change via environment variables (e.g., `force user`).

I suspect this won't be possible, so please don't do too much digging. The thing I'm trying to avoid here is again related to app updates not bringing down changes we may need to make to this file at some point down the road.

I'm going to make a couple small changes to the compose file and push them.

nmfretz:
Alrighty @joaovictor-local I made two main changes:

1) I changed the smb.conf bind mount path on the host to `${APP_DATA_DIR}/data/samba/smb.conf` https://github.com/getumbrel/umbrel-apps/pull/1161/commits/541476aefb242163dfba7eab5c45ab4bd8f41539

2) I changed the SMB port from 445 to 446. 
    This is needed because the [Back That Mac Up app](https://github.com/getumbrel/umbrel-apps/blob/master/back-that-mac-up/docker-compose.yml) is already using 445. If a user goes to install both apps then the second one that is being installed will fail because port 445 will already be in use.
    
    This means that the instructions in the `index.html` will need to change to reference `smb://umbrel.local:446`

samba/umbrel-app.yml
```diff
@@ -0,0 +1,33 @@
...
+
+
+  Samba is Free Software licensed under the GNU General Public License, the Samba project is a member of the Software Freedom Conservancy.
+
+
+  Since 1992, Samba has provided secure, stable and fast file and print services for all clients using the SMB/CIFS protocol, such as all versions of DOS and Windows, OS/2, Linux and many others.
+
+
+  Samba is an important component to seamlessly integrate Linux/Unix Servers and Desktops into Active Directory environments. It can function both as a domain controller or as a regular domain member.
+
```
sharknoon, line 19: This is the standard text from the website. I have enhanced the text to be more end user friendly. Feel free to make additional changes if you like:
```suggestion
📂 Simplify your file sharing with Samba!

Samba is a powerful tool that makes it easy to share files across different devices on your network. It allows your Umbrel device to seamlessly share files with Windows, macOS, Linux, and more. Whether you're sharing documents, media files, or backups, Samba makes it simple and secure.


📄 Effortless File Access

Access and manage the files on your Umbrel device from any computer on your network. Say good bye to the complicated SSH access!


🖥️ Cross-Platform Compatibility

Samba works with all major operating systems, including Windows, macOS, and Linux. Share your files without worrying about compatibility issues.


🛡️ Secure and Reliable

Samba has been providing secure and stable file sharing for over 30 years. Trust it to keep your files safe and accessible.
```

samba/data/samba/smb.conf
sharknoon, line 1: I have checked the default `smb.conf` in the container and it seems like the only difference is the `force user` and `force group` keys. Both of which can set by environment variables. The only reason I can think of to keep this file then in the repository, is to allow advanced users to edit this file. I am not sure, maybe we can remove this file altogether and tell advanced users to modify the `docker-compose.yml` directly.

samba/data/public/index.html
```diff
@@ -0,0 +1,333 @@
...
+        </div>
+        <p class="text-muted">
+            <hr />
+        <h2>How to Connect to the Umbrel SMB Server</h2>
+        <p>Follow these steps to connect to the Umbrel SMB Server:</p>
+
+        <h3>For Windows Users</h3>
+        <ol>
+            <li>Open <strong>File Explorer</strong>.</li>
+            <li>In the address bar, type <code>\\umbrel.local</code> and press <strong>Enter</strong>.</li>
```
sharknoon, line 285:
```suggestion
            <li>In the address bar, type <code>\\umbrel.local:446</code> and press <strong>Enter</strong>.</li>
```

samba/data/public/index.html
```diff
@@ -0,0 +1,333 @@
...
+            </li>
+            <li>You should now see all of your umbrelOS data like downloads, movies, music, etc...</li>
+        </ol>
+
+        <h3>For macOS Users</h3>
+        <ol>
+            <li>Open <strong>Finder</strong>.</li>
+            <li>In the menu bar, click on <strong>Go</strong> and then select <strong>Connect to Server...</strong>.
+            </li>
+            <li>In the Server Address field, type <code>smb://umbrel.local</code> and click
```
sharknoon, line 298:
```suggestion
            <li>In the Server Address field, type <code>smb://umbrel.local:446</code> and click
```

samba/data/public/index.html
```diff
@@ -0,0 +1,333 @@
...
+                default
+                credentials).
+            </li>
+            <li>You should now see all of your umbrelOS data like downloads, movies, music, etc...</li>
+        </ol>
+
+        <h3>For Linux Users</h3>
+        <ol>
+            <li>Open a file manager (such as <strong>Nautilus</strong> or <strong>Dolphin</strong>).</li>
+            <li>In the address bar, type <code>smb://umbrel.local</code> and press <strong>Enter</strong>.</li>
```
sharknoon, line 311:
```suggestion
            <li>In the address bar, type <code>smb://umbrel.local:446</code> and press <strong>Enter</strong>.</li>
```

samba/data/public/index.html
```diff
@@ -0,0 +1,333 @@
...
+        <p>If you encounter any issues, make sure that the Umbrel SMB Server is running and that you have the
+            correct
+            permissions to access it. Additionally, ensure that your firewall or antivirus software is not blocking
+            the
+            connection.</p>
+
+        <p>For more detailed configuration information, <a href="smb.conf">click here to check your config file</a>.
+            You can change the config by opening the terminal (Settings -> Advanced Settings -> Terminal) and manually
+            modifying the smb file (e.g. with nano umbrel/app-data/samba/data/smb.conf).
+        </p>
```
sharknoon, line 328: If we decide to leave out the `smb.conf` altogether, we should change this description and tell users to edit the `docker-compose.yml` directly in order to add a custom `smb.conf` file.

sharknoon (changes requested):

sharknoon:
@joaovictor-local 👀

joaovictor-local:
> @joaovictor-local 👀

I will be doing it today yet (UTC-3).
🙂

samba/data/samba/smb.conf
joaovictor-local: Let's remove it for now and when we had a good UI to edit it we can add it back.

joaovictor-local:
@sharknoon the only problem with the `docker-compose.yml` edit it that it will be erased after a update and the user may forget that we need to put it back manually. I have changed the `force user` and `force group` keys just for informative proposal because it will always use the values in the environment variables.

sharknoon:
@joaovictor-local thank you for your commits 😀 Yes, you are right, changes to the `docker-compose.yml` will be erased after an update.

### `smb.conf`
I am thinking do we need to have a `smb.conf` exposed to the user? What are the use cases for a user to edit this file? I am not an expert in Samba, maybe you can clarify it a bit for me, thanks! 😊

### Container Image
Would it be possible for you to pack the UI part inside a container as @nmfretz suggested it? That way it is easier to update and maintain. We would also have a clear line of responsibilities (everything feature/app releated inside the container, everything hosting/umbrel related in the data directory/docker compose).

### Updated Port
Other than that we had to change to port from the standard `445` to `446`, because that port was already allocated by `Back That Mac Up`. Therefore we also need to change the port in the UI.

Thank you so much for the great work you are doing and I can't wait to see this app on the app store!

joaovictor-local:
> @joaovictor-local thank you for your commits 😀 Yes, you are right, changes to the `docker-compose.yml` will be erased after an update.
> ### `smb.conf`
> 
> I am thinking do we need to have a `smb.conf` exposed to the user? What are the use cases for a user to edit this file? I am not an expert in Samba, maybe you can clarify it a bit for me, thanks! 😊
> ### Container Image
> 
> Would it be possible for you to pack the UI part inside a container as @nmfretz suggested it? That way it is easier to update and maintain. We would also have a clear line of responsibilities (everything feature/app releated inside the container, everything hosting/umbrel related in the data directory/docker compose).
> ### Updated Port
> 
> Other than that we had to change to port from the standard `445` to `446`, because that port was already allocated by `Back That Mac Up`. Therefore we also need to change the port in the UI.
> 
> Thank you so much for the great work you are doing and I can't wait to see this app on the app store!

1. You may want to change it to make it faster.
https://www.reddit.com/r/OpenMediaVault/comments/11gwi1g/significant_samba_speedperformance_improvement_by/

2. I was working on it but I had to take a break. Take a lot in the last commit and also: https://github.com/joaovictor-local/static-web-server-samba and https://hub.docker.com/repository/docker/loficafe/static-web-server-samba/general.

3. @nmfretz change it for me :)

sharknoon:
Hey @joaovictor-local

I am sorry to answer so late, I haven't forgotten your PR.

Great work on the docker image, it looks awesome! The Ports in the UI are also changed 😃 

It looks good to me 👍

nmfretz:
Thanks for the ping @joaovictor-local and very sorry for the delay on my end while I was working on other things. We'll get gallery assets made up and then send this live to the app store. Really really great work!

sharknoon (approved):

loralb:
Hello, sorry for the ping, but how long do gallery assets usually take to generate?

nmfretz:
@joaovictor-local gallery assets are now live! 
https://github.com/getumbrel/umbrel-apps-gallery/tree/master/samba 

Thanks so much for your patience and for your excellent work in bringing Samba to the app store. Going live 🎉

image

a4004:
Hey, just came across your project due to an issue raised in the Discord @joaovictor-local 

You may want to consider using the default SMB port 445 instead of using 446 because unfortunately for some bizzare reason, Windows is unable to connect to the fileshare when using non-standard SMB/CIFS port numbers. This has been the case for one Umbrel user today, and I've experienced the same issue when testing.

I know this might cause a port conflict with the Back That Mac Up app so that could be a non-solution, but something to consider.

It is a known issue with Windows itself, dating back to Windows 7 as far as I could see so maybe changing the instructions at the very least, to suggesting setting up something like [CIFS over SSH](https://jjkeijser.github.io/cifs-over-ssh/) or the answer from the user *Hashbron* on [this thread](https://superuser.com/questions/702948/how-to-mount-a-samba-share-on-non-standard-port),

All in all nice app! 🚀

sharknoon:
@joaovictor-local Could you update your image `loficafe/static-web-server-samba` to change the usage instructions for the windows version? As @a4004 pointed out, the instructions don't work on windows. Maybe you could replace the windows description with the suggestions by @a4004.

As soon as the new umbrelOS app framework will be released with proper handling for port conflicts, this is unfortunately the only solution we can give.

DJMo13:
Hello, would it be possible to create an incomparability tab in the app store and make back-my-mac-up and samba run on the same port 445?

PR #1168, opened 2024-06-22
" and end with "

Ahmed262111:
". For example, if the changes include adding a new feature, the title could be "

Add new feature

".

The description of the pull request should provide more details about the changes made, but still be concise. It should also start with "

" and end with "


nmfretz:
Hi @Ahmed262111, this looks like an accidental PR that you've opened into the add-widgets branch? I'm closing this, but please let me know if you were trying to do something else.

PR #1170, opened 2024-06-23
Update PeerSwap Web UI to v1.5.7

Impa10r:
Adds channel Auto Fee functionality plus various enhancements.

image

### DIFF
--- a/peerswap/docker-compose.yml
+++ b/peerswap/docker-compose.yml
@@ -7,13 +7,14 @@ services:
       APP_PORT: 1984
 
   web:
-    image: ghcr.io/impa10r/peerswap-web:v1.4.7@sha256:9bc6aa9ab40e5762a5b6fdac530f5739cb85f637621cfc9a0ed412af8452b9d2
+    image: ghcr.io/impa10r/peerswap-web:v1.5.6@sha256:861c677abe7c3adf9d1ca8c3fe478cc59770020edec946b78d07b51337cde5e2
     user: "1000:1000"
     restart: on-failure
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for the update @Impa10r. I'm just re-syncing Elements Core to test.

nmfretz:
Hey @Impa10r, I changed the environment variable `NO_HTTPS` from:
```yaml
NO_HTTPS: true
```

to

```yaml
NO_HTTPS: "true"
``` 

Can you please test to see that peerswap is consuming this correctly?

The reason I did this is that I noticed Docker Compose v1 and v2 treat `boolean` values differently:

V1 does not allow booleans for environment keys: https://docs.docker.com/compose/release-notes/#170

So users running umbrelOS 0.5.4 (which uses Compose v1.29.2) will get a broken install/update, with this error:
`services.web.environment.NO_HTTPS contains true, which is an invalid type, it should be a string, number, or a null`

V2, however, does allow certain booleans:
- string interpolation in 2.0: https://docs.docker.com/compose/release-notes/#compose-file-version-2x
- boolean values of `true` and `false` allowed as of version 2.17.0: https://docs.docker.com/compose/release-notes/#2170

umbrelOS 1.x uses Compose v2.27.1, so the original boolean value you used works fine.

Please let me know if this change works for you and then I'll merge.

Impa10r (commented):
Thanks! It is consumed correctly.

nmfretz:
Perfect, thanks for checking! Going live.

PR #1176, opened 2024-06-25
Fixed booleans for umbrelOS 0.5

sharknoon:
Hello,

Docker Compose V1 does not allow booleans inside the `environment`, `labels` or the `extra_hosts` key.

This PR fixes that issue to get those apps running on umbrelOS 0.5 .

Kind reagards

Sharknoon

### DIFF
--- a/mealie/docker-compose.yml
+++ b/mealie/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     restart: on-failure
     environment:
       # Allow additional user sign-up without token
-      ALLOW_SIGNUP: true
+      ALLOW_SIGNUP: "true"
       PUID: 1000
       PGID: 1000
       MAX_WORKERS: 1
--- a/ntfy/docker-compose.yml
+++ b/ntfy/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     environment:
       APP_HOST: ntfy_app_1
       APP_PORT: 80
-      PROXY_AUTH_ADD: false
+      PROXY_AUTH_ADD: "false"
 
   app:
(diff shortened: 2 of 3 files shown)


nmfretz:
Wonderful, thanks very much @sharknoon 🙏. I have tested on umbrelOS 0.5 and 1.2.1.

## Adding some context for our records:

umbrelOS 0.5 uses Compose V1 (1.29.2 by default if a user doesn't already have it installed or hasn't updated it):
```
$ docker-compose version
docker-compose version 1.29.2, build unknown
```

umbrelOS 1.x uses Compose V2. As of umbrelOS 1.2.1  we are using Compose v2.27.1:
```
$ docker compose version
Docker Compose version v2.27.1
```

In Compose V1, booleans cannot be used as values for the `environment`, `labels`, and `extra_hosts` keys.
Docker Compose 1.7.0 https://docs.docker.com/compose/release-notes/#170:
> Booleans are no longer allows as values for mappings in the Compose file (for keys environment, labels and extra_hosts). Previously this was a warning. Boolean values should be quoted so they become string values.

Compose errors out with something like: 
```
services.web.environment.NO_HTTPS contains true, which is an invalid type, it should be a string, number, or a null
```

So in umbrelOS 0.5, installs/updates that use boolean values for these keys will fail. See  https://github.com/getumbrel/umbrel-apps/pull/1170#issuecomment-2187890729 for example.

But Compose V2 handles booleans for these keys, so umbrelOS 1.x app installs/updates won't error out when using boolean values for these keys:
- v2.0 converts booleans to strings: https://docs.docker.com/compose/release-notes/#compose-file-version-2x
- v2.17.0 supports booleans: https://docs.docker.com/compose/release-notes/#2170

sharknoon:
You are very welcome 😃

PR #1178, opened 2024-06-27
App Submission: Budibase

sharknoon:
# App Submission

### App name
Budibase

### 256x256 SVG icon
logo

### Gallery images
Bildschirmfoto am 2024-06-27 um 21 19 41
Bildschirmfoto am 2024-06-27 um 21 20 08
Bildschirmfoto am 2024-06-27 um 21 21 09
Bildschirmfoto am 2024-06-27 um 21 22 00

### I have tested my app on:
- [ ] umbrelOS on Linux VM


sharknoon:
Unfortunately setting `user: "1000:1000"` in the `docker-compose.yml` prevents the app from starting.

joaovictor-local (approved):
I installed on umbrelOS 1.2.1 (Intel N100). 
Thank you, @sharknoon! I already have some use cases for it :smile:

nmfretz:
Fantastic work @sharknoon! This one was fun to play with.

Reviewed and tested. Let's get gallery assets made up.

nmfretz:
@sharknoon, actually one thing that I think we should do:

Right now the single bind mount is set to `${APP_DATA_DIR}/data`. This makes sense because budibase is an all-in-one image with a single container directory to bind to the host. But I think we should bind to `${APP_DATA_DIR}/data/budibase` instead. The reason being that this will future proof us in the event that a new container is added outside the all-in-one image in the future. For example, an nginx container or some umbrel specific container that requires a bind mount (I dunno, these aren't the greatest example). That means we could just bind `${APP_DATA_DIR}/data/<new-container>` and not worry at all about migrating the budibase data into it's own dedicated directory, or worrying about the order of the bind mounts to avoid conflict.

What do you think?

sharknoon:
@nmfretz Yes, that is a good foresight. I have changed and tested it 😀

nmfretz:
Another app ready to go from the legendary @sharknoon! 
Gallery assets: https://github.com/getumbrel/umbrel-apps-gallery/tree/master/budibase

Going live now.

image

PR #1179, opened 2024-06-27
App Submission: Duplicati

sharknoon:
# App Submission

### App name
Duplicati

### 256x256 SVG icon
logo

### Gallery images
Bildschirmfoto am 2024-06-27 um 23 11 58
Screenshot 2024-06-27 230700
Bildschirmfoto am 2024-06-27 um 23 11 18
Bildschirmfoto am 2024-06-27 um 23 11 23

### I have tested my app on:
- [ ] umbrelOS on Linux VM

### DIFF
--- a/duplicati/docker-compose.yml
+++ b/duplicati/docker-compose.yml
@@ -1,0 +1,18 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: duplicati_duplicati_1
+      APP_PORT: 8200
+      PROXY_AUTH_ADD: "false"
+  
+  duplicati:
+    image: duplicati/duplicati:2.0.8.1_beta_2024-05-07@sha256:97b8ef7bcaa095793ef9f39049851f987860a8c832aaf67ef977ff398f9e996e
+    entrypoint: ["/usr/bin/duplicati-server", "--webservice-port=8200", "--webservice-interface=any", "--webservice-password=${APP_PASSWORD}", "--webservice-allowed-hostnames=*"] 
+    user: "1000:1000"
+    restart: on-failure
+    stop_grace_period: 1m
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${UMBREL_ROOT}/data/storage:/umbrel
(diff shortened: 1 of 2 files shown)


nmfretz:
Another excellent submission @sharknoon! Let's get gallery assets finalized.

In the meantime, here's a couple things to think about that I would love your thoughts on:

1) Password
It looks like the web UI let's us change the password, which obviously won't persist on app restart because we set it via the entrypoint (which I think we are required to do in order to even have a password). I'll look into whether there is something we can change this, but I think we are probably stuck with this behaviour.

2) Adding app-data directory
RIght now we bind `${UMBREL_ROOT}/data/storage:/umbrel` so that users can choose to back up directories and files within `/home/home/umbrel/data/storage`. This allows users to backup things like media, books, anything they store in that directory, etc. I'm imagining a lot of users would also want to back up app specific data (e.g., nextcloud, nostr relay database, app config files, etc). In order to do this they need access to `/home/home/umbrel/app-data`. 
so users can backup data from specific apps (e.g., nextcloud, or maybe their qbittorrent config, etc). I'm currently thinking on the implications, but if we want to do this then we could change the bind mounts to something like:

```yaml
volumes:
      - ${APP_DATA_DIR}/data:/data
      - ${UMBREL_ROOT}/data/storage:/umbrel-storage
      - ${UMBREL_ROOT}/app-data:/umbrel-app-data
```

nmfretz:
@sharknoon, gallery assets are live: https://github.com/getumbrel/umbrel-apps-gallery/tree/master/duplicati

Let's take another look at the password behaviour and then launch this!

image

sharknoon:
Thank you for your reply @nmfretz :)

I have looked around in the documentation and found out, that I only have to set the UI password once: `This option is saved so you do not need to set it on each run. Setting an empty value disables the password.`. 

Maybe we could use a hook to run after the install and set the password by running the container and if the user changes it in the UI, the change would persist. But I don't know if the hook would rerun after an update.

But I think this will add a lot of complexity just for being able to change the password. I would recommend against that.

@nmfretz What is your opinion?

nmfretz:
Thanks for looking into that @sharknoon. I agree with your recommendation to keep things simple here. Let's launch Duplicati as-is, and if we get significant feedback on this then we can use lifecycle hooks to change the behaviour.

Going live 🎉

PR #1185, opened 2024-07-03
Add Home Assistant Fusion UI

joaovictor-local:
# App Submission
### App name

Home Assistant Fusion UI

### 256x256 SVG icon
logo

### Gallery images

preview

### DIFF
--- a/home-assistant-fusion-ui/docker-compose.yml
+++ b/home-assistant-fusion-ui/docker-compose.yml
@@ -4,11 +4,10 @@ services:
   web:
     image: ghcr.io/matt8707/ha-fusion@sha256:bd396be1a077374dccadb47656b95137f93f15b7c76547a5ff29316dda9ea903
     restart: on-failure
-    network_mode: bridge
     user: 1000:1000
     ports:
       - 5023:5050
     volumes:
       - ${APP_DATA_DIR}/data/fusion:/app/data
     environment:
-      HASS_URL: http://${UMBREL_LOCAL_IP}:8123
+      HASS_URL: http://${DEVICE_DOMAIN_NAME}:8123
--- a/home-assistant-fusion-ui/docker-compose.yml
+++ b/home-assistant-fusion-ui/docker-compose.yml
@@ -11,4 +11,4 @@ services:
     volumes:
       - ${APP_DATA_DIR}/data/fusion:/app/data
     environment:
-      HASS_URL: http://10.32.0.1:8123
+      HASS_URL: http://${UMBREL_LOCAL_IP}:8123
--- a/home-assistant-fusion-ui/docker-compose.yml
+++ b/home-assistant-fusion-ui/docker-compose.yml
@@ -1,0 +1,14 @@
+version: "3.7"
+
+services:
+  web:
+    image: ghcr.io/matt8707/ha-fusion@sha256:bd396be1a077374dccadb47656b95137f93f15b7c76547a5ff29316dda9ea903
+    restart: on-failure
+    network_mode: bridge
+    user: 1000:1000
+    ports:
+      - 5023:5050
+    volumes:
+      - ${APP_DATA_DIR}/data/fusion:/app/data
+    environment:
+      HASS_URL: http://10.32.0.1:8123
--- a/home-assistant-fusion-ui/docker-compose.yml
+++ b/home-assistant-fusion-ui/docker-compose.yml
@@ -11,3 +11,5 @@ services:
       - ${APP_DATA_DIR}/data/fusion:/app/data
     environment:
       HASS_URL: http://${DEVICE_DOMAIN_NAME}:8123
+    extra_hosts:
+      - "${DEVICE_DOMAIN_NAME}:${APP_HAFUSION_MACHINE_IP}"
--- a/home-assistant-fusion-ui/docker-compose.yml
+++ b/home-assistant-fusion-ui/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   web:
-    image: ghcr.io/matt8707/ha-fusion@sha256:bd396be1a077374dccadb47656b95137f93f15b7c76547a5ff29316dda9ea903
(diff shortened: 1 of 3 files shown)


marciasmedia:
Nice :)

sharknoon:
And here is another fantastic submission of @joaovictor-local 🎉 Thank you so much for your hard work on bringing new apps to umbrelOS!

I have some suggestions:
- Maybe we can replace the `UMBREL_LOCAL_IP` with `NETWORK_IP`, as this environment variable will automatically be set by umbrelOS. See an example [here](https://github.com/getumbrel/umbrel-apps/blob/2a69c7d83eb0708ac58ba9daf94b371165c6fa02/nextcloud/docker-compose.yml#L44). I know this is hard to find, because it isn't documented. If that turns out to be working, we can skip the `exports.sh` file altogether 😀
- Consider adding the `app_proxy` infront of the application, as it enhances security by only redirecting HTTP traffic. Fusion UI probably has its own authentication, so you could disable the Proxy Authentication with `PROXY_AUTH_ADD: "false"`
- Do we need `network_mode: bridge`? I always thought, this is the default, but I could be wrong

Great work on reducing the user permissions to `1000:1000`!

joaovictor-local:
@sharknoon Thank you @sharknoon :) 

1. I just tried but it didn't worked because `NETWORK_IP` returns the docker internal IP that is not accessible for the client browser, but we can use `DEVICE_DOMAIN_NAME` instead of exporting the network IP. I hope they add some new integration options, I am not a fan of oauth for this this scenario. The best implementation for me would be something that would allow us to only need `auth_proxy`.

| image  |
|---|
| ha-fusion usage oauth  |


2. I also tried to use the `app_proxy` but it breaks the camera widget/object with 503 gateway time out, and I think it because that would be the moment where `ha-fusion` should be using `NETWORK_IP` for it's internal request but it's not possible to define at the moment.

```yaml
  app_proxy:
    environment:
      APP_HOST: home-assistant-fusion_web_1
      APP_PORT: 5050
      PROXY_AUTH_ADD: "false"
```

3. You are right, it didn't. I will remove.

sharknoon:
Thank you for your implementations 😀 If using the proxy causes connection issues, then I also think we do not need to use it. Especially because this app has it's own authentication.

Does your camera work now if you are using `DEVICE_DOMAIN_NAME` instead of your the ip address?

nmfretz:
Nice work @joaovictor-local!

I just tested and we are almost there. @sharknoon was on the right track with using `NETWORK_IP`... suuuuuper close, but what we want here is `GATEWAY_IP`:

- NETWORK_IP is the network/subnet identifier (e.g., 10.21.0.0) which used here is just specifying the start of the Docker network IP range.

- GATEWAY_IP is the actual IP address of the Docker network’s gateway (e.g., 10.21.0.1), which is what we want.


> I just tried but it didn't worked because NETWORK_IP returns the docker internal IP that is not accessible for the client browser, but we can use DEVICE_DOMAIN_NAME instead of exporting the network IP.

^ So for this, the browser isn't making a direct request to NETWORK_IP (or GATEWAY_IP in our case now). Instead it is making a request to something like `http://umbrel.local:5023/api/...`. The actual communication between fusion api and home-assistant will use the internal docker IP.

-------------------------
I've tested with `HASS_URL: http://${GATEWAY_IP}:8123` by streaming a video file over HTTP from my laptop and connecting via Home Assistant:

image

image

Can you please change the HASS_URL to `HASS_URL: http://${GATEWAY_IP}:8123` and give it a test?

We'll get the gallery assets prepared in the meantime.

nmfretz:
EDIT: @joaovictor-local @sharknoon I'm actually wrong about `GATEWAY_IP` fixing our issue. 
It does fix the camera issue, BUT it completely messes up the initial HA access request which occurs via the Browser and so must be `umbrel.local:8123`. I missed this when I was playing around, because I had already installed HA-Fusion and given it access to HA, so it never had to try this again.

HASS_URL is used for this:
`http://umbrel.local:8123/auth/authorize?response_type=code&redirect_uri=<token>`
image

If we use `GATEWAY_IP` then we the Browser tries to access `http:/10.21.0.1:8123/auth/authorize?response_type=code&redirect_uri=<token>` which obviously won't work.

Apologies for missing this @joaovictor-local, I see you were getting at something similar here: https://github.com/getumbrel/umbrel-apps/pull/1185#issuecomment-2210867355

So we're stuck using `HASS_URL: http://${DEVICE_DOMAIN_NAME}:8123` but then have issues connecting to cameras:

Error occurred while trying to proxy: umbrel.local:5023/api/camera_proxy/camera.192_168_4_71?token=1baf2552af54b4c921d4ebc29b246122060e3265d12481fe73b4cb73d770c2de&date=1726039586464

Interestingly, I just tried testing this commit https://github.com/getumbrel/umbrel-apps/pull/1185/commits/98515b2eab7d115c6d11b8186c2f5be38da76e70 @joaovictor-local where you're using the local IP and the camera works using this, BUT unfortunately  you have to be accessing HA-Fusion from `<your-local-ip>:5023` for HA-Fusion to work at all. If you are accessing from http://umbrel.local:5023 then you'd be hit with CORS errors. This is an issue because the vast majority of people will be accessing HA-Fusion from http://umbrel.local:5023.

nmfretz:
@sharknoon @joaovictor-local I managed to get it working. Can you please give it a test as well?

I pushed some changes to finalize the submission, but the main commit with the fix is https://github.com/getumbrel/umbrel-apps/pull/1185/commits/798dba6e4c9da4010fa6255e1e08b4dac13184c7, which adds
```yaml
extra_hosts:
      - "${DEVICE_DOMAIN_NAME}:${APP_HAFUSION_MACHINE_IP}"
```

sharknoon:
@nmfretz I have tested it on my Raspberry Pi and it works perfectly fine :)

nmfretz:
Thanks for testing @sharknoon! And thanks again for this submission @joaovictor-local!

Gallery assets are ready and we are going live:

image

PR #1194, opened 2024-07-08
Update all Servarr apps and download clients to auto-configure

nmfretz:
This PR updates Radarr, Sonarr, Lidarr, Readarr, Prowlarr, qBittorrent, and SABnzbd.

In addition to bumping the versions of these apps, this PR also makes it so that the Servarr and download client ecosystem auto-configures upon installation (this was already the case for Servarr apps and Transmission). Auto-configuration of Overseerr, Jellyseerr, and Autobrrr will be added in the future.

### Radarr, Sonarr, Lidarr, Readarr:
- The media-app-configurator will now connect these apps to qBittorrent and SABnzbd if they are installed. These download clients are optional and configuration will only be attempted if they are installed on a user's umbrel.
- This will not change a user's existing configuration if they have already connected the download client.
- User's can make changes to the configurations after install (e.g., name of client, priority, tags, etc) and we won't overwrite their changes.

### Prowlarr:
- The media-app-configurator will now connect qBittorrent and SABnzbd if they are installed. These download clients are optional and configuration will only be attempted if they are installed on a user's umbrel.
- This will not change a user's existing configuration if they have already connected the download client.
- User's can make changes to the configurations after install (e.g., name of client, priority, tags, etc) and we won't overwrite their changes.
- The Flaresolverr container is now included (Thanks to @Julienpeps https://github.com/getumbrel/umbrel-apps/pull/1042). Flaresolverr is automatically configured to be available. To use it, simply add the “flaresolverr” tag to any indexers that require it.

### qBittorrent
- qBittorrent restarts any Servarr apps (e.g., Radarr) after qBittorrent is installed so that the Servarr app runs the auto-configuration. 
- In order to allow connections coming in from Servarr apps, we set `WebUI\AuthSubnetWhitelistEnabled` to `true` and add the umbrel Docker network subnet to `WebUI\AuthSubnetWhitelist`.
- User's can make changes to the configurations after install and we won't overwrite their changes.

### SABnzbd 
- SABnzbd restarts any Servarr apps (e.g., Radarr) after SABnzbd is installed so that the Servarr app runs the auto-configuration. 
- The download directories are now set correctly at install without needing to modify them through the setup wizard, and the sabnzbd container name has been added to `host_whitelist` to allow connections coming in from Servarr apps (Thanks to @fabricionaweb for initiating this https://github.com/getumbrel/umbrel-apps/pull/1153)
- Categories are set up for Prowlarr (`prowlarr-fallback`), Radarr (`movies`), Sonarr (`shows`), Readarr (`books`), and Lidarr (`music`) so that the media-app-configurator can connect Servarr apps using these categories (Connecting to SABnzbd requires that a category set in a Servarr app is an exact match to a category set in SABnzbd).
- User's can make changes to the configurations after install and we won't overwrite their changes.
- Users who already have SABnzbd **AND** have not previously set up SABnzbd with media apps, should uninstall and reinstall SABnzbd to avoid any potential category conflicts.

Tested app updates and fresh installs.

### DIFF
--- a/lidarr/docker-compose.yml
+++ b/lidarr/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api/*"
 
   server:
-    image: linuxserver/lidarr:2.1.7@sha256:a695aec5a08032c1a15e3d4b2c784952fe1f86071c377abf3ad19413657f4cbe
+    image: linuxserver/lidarr:2.3.3@sha256:32ddd032d57d7e996052b0fe68c651e88f87909eaab26935dc0160ea468a936e
     environment:
       - PUID=1000
       - PGID=1000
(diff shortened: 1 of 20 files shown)


sharknoon:
@nmfretz This looks awesome! Really solid work autoconfiguring those apps. I have one question: SABnzbd and qBittorrent are optional. Is there a reason, why Transmission is not optional? I personally only use the Usenet and not Torrent downloads. Maybe this can be optional too?

Thanks again for your big effort making configuration automated 😀

nmfretz:
Thanks @sharknoon! Hopefully we can continue to make the entire ecosystem even easier for users to bootstrap (e.g., Overseerr, Jellyseerr, and Autobrrr)... it's quite confusing for people who are new to Servarr (and even those who are experienced, to be honest).

>  I have one question: SABnzbd and qBittorrent are optional. Is there a reason, why Transmission is not optional? I personally only use the Usenet and not Torrent for my downloads. Maybe this can be optional too?

I completely agree with you. I also only use Usenet and it is a bit weird having Transmission still available as an app.

As background, this is a holdover from when the Servarr apps were first brought to the App Store. At the time there was no compatible download client in the app store yet, so we added Transmission at the same time. It then made sense to auto-configure everything to the only download client available (Transmission) to make it easy for users.

Now with the additional of more download clients, and autoconfiguration to any of them (from this PR) we should take away Transmission as a dependency for the Servarr apps as you suggest. I'm going to merge this PR and then we can remove Transmission as a dependency in a future update (will require changes to the media app configurator).

PR #1195, opened 2024-07-11
Update Tailscale 1.68.1

yevmoroz:
Bump Tailscale to 1.68.1
Tested on: Raspberry Pi 4, umbrelOS 1.1.2 ✔️


nmfretz:
Excellent, thanks very much for this @yevmoroz! Tested on arm64 and amd64. Going live.

 Noting this here for others who may make there way to this PR:
 The current latest Tailscale version is 1.68.2 according to https://github.com/tailscale/tailscale/releases, but no official Docker image is available for that version yet (not even the `latest` tag).

PR #1196, opened 2024-07-11
App Submission: Omada Controller

hugofnm:
# App Submission

### App name
Omada Controller

### 256x256 SVG icon

https://www.tp-link.com/res/images/icons/tag-omada.svg

### Gallery images

Screenshot1
Screenshot2
Screenshot3

### I have tested my app on:
- [X] umbrelOS on a Raspberry Pi
- [ ] umbrelOS on Linux VM

### DIFF
--- a/omada-controller/docker-compose.yml
+++ b/omada-controller/docker-compose.yml
@@ -1,0 +1,28 @@
+version: "3.7"
+
+services:
+  server:
+    image: mbentley/omada-controller:5.14@sha256:ce429a64b11bcc8f86fdc5aa1990a54a158de8ce83f92c316bb556f29f732afd
+    init: true
+    restart: unless-stopped
+    stop_grace_period: 60s
+    network_mode: host # Controller needs lots of ports for device discovery
+    ulimits:
+      nofile:
+        soft: 4096
+        hard: 8192
+    environment:
+      - PUID=1000
+      - PGID=1000
+      - MANAGE_HTTP_PORT=8088
+      - MANAGE_HTTPS_PORT=8043
+      - PORTAL_HTTP_PORT=8088
+      - PORTAL_HTTPS_PORT=8043
+      - SHOW_SERVER_LOGS=true
+      - SHOW_MONGODB_LOGS=false
+      - SSL_CERT_NAME="tls.crt"
+      - SSL_KEY_NAME="tls.key"
+    volumes:
+      - ${APP_DATA_DIR}/data:/opt/tplink/EAPController/data
+      - ${APP_DATA_DIR}/logs:/opt/tplink/EAPController/logs
+      - ${APP_DATA_DIR}/work:/opt/tplink/EAPController/work
--- a/omada-controller/umbrel-app.yml
+++ b/omada-controller/umbrel-app.yml
@@ -1,0 +1,42 @@
+manifestVersion: 1
+id: omada-controller
+name: Omada Controller
+tagline: On-Premises Omada SDN Controller
+icon: https://cdn.jsdelivr.net/gh/walkxcode/dashboard-icons/svg/omada.svg
+category: Networking
+version: "5.14"
+port: 8088
+description: >-
+  Omada Software Controller
+
+
+  A Professional and Free On-Premises Omada SDN Controller
+
+  - Centralized Management : Up to 1,500 Omada access points, JetStream switches, and Omada routers.
+
+  - Cloud Access : Manage and monitor with the Omada app or Web UI from anywhere, anytime.
+
+  - Free of Charge : Download and use locally or from the cloud without additional expense.
+
+  - Easy and Intelligent Network Monitoring : The easy-to-use dashboard makes it simple to see the real-time network status and traffic distribution.
+
+  - Real-Time Network Topology : Helps IT admins quickly see and troubleshoot connections at a glance.
+
+  - Easier Network Maintenance : WiFi heatmap simulator, visualizable network report, and batch & multi-site management benefit network maintenance.
+
+
+  ⚠️ - This app can take 15-20 minutes to start after installation.
+
+developer: TP-Link / Matt Bentley
+website: https://github.com/mbentley/docker-omada-controller
+submitter: hugofnm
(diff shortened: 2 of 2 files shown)


omada-controller/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+      - PORTAL_HTTP_PORT=8088
+      - PORTAL_HTTPS_PORT=8043
+      - SHOW_SERVER_LOGS=true
+      - SHOW_MONGODB_LOGS=false
+      - SSL_CERT_NAME="tls.crt"
+      - SSL_KEY_NAME="tls.key"
+    volumes:
+      - ${APP_DATA_DIR}/data:/opt/tplink/EAPController/data
+      - ${APP_DATA_DIR}/logs:/opt/tplink/EAPController/logs
+      - ${APP_DATA_DIR}/work:/opt/tplink/EAPController/work
```
sharknoon, line 28: Due to permission issues on umbrelOS, please create a `.gitignore` file for each mounted directory:
- `omada-controller/data/.gitignore`
- `omada-controller/logs/.gitignore`
- `omada-controller/logs/.gitignore`
The problem is that on older versions of umbrelOS (again backwards compatibility) the permissions of automatically created repositories are wrong. By supplying a `.gitignore` file, those directories are getting created by installing the app automatically.

sharknoon (changes requested):
Thank you so much @hugofnm for your awesome submission! I have left some suggestions for this PR. Sorry that you had to wait so long, but we have just finished our linter for any new app submissions. You will see it when you are adding new commits to this PR :)

hugofnm:
Thanks @sharknoon for the suggestions ! 
Should be good to go :)

sharknoon:
Awesome work @hugofnm! Our new linter is also happy 😀
@nmfretz I think this app is ready to go

nmfretz:
Thanks for the ping and review @sharknoon 🤝.
 
And great work packaging Omada Controller @hugofnm!

The only outstanding task is to avoid potential port clashes with Omada Controller running in host networking mode. It’s on the umbrelOS roadmap to mitigate port clashes automatically, but this currently isn't implemented, and port `8088` is in use by Element: https://github.com/getumbrel/umbrel-apps/blob/master/element/umbrel-app.yml

This means if a user tries to run both Omada Controller and Element on the same device, there will be a port clash, causing one of the apps to fail to launch (or fail to install).

To mitigate this, here are the ports Omada Controller uses:
https://github.com/getumbrel/umbrel-apps/blob/master/element/umbrel-app.yml

We don’t need to worry about UDP ports, as they can be shared. The only one we need to plan for right now is `8088` I believe.

Here are the env vars we have access to:
https://github.com/getumbrel/umbrel-apps/blob/master/element/umbrel-app.yml

So we can use `MANAGE_HTTP_PORT` to change 8088 to a free port.

hugofnm:
Hi @nmfretz 👋,
I've changed the HTTP port 8088 -> 8078 as suggested.

Looking at the official documentation (https://www.tp-link.com/us/support/faq/3281/), it seems that Omada Controller uses about 10 different TCP ports for device discovery, web management, ...
Is there a spreadsheet or something like that to check port availability and avoid conflicts ?

Also, should this app stay on host mode or should I expose each port individually?

nmfretz:
> I've changed the HTTP port 8088 -> 8078 as suggested.
> Also, should this app stay on host mode or should I expose each port individually?

Thanks for doing that @hugofnm. With this port change, let's leave this in host network mode because Omada will need this for device discovery.

Gallery assets are here: https://github.com/getumbrel/umbrel-apps-gallery/tree/master/omada-controller

@hugofnm @sharknoon, I have made some small tweaks and have tested the app. One thing to look into before this goes live is the ssl cert experience. Currently, users who install Omada Controller will be met with a warning in their browser that the connection is not secure.

image

Do you happen to know whether we can avoid this entirely via compose configuration? We should really try to avoid encouraging users to click past these warnings, but as a last resort we could include instructions in the app's description.

sharknoon:
@hugofnm Is there a way to disable the HTTPS endpoint completly and only use the HTTP one?

hugofnm:
Well... I'm afraid it's not possible to disable the HTTPS endpoint @sharknoon
Omada Controller is closed-source and documentation is rare...
The only "way around" it's to disable HTTPS redirect, but you can only disable it in the Web UI.

image

sharknoon:
Thank you for your answer @hugofnm. I guess we have to accept the warning screen then, until umbrelOS supports domains and therefore HTTPS. Whats your opinion @nmfretz?

nmfretz:
> Well... I'm afraid it's not possible to disable the HTTPS endpoint @sharknoon
Omada Controller is closed-source and documentation is rare...
The only "way around" it's to disable HTTPS redirect, but you can only disable it in the Web UI.

Darn, that's a shame. This means that the flow for a user installing Omada Controller would be:
- Open the app, which will redirect to https. User will be met with a scary warning that we shouldn't be encouraging users to click past.
- Click past the warning.
- Set up an account
- Edit the settings in the UI to disable https redirects.
- Restart the app from the umbrelOS homescreen (it's required to restart the omada container apparently when changing the redirect setting).

Great work packaging this app @hugofnm. The effort is super appreciated. Unfortunately, I think we should wait to bring this to the official app store until after umbrelOS has integrated HTTPS support, which is on the horizon. That way users of all technical skill levels can easily use the app and we aren't providing a poor experience as a result of the OS.

Would you prefer that we convert this PR to draft but keep it open, or would you prefer to close this PR and reopen once HTTPS support is available?

In the meantime people could still run Omada Controller on their Umbrel by either:
1) Installing it from a Community App Store https://github.com/getumbrel/umbrel-community-app-store, or
2) Installing it via the Portainer app: https://apps.umbrel.com/app/portainer

hugofnm:
Sure, I'll convert this PR ito draft until Umbrel supports HTTPS !

nmfretz:
Sounds good, thanks @hugofnm 🤝

nmfretz:
Reminder for us: gallery assets have already been prepared for Omada Controller

PR #1217, opened 2024-07-17
Update uptime-kuma to 1.23.13

theeldarka:


nmfretz:
Thanks very much for this update @theeldarka! Tested and going live.

PR #1218, opened 2024-07-17
Update Jellyseer from 1.7.0 to 1.9.2

theeldarka:


nmfretz:
Thanks for this update as well @theeldarka! Much appreciated. Tested and going live.

PR #1223, opened 2024-07-20
Ignore Proxy auth for usage with apps

MalteKiefer:
Without this line, no mobile app can recognize paperless and a usage with mobile apps is not possible

### DIFF
--- a/paperless/docker-compose.yml
+++ b/paperless/docker-compose.yml
@@ -24,7 +24,7 @@ services:
       POSTGRES_PASSWORD: paperless
 
   webserver:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @MalteKiefer. I think this is a good solution considering Paperless has its own in-built authentication already.

The other solution would be to keep the proxy's authentication, but whitelist certain routes with the `PROXY_AUTH_WHITELIST` environment variable for the `app_proxy` container.

So for example, if a mobile app interacts with paperless by something like `http://umbrel.local:2349/api/some-endpoint`, we could add the following to the `app_proxy` container's environment variables to whitelist just the `/api/*` route:

`PROXY_AUTH_WHITELIST: "/api/*"`

In order for existing Paperless-ngx installs to bring down this change, we need to update the app so that users see that an app update is available. I will make the necessary changes to this PR.

nmfretz:
Updated Paperless to latest version and tested.

PR #1224, opened 2024-07-20
Update Gitea to 1.22.1

MalteKiefer:


nmfretz:
Thanks for this @MalteKiefer.

I have:
- tested on arm64 and x86 for both fresh install and app update
- tweaked the release notes
- Checked mariadb release notes for breaking changes that we need to be aware of when updating the database.

PR #1226, opened 2024-07-20
Seafile update to 11.0.9

MalteKiefer:


nmfretz:
Thanks very much for this update as well @MalteKiefer.

Tested and going live.

PR #1227, opened 2024-07-20
Filebrowser update to 2.30.0

MalteKiefer:


nmfretz:
Thanks for this update as well @MalteKiefer 🙌.

I have added release notes for version 2.30.0 and tested. Going live.

PR #1228, opened 2024-07-20
Element update to 1.11.71

MalteKiefer:


nmfretz:
Hi @MalteKiefer. Do you run Element (and possibly Synapse) on your Umbrel? This update bumps Element up several versions and I'd love to have someone help test properly to make sure the app doesn't break for existing installs.

I'm thinking we should update Synapse (matrix server) at the same time and make sure the apps are compatible.

MalteKiefer:
Yes I use it.
Can send some screens to prove that it works, when this would help?

nmfretz:
Thanks @MalteKiefer. I have tested and the update breaks compatibility with the current Synapse app (suuuuuuuper old) on the app store, so I have updated Synapse so that we retain compatibility.

I have done the following to finalize:
- bumped the Synapse app to 1.55.0: https://github.com/getumbrel/umbrel-apps/pull/1357
- bumped the element image to the latest and retested
- added release notes
- clarified the app description and included up-to-date connection instructions.

Going live 🎉

PR #1229, opened 2024-07-20
App Submission: Pingvin Share

MalteKiefer:
# App Submission

### App name
Pingvin Share

### 256x256 SVG icon
logo
Sorry, no experience to create a SVG from an PNG

...

### Gallery images
Pingvin_03
Pingvin_02
Pingvin_01

### I have tested my app on:
- [X] umbrelOS on Linux VM

### DIFF
--- a/pingvin-share/docker-compose.yml
+++ b/pingvin-share/docker-compose.yml
@@ -1,0 +1,27 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: pingvin-share_web_1
+      APP_PORT: 6453
+      PROXY_AUTH_ADD: "false"
+  
+  web:
+    image: stonith404/pingvin-share:v0.27.0@sha256:ebcd050ddcf0fc6e2f14944e66f478825697479cb79feaec7d989fa4a45a0635
+    restart: on-failure
+    environment:
+      - PUID=1000
+      - PGID=1000
+    ports:
+      - 6453:3000
+    volumes:
+      - "${APP_DATA_DIR}/data:/opt/app/backend/data"
+      - "${APP_DATA_DIR}/data/images:/opt/app/frontend/public/img"
+    depends_on:
+      clamav:
+        condition: service_healthy
+
+  clamav:
+    restart: on-failure
+    image: clamav/clamav
--- a/pingvin-share/docker-compose.yml
+++ b/pingvin-share/docker-compose.yml
@@ -16,10 +16,3 @@ services:
     volumes:
       - "${APP_DATA_DIR}/data:/opt/app/backend/data"
       - "${APP_DATA_DIR}/data/images:/opt/app/frontend/public/img"
-    depends_on:
-      clamav:
-        condition: service_healthy
-
-  clamav:
-    restart: on-failure
-    image: clamav/clamav
--- a/pingvin-share/docker-compose.yml
+++ b/pingvin-share/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   app_proxy:
     environment:
       APP_HOST: pingvin-share_web_1
-      APP_PORT: 6453
+      APP_PORT: 3000
       PROXY_AUTH_ADD: "false"
   
   web:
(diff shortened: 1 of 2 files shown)


pingvin-share/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+    image: stonith404/pingvin-share:v0.27.0@sha256:ebcd050ddcf0fc6e2f14944e66f478825697479cb79feaec7d989fa4a45a0635
+    restart: on-failure
+    environment:
+      - PUID=1000
+      - PGID=1000
+    ports:
+      - 6453:3000
+    volumes:
+      - "${APP_DATA_DIR}/data:/opt/app/backend/data"
+      - "${APP_DATA_DIR}/data/images:/opt/app/frontend/public/img"
```
sharknoon, line 20: Due to privilege issues on older umbrelOS devices, we need to add those directories to the app. Please create the following files:
- `pingvin-share/data/images/.gitkeep`
- `pingvin-share/data/backend/.gitkeep`

The problem is that on older versions of umbrelOS (again backwards compatibility) the permissions of automatically created repositories are wrong. By supplying a .gitkeep file, those directories are getting created by installing the app automatically.
Also maybe it would be better to save the backend data separately to prevent the clashing of files:
```suggestion
      - "${APP_DATA_DIR}/data/backend:/opt/app/backend/data"
      - "${APP_DATA_DIR}/data/images:/opt/app/frontend/public/img"
```

sharknoon (changes requested):
Thank you so much @MalteKiefer for your awesome app submission! I have left some suggestions below.

nmfretz:
@MalteKiefer friendly ping to see if you are still working on this. The main issue noted by @sharknoon is that the `clamav` image is only available for `linux/amd64` architectures so the app would be broken for users running umbrelOS on a raspberry pi.

nmfretz:
Thinking about this more, it seems like Pingvin Share is really only useful when it is exposed to the internet. The easiest way to do this on umbrelOS right now is to:
1. Install the Cloudflare Tunnel app
2. Set up a Cloudflare account outside of umbrelOS and purchase+add a domain
3. Set up a tunnel to the pingvin share `web` container, bypassing the app proxy.

This is certainly possible for technical users to do, but I think realistically what will happen is that an average user sees this cool looking file share app, installs it, and then becomes frustrated that it doesn't work out of the box.

I propose we convert this to draft and then launch it once umbrelOS has a simple way for users to expose apps to the internet (on the horizon). That way when this hits the official app store, there will be the option for it to work out-of-the-box for users.

In the meantime, technical users could still run Pingvin Share on their Umbrel by either:
- Installing it from a Community App Store https://github.com/getumbrel/umbrel-community-app-store, or
- Installing it via the Portainer app: https://apps.umbrel.com/app/portainer

dennysubke:
With a Cloudflare tunnel on an Umbrel Home everything works as it should. I have packed Pingvin Share in my Community Store: https://github.com/dennysubke/umbrel-community-app-store

dennysubke:
@MalteKiefer 
I have adopted pingvin share in [my community appstore](https://github.com/dennysubke/dennys-umbrel-app-store). Everything is going great. However, the app only makes sense in conjunction with Coudflare.


Screenshot Pingvin Share UmbrelStore



> I propose we convert this to draft and then launch it once umbrelOS has a simple way for users to expose apps to the internet (on the horizon). That way when this hits the official app store, there will be the option for it to work out-of-the-box for users.

What do you mean by that, @nmfretz ?

nmfretz:
Great to know it's working well on your community app store @dennysubke!

On revisiting this, I think we should send this to the app store with a note at the top of the app description saying that Pingvin Share needs to be exposed to the internet for full functionality. We do this for some other apps (e.g., Plausible), so it doesn't make sense to single out Pingvin Share for this.

We will get gallery assets prepared and I will finalize this submission.

 > What do you mean by that, @nmfretz ?
 
 I can't give away too much at this point, but we have plans to allow users to easily expose apps to the internet. As noted above though, we shouldn't be waiting for this to bring this app to the app store. Thanks for your input @dennysubke!

nmfretz:
I have gone through and finalized this submission. I have tested on x86 and arm64, including exposing Pingvin Share to the web using a Cloudflare tunnel.

image

PR #1230, opened 2024-07-20
Vikunja update to 0.24.1

MalteKiefer:


nmfretz:
Thanks very much for all the app updates and submissions @MalteKiefer! Very much appreciated. We will take a look at these shortly. Please ignore the random close/open from me testing a workflow.

MalteKiefer:
External port is need, there is a mobile app for Vikunja.

nmfretz (commented):
Thanks again for this Vikunja update @MalteKiefer. Looks like the developers are simplifying the architecture for Docker Compose deployments which is nice.

I've left some review comments below with suggested changes. In addition to those changes, we can also delete the entire `proxy` subdirectory within `vikunja/data` since the new compose architecture no longer requires nginx.

Once addressed, we can test to make sure the app is working well from fresh installs and app updates.

nmfretz:
I have gone ahead and made the required changes and tested Vikunja as a fresh install and update from v0.22.1 with data persistence. Thanks again for this PR!

PR #1232, opened 2024-07-21
App Submission: Navidrome

owmsoton:
# App Submission

### App name

Navidrome

### 256x256 SVG icon

navidrome

### Gallery images

light-albums-view

ss-desktop-player

ss-mobile-album-view

ss-mobile-login

ss-mobile-player

### I have tested my app on:
- [x] umbrelOS on Linux VM


navidrome/umbrel-app.yml
```diff
@@ -0,0 +1,59 @@
...
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+releaseNotes: >-
+
+  This is mostly a bug-fix version, with focus on some Jukebox issues, but also other general improvements and fixes.
+
+  Full release notes are available at https://github.com/navidrome/navidrome/releases/tag/v0.52.5
+submitter: owmsoton
+submission:
```
sharknoon, line 59: I also tend to forget to add the submission field afterwards :)
```suggestion
submission: https://github.com/getumbrel/umbrel-apps/pull/1232
```

navidrome/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+services:
+  navidrome:
+    image: deluan/navidrome:latest
```
sharknoon, line 3: We try to use not the lastest tag to make it more clear which image tag the app uses. The digest at the end is needed to ensure reproducibility in case a tag gets changes (happens for the latest tag all the time)

```suggestion
    image: deluan/navidrome:0.52.5@sha256:b154aebe8b33bae82c500ad0a3eb743e31da54c3bfb4e7cc3054b9a919b685c7
```

navidrome/umbrel-app.yml
```diff
@@ -0,0 +1,59 @@
...
+  - 2.jpg
+  - 3.jpg
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+releaseNotes: >-
+
+  This is mostly a bug-fix version, with focus on some Jukebox issues, but also other general improvements and fixes.
+
+  Full release notes are available at https://github.com/navidrome/navidrome/releases/tag/v0.52.5
```
sharknoon, line 57: For initial app submissions release notes are not required. Those are being displayed to the user when there is a new update available.
```suggestion
releaseNotes: ""
```

navidrome/umbrel-app.yml
```diff
@@ -0,0 +1,59 @@
...
+repo: https://github.com/navidrome/navidrome
+support: https://github.com/navidrome/navidrome/discussions
+port: 4533
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
+path: ""
+defaultUsername: ""
+defaultPassword: ""
```
sharknoon, line 52: If you don't have a default username and password, you can omit those
```suggestion
```

navidrome/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+    restart: on-failure
+    environment:
+      - PUID=1000
+      - PGID=1000
+    ports:
+      - "4533:4533"
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${UMBREL_ROOT}/data/storage/music:/music:rw
```
sharknoon, line 12: Please create a file called `navidrome/data/.gitkeep`.
The problem is that on older versions of umbrelOS the permissions of automatically created repositories are wrong. By supplying a .gitkeep file, those directories are getting created by installing the app automatically.

navidrome/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+services:
+  navidrome:
+    image: deluan/navidrome:latest
+    restart: on-failure
+    environment:
+      - PUID=1000
+      - PGID=1000
```
sharknoon, line 7: Nice work restricting the user permissions :)

navidrome/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+    ports:
+      - "4533:4533"
```
sharknoon, line 9: If this port is only being used to serve http(s) (which I think it does) it is better to use the app proxy (see line 2). In this case we can omit those ports here:
```suggestion
```

navidrome/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+services:
```
sharknoon, line 1: I suggest using the app proxy instead of exposing the http port directly. This makes it more secure as the proxy is only for HTTP traffic and everything else can't get through
```suggestion
services:
  app_proxy:
    environment:
      APP_HOST: navidrome_navidrome_1
      APP_PORT: 4533
      PROXY_AUTH_ADD: "false"
```

navidrome/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+    restart: on-failure
+    environment:
+      - PUID=1000
+      - PGID=1000
+    ports:
+      - "4533:4533"
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${UMBREL_ROOT}/data/storage/music:/music:rw
+    container_name: navidrome_1
```
sharknoon, line 13: Is there a specific reason to add the container_name? If not we may remove it :)

sharknoon (changes requested):
Thank you so much for your awesome app submission @owmsoton! I have left you some suggestions below:

owmsoton:
@sharknoon Thank you for all your help!

All your suggestions were very helpful and I learned a lot. I ended up improving some details and I believe the corrections were well applied.

I look forward to the next steps.

Thanks again.

navidrome/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+      - PUID=1000
+      - PGID=1000
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${UMBREL_ROOT}/data/storage/music:/music:rw
+
+# You do not need to expose the port that your app's web server is listening on if you're using the app_proxy service.
+      # This is handled by the APP_HOST and APP_PORT environment variables in the service above.
+      #
+      # If you need to expose additional ports, you can do so like this, replacing <port> with the port number:
```
sharknoon, line 20: You accidentially left some comments in here
```suggestion
```

navidrome/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+      APP_PORT: 4533
+      PROXY_AUTH_ADD: "false"
+  server:
+    image: deluan/navidrome:0.52.5@sha256:b154aebe8b33bae82c500ad0a3eb743e31da54c3bfb4e7cc3054b9a919b685c7
+    restart: on-failure
+    environment:
+      - PUID=1000
+      - PGID=1000
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
```
sharknoon, line 14: Please create a file called `navidrome/data/.gitkeep`.
The problem is that on older versions of umbrelOS the permissions of automatically created repositories are wrong. By supplying a .gitkeep file, those directories are getting created by installing the app automatically.

navidrome/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+services:
+  app_proxy:
+    environment:
+      APP_HOST: navidrome_server_1
+      APP_PORT: 4533
+      PROXY_AUTH_ADD: "false"
+  server:
```
sharknoon, line 7: Maybe add a blank here to make the readability a bit better, but this is just a very minor nitpick
```suggestion
      PROXY_AUTH_ADD: "false"

  server:
```

sharknoon (changes requested):
Awesome update @owmsoton! You really made a lot of great changes :) I have left 3 suggestions below, after that I think we are ready to go :)

navidrome/docker-compose.yml
sharknoon: Or did you add them on purpose?

owmsoton: It really was an accident, I only realized it with your warning.

navidrome/docker-compose.yml
owmsoton: this change looks good.

navidrome/docker-compose.yml
owmsoton: now this worked, file created

owmsoton:
@sharknoon, Just a few small tweaks, but in the end, they make a huge difference for everyone.

Thanks again!

sharknoon:
Awesome work @owmsoton, thank you so much 👍 @nmfretz I think this app submission is ready to go 😃

nmfretz:
Excellent, thanks for this submission @owmsoton! And for the excellent review @sharknoon. We will get gallery assets prepared and then test the app before going live.

I noticed one minor change we can make:

In the `server` containers volumes you should be able to remove the `rw` so that the mount is just:
```yaml
- ${UMBREL_ROOT}/data/storage/music:/music
```

And then in the umbrel-app.yml we need to add the following for proper permissions to `${UMBREL_ROOT}/data/storage`
```yaml
permissions:
  - STORAGE_DOWNLOADS
```

See Plex for example: https://github.com/getumbrel/umbrel-apps/blob/master/plex/umbrel-app.yml

Can you please make this change and let us know if Navidrome is working correctly?

owmsoton:
> Can you please make this change and let us know if Navidrome is working correctly?

@nmfretz Application working perfectly with your management.

Thanks to everyone for the help and guidance.

nmfretz:
Thanks again for this submission @owmsoton! I have tested and made some final tweaks. The biggest change is that I have changed the bind mount for the music folder on the host such that it is the same folder used by other music apps in the umbrelOS app store. This means, for example, that you can download music files with Lidarr and they will automatically appear in Navidrome.
 
Gallery assets: https://github.com/getumbrel/umbrel-apps-gallery/tree/master/navidrome

image

Going live to the app store 🎉🎧

PR #1233, opened 2024-07-21
Immich update to 1.109.2

MalteKiefer:


nmfretz:
Thanks for updating Immich @MalteKiefer! 

I have added back in the client/server compatibility warning that we have been permanently including in the release notes:
```yaml
⚠️ As usual, please check that your mobile app is compatible with this release of Immich.
```

Immich frequently breaks compatibility between updates, so this warning prevents users going down a rabbit-hole trying to troubleshoot issues with their server when the issue is usually resolved by updating their mobile app.

Tested and going live.

PR #1242, opened 2024-07-24
Add mempool widget

mayankchhabra:
Screenshot 2024-07-24 at 8 34 23 PM


nmfretz:
Tested on arm64 and x86! Working beautifully.

PR #1243, opened 2024-07-24
btcpay-server update to 1.13.5

VnUgE:
Updated btcpay-server to 1.13.5. All other dependencies (postgress and nbxplorer) were verified up-to-date!

### DIFF
--- a/btcpay-server/docker-compose.yml
+++ b/btcpay-server/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   nbxplorer:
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks very much for this update @VnUgE! And thanks for going through and summarizing the release notes 🤝.

I noticed that the `nbxplorer` image actually has a more recent version we can update to (v2.5.5). Can you please update that image as well and then I will test on arm64 and x86?

VnUgE:
I totally mistook the 2 for a 5! Glad you caught it

nmfretz:
Perfect, thanks for updating @VnUgE!

I have tested and it is working well. Going live.

jimbrend:
gj yall :)

Issue #1086, opened 2024-05-12
 [App Request] Nextcloud Office


anthony-robin:
Hi Umbrel community 👋 

I would love to have a pretty and modern Office solution on Umbrel without going through an external server. As Nextcloud is already available in the App store, I think [Nextcloud Office](https://github.com/nextcloud/richdocuments) would be a very good fit to fill the lack of similar applications !

The documentation explains how Nextcloud Office could be integrated with Nextcloud: https://nextcloud.com/blog/how-to-install-nextcloud-office/
Here is another issue referencing a similar request: https://github.com/getumbrel/umbrel-apps/issues/575

Do you think this solution is doable and easy to add to Umbrel as a side application of Nextcloud ?

Thank you :)

anthony-robin:
Closed by #1288

PR #1288, opened 2024-07-29
App Submission: ONLYOFFICE Document Server for Nextcloud

a4004:
# App Submission

### 256x256 SVG icon
Icon

### I have tested my app on:
- [X] umbrelOS on Linux VM

### DIFF
--- a/onlyoffice-nextcloud/docker-compose.yml
+++ b/onlyoffice-nextcloud/docker-compose.yml
@@ -9,9 +9,9 @@ services:
     image: a4004/onlyoffice-nextcloud-web:1.0.4@sha256:774bd9bda641e6be49286921637840efba050fdfd022934dfd9ee9b6ec8d0a63
     restart: on-failure
     environment:
-      - DOCS_ADDRESS=http://$DEFAULT_INTERFACE_IP:$DOCSERVER_PORT
-      - DOCS_INTERNAL_ADDRESS=http://$DEFAULT_INTERFACE_IP:$DOCSERVER_PORT
-      - NEXTCLOUD_INTERNAL_ADDRESS=http://$DOCKER_INTERFACE_IP:$APP_NEXTCLOUD_PORT
+      - DOCS_ADDRESS=http://$DEVICE_DOMAIN_NAME:$DOCSERVER_PORT
+      - DOCS_INTERNAL_ADDRESS=http://onlyoffice-nextcloud_documentserver_1
+      - NEXTCLOUD_INTERNAL_ADDRESS=http://$GATEWAY_IP:$APP_NEXTCLOUD_PORT
       - NEXTCLOUD_WEB_URL=http://$DEVICE_DOMAIN_NAME:$APP_NEXTCLOUD_PORT
 
   documentserver:
@@ -19,7 +19,7 @@ services:
     restart: on-failure
     stop_grace_period: 1m
     ports:
-      - 5672:80
+      - "${DOCSERVER_PORT}:80"
     environment:
       - JWT_ENABLED=false
       - ONLYOFFICE_HTTPS_HSTS_ENABLED=false
--- a/onlyoffice-nextcloud/docker-compose.yml
+++ b/onlyoffice-nextcloud/docker-compose.yml
@@ -11,7 +11,7 @@ services:
     environment:
       - DOCS_ADDRESS=http://$DEVICE_DOMAIN_NAME:$DOCSERVER_PORT
       - DOCS_INTERNAL_ADDRESS=http://onlyoffice-nextcloud_documentserver_1
-      - NEXTCLOUD_INTERNAL_ADDRESS=http://$GATEWAY_IP:$APP_NEXTCLOUD_PORT
+      - NEXTCLOUD_INTERNAL_ADDRESS=http://nextcloud_web_1
       - NEXTCLOUD_WEB_URL=http://$DEVICE_DOMAIN_NAME:$APP_NEXTCLOUD_PORT
 
   documentserver:
--- a/onlyoffice-nextcloud/docker-compose.yml
+++ b/onlyoffice-nextcloud/docker-compose.yml
@@ -1,0 +1,33 @@
+version: '3.9'
+services:
+  app_proxy:
+    environment:
+      APP_HOST: onlyoffice-nextcloud_web_1
+      APP_PORT: 3000
+
+  web:
+    image: a4004/onlyoffice-nextcloud-web:1.0.4@sha256:774bd9bda641e6be49286921637840efba050fdfd022934dfd9ee9b6ec8d0a63
+    restart: on-failure
+    environment:
+      - DOCS_ADDRESS=http://$DEFAULT_INTERFACE_IP:$DOCSERVER_PORT
+      - DOCS_INTERNAL_ADDRESS=http://$DEFAULT_INTERFACE_IP:$DOCSERVER_PORT
+      - NEXTCLOUD_INTERNAL_ADDRESS=http://$DOCKER_INTERFACE_IP:$APP_NEXTCLOUD_PORT
+      - NEXTCLOUD_WEB_URL=http://$DEVICE_DOMAIN_NAME:$APP_NEXTCLOUD_PORT
+
+  documentserver:
+    image: onlyoffice/documentserver:8.1@sha256:9b74cb0e05580bced87f9af69de68d5f4c9e5b3340448ddc94333f5f7f4eba85
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - 5672:80
+    environment:
+      - JWT_ENABLED=false
+      - ONLYOFFICE_HTTPS_HSTS_ENABLED=false
+      - USE_UNAUTHORIZED_STORAGE=true
+      - ALLOW_PRIVATE_IP_ADDRESS=true
+      - ALLOW_META_IP_ADDRESS=true
+    volumes:
+      - ${APP_DATA_DIR}/data/logs:/var/log/onlyoffice
+      - ${APP_DATA_DIR}/data/certs:/var/www/onlyoffice/Data
+      - ${APP_DATA_DIR}/data/cache:/var/lib/onlyoffice
+      - ${APP_DATA_DIR}/data/db:/var/lib/postgresql
--- a/onlyoffice-nextcloud/exports.sh
+++ b/onlyoffice-nextcloud/exports.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+export DOCSERVER_PORT=5672
+export APP_DOCSERVER_PORT=3014
+
+export DEFAULT_INTERFACE_IP=$(ip addr show $(ip route | grep default | awk '{print $5}') | grep 'inet ' | awk '{print $2}' | cut -d/ -f1)
+export DOCKER_INTERFACE_IP=$(ip addr show docker0 | grep 'inet ' | awk '{print $2}' | cut -d/ -f1)
--- a/onlyoffice-nextcloud/exports.sh
+++ b/onlyoffice-nextcloud/exports.sh
@@ -1,7 +1,3 @@
 #!/usr/bin/env bash
 
 export DOCSERVER_PORT=5672
-export APP_DOCSERVER_PORT=3014
-
-export DEFAULT_INTERFACE_IP=$(ip addr show $(ip route | grep default | awk '{print $5}') | grep 'inet ' | awk '{print $2}' | cut -d/ -f1)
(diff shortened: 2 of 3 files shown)


a4004:
Gallery source images:

Document
nextcloud-light
Spreadsheet
Presentation
Forms DOCXF

onlyoffice-nextcloud/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      APP_HOST: onlyoffice-nextcloud_web_1
+      APP_PORT: 3000
+
+  web:
+    image: a4004/onlyoffice-nextcloud-web:1.0.4@sha256:774bd9bda641e6be49286921637840efba050fdfd022934dfd9ee9b6ec8d0a63
+    restart: on-failure
+    environment:
+      - DOCS_ADDRESS=http://$DEFAULT_INTERFACE_IP:$DOCSERVER_PORT
+      - DOCS_INTERNAL_ADDRESS=http://$DEFAULT_INTERFACE_IP:$DOCSERVER_PORT
+      - NEXTCLOUD_INTERNAL_ADDRESS=http://$DOCKER_INTERFACE_IP:$APP_NEXTCLOUD_PORT
```
nmfretz, line 14: I think we can simplify this a bit and avoid needing to run ip-related commands in the `exports.sh`. I'm not very familiar with the Nextcloud/ONLYOFFICE ecosystem, so please let me know if anything I suggest here doesn't make sense or has unintended consequences:

1) `DOCS_ADDRESS=http://$DEFAULT_INTERFACE_IP:$DOCSERVER_PORT`
I'm thinking that most users will be accessing Nextcloud via `umbrel.local` not their IP address, so if they are using their local IP address in the nextcloud/onlyoffice config they will get hit with cors errors when clicking on a document in Nextcloud from `http://umbrel.local:8081` and it tries to make a request to `$DEFAULT_INTERFACE_IP:5672`. 

Shall we change this to `DOCS_ADDRESS=http://$DEVICE_DOMAIN_NAME:$DOCSERVER_PORT` (which will be http://umbrel.local:5672) and then include instructions saying that if a user is accessing Nextcloud from a different hostname/ip, they should change make the appropriate change to `ONLYOFFICE Docs address` in their Nextcloud settings?

2) `DOCS_INTERNAL_ADDRESS=http://$DEFAULT_INTERFACE_IP:$DOCSERVER_PORT`
This is for internal requests to the `documentserver` container, so we can just use the container name here and Docker's internal DNS will resolve it to the correct IP address. So I think we can change this to `DOCS_INTERNAL_ADDRESS=http://onlyoffice-nextcloud_documentserver_1:$DOCSERVER_PORT`.
The user will then copy-paste `http://onlyoffice-nextcloud_documentserver_1` into the ONLYOFFICE settings in Nextcloud.

3) `NEXTCLOUD_INTERNAL_ADDRESS=http://$DOCKER_INTERFACE_IP:$APP_NEXTCLOUD_PORT`
I think we can get away without the export.sh ip command here as well. We actually have access to the gateway IP for the umbrel docker network as an env var. Check it out: https://github.com/getumbrel/umbrel/blob/570acdabcab13ac237fba992f491d126fff30810/packages/umbreld/source/modules/apps/legacy-compat/app-script#L10
So we should be able to use `NEXTCLOUD_INTERNAL_ADDRESS=http://$GATEWAY_IP:$APP_NEXTCLOUD_PORT` which will yield `http://10.21.0.1:8081/` for the user to copy-paste into the ONLYOFFICE settings in Nextcloud.

So a user would end up with this:

image

onlyoffice-nextcloud/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      - NEXTCLOUD_WEB_URL=http://$DEVICE_DOMAIN_NAME:$APP_NEXTCLOUD_PORT
+
+  documentserver:
+    image: onlyoffice/documentserver:8.1@sha256:9b74cb0e05580bced87f9af69de68d5f4c9e5b3340448ddc94333f5f7f4eba85
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - 5672:80
```
nmfretz, line 22: might as well use `- "${DOCSERVER_PORT}:80"` from exports.sh

onlyoffice-nextcloud/exports.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+export DOCSERVER_PORT=5672
+export APP_DOCSERVER_PORT=3014
```
nmfretz, line 4: Is this still needed, or can it be removed?

onlyoffice-nextcloud/exports.sh
```diff
@@ -0,0 +1,7 @@
...
+
+export DEFAULT_INTERFACE_IP=$(ip addr show $(ip route | grep default | awk '{print $5}') | grep 'inet ' | awk '{print $2}' | cut -d/ -f1)
+export DOCKER_INTERFACE_IP=$(ip addr show docker0 | grep 'inet ' | awk '{print $2}' | cut -d/ -f1)
```
nmfretz, line 7: If we end up going with the simplifications here (), then we can remove these exports.

nmfretz (commented):
Really awesome work on this @a4004. Your `onlyoffice-nextcloud-web` frontend is excellent, and instructions are super clear 👌

I've left some suggestions below, mostly around how we can simplify the environment variables/exports script.

onlyoffice-nextcloud/exports.sh
a4004: To elaborate on the name change, it is a good idea and will improve the appearance of the app, but I also think some improvements can be made to the whole app in light of that.

I'm also consdering adding a startup screen, similar to how the Bitcoin Node and Lightning Node apps have it with a progress bar, so if a user clicks on the ONLYOFFICE Docs app, they will be redirected to the actual app within Nextcloud in a few seconds with the opportunity to cancel and view the setup instructions instead. 

Ideally, there would be some configuration detection mechanism so it doesn't redirect upon the very first launch and only when it knows Nextcloud has been configured, but I think that's for a future update for now.

a4004:
Thanks for the suggestions @nmfretz! Happy with all of these:
- https://github.com/getumbrel/umbrel-apps/pull/1288#discussion_r1699463825 (name change)
- https://github.com/getumbrel/umbrel-apps/pull/1288#discussion_r1699506512 (using env for port)
- https://github.com/getumbrel/umbrel-apps/pull/1288#discussion_r1699507301 (we can remove the export)

I did some testing with the new configuration and there does seem to be an issue with using the container name for the *ONLYOFFICE Docs address for internal requests from the server* property. 

`onlyoffice-nextcloud_documentserver_1` correctly resolves on my machine howerver when the container IP is hit directly, it yields a HTTP/0.9 response that Nextcloud and Curl refuse to accept. 

Hitting the `docker0` or gateway IP (bridge device) works perfectly fine but using the the name or the container's actual IP from Docker doesn't work in my case.

So as a quick solution, we might be able to just use the `$GATEWAY_IP` for both the `DOCS_INTERNAL_ADDRESS` and `NEXTCLOUD_INTERNAL_ADDRESS`.

a4004:
To elaborate on the name change, it is a good idea and will improve the appearance of the app, but I also think some improvements can be made to the whole app in light of that.

I'm also consdering adding a startup screen, similar to how the Bitcoin Node and Lightning Node apps have it with a progress bar, so if a user clicks on the ONLYOFFICE Docs app, they will be redirected to the actual app within Nextcloud in a few seconds with the opportunity to cancel and view the setup instructions instead. 

Ideally, there would be some configuration detection mechanism so it doesn't redirect upon the very first launch and only when it knows Nextcloud has been configured, but I think that's for a future update for now.

a4004:
Tested it on my instance and all looks good now. Should be ready to go.

onlyoffice-nextcloud/docker-compose.yml
joaovictor-local: @nmfretz  Is it not better to use something like `http://nextcloud_web_1:8081`? We are using container hostname reference on torrenting apps like Prowlarr, Radarr and Sonarr.

image

a4004: That's actually a good point. 

I have tested both `http://nextcloud_web_1/` and `http://nextcloud_web_1:8081/` but it doesn't appear to work for some reason. Nextcloud is throwing an error `Error when trying to connect (Error occurred in the document service: Error while downloading the document file to be converted.) (version 8.1.1.26)`.

Perhaps I'm doing something wrong here or it's not set up for this type of communication, not too sure but I agree with the idea of using container hostnames over IPs for sure.

nmfretz: @joaovictor-local @a4004, ya the issue here is that the Nextcloud app doesn't have the `web` container name as a trusted proxy in its `config.php`, so instead we are using an IP that it does trust (e.g., `10.21.0.1` or `172.17.0.1`).

https://github.com/getumbrel/umbrel-apps/blob/49580db282c58ac50f077f2d83f998f71f460d17/nextcloud/docker-compose.yml#L47
https://github.com/getumbrel/umbrel-apps/blob/49580db282c58ac50f077f2d83f998f71f460d17/nextcloud/exports.sh#L3-L4

For this to work, the Nextcloud config.php would have to include `nextcloud_web_1/` under trusted_domains. Then in the field for `Server address for internal requests from ONLYOFFICE Docs` a user would use `http://nextcloud_web_1/`, without a port specified since we want the default port 80 **inside the web container** (not 8081).

I do really like the simplicity of using container names @joaovictor-local, so we could consider the following:

- Update the Nextcloud app to include `nextcloud_web_1/` in the config.php
- Change the env var in ONLYOFFICE Docs to `NEXTCLOUD_INTERNAL_ADDRESS=http://nextcloud_web_1`
- Include a note in the `a4004/onlyoffice-nextcloud-web` UI (or maybe just the app description) telling users to update Nextcloud if they haven't.

What do you guys think?

nmfretz:
Thanks @a4004!

> Tested it on my instance and all looks good now. Should be ready to go.

Looks like you may have accidentally pushed the changes to a different branch https://github.com/a4004/umbrel-apps/commit/3daa6b91fcbb5b2c147f159e3b5ce2b7c40fd03b

> onlyoffice-nextcloud_documentserver_1 correctly resolves on my machine howerver when the container IP is hit directly, it yields a HTTP/0.9 response that Nextcloud and Curl refuse to accept.

Just noting for our records that `onlyoffice-nextcloud_documentserver_1` does indeed work as long as port 80 is hit (and not 8081 which is for the app proxy).

> I'm also consdering adding a startup screen, similar to how the Bitcoin Node and Lightning Node apps have it with a progress bar, so if a user clicks on the ONLYOFFICE Docs app, they will be redirected to the actual app within Nextcloud in a few seconds with the opportunity to cancel and view the setup instructions instead.

This is a cool idea, but honestly I think there's a lot of benefit to having this app just stay on the instructions page because:
1) Opening the app in Nextcloud just means viewing the admin settings for onlyoffice (is that right?).
2) The onlyoffice integration in nexcloud is just used automatically when clicking on files. You don't actually go to some main onlyoffice homescreen.
3) If a user needs to troubleshoot connection issues they can just open your onlyoffice-nextcloud app to see instructions.

That being said, this is your app, so it's totally up to you!

Also, @joaovictor-local @a4004 I've responded here with a suggestion on how we could use the nextcloud web container name: https://github.com/getumbrel/umbrel-apps/pull/1288#discussion_r1703459924

nmfretz:
Thanks for pushing the changes @a4004. Absolutely fantastic work on this app! Going live to the app store 🎉.

FYI, I have done the following to finalize:
- Changed `NEXTCLOUD_INTERNAL_ADDRESS` to use the nextcloud web container name and Docker dns (i.e., `http://nextcloud_web_1`). To make sure this works with the Nextcloud app on umbrelOS I have updated Nextcloud to include nextcloud_web_1 as a trusted domain for both new installs and pre-29.0.4 apps that update https://github.com/getumbrel/umbrel-apps/pull/1297
- updated the tagline and app description to make it clear that this is for Nextcloud integration.

image

onlyoffice-nextcloud/docker-compose.yml
nmfretz: Update: I have updated Nextcloud to allow connection via nextcloud_web_1 https://github.com/getumbrel/umbrel-apps/pull/1297

anthony-robin:
Thank you very much for making ONLYOFFICE available to Umbrel and, furthermore, associated to NextCloud :heart: :tada: 

This PR closes #1086

emtreulapollaguera:
Hello,

This app needs update to new Nexcloud version 35.

The compatibility is broken since Umbrel Nextcloud has been updated to version 35.

Issue #1292, opened 2024-08-01
Since the update BTCPay to 1.13.5 the app no ​​longer works - "BTC node synching - Name or service not known (nbxplorer:32838)"


JurGi7:
According to Dennis, the problem seems to be in umbrel-apps.

The problem is described here: https://github.com/btcpayserver/btcpayserver/issues/6128

nmfretz:
Thanks for bringing this to our attention @JurGi7. And to @ElucGeek for the btcpayserver repo issue and @dennisreimann for the PR fix 🤝

PR #1293, opened 2024-08-01
BTCPay Server: Remove deprecated NBXplorer options

dennisreimann:
Fixes getumbrel/umbrel-apps#1292 and btcpayserver/btcpayserver#6128.


nmfretz:
Thanks for this fix @dennisreimann! Really appreciate it. Interestingly, I was not able to reproduce the issue when testing just now.

I have updated the app manifest so that current installs see an available update.

Tested and going live.

PR #1296, opened 2024-08-05
App Submission: NPM (Nginx Proxy Manager)

sahilph:
# App Submission

### App name
Nginx Proxy Manager

> ⚠️ Please take care while exposing apps to the internet.
  
> ⚠️ Note: This app listen on ports 40080 and 40443. In order to make the apps proxied by NPM accessible from internet, users need to port forward 80 to 40080 and 443 to 40443 on their router.⚠️

[DETAILED GUIDE](https://github.com/sahilph/sahil-umbrel-apps/wiki/How-to-use-Nginx-Proxy-Manager-on-your-Umbrel)

### 256x256 SVG icon
```
https://nginxproxymanager.com/logo.svg
```

### Gallery images

Image 1:

Image 2:

Image 3:

Image 4:

Note: These screenshots are taken from NPM website.

### I have tested my app on:
- [x] umbrelOS on Linux VM

I do not have Umbrel Home or Raspberry Pi, hence couldn't test it on them. However I have tested it on my Linux Machine (Umbrel OS 1.2.2) and it works fine.

---

For those who want to try it now, Nginx Proxy Manager is also available on [ my personal app store](https://github.com/sahilph/sahil-umbrel-apps)

### DIFF
--- a/nginx-proxy-manager/docker-compose.yml
+++ b/nginx-proxy-manager/docker-compose.yml
@@ -1,0 +1,22 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: nginx-proxy-manager_web_1
+      APP_PORT: 81
+  web:
+    image: jc21/nginx-proxy-manager:2.11.3@sha256:5bd2aae54ac0b6f4ef6777143ab736baf5366807894a4c7fca7bfa2fbaf1489c
+    hostname: nginx-proxy-manager_web_1
+    restart: on-failure
+    ports:
+      - 443:443
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${APP_DATA_DIR}/letsencrypt:/etc/letsencrypt
+    environment:
+      PUID: 1000
+      PGID: 1000
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+      - "${APP_DOMAIN}:host-gateway"
--- a/nginx-proxy-manager/docker-compose.yml
+++ b/nginx-proxy-manager/docker-compose.yml
@@ -23,5 +23,7 @@ services:
   docker-host:
     image: qoomon/docker-host:3.3.1@sha256:ec9d214d7fa88932e9186ca2f4419d44661376929931248525b08e7fa9e270f7
     hostname: $DEVICE_DOMAIN_NAME
-    cap_add: [ 'NET_ADMIN', 'NET_RAW' ]
+    cap_add:
+      - NET_ADMIN
+      - NET_RAW
     restart: on-failure
--- a/nginx-proxy-manager/docker-compose.yml
+++ b/nginx-proxy-manager/docker-compose.yml
@@ -21,7 +21,7 @@ services:
       PGID: 1000
   
   docker-host:
-    image: qoomon/docker-host:3@sha256:ec9d214d7fa88932e9186ca2f4419d44661376929931248525b08e7fa9e270f7
+    image: qoomon/docker-host:3.3.1@sha256:ec9d214d7fa88932e9186ca2f4419d44661376929931248525b08e7fa9e270f7
     hostname: umbrel.local
     cap_add: [ 'NET_ADMIN', 'NET_RAW' ]
     restart: on-failure
--- a/nginx-proxy-manager/docker-compose.yml
+++ b/nginx-proxy-manager/docker-compose.yml
@@ -19,6 +19,9 @@ services:
     environment:
       PUID: 1000
       PGID: 1000
-    extra_hosts:
-      - "host.docker.internal:host-gateway"
-      - "${APP_DOMAIN}:host-gateway"
+  
+  docker-host:
+    image: qoomon/docker-host@sha256:ec9d214d7fa88932e9186ca2f4419d44661376929931248525b08e7fa9e270f7
+    hostname: umbrel.local
+    cap_add: [ 'NET_ADMIN', 'NET_RAW' ]
+    restart: on-failure
--- a/nginx-proxy-manager/docker-compose.yml
+++ b/nginx-proxy-manager/docker-compose.yml
@@ -5,6 +5,7 @@ services:
     environment:
       APP_HOST: nginx-proxy-manager_web_1
       APP_PORT: 81
+      
   web:
     image: jc21/nginx-proxy-manager:2.11.3@sha256:5bd2aae54ac0b6f4ef6777143ab736baf5366807894a4c7fca7bfa2fbaf1489c
     hostname: nginx-proxy-manager_web_1
--- a/nginx-proxy-manager/docker-compose.yml
+++ b/nginx-proxy-manager/docker-compose.yml
@@ -10,10 +10,11 @@ services:
     hostname: nginx-proxy-manager_web_1
     restart: on-failure
     ports:
-      - 443:443
+      - 40080:80
+      - 40443:443
     volumes:
-      - ${APP_DATA_DIR}/data:/data
-      - ${APP_DATA_DIR}/letsencrypt:/etc/letsencrypt
+      - ${APP_DATA_DIR}/data/npm-data:/data
+      - ${APP_DATA_DIR}/data/letsencrypt:/etc/letsencrypt
     environment:
       PUID: 1000
       PGID: 1000
--- a/nginx-proxy-manager/docker-compose.yml
+++ b/nginx-proxy-manager/docker-compose.yml
@@ -22,6 +22,6 @@ services:
   
   docker-host:
     image: qoomon/docker-host:3.3.1@sha256:ec9d214d7fa88932e9186ca2f4419d44661376929931248525b08e7fa9e270f7
-    hostname: umbrel.local
+    hostname: $DEVICE_DOMAIN_NAME
     cap_add: [ 'NET_ADMIN', 'NET_RAW' ]
     restart: on-failure
--- a/nginx-proxy-manager/docker-compose.yml
+++ b/nginx-proxy-manager/docker-compose.yml
@@ -21,7 +21,7 @@ services:
       PGID: 1000
   
   docker-host:
-    image: qoomon/docker-host@sha256:ec9d214d7fa88932e9186ca2f4419d44661376929931248525b08e7fa9e270f7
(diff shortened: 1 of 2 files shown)


nginx-proxy-manager/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+  app_proxy:
+    environment:
+      APP_HOST: nginx-proxy-manager_web_1
+      APP_PORT: 81
+  web:
+    image: jc21/nginx-proxy-manager:2.11.3@sha256:5bd2aae54ac0b6f4ef6777143ab736baf5366807894a4c7fca7bfa2fbaf1489c
+    hostname: nginx-proxy-manager_web_1
+    restart: on-failure
+    ports:
+      - 443:443
```
nmfretz, line 13: Nice job realizing that port 80 is used by umbrelOS 👌

What we can actually do here is map ports 80 and 443 from the `nginx-proxy-manager` container to different ports on the host. So for example:
```suggestion
    ports:
      - 40080:80
      - 40443:443
```

Then in the app description we could clearly state that users need to port forward 80 to 40080 and 443 to 40443 on their router.

This way we completely avoid current clashes with port 80 and future clashes with port 443. What do you think?

nginx-proxy-manager/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${APP_DATA_DIR}/letsencrypt:/etc/letsencrypt
```
nmfretz, line 16: Let's put both of these inside the `${APP_DATA_DIR}/data` directory. So something like:
```suggestion
    volumes:
      - ${APP_DATA_DIR}/data/npm-data:/data
      - ${APP_DATA_DIR}/data/letsencrypt:/etc/letsencrypt
```

You'll need to make the equivalent changes to the subdirectories committed in this PR

nginx-proxy-manager/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+    environment:
+      PUID: 1000
+      PGID: 1000
```
nmfretz, line 19: Excellent work running this as non-root 🤝

nginx-proxy-manager/umbrel-app.yml
```diff
@@ -0,0 +1,47 @@
+manifestVersion: 1
+id: nginx-proxy-manager
+name: Nginx Proxy Manager
+tagline: Expose your services easily and securely
+icon: https://nginxproxymanager.com/logo.svg
```
nmfretz, line 5: This can be deleted entirely. We will host the icon at https://github.com/getumbrel/umbrel-apps-gallery

nginx-proxy-manager/umbrel-app.yml
```diff
@@ -0,0 +1,47 @@
...
+      
+      - Advanced Nginx configuration available for super users
+      
+      - User management, permissions and audit log
+      
+  
+  ⚠️ Please take care while exposing apps to the internet.
+  
+  ⚠️ Note: Since Port 80 is used my Umbrel itself, it will not be handled by Nginx Proxy Manager. Hence do not to forward Port 80 from your router. Only forward Port 443. ⚠️
+
```
nmfretz, line 31: We can update this with instructions on which ports to forward if we bind 40443 and 40080. I'll do this once the other suggestions are committed.

nginx-proxy-manager/umbrel-app.yml
```diff
@@ -0,0 +1,47 @@
...
+      - Advanced Nginx configuration available for super users
+      
+      - User management, permissions and audit log
+      
+  
+  ⚠️ Please take care while exposing apps to the internet.
+  
+  ⚠️ Note: Since Port 80 is used my Umbrel itself, it will not be handled by Nginx Proxy Manager. Hence do not to forward Port 80 from your router. Only forward Port 443. ⚠️
+  
+developer: Jamie Curnow (jc21.com)
```
nmfretz, line 32: We can remove the website link here
```suggestion
developer: Jamie Curnow
```

nginx-proxy-manager/umbrel-app.yml
```diff
@@ -0,0 +1,47 @@
...
+website: https://nginxproxymanager.com/
+submitter: sahilph
+submission: https://github.com/getumbrel/umbrel-apps/pull/1296
+repo: https://github.com/sahilph/sahil-umbrel-apps/
+support: https://github.com/sahilph/sahil-umbrel-apps//issues
```
nmfretz, line 37: These should be changed to the nginx-proxy-manager github:
```suggestion
repo: https://github.com/NginxProxyManager/nginx-proxy-manager
support: https://github.com/NginxProxyManager/nginx-proxy-manager/issues
```

nginx-proxy-manager/umbrel-app.yml
```diff
@@ -0,0 +1,47 @@
...
+gallery:
+  - https://nginxproxymanager.com/screenshots/login.png
+  - https://nginxproxymanager.com/screenshots/dashboard.png
+  - https://nginxproxymanager.com/screenshots/proxy-hosts.png
+  - https://nginxproxymanager.com/screenshots/proxy-hosts-add.png
```
nmfretz, line 42: We'll host the gallery assets here https://github.com/getumbrel/umbrel-apps-gallery, so this can be changed to:
```suggestion
gallery:
  - 1.jpg
  - 2.jpg
  - 3.jpg
```

nmfretz (commented):
Really great addition @sahilph, thanks for working on this!

This is almost ready to go, there are just a few items to address below. We'll start working on gallery assets.

nginx-proxy-manager/docker-compose.yml
sahilph: Yes, this  should be fine. Forwarding both 80 and 443 ports will make HTTP-01 challenge work as well.

nginx-proxy-manager/umbrel-app.yml
sahilph: Ok, will do

sahilph:
@nmfretz 
I have made the requested changes. Please check.

nmfretz:
> @nmfretz
I have made the requested changes. Please check.

Thanks for making those changes @sahilph! I have pushed some changes to finalize the app (app description, UI port, etc). We should have the gallery assets ready soon and can then go live.

In the meantime, would you be able to confirm if using `umbrel.local` as a forward hostname / ip when configuring a proxy host works for you?

I have just been doing some testing and I can't get this to work even though you have added `"${APP_DOMAIN}:host-gateway"` to the `extra_hosts` directive in the compose file. What this does is add `172.17.0.1   umbrel.local` to the container's `/etc/hosts` file. But even so, it does not appear to work.

Using Jellyfin as an example:

1. Forward Hostname / IP = `umbrel.local` | port = `8096` --> I am hit with a 502 bad gateway
2.  Forward Hostname / IP = `127.17.0.1` | port = `8096` --> This works
3.  Forward Hostname / IP = `10.21.0.1` | port = `8096` --> This works. 10.21.0.1 is the Gateway IP of the main Docker umbrel network.
4. Forward Hostname / IP = jellyfin_server_1 | port = `8096` --> This works.

The only reason 2 and 3 work in this case is because our app_proxy container sitting in front of Jellyfin has the authentication disabled (Jellyfin has it's own auth):
https://github.com/getumbrel/umbrel-apps/blob/92d5832bc9bb225ac242bdc6b8cab0b25a21f437/jellyfin/docker-compose.yml#L8

For any apps that sit behind the app_proxy with authentication enabled (by default), they would need to proxy using case 4 above, where they bypass the the app_proxy and instead proxy directly to the main app container. If you want to test that out try proxying to the Radarr app using 127.17.0.1 or 10.21.0.1 and it should not work.

sahilph:
> Forward Hostname / IP = `umbrel.local` | port = `8096` --> I am hit with a 502 bad gateway

Ahh, this is interesting. I get the same 502 bad gateway. I was surprised by this, considering `/etc/hosts` already had `umbrel.local` in it.

It turns out that nginx ignores `/etc/hosts` when using proxy pass.
Related issue: https://github.com/NginxProxyManager/nginx-proxy-manager/issues/1225

So one option is to ask users to directly use the IP `172.17.0.1` instead of  `umbrel.local`. But not sure if IP is always the same for everyone. Also it not really that convenient as well 

Hence, as a workaround, I found a docker image [qoomon/docker-host](https://github.com/qoomon/docker-host) which basically accepts traffic on all ports and forwards it to the host.  With this users will be able to use  `umbrel.local` as Forward hostname.

I have added a new commit please check.

sahilph:
> For any apps that sit behind the app_proxy with authentication enabled (by default), they would need to proxy using case 4 above, where they bypass the the app_proxy and instead proxy directly to the main app container. If you want to test that out try proxying to the Radarr app using 127.17.0.1 or 10.21.0.1 and it should not work.

Yes thats correct, because I think the app_proxy redirects to `domain.name:2000` to ask for authentication and it wont work as port 2000 will not be accessible from the internet.  Maybe we can add a note about this.

nmfretz:
>It turns out that nginx ignores /etc/hosts when using proxy pass.
Related issue: https://github.com/NginxProxyManager/nginx-proxy-manager/issues/1225

Ah I see, good sleuthing.

> So one option is to ask users to directly use the IP 172.17.0.1 instead of umbrel.local. But not sure if IP is always the same for everyone. Also it not really that convenient as well
>
>Hence, as a workaround, I found a docker image [qoomon/docker-host](https://github.com/qoomon/docker-host) which basically accepts traffic on all ports and forwards it to the host. With this users will be able to use umbrel.local as Forward hostname.

Great find with `docker-host`! A bit unfortunate that we need to use it because the container needs elevated capabilities to run (NET_ADMIN and NET_RAW), but we currently have precedence for this in the app store with other apps like adguard/pihole/wireguard, so we should allow this here for the functionality it provides. In the future we can look into locking this down more.

I have added a dynamic environment variable for hostname (https://github.com/getumbrel/umbrel-apps/pull/1296/commits/9151cf1ae786b65a4f07abb58a6beadc0e8ad759) and have tested the app, including using umbrel.local as a forward hostname. Going live! Thanks again for all your hard work bringing this to the app store 🚀

image

sahilph:
Awesome. Thanks 👍

PR #1297, opened 2024-08-06
Update Nextcloud to  v29.0.4

nmfretz:
This is a major upgrade from version 28.0.3 to 29.0.4. The Nextcloud app on the Umbrel app store will now receive updates in alignment with the official Nextcloud release schedule moving forward.

- For new installs: added the nextcloud web container name to  the`NEXTCLOUD_TRUSTED_DOMAINS` env var so that apps like ONLYOFFICE Docs can communicate via Docker dns
- For existing installs: added a post-update hook that adds the nextcloud web container name to trusted_domains in config.php.

Tested fresh install and update.


dennysubke:
Thanks for the update and the good work.

After the update I get the following error messages:

**- Your data directory and files are probably accessible from the Internet. The .htaccess file is not working. It is strongly recommended that you configure your web server so that the data directory is no longer accessible from the Internet or that you move it out of the document root directory of the web server.

- Your "trusted_proxies" setting is not set correctly, it should be an array of IP addresses - optionally with a range in CIDR notation. You can find more information in the documentation.**

I use Cloudflare and have already entered all IP address ranges in the Config.php under Trusted Proxies. Any ideas?

nmfretz:
Thanks for reporting and reaching out @dennysubke. Both of these error messages are likely related to the trusted_proxies array in the config.php file. It seems to be a pretty common Nextcloud issue in v29 from looking at this thread: https://help.nextcloud.com/t/after-upgrade-from-28-to-29-i-have-data-directory-and-your-files-are-probably-accessible-from-the-internet/189882/42

Is there anything in there that helps explain your situation?

nmfretz:
Moving discussion to https://community.umbrel.com/t/nextcloud-update-to-v29-0-4-error-messages-and-security-warnings/18705.

PR #1328, opened 2024-08-11
Bump LNBits to 0.12.10

mroxso:
This bumps LNBits to the latest version

### DIFF
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   web:
-    image: lnbits/lnbits:0.12.2@sha256:ce11b1127ab9a6314ed5a57792dcd938ee189c2f549888f6cdda21a181a12fc9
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this update of LNBits @mroxso!

There's a few things we need to do before merging to the app store:

1) Use the multi-arch digest for lnbits Docker image so that both x86(amd64) and arm64 users are supported. Right now, the image digest used is only for `linux/amd64`, so arm64 users who attempt to install/update LNbits will end up with a broken install.

2) Add release notes details to `umbrel-app.yml`

3) Test the app to make sure fresh installs work and that data is persisted for app updates (e.g., we need to make sure users will not lose funds).

Would you like to take the lead on this, or would you be okay with me pushing changes to your branch?

nmfretz:
Thanks again for this PR @mroxso. I am closing this in favour of this one here where I have asked the LNbits team for advice on the update:  https://github.com/getumbrel/umbrel-apps/pull/1370

Issue #1358, opened 2024-08-14
[Update Request] LNbits to 0.12.10


Ghooosstt:
Hello,

Is it possible to update the LNbits version from 0.12.2 to 0.12.10?
https://github.com/lnbits/lnbits/releases/tag/0.12.10

Thank you

nmfretz:
Thanks @Ghooosstt. Closing as completed by your PR https://github.com/getumbrel/umbrel-apps/pull/1370 💪

PR #1370, opened 2024-08-15
Update LNbits to 0.12.10 #1358

Ghooosstt:
Update LNbits to the latest version

### DIFF
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   web:
-    image: lnbits/lnbits:0.12.2@sha256:ce11b1127ab9a6314ed5a57792dcd938ee189c2f549888f6cdda21a181a12fc9
+    image: lnbits/lnbits:0.12.10@sha256:7baf6f5bc78707f9923968b8f5ec9142cb589841e2b58736dbf96585531941e0
     init: true
     restart: on-failure
     stop_grace_period: 1m
--- a/lnbits/docker-compose.yml
+++ b/lnbits/docker-compose.yml
@@ -33,4 +33,3 @@ services:
       LNBITS_DISABLED_EXTENSIONS: "amilk"
       LNBITS_ADMIN_EXTENSIONS: "ngrok"
       LNBITS_ADMIN_UI: "true"
-      SUPER_USER: "$APP_PASSWORD"
--- a/lnbits/umbrel-app.yml
+++ b/lnbits/umbrel-app.yml
@@ -26,7 +26,7 @@ gallery:
   - 3.jpg
 path: ""
 defaultUsername: ""
-deterministicPassword: true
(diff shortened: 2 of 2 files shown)


Ghooosstt:
This is the latest version of LNbits and it has a lot of changes/updates, but I don't know how to test it on my umbrel which works on my raspberry pi 5. Anyone is able to check if the change of the Docker image is sufficient for this to work or if any further modifications are needed?

nmfretz:
Thanks very much for this update @Ghooosstt. And sorry for the delay on my end while I've been away. 

I have tested and this is working well for me. However, before updating this I just want to bring in the LNbits team to check something:

@motorina0, I understand that LNbits allows super user registration as of 0.12.0. Prior to 0.12.0 you guys added a way to create a superuser by using the environment variable `SUPER_USER` in the docker-compose.yml and then having the user navigate to something like `umbrel.local:3007/uuidv4/<your-lnbits-password>`. This workaround allowed umbrel users to have an admin account that they could manage extensions from.

Context: https://github.com/lnbits/lnbits/pull/1688

What I'm wondering is if we are now okay to remove the `SUPER_USER` environment variable from the [docker-compose.yml ](https://github.com/getumbrel/umbrel-apps/blob/b395eff870d36287b8608e071fbb4e490348ecc3/lnbits/docker-compose.yml#L36)and get rid of the current superuser instructions:

image

In my testing, removing the above items works fine because:

1) **Fresh** installs (of 0.12.0 and up) are prompted to create a superuser account through the UI:

image

2) Legacy installs on umbrelOS that were installed pre 0.12.0 can still access their super user accounts by either:
a) logging in through the UI with the appropriate `user ID` from their legacy admin account:
image

b) using either the old format from bookmarking back in the day:
`http://umbrel.local:3007/wallet?usr=<user-id>&wal=<wallet-id>

or even using the uuidv4 method:
`umbrel.local:3007/uuidv4/<password from the app store>`

Again, this works fine in my testing, but I'd appreciate confirmation or letting me know if I've missed something.

Ghooosstt:
Thank you for your help on this pull request @nmfretz! 🧡
I hope @motorina0 will approve your analysis!

nmfretz:
You're welcome @Ghooosstt. I've contacted Vlad offline and he can take a look at this early next week.

motorina0:
@nmfretz you are correct. The new `fresh install` page will allow the user to directly create the admin account.
Therefore the docker environment is not required anymore.

LGTM, thanks for updating LNbits!

motorina0 (approved):

nmfretz:
Excellent, thanks very much for the review and confirmation @motorina0. I have cleaned up the legacy superuser items from the compose file and app manifest, and added a reminder in the release notes for legacy LNbits installs on how to access their admin accounts.

Tested again. Going live.

Ghooosstt:
Thank you so much @nmfretz and @motorina0 !!

PR #1376, opened 2024-08-24
App Submission - Zabbix

DavidGarciaCat:
I created this Zabbix app for my [community store](https://github.com/deralia/umbrel-app-store), so I could add it to my Umbrel OS easily.

Given that there are other network tools and that Zabbix is one of the most useful monitoring tools available, it might be interesting to have it here, too.

Feedback is more than welcome!


zabbix/data/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
sharknoon, line 2: I am not sure, what this file does. If you only want to keep the data directory when syncing it with git, please replace this file with an empty `.gitkeep` file.

zabbix/umbrel-app.yml
```diff
@@ -0,0 +1,36 @@
...
+icon: https://raw.githubusercontent.com/deralia/umbrel-app-store/master/deralia-zabbix/assets/icon.png
+version: "7.0.3"
+description: >-
+  A Zabbix Server + Frontend interface to monitor your network.
+
+
+  The first time running Zabbix might take a few minutes because it needs to initialize the database. Please be patient.
+
+
+  The Zabbix Agent listens on port 10050, the Zabbix Server listens on port 10051 and Zabbix Frontend listens on port 10052.
```
sharknoon, line 15: Good catch on the two newlines 😀

sharknoon (changes requested):
Thank you @DavidGarciaCat for your App Submission! Really good work.
I have only one question in this review (see below). I will test the app and give you feedback.

zabbix/data/.gitignore
DavidGarciaCat: Actually, there is no official convention or usage for `.gitkeep`, while it has already been highlighted in other projects and repositories that the correct way to keep a folder is using `.gitignore`.

This has been discussed, for example, in Sulu and Symfony repositories (that I can remember).

May I ask if there is an “official” reason to use `.gitkeep` instead of using the official `.gitignore` files, as these ones are the only ones with official support for Git?

zabbix/umbrel-app.yml
DavidGarciaCat: Thanks, I guess :-)

I noticed this “requirement” while testing it for my community repo

Resolving this conversation and waiting for more feedback on the other one

zabbix/data/.gitignore
DavidGarciaCat: And here, we can see guides for attributes or ignore, but not for keep

https://git-scm.com/docs

sharknoon: Thank you for the comprehensive answer @DavidGarciaCat. I didn't knew, that a `.gitkeep` file is just a convention but not actually implemented in git itself. I would propose to use `.gitkeep` instead of `.gitignore` for the following reasons:
- Consistency between other apps: All other apps use `.gitkeep` instead of `.gitignore`
- The linter used in this repository searches for `.gitkeep` files in directories
In general, both would work, I have found a good Stack Overflow article about that topic [here](https://stackoverflow.com/a/7229996)

DavidGarciaCat: Given the idea of keeping a folder there is to ensure that folder path is synchronised with Git, do you want me to add the `.gitignore` file in the root of the app and target the data, but exclude the `.gitkeep` file so it remains here?

Maybe it is a silly question, but I don't 100% understand yet how apps are handled after they have been downloaded fro the app store.

sharknoon: Sorry hat I haven't clarified that. When Docker starts a docker compose file with a volume binding, it automatically creates those directories. But those are being created with the same user as docker runs (probably root). This can lead to permission errors, if a container with a different user tries to access that folder.

To solve this, we are creating the folder beforehand in the cloning step via git. This is executed with a local user and therefore the permissions are set correctly.

DavidGarciaCat: I understand there is no need to move the `.gitignore` to the root folder and target the `data` folder we want to keep as part of the git synching. I will push one more commit today, as soon as I can.

DavidGarciaCat: Just pushed the requested update. Resolving this conversation.

sharknoon:
Thank you so much @DavidGarciaCat for your contribution 😃
@nmfretz I think this app is ready 👍

sharknoon:
Thank you so much @DavidGarciaCat for your work! Two things I have noticed:
- A `.gitkeep` file is missing for `data/postgres`
- The use of the standard container user. Can you try to add `user: 1000:1000` to the containers? It could be that some containers won't start with a different user, but those who do work fine with the new user are properly sandboxed.

sharknoon:
Thank you for your quick response 😀 
Yes, please move it to `data/postgres` instead. In the current case, docker would create the `postgres` directory by itself, with wrong permissions. Therefore we unfortunately have to create for each empty directory a `.gitkeep` file.
Thanks for checking the UID / GID

zabbix/data/.gitignore
nmfretz: Thanks for the .gitkeep/.gitignore notes @DavidGarciaCat!

In the case of umbrelOS we've gone with the .gitkeep convention and need to stick with this for the current umbrelOS implementation. This is because umbrelOS explicitly looks for `.gitkeep` files and deletes them so that a dotfile is not present in a directory that would prevent an app from working correctly. For example, many databases will not initialize properly if they detect a dotfile present.

https://github.com/getumbrel/umbrel/blob/7103e98909041bb8e15702b4b2e3e7f78da4b351/packages/umbreld/source/modules/apps/app-repository.ts#L85-L87

DavidGarciaCat: Thanks, Nathan, for pointing to this line in the code management.

DavidGarciaCat:
@sharknoon I was checking the note for the user, and running `cat /etc/passwd` I can see there's no `1000` user we can use here.

However, running `whoami`, I can see the primary user seems to be `zabbix` with UID `1997`:

```
$ whoami
zabbix

$ cat /etc/passwd
...
zabbix:x:1997:1995:Zabbix monitoring system:/var/lib/zabbix/:/sbin/nologin
```

This user is already set as the default user for the container when pulling the images.

Do we still need to make adjustments for the containers?

sharknoon:
Hi @DavidGarciaCat 
thank you for the update! If there is already a user other than root, this is totally fine 😀 Good catch!

DavidGarciaCat:
@sharknoon, thanks for your confirmation

@nmfretz, I think all feedback has been implemented - are we ready to go live with this new app?

zabbix/docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+    hostname: zabbix-frontend
+    restart: on-failure
+    environment:
+      DB_SERVER_HOST: zabbix-database
+      POSTGRES_USER: zabbix
+      POSTGRES_PASSWORD: zabbix
+      ZBX_SERVER_HOST: zabbix-server
+      PHP_TZ: UTC
+    ports:
+      - "10052:8080"
```
nmfretz, line 18: We can add our app_proxy container as the first one in this compose file and point it towards zabbix_zabbix-frontend_1 port 8080. We can then remove this port mapping. See Radarr as an example:

https://github.com/getumbrel/umbrel-apps/blob/master/radarr/docker-compose.yml#L4-L8

zabbix/umbrel-app.yml
```diff
@@ -0,0 +1,39 @@
...
+support: https://www.zabbix.com/support
+port: 10052
+gallery:
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/01.png
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/02.png
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/03.png
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/04.png
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/05.png
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/06.png
+path: "/admin/"
```
nmfretz, line 32: This makes it so that when a user clicks on the Zabbix icon they are taken to `http://umbrel.local:10052/admin/`. But this path results in a 404 response. I haven't tested, but you can likely just change this to 
```suggestion
path: ""
```

nmfretz (commented):
Thanks for the submission @DavidGarciaCat. I have left a review below with some items to address.

Let @sharknoon and I know if anything is confusing!

zabbix/umbrel-app.yml
```diff
@@ -0,0 +1,39 @@
+manifestVersion: 1.1
+id: zabbix
+category: networking
+name: Zabbix
+tagline: The all-in-one, open-source solution that lets you monitor anything
+icon: https://raw.githubusercontent.com/deralia/umbrel-app-store/master/deralia-zabbix/assets/icon.png
```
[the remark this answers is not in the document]
DavidGarciaCat: I have committed and pushed the requested change for the icon

zabbix/umbrel-app.yml
```diff
@@ -0,0 +1,39 @@
...
+repo: https://github.com/zabbix/zabbix
+support: https://www.zabbix.com/support
+port: 10052
+gallery:
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/01.png
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/02.png
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/03.png
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/04.png
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/05.png
+  - https://raw.githubusercontent.com/getumbrel/umbrel-apps/master/zabbix/assets/06.png
```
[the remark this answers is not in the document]
DavidGarciaCat: I have committed and pushed the requested change for the gallery

zabbix/umbrel-app.yml
```diff
@@ -0,0 +1,39 @@
...
+path: "/admin/"
+defaultUsername: "Admin"
+defaultPassword: "zabbix"
+torOnly: false
+releaseNotes: >-
+
```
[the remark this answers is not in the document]
DavidGarciaCat: I have committed and pushed the requested change for the release notes

zabbix/umbrel-app.yml
DavidGarciaCat: Nice catch! That was here because I used Pi-Hole's settings as the template file for me. Pushing now!

zabbix/docker-compose.yml
DavidGarciaCat: I can apply this update, but then I think I must highlight the fact that there are other apps (like pi-hole) not following this convention. If we need to apply this update here then it must also be implemented in other containers for consistency.

Screenshot 2024-09-13 at 19 13 00

zabbix/docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+    image: zabbix/zabbix-server-pgsql:7.0.3-alpine@sha256:d6235c964a025d7734c5dad3a5c94e0c016ac3fde7ac4e6ae2101cb7d336b812
+    hostname: zabbix-server
+    restart: on-failure
+    init: true
+    environment:
+      DB_SERVER_HOST: zabbix-database
+      POSTGRES_USER: zabbix
+      POSTGRES_PASSWORD: zabbix
+    ports:
+      - "10050:10050"
```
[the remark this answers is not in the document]
DavidGarciaCat: Zabbix Server Port should be open as it is used to get comms from the installed agents in the monitored devices. This one should remain open/exposed.

zabbix/docker-compose.yml
DavidGarciaCat: Fixed for this new App

DavidGarciaCat: https://github.com/getumbrel/umbrel-apps/issues/1463

nmfretz:
Thanks again for your hard work on Zabbix @DavidGarciaCat! I have pushed some commits to finalize this submission and have added gallery assets:

image

Going live to the app store 🎉

zabbix/umbrel-app.yml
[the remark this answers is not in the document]
nmfretz: Thanks. I have also removed the assets from this repo https://github.com/getumbrel/umbrel-apps/pull/1376/commits/2c6174b10b3dd3909b5b31d2c8af2879fe8f069d

zabbix/docker-compose.yml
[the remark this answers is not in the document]
nmfretz: Thanks 👌

zabbix/docker-compose.yml
nmfretz: Discussed and closed https://github.com/getumbrel/umbrel-apps/issues/1463

DavidGarciaCat:
Thank you, @nmfretz! It has been my pleasure to work providing a new tool that I believe might be very useful, especially for SME users.

I also appreciate your feedback and will try to remember it for other potential submissions.

DavidGarciaCat:
Hey, @nmfretz!

As part of the getting ready process you guys ran, there was a typo in the description/setup

Please refer to this new PR to fix it https://github.com/getumbrel/umbrel-apps/pull/1496

Cheers!

nmfretz:
Thanks for the eagle-eyes @DavidGarciaCat! Merged.

PR #1406, opened 2024-08-28
Update ordinals to v0.19.1

nmfretz:
This is a major version update, which requires a full re-index of ord (handled by our pre-start script).

Currently testing.


twosatsmaxi (approved):

nmfretz:
Tested on x86 and arm64. Going live. Good luck with the re-index everyone 🫡

PR #1409, opened 2024-08-30
App Submission: AlbyHub v1.10.1

bumi:
AlbyHub is Alby's new self-custodial wallet. It is build around the NWC protocol to allow users connect their Umbrel lightning nodes to the growing number applications supporting NWC for payments. Most known are the various nostr apps to enable zapping.
Alby Hub is also the continuation of the Nostr Wallet Connect application with is already in the store.  

### Links: 

+ https://nwc.dev/
+ https://albyhub.com/

# App Submission

### App name
Alby Hub

### 256x256 SVG icon

https://svgur.com/s/19oF

#### PNG:

connections
wallet
node
store

#### JPG:

connections
wallet
node
store

### I have tested my app on:
- [x] umbrelOS on Linux VM

### DIFF
--- a/albyhub/docker-compose.yml
+++ b/albyhub/docker-compose.yml
@@ -19,4 +19,4 @@ services:
       COOKIE_SECRET: ${APP_SEED}
       LOG_EVENTS: "true"
     restart: on-failure
-    stop_grace_period: 1m
+    stop_grace_period: 3m
--- a/albyhub/docker-compose.yml
+++ b/albyhub/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       LND_CERT_FILE: "/lnd/tls.cert"
       LND_MACAROON_FILE: "/lnd/data/chain/bitcoin/$APP_BITCOIN_NETWORK/admin.macaroon"
       WORK_DIR: "/data/albyhub"
-      COOKIE_SECRET: "742375a0f444d6de9c692aa83ac43c10"
+      COOKIE_SECRET: ${APP_SEED}
       LOG_EVENTS: "true"
     restart: on-failure
     stop_grace_period: 1m
--- a/albyhub/docker-compose.yml
+++ b/albyhub/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       APP_PORT: 8080
       PROXY_AUTH_ADD: "false"
   server:
-    image: ghcr.io/getalby/hub:v1.6.0@sha256:45ec367860a8f42c6f66317dd5889d8e2af8b0df1e0bc59b33e67c08770e4e17
+    image: ghcr.io/getalby/hub:v1.8.0@sha256:527ec1f06bd37cc112524603428a63f6ec4abb6d1c646c09623e63b430c12a5a
     volumes:
       - ${APP_DATA_DIR}/data:/data
       - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
--- a/albyhub/docker-compose.yml
+++ b/albyhub/docker-compose.yml
@@ -1,0 +1,22 @@
+version: "3.7"
+services:
+  app_proxy:
+    environment:
+      APP_HOST: albyhub_server_1
+      APP_PORT: 8080
+      PROXY_AUTH_ADD: "false"
+  server:
+    image: ghcr.io/getalby/hub:v1.6.0
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
+    environment:
+      LN_BACKEND_TYPE: "LND"
+      LND_ADDRESS: $APP_LIGHTNING_NODE_IP:$APP_LIGHTNING_NODE_GRPC_PORT
+      LND_CERT_FILE: "/lnd/tls.cert"
+      LND_MACAROON_FILE: "/lnd/data/chain/bitcoin/$APP_BITCOIN_NETWORK/admin.macaroon"
+      WORK_DIR: "/data/albyhub"
+      COOKIE_SECRET: "742375a0f444d6de9c692aa83ac43c10"
+      LOG_EVENTS: "true"
+    restart: on-failure
+    stop_grace_period: 1m
--- a/albyhub/docker-compose.yml
+++ b/albyhub/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       APP_PORT: 8080
       PROXY_AUTH_ADD: "false"
   server:
-    image: ghcr.io/getalby/hub:v1.6.0
+    image: ghcr.io/getalby/hub:v1.6.0@45ec367860a8f42c6f66317dd5889d8e2af8b0df1e0bc59b33e67c08770e4e17
     volumes:
       - ${APP_DATA_DIR}/data:/data
       - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
--- a/albyhub/docker-compose.yml
+++ b/albyhub/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       APP_PORT: 8080
       PROXY_AUTH_ADD: "false"
   server:
-    image: ghcr.io/getalby/hub:v1.6.0@45ec367860a8f42c6f66317dd5889d8e2af8b0df1e0bc59b33e67c08770e4e17
+    image: ghcr.io/getalby/hub:v1.6.0@sha256:45ec367860a8f42c6f66317dd5889d8e2af8b0df1e0bc59b33e67c08770e4e17
     volumes:
       - ${APP_DATA_DIR}/data:/data
       - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
--- a/albyhub/docker-compose.yml
+++ b/albyhub/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       APP_PORT: 8080
       PROXY_AUTH_ADD: "false"
   server:
-    image: ghcr.io/getalby/hub:v1.8.0@sha256:527ec1f06bd37cc112524603428a63f6ec4abb6d1c646c09623e63b430c12a5a
+    image: ghcr.io/getalby/hub:v1.10.1@sha256:155ced1c6cbabd201ceb391dfae3f07c2f1c5e57c8ea9561687bef7d8c981ce1
     volumes:
       - ${APP_DATA_DIR}/data:/data
       - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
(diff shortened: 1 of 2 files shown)


bumi:
what is the warning with the container user? The app needs to be able to write to the data directory. for this root is needed, isn't it?

nmfretz:
Have been looking forward to this one @bumi! I will review and test tomorrow.

> what is the warning with the container user?

Ideally, we want containers to run as non-root as part of app sandboxing measures, but it is not a strict requirement right now which is why the github actions linter is showing a severity indicator of "info" here. If the `getalby/hub` container requires a root user then this is fine for now.

> The app needs to be able to write to the data directory. for this root is needed, isn't it?

On app install, umbrelOS will create any host bind mounts that are committed to this directory (in this case `${APP_DATA_DIR}/data`) with user:group 1000:1000 ownership. So if `getalby/hub` were to run as 1000:1000 it could write to the data directory.

albyhub/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
+    environment:
+      LN_BACKEND_TYPE: "LND"
+      LND_ADDRESS: $APP_LIGHTNING_NODE_IP:$APP_LIGHTNING_NODE_GRPC_PORT
+      LND_CERT_FILE: "/lnd/tls.cert"
+      LND_MACAROON_FILE: "/lnd/data/chain/bitcoin/$APP_BITCOIN_NETWORK/admin.macaroon"
+      WORK_DIR: "/data/albyhub"
+      COOKIE_SECRET: "742375a0f444d6de9c692aa83ac43c10"
```
nmfretz, line 19: If you want this to be unique to a user, you can utilize `${APP_SEED}` here.
```suggestion
      COOKIE_SECRET: ${APP_SEED}
```

APP_SEED is a unique 256 bit long hex string (128 bits of entropy) deterministically derived from a user's umbrelOS seed and your app's ID.

albyhub/umbrel-app.yml
```diff
@@ -0,0 +1,29 @@
...
+name: Alby Hub
+tagline: Your Hub to connect to the Bitcoin world
+category: finance
+version: "1.6.0"
+port: 59000
+description: >-
+  Alby Hub is your gateway to Bitcoin world and allows you to use your Umbrel lightning node with the growing number of exciting bitcoin and nostr applications.
+  Connect any bitcoin app to your Hub for self-custodial payments anywhere — websites, mobile apps, desktop apps, games, CLI tools and more.
+
+  Use Alby Hub to manage your lightning channels through one sleek, user-friendly interface.
```
nmfretz, line 12: To get whitespace between the two paragraphs we need to add two newlines:
```suggestion
description: >-
  Alby Hub is your gateway to Bitcoin world and allows you to use your Umbrel lightning node with the growing number of exciting bitcoin and nostr applications.
  Connect any bitcoin app to your Hub for self-custodial payments anywhere — websites, mobile apps, desktop apps, games, CLI tools and more.


  Use Alby Hub to manage your lightning channels through one sleek, user-friendly interface.
```

This will now render nicely like this:

image

albyhub/umbrel-app.yml
```diff
@@ -0,0 +1,29 @@
+manifestVersion: 1
+id: albyhub
+name: Alby Hub
+tagline: Your Hub to connect to the Bitcoin world
+category: finance
```
nmfretz, line 5: I just noticed that the nostr wallet connect app is also listed in the `finance` category.

This means that both of the alby apps end up here in the app store:
image

Instead of in the `bitcoin` category with related apps:
image

It's totally up to you where you want this to render. If you'd like to switch to the bitcoin category then you can do:
```suggestion
category: bitcoin
```

albyhub/umbrel-app.yml
```diff
@@ -0,0 +1,29 @@
...
+website: https://albyhub.com
+submitter: Alby
+submission: https://github.com/getumbrel/umbrel-apps/pull/1409
+repo: https://github.com/getAlby/hub
+support: https://github.com/getAlby/hub/issues
+gallery:
+  - 1.png
+  - 2.png
+  - 3.png
+  - 4.png
```
nmfretz, line 23: Can you please change to:
```suggestion
gallery:
  - 1.jpg
  - 2.jpg
  - 3.jpg
  - 4.jpg
```

Don't worry about converting the actual images, I'll do that when I commit them to https://github.com/getumbrel/umbrel-apps-gallery

albyhub/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+version: "3.7"
+services:
+  app_proxy:
+    environment:
+      APP_HOST: albyhub_server_1
+      APP_PORT: 8080
+      PROXY_AUTH_ADD: "false"
+  server:
+    image: ghcr.io/getalby/hub:v1.6.0@sha256:45ec367860a8f42c6f66317dd5889d8e2af8b0df1e0bc59b33e67c08770e4e17
+    volumes:
```
nmfretz, line 10: If this container can run as a non-root user (which should be possible based on this Dockerfile https://github.com/getAlby/hub/blob/master/Dockerfile) then we can change this to:

```suggestion
  server:
    image: ghcr.io/getalby/hub:v1.6.0@sha256:45ec367860a8f42c6f66317dd5889d8e2af8b0df1e0bc59b33e67c08770e4e17
    user: "1000:1000"
    volumes:
```

When the app is installed, umbrelOS creates the data directory that's commited to this repo with 1000:1000 ownership/permissions, so a `getalby/hub` container running as user 1000 will have write permissions to `${APP_DATA_DIR}/data`

nmfretz (commented):
I've tested and this is working really well @bumi. Fantastic work by the whole alby team 🐝

I've left some comments/suggestions below. After they are addressed we can go live.

nmfretz:
Hey @bumi, just pinging you again in case you missed the above review from last week.

nmfretz:
@bumi - I forgot to ask: with the addition of Alby Hub should we be removing Nostr Wallet Connect from the app store?
https://github.com/getumbrel/umbrel-apps/tree/master/alby-nostr-wallet-connect

If so, what I can do is add `disabled: true` to the umbrel-app.yml of nostr wallet connect and it will make it so that existing installs can continue to use the app, but users will no longer see the app as available in the app store.

albyhub/docker-compose.yml
bumi: My docker and docker compose knowledge is limited. We do not want to change the image now as it is used in various places and I would be worried to break other deployments. 

The app only writes to the configured volume. (log files and SQLite DB files) 
If we configure this user here what does that exactly mean? can you point me to some docs? 

It also binds to ports > 1024 - 8080 for the webapp and the lightning port 9375 (if the embedded LDK node is used and not the default Umbrel LND node) - so no root is needed for that.

bumi:
@nmfretz thanks a lot for the feedback and improvements. 

> should we be removing Nostr Wallet Connect from the app store?
I am not sure what's best here. We are not actively working on it anymore and for us Alby Hub is the successor of that old NWC app. We recommend people to update to Alby Hub. So to me `disabled: true` sounds good - but some people also still liked the original NWC app it seemed.

bumi:
fyi. we have one patch release coming this week which I would like to have in this release.

nmfretz:
> I am not sure what's best here. We are not actively working on it anymore and for us Alby Hub is the successor of that old NWC app. We recommend people to update to Alby Hub. So to me disabled: true sounds good - but some people also still liked the original NWC app it seemed.

If you'd like to keep NWC available for new users still then perhaps for now we can just add a note at the top of NWC's app description saying that it is recommended to use the successor app Alby Hub instead. Although I imagine that since you're not working on NWC anymore, there will come a time when you'd want the app retired (e.g., it's not receiving security updates, etc).

> fyi. we have one patch release coming this week which I would like to have in this release.

Great, thanks for letting me know. Ping me when it's ready and I'll give it a final test before we go live!

albyhub/docker-compose.yml
nmfretz: No worries. The `user` key in a docker-compose.yml file allows you to run the container as a certain user WITHOUT having to change the underlying image.

Setting `user` in the docker-compose.yml overrides the user used to run the container process. If you don't set a `user` in the docker-compose.yml file then Docker uses whatever has been defined in the Dockerfile with the `USER` instruction. In the case of your [Dockerfile](https://github.com/getAlby/hub/blob/master/Dockerfile), the lack of `USER` instruction means that by default it runs as root.

As long as the container can write to it's log and db files as user 1000:1000 then we should be good. I can give this a test for you.

Docs:
https://docs.docker.com/reference/dockerfile/#user
https://docs.docker.com/reference/compose-file/services/#user

nmfretz:
Here's a checklist for us:

- [ ] Update `ghcr.io/getalby/hub` image tag and digest in docker-compose.yml with latest patch release (@bumi)
- [ ] Update the `version` in the umbrel-app.yml to match the patch release version (@bumi)
- [ ] Test user 1000:1000 (@nmfretz)
- [ ] Check gallery assets resolution (@nmfretz)

nmfretz:
Hey @bumi, I see that hub is now on v1.10.1. Would you like to go live with that version?

bumi:
@nmfretz yes, v1.10.1 is it! 
so if you could help or decide on the user  1000:1000 then I think we're good to go. 🙏

nmfretz:
Excellent, thanks again for all your work bringing Alby Hub to the Umbrel app store @bumi!

I have tested the new image with user 1000:1000 and everything is working well (e.g., app install, sending/receiving, connection to alby go, connection to apps, etc). If at some point it becomes apparent that there is a permissions issue for something (very unlikely) then the fix is very simple in this direction (i.e., going from user 1000 back to root). All we would need to do is remove this line from the compose file:
`user: 1000:1000`

I don't have permissions to push a commit to `getAlby/umbrel-apps.git` so I'll add `user: 1000:1000` as a suggestion below. Once it is committed we can go live.

image

albyhub/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+version: "3.7"
+services:
+  app_proxy:
+    environment:
+      APP_HOST: albyhub_server_1
+      APP_PORT: 8080
+      PROXY_AUTH_ADD: "false"
+  server:
+    image: ghcr.io/getalby/hub:v1.10.1@sha256:155ced1c6cbabd201ceb391dfae3f07c2f1c5e57c8ea9561687bef7d8c981ce1
```
nmfretz, line 9:
```suggestion
  server:
    image: ghcr.io/getalby/hub:v1.10.1@sha256:155ced1c6cbabd201ceb391dfae3f07c2f1c5e57c8ea9561687bef7d8c981ce1
    user: 1000:1000
```

nmfretz:
Perfect, thanks @bumi. Going live 🐝

MoritzKa:
Amazing. Thanks for publishing Alby Hub, @nmfretz.
We have a few users asking if they can simply update their Alby Hub installed from the community app store with the version listed on the official app store. 
Are all NWC connections available?
Can they update using the official app store listing?
Should they know something before doing an update?

Would appreciate any hints.

bumi:
I don't think people can update from the community store actually. it's a different app

PR #1430, opened 2024-09-05
Moved just-download to codeberg

highghlow:
I have moved just-download to codeberg: https://codeberg.org/highghlow/just-download

That's the only change


nmfretz:
Excellent, thanks @highghlow. With the version change, existing installs will be notified of an available update, so I added release notes.

Tested and going live.

PR #1444, opened 2024-09-08
App Submission QEMU guest agent

phaus:
# App Submission

### App name

QEMU guest agent

### Version

9.0.2

### 256x256 SVG icon

monitor-fur-virtuelle-maschinen

[Hypervisor Icons created by Kalashnyk - Flaticon](https://www.flaticon.com/de/kostenlose-icons/hypervisor)

### Gallery images

It is a background daemon, so there is now UI at all.

### I have tested my app on:
- [x] umbrelOS on Linux VM


nmfretz:
Hey @phaus, thanks for this submission!

I am not familiar with QEMU guest agent. Can you please let me know your use-case for this and how it makes running umbrelOS via QEMU better? Specifically, is there something that is not working out-of-the-box with QEMU that we can support on the umbrelOS side?

phaus:
Hi @nmfretz  thanks for answering.
My use-case is, that I have a NAS running here at home and I did setup umbrel as VM on that NAS.
The QEMU guest agent is basically a small software agent, that can exchange data with the VM host via a serial connection.
With that, the VM Host is able to get the Client IP, but it is also possible to trigger a clean shutdown/reboot from the VM Host.

I minimal impl. in Go can be seen here:
https://github.com/bradfitz/qemu-guest-kragent/blob/main/kragent.go

Since I created this ticket, I already found out, that you can use your own App Store in Umbrel, so I just did that and [added this Application](https://github.com/phaus/app-store/tree/master/phaus-qemu-guest-agent) to it, to have a working Agent Setup running after an umbrel update :-).

nmfretz:
Hey @phaus, thanks for describing your use case. Really appreciate it.

We've talked about this internally, and with the upcoming changes to our apps framework this app would end up breaking. This is because it requires a level of network/system access that we can't allow in the app store in order to maintain proper sandboxing/security for apps.

With that in mind, I'm going to close this submissions. But thanks a ton for your efforts packaging and submitting this. 

> Since I created this ticket, I already found out, that you can use your own App Store in Umbrel, so I just did that
Great job on finding a workaround for now! Glad that's working for you.

We could potentially add better QEMU functionality to the core OS if there was a strong use case for it and it doesn’t have downsides for non QEMU users. I've opened an issue on the main umbrel repo to discuss this (https://github.com/getumbrel/umbrel/issues/1922). It would be wonderful if you could drop by that Issue and share any thoughts and some additional details on use case. Hopefully we can get some others to chime in and we can see if there's strong enough use-cases and interest to potentially add more QEMU functionality to the umbrelOS.

PR #1459, opened 2024-09-09
Update ordinals to version v0.20.0

nmfretz:
This is a major version update, which requires a full re-index of ord (handled by our pre-start script).


nmfretz:
Reviewed and tested

PR #1461, opened 2024-09-11
Update Bitcoin Node to v27.1

nmfretz:
Awaiting `getumbrel/bitcoind:v27.1` image build. New bitcoind image must be added to the compose file and the app tested.

This PR updates the Bitcoin Node app to use the latest Bitcoin Core version 27.1. The previous [lncm/bitcoind](https://github.com/lncm/docker-bitcoind) image has been replaced with [getumbrel/bitcoind](https://github.com/getumbrel/docker-bitcoind).

### DIFF
--- a/bitcoin/docker-compose.yml
+++ b/bitcoin/docker-compose.yml
@@ -40,7 +40,7 @@ services:
         ipv4_address: $APP_BITCOIN_IP
 
   bitcoind:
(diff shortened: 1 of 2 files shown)


nmfretz:
Tested on x86 and arm64
Tested for fresh install and app update
Tested on umbrelOS 0.5 and 1.x

PR #1482, opened 2024-09-16
Update Plex to v1.41.0.8992

nmfretz:


nmfretz:
Reviewed and tested on x86 and arm64.

PR #1495, opened 2024-09-17
Updated to latest databag container

balzack:
container changes consist of:
- updated to latest go tools with security patches
- supporting push notification repeater instead of deprecated firebase API


nmfretz:
Thanks for keeping databag up-to-date @balzack! Reviewed and tested.

PR #1496, opened 2024-09-18
Bugfix: Update invalid Zabbix Agent service name

DavidGarciaCat:
Zabbix Agent has its own service, so we need to target the Agent's service instead of the Server

As you can see, Zabbix reports an error (in red) when targeting the invalid service name, but it resolves successfully when targeting the updated one.

Screenshot 2024-09-18 at 09 38 13

Screenshot 2024-09-18 at 09 38 43

### DIFF
--- a/zabbix/umbrel-app.yml
+++ b/zabbix/umbrel-app.yml
@@ -30,7 +30,7 @@ description: >-
   ⚙️ Configuration:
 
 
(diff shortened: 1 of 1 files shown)


sharknoon:
Thank you @DavidGarciaCat for the bugfix :)
@nmfretz looks good to me

DavidGarciaCat:
> Thank you @DavidGarciaCat for the bugfix :) @nmfretz looks good to me

extra testing never hurts!

PR #1506, opened 2024-09-19
Update noStrudel to v0.41.0

hzrd149:
upstream release https://github.com/hzrd149/nostrudel/blob/master/CHANGELOG.md#0410

### DIFF
--- a/nostrudel/docker-compose.yml
+++ b/nostrudel/docker-compose.yml
@@ -12,7 +12,7 @@ services:
     environment:
       CORSANYWHERE_REQUIRE_HEADERS: "host"
   web:
-    image: ghcr.io/hzrd149/nostrudel:0.38.3@sha256:cda4e5964e2f24f118547189819465682997fe8b3d6e0b688a3ae0685fed00da
+    image: ghcr.io/hzrd149/nostrudel:0.41.0
     restart: on-failure
     stop_grace_period: 1m
     init: true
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for the noStrudel update @hzrd149!  

> If the docker image supports both amd64 and arm64 architectures then its going to have two hashes. which one should be used?

There's actually another hash for the multi-arch image manifest. There are quite a few ways to get it, but here's a couple:

1) On ghrc.io, if you've published a multi-arch image then it will be the digest shown under the tag:

image

You can also see it at the link you provided (circled in yellow):

image


2) If you pull the Docker image by tag, it will give you the multi-arch manifest digest:

```
$ sudo docker pull ghcr.io/hzrd149/nostrudel:0.41.0
[sudo] password for umbrel:
0.41.0: Pulling from hzrd149/nostrudel
...
Digest: sha256:94936037a0370d3c63bcc1beba62d18c20bd9b723073e695549cd75d2076ff2f
Status: Downloaded newer image for ghcr.io/hzrd149/nostrudel:0.41.0
ghcr.io/hzrd149/nostrudel:0.41.0
```

> I don't quite understand the requirement for the `@<sha256>` on the docker image

What this does is lock us in to the exact Docker image that we've tested in this PR. Docker gives precedence to the digest over the tag, so even if another image is pushed to the 0.41.0 tag we'll still grab the one associated with digest `sha256:94936037a0370d3c63bcc1beba62d18c20bd9b723073e695549cd75d2076ff2f` that we've tested here.

nmfretz:
@hzrd149 I have added the multi-arch manifest digest. I have also tweaked the release notes formatting, which can be finicky, so that we get:

image

Instead of this:

image

Tested and going live. Thanks again for the update!

PR #1509, opened 2024-09-21
Update morphos to v0.6.0

danvergara:
Update morphos to v0.6.0 https://github.com/danvergara/morphos/releases/tag/v0.6.0


nmfretz:
Thanks for this update @danvergara! I have tested and it is working well. Thanks for using a non-root user 🤝.

One thing you could consider in the future, is not defining the user in your Dockerfile, and instead using the `user` directive in the docker-compose.yml to assign any user you desire.

PR #1510, opened 2024-09-21
App Submission - Simplexmq

arguser:
# App Submission

### App name
SimpleXMQ
...

### 256x256 SVG icon

...

### Gallery images
...

### I have tested my app on:
- [ ] umbrelOS on Linux VM


arguser:
Looks like I will have to build a multi-architecture image for this...

nmfretz:
@arguser thanks for submitting SimpleX 🎉. This one's been pretty heavily requested.

nmfretz:
Hey @arguser, thanks for your patience 🙏.

Before I dig in too far, do you currently have Simplexmq working?

I just took a quick scan through the PR and noticed that both the smp and xftp services are set up as Tor only:
e.g.,
```yaml
environment:
      ADDR: ${APP_SIMPLEX_SERVER_SMP_HIDDEN_SERVICE}
      PASS: ${APP_PASSWORD}
```

But I don't think the current `torrc-template` you have will set things up correctly:
```
# Enable log (otherwise, tor doesn't seem to deploy onion address)
Log notice file /tmp/notices.log
# Enable single hop routing (2 options below are dependencies of the third) - It will reduce the latency at the cost of lower anonimity of the server - as SMP-server onion address is used in the clients together with public address, this is ok. If you deploy SMP-server with onion-only address, you may want to keep standard configuration instead.
SOCKSPort 0
HiddenServiceNonAnonymousMode 1
HiddenServiceSingleHopMode 1
# smp-server hidden service host directory and port mappings
HiddenServiceDir /data/app-$APP_ID-smp
HiddenServicePort $APP_SIMPLEX_SERVER_SMP_PORT simplexmq_smp-server:$APP_SIMPLEX_SERVER_SMP_PORT
# xftp-server hidden service host directory and port mappings
HiddenServiceDir /data/app-$APP_ID-xft
HiddenServicePort $APP_SIMPLEX_SERVER_XFTP_PORT simplexmq_xftp-server:$APP_SIMPLEX_SERVER_XFTP_PORT
```

The container names for the smp and xftp services will need a `_1` at the end of them (e.g., `simplexmq_smp-server_1`) in order for Docker's internal DNS to resolve an IP. Also, I noticed that `HiddenServiceDir /data/app-$APP_ID-xft` is missing a `p` at the end (`xftp`) which is how it's defined in your `exports.sh`.

I suspect that the tor hidden services won't actually be created when the app starts and so you'll get `notyetset.onion` passed into your smp and xftp containers as the env var for `ADDR` (based on the `exports.sh` logic).

-----------------------------
Also, I'm not familiar with SimpleX chat or running a smp/xftp server. Will the server UI show users their smp and xftp connection details? Or is Simplexmq still in early-days and users would be required to use CLI?

arguser:
Hey @nmfretz yes I had it running via TOR. I probably made mistakes while moving the code from my community-apps repo, let me check this out.
Not using TOR would require port forwarding or other tooling for using the sever as a relay, we could discuss what's the best way to go.

The web doesn't show the addresses, I have to figure out if there's a configuration for that. So user should check the terminal.

jeremycady:
I'd love a non-TOR option as well. 

Umbrel comes with the abilty to easily integrate Cloudflare Tunnels or to use Tailscale.

I would think it would be pretty simple to shoot a domain to that port and allow others to use it.

> Hey @nmfretz yes I had it running via TOR. I probably made mistakes while moving the code from my community-apps repo, let me check this out. Not using TOR would require port forwarding or other tooling for using the sever as a relay, we could discuss what's the best way to go.
> 
> The web doesn't show the addresses, I have to figure out if there's a configuration for that. So user should check the terminal.

arguser:
@nmfretz Update! since v6.1.0 we can show the needed information on the simplex web! https://simplex.chat/docs/server.html#server-information-page

I've built images for arm/amd of v6.2.0 and trying to figure out how to initialize the addresses on the container start. On the new version SMP is not starting, probably due to configuration, I'll come back when progress is made.

nmfretz:
> @nmfretz Update! since v6.1.0 we can show the needed information on the simplex web!

@arguser Perfect, that's a huge win for usability! Ping me when you're ready for review.


> I'd love a non-TOR option as well.
>
> Umbrel comes with the abilty to easily integrate Cloudflare Tunnels or to use Tailscale.
>
> I would think it would be pretty simple to shoot a domain to that port and allow others to use it.

@jeremycady good idea. @arguser once you have 6.2.0 up and running we can test configuring with a local hostname and then tunneling from cloudflare.

arguser:
@nmfretz got it working, had to do some hacky things and the server information doesn't have a proper server addresses fields, so for the time being I'm using others available fields. Tell me what you think!

arguser:
@nmfretz let me know what you think!

nmfretz:
Thanks @arguser! Sorry for the delay, will check this out soon.

nmfretz:
> @nmfretz got it working, had to do some hacky things and the server information doesn't have a proper server addresses fields, so for the time being I'm using others available fields. Tell me what you think!

Ha, very clever! And great work getting the servers up and running on start.

Looking at the UI, I think it might be too confusing for users to see the server addresses listed under the `Web site` and `Hosting provider` headings. From experience with other apps, things need to be either very simple or else conform to tutorials that users will find online.

Also, I think we'd want to show both Tor and Clearnet addresses for the servers, right?
 
image

What do you think about these two options? 
 
1) We ship the app with the default UI and just include general instructions in the app description on how to find server details. For example, linking to [official docs](https://simplex.chat/docs/server.html#documentation) and then saying you can access the info through Files (currently in beta) so they don't even have to use the terminal. This way when users search for SimpleX tutorials, the UI will look the same and the general way they concatenate the server details will be the same.

or

2) Create a simple frontend like was done for the [Samba app](https://github.com/getumbrel/umbrel-apps/pull/1161#issuecomment-2193737610) that just shows the server details for tor and clearnet. This would be the easiest UX for users, but then it would add additional maintenance overhead like if the connection details move to a different place and we can't update the app without fixing the ui container. And also, the app wouldn't have the official UI.

What do you think of either of those?

arguser:
@nmfretz Thanks!

Not sure if I can make it both work together, or should show TOR only when enabled System Wise?

Regarding the information display I would go with **option 2** but I also guess it could be possible to get those fields on the official UI.

nmfretz:
> Not sure if I can make it both work together, or should show TOR only when enabled System Wise?

Ah, is it not possible to have a tor container running alongside simplex in the same compose file and then have both tor and clearnet addresses available?

> Regarding the information display I would go with option 2 but I also guess it could be possible to get those fields on the official UI.

If it's not too hard to change the fields in the official UI then this would be a great solution.

PR #1543, opened 2024-09-26
Gitlab App - How to change the `root` password

DavidGarciaCat:
This is updating the description of the app so we all know how to change the password for the `root` user when installing the app for the very first time.

I have tested it with my local install (umbrel home) and I got it from this page: https://stackoverflow.com/questions/55747402/docker-gitlab-change-forgotten-root-password

### DIFF
--- a/gitlab/umbrel-app.yml
+++ b/gitlab/umbrel-app.yml
@@ -9,6 +9,9 @@ description: >-
   ⚠️ This app is RAM-intensive (+6GB recommended) and can take 15-20 minutes to start after installation on low-powered devices.
 
 
+  🔐 After installing the App, you can create a new user account via web interface, or you can connect via SSH/Terminal and change the `root` password by running the `gitlab-rake "gitlab:password:reset[root]"` command
+
+
   GitLab is a comprehensive, web-based DevOps lifecycle tool that provides a robust platform for managing projects and repositories. It offers a wide range of features that cater to the entire software development lifecycle, including version control, CI/CD, code review, issue tracking, and more. All operations are performed in a secure, collaborative environment, ensuring your code and data are safe and accessible.
 
   


nmfretz:
Thanks for this PR @DavidGarciaCat. Changing both the username and password for the `root` user is all available from within the app's UI, which is something fairly common across apps that people would likely find just by exploring.

With that in mind, should we refrain from adding additional instructions in the app description, especially ones that require terminal commands?

Changing username:
image

Changing password:
image

nmfretz:
@DavidGarciaCat I'm closing this PR, but linking to it here for our consideration when updating Gitlab to the latest release:
https://github.com/getumbrel/umbrel-apps/pull/1564

DavidGarciaCat:
Hey, @nmfretz 

According to the [Umbrel App yaml file](https://github.com/getumbrel/umbrel-apps/blob/ce539448bb6b40c96a424cec88c6dba63db12b56/gitlab/umbrel-app.yml#L61), we only see the default username (root) but no password. Shall I assume there's no password? If that't the case, then maybe we should add a note for this:

https://github.com/getumbrel/umbrel-apps/blob/ce539448bb6b40c96a424cec88c6dba63db12b56/gitlab/umbrel-app.yml#L61

nmfretz:
Ah, okay I see the confusion now! 

> According to the [Umbrel App yaml file](https://github.com/getumbrel/umbrel-apps/blob/ce539448bb6b40c96a424cec88c6dba63db12b56/gitlab/umbrel-app.yml#L61), we only see the default username (root) but no password. Shall I assume there's no password?

There actually is a password and it is shown to the user. It is a deterministically derived password that is unique to a user's umbrel installation + app (https://github.com/getumbrel/umbrel/blob/7103e98909041bb8e15702b4b2e3e7f78da4b351/packages/umbreld/source/modules/apps/legacy-compat/app-script#L69-L88).

In the `umbrel-app.yml` file it is assigned here:
https://github.com/getumbrel/umbrel-apps/blob/9dc50f8b79c90e982c3823ed226b69bd97eaf3f1/gitlab/umbrel-app.yml#L60

And in the compose file it is assigned here:
https://github.com/getumbrel/umbrel-apps/blob/9dc50f8b79c90e982c3823ed226b69bd97eaf3f1/gitlab/docker-compose.yml#L24

But those details themselves don't matter to users because they don't need to know or care about an app's underlying config files. Instead they'll see the password in the app store when they install the app:

image

and as well if they right click an app icon and select "show default credentials"

image

image

In a future umbrelOS update we also plan to have a notification come up the first time an app is clicked letting the user know of the default credentials.

DavidGarciaCat:
Ah !! I see, this is the same approach used for Pi-hole.

OK, thanks for the update. And yes, that makes this PR completely unnecessary.

Thanks!

Screenshot 2024-10-03 at 14 10 26

PR #1549, opened 2024-09-26
Update default `docker-compose.yml`

Zoobdude:
`the attribute version is obsolete, it will be ignored, please remove it to avoid potential confusion`

[Reference in the compose-spec](https://github.com/compose-spec/compose-spec/blob/main/spec.md#version-and-name-top-level-elements)

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,8 +100,6 @@ We'll now create a `docker-compose.yml` file in this directory to define our app
 Let's copy-paste the following `docker-compose.yml` template into a text editor and modify it according to our app.
 
 ```yml
-version: "3.7"
-
 services:
   app_proxy:
     environment:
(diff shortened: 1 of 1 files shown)


nmfretz:
Thanks for noticing this @Zoobdude and for submitting a fix!

We are in the process of re-doing our app ecosystem on umbrelOS, which will include bringing compose files up to spec. 

In the meantime though, we need to keep this legacy version attribute as-is for compatibility with umbrelOS 0.5 which is using a version of Compose that requires it (this version also requires that our app_proxy compose file and the app's main compose file have the exact same version or Compose will error out). Both umbrelOS 0.5 and 1.x use this repo, so removing the version attribute would break all umbrelOS 0.5 instances.

Closing this issue, noting that we are aware of the spec. Thanks again for the sharp eyes.

PR #1554, opened 2024-09-28
Update homebridge to 2024-09-03

JoseMoranUrena523:
This is a pull request to update homebridge to version 2024-09-03. PR must be reviewed and tested before merging.

### DIFF
--- a/homebridge/docker-compose.yml
+++ b/homebridge/docker-compose.yml
@@ -1,8 +1,8 @@
-version: "3.7"
+version: "2"
 
 services:
   server:
-    image: homebridge/homebridge:2024-06-27@sha256:be09c4402c351520069014ba77a6d12f22021294c04125b827a8e7cb5b79b9da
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this PR @JoseMoranUrena523! Good catch on the newer Docker image. Homebridge docker tags are released on a different schedule than the main homebridge release which leads to some confusion. For instance...

The previous image tag "2024-06-27" shows homebridge as already being on 1.8.4

image

image

Which strangely doesn't match up with homebridge's release documentation which shows that Docker tag as including v1.8.3: https://github.com/homebridge/docker-homebridge/releases

What I've done is kept the version as 1.8.4 in the app store with the suffix `-ui-update`. This change to the manifest means that users will get notified of an update while still seeing that Homebridge is 1.8.4.

I added release notes explaining that this release updates the Homebridge UI (v4.56.4 --> v4.57.1)

Tested and going live.

JoseMoranUrena523:
Hello @nmfretz! You’re welcome on the PR. Do you know when issues and pull requests on [getumbrel/umbrel](https://github.com/getumbrel/umbrel) will be reviewed as well as job applications?

PR #1563, opened 2024-10-01
App Submission: Enclosed

Zoobdude:
# App Submission

## App name
Enclosed

## App logo
https://svgshare.com/s/1BQT

## Gallery
1.jpg:
image

2.jpg:
image

3.jpg:
image

I have tested my app on:
[] umbrelOS on a Raspberry Pi
[] umbrelOS on an Umbrel Home
[] umbrelOS on Linux VM

### DIFF
--- a/enclosed/umbrel-app.yml
+++ b/enclosed/umbrel-app.yml
@@ -17,21 +17,12 @@ description: >-
 
   You can set up https using the Cloudflare Tunnel app available in the app store.
   Detailed instructions for configuring the tunnel can be found here: https://github.com/Radiokot/umbrel-cloudflared/wiki/How-to-set-up-Cloudflare-Tunnel-on-your-Umbrel
-releaseNotes: >-
-  This release includes various bug fixes and improvements, including:
-  
-  - Fix navbar layout on mobile
-  - Preserve line breaks
 developer: Corentin Thomasset
 website: https://enclosed.cc/
 dependencies: []
 repo: https://github.com/CorentinTh/enclosed
 support: https://github.com/CorentinTh/enclosed/discussions
 port: 8787
-gallery:
-  - 1.jpg
-  - 2.jpg
-  - 3.jpg
 path: ""
 defaultUsername: ""
 defaultPassword: ""
--- a/enclosed/umbrel-app.yml
+++ b/enclosed/umbrel-app.yml
@@ -17,12 +17,14 @@ description: >-
 
   You can set up https using the Cloudflare Tunnel app available in the app store.
   Detailed instructions for configuring the tunnel can be found here: https://github.com/Radiokot/umbrel-cloudflared/wiki/How-to-set-up-Cloudflare-Tunnel-on-your-Umbrel
+releaseNotes: 
 developer: Corentin Thomasset
 website: https://enclosed.cc/
 dependencies: []
 repo: https://github.com/CorentinTh/enclosed
 support: https://github.com/CorentinTh/enclosed/discussions
 port: 8787
+gallery:
 path: ""
 defaultUsername: ""
 defaultPassword: ""
--- a/enclosed/umbrel-app.yml
+++ b/enclosed/umbrel-app.yml
@@ -26,7 +26,10 @@ dependencies: []
 repo: https://github.com/CorentinTh/enclosed
 support: https://github.com/CorentinTh/enclosed/discussions
 port: 8788
-gallery: []
+gallery:
+- 1.jpg
+- 2.jpg
+- 3.jpg
 path: ""
 submitter: Zoobdude
 submission: https://github.com/getumbrel/umbrel-apps/pull/1563
--- a/enclosed/umbrel-app.yml
+++ b/enclosed/umbrel-app.yml
@@ -17,14 +17,14 @@ description: >-
 
   You can set up https using the Cloudflare Tunnel app available in the app store.
   Detailed instructions for configuring the tunnel can be found here: https://github.com/Radiokot/umbrel-cloudflared/wiki/How-to-set-up-Cloudflare-Tunnel-on-your-Umbrel
-releaseNotes: 
+releaseNotes: ""
 developer: Corentin Thomasset
 website: https://enclosed.cc/
 dependencies: []
 repo: https://github.com/CorentinTh/enclosed
 support: https://github.com/CorentinTh/enclosed/discussions
 port: 8787
-gallery:
+gallery: []
 path: ""
 defaultUsername: ""
 defaultPassword: ""
--- a/enclosed/umbrel-app.yml
+++ b/enclosed/umbrel-app.yml
@@ -35,4 +35,4 @@ path: ""
 defaultUsername: ""
 defaultPassword: ""
 submitter: Zoobdude
-submission: https://github.com/getumbrel/umbrel/pull/334
+submission: https://github.com/getumbrel/umbrel-apps/pull/1563
--- a/enclosed/umbrel-app.yml
+++ b/enclosed/umbrel-app.yml
@@ -23,7 +23,7 @@ website: https://enclosed.cc/
 dependencies: []
 repo: https://github.com/CorentinTh/enclosed
 support: https://github.com/CorentinTh/enclosed/discussions
-port: 8787
+port: 8788
 gallery: []
 path: ""
 defaultUsername: ""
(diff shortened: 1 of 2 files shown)


Zoobdude:
Will pick a different port, set relevant env vars to make it work and test on RPI

dennysubke:
@Zoobdude Very good app! Thanks for the tip! Works perfectly with Cloudflare. I took it to my community store: 
Screenshot

nmfretz:
Thanks for submitting Enclosed @Zoobdude! This looks great.

FYI, the next port number of `8788` is free! All you'll need to do is change the `port` in the `umbrel-app.yml`. You can leave the docker-compose.yml file as-is since the app_proxy container will handle things for you.

enclosed/umbrel-app.yml
```diff
@@ -0,0 +1,32 @@
...
+
+  You can set up https using the Cloudflare Tunnel app available in the app store.
+  Detailed instructions for configuring the tunnel can be found here: https://github.com/Radiokot/umbrel-cloudflared/wiki/How-to-set-up-Cloudflare-Tunnel-on-your-Umbrel
+releaseNotes: ""
+developer: Corentin Thomasset
+website: https://enclosed.cc/
+dependencies: []
+repo: https://github.com/CorentinTh/enclosed
+support: https://github.com/CorentinTh/enclosed/discussions
+port: 8787
```
nmfretz, line 26: This is the value to change. No other change should be necessary. For example:

```suggestion
port: 8788
```

Zoobdude:
Thanks for the reply, didn't know the app_proxy could just change the port, that's pretty handy.

enclosed/umbrel-app.yml
```diff
@@ -0,0 +1,32 @@
...
+developer: Corentin Thomasset
+website: https://enclosed.cc/
+dependencies: []
+repo: https://github.com/CorentinTh/enclosed
+support: https://github.com/CorentinTh/enclosed/discussions
+port: 8788
+gallery: []
+path: ""
+defaultUsername: ""
+defaultPassword: ""
```
sharknoon, line 30: Default username and passwords are not needed, if they are empty
```suggestion
```

sharknoon (changes requested):
Thank you @Zoobdude for the amazing app submission. I have only one very minor suggestion (see below). I have also tested the app and it is working perfectly :)

sharknoon:
Wow that was fast @Zoobdude :) 
@nmfretz this app is ready from my side

Zoobdude:
Just waiting to be back at my desk with my 4k monitor for the screenshots. Don't think my 2013 13" MacBook Air will cut it xD

Zoobdude:
Hopefully those screenshots are alright!

sharknoon:
Yes, they are looking good :) Thank you

Zoobdude:
Add app logo

nmfretz:
Wonderful, thanks again for this submission @Zoobdude! And thanks for reviewing @sharknoon.

I have pushed a few changes to finalize the submission. We'll get the gallery assets created and then this is good to go live!

-----------------------------
Context for commit https://github.com/getumbrel/umbrel-apps/pull/1563/commits/ea086b7d3eef0b6aec90826561958fee01bc2cef:

During testing, I ran into a permissions issue when creating a note:
```
{"level":50,"time":1728961593126,"pid":1,"hostname":"ef7a1975224b","namespace":"middlewares:error","error":{"errno":-13,"code":"EACCES","syscall":"open","path":"/app/.data/01ja728fs5y3x8ffeqzdcemva6"},"msg":"EACCES: permission denied, open '/app/.data/01ja728fs5y3x8ffeqzdcemva6'"}
```

This is because the uid/gid of the nonroot user wasn't being set correctly in the docker-compose.yml:

```
/app $ whoami
nonroot
/app $ id
uid=100(nonroot) gid=101(nonroot) groups=101(nonroot)
/app $ ls -la
total 284
drwxr-xr-x    1 nonroot  nonroot       1024 Sep 29 20:36 .
drwxr-xr-x    1 root     root          1024 Oct 15 03:01 ..
drwxr-xr-x    2 node     node          1024 Oct 15 02:58 .data
-rw-r--r--    1 nonroot  nonroot     285568 Sep 29 20:33 index.cjs
drwxr-xr-x    1 nonroot  nonroot       1024 Sep 29 20:33 public
```

So the container was trying to write to a data directory owned by 1000:1000 (node) while operating as 100:101

Using the `user` directive in the compose file to set the user to 1000:1000, instead of using environment variables is the fix:

```
/app $ whoami
node
/app $ id
uid=1000(node) gid=1000(node) groups=1000(node)
/app $ ls -la
total 284
drwxr-xr-x    1 nonroot  nonroot       1024 Sep 29 20:36 .
drwxr-xr-x    1 root     root          1024 Oct 15 03:14 ..
drwxr-xr-x    2 node     node          1024 Oct 15 02:58 .data
-rw-r--r--    1 nonroot  nonroot     285568 Sep 29 20:33 index.cjs
drwxr-xr-x    1 nonroot  nonroot       1024 Sep 29 20:33 public
```

Zoobdude:
I must admit, until using enclosed/umbrel I hadn't ever used alternative users in a docker container. Thanks for the explanation, I appreciate it!

nmfretz:
Gallery assets are now ready and we can go live . Thanks again for all your efforts on this @Zoobdude! 

image

Oni-giri:
Hi @Zoobdude could you share the internal url that we need to setup in cloudflare tunnel? I can't get it to work.

Zoobdude:
Hey @Oni-giri, I believe that would be http://localhost:8788

Oni-giri:
> Hey @Oni-giri, I believe that would be http://localhost:8788

Thanks a lot for the answer!! Unfortunately, this doesn't seem to work: check the url: enclosed8788 _dot_ yakitori.capital.
Capture d’écran 2024-11-10 à 01 50 45

Here are the params^

I'm running umbrelOS 1.2.2

I tried to use app_host instead, but it doesn't work. It seems that the port specified in the docker-compose is different than the one in the umbrel-app, maybe that's the reason?

Zoobdude:
> I'm running umbrelOS 1.2.2
> 
> I tried to use app_host instead, but it doesn't work. It seems that the port specified in the docker-compose is different than the one in the umbrel-app, maybe that's the reason?

The reason two different ports are used is because 8787 is already used by another app available on the Umbrel app store. Therefore the next port number was allocated to allow someone to deploy both apps at the same time.

This port re-mapping is handled by the Umbrel app proxy, this is why it is different to what is in the compose file, which is as close to the original as possible.

Just to check, are you sure your `cloudflared` daemon is running correctly? This should be in the form of this Umbrel app - https://apps.umbrel.com/app/cloudflared

The status of this can be found on the cloudflared dashboard.

Maybe you could try proxying another app down it?

Oni-giri:
> Just to check, are you sure your `cloudflared` daemon is running correctly? This should be in the form of this Umbrel app - https://apps.umbrel.com/app/cloudflared
> 
> The status of this can be found on the cloudflared dashboard.
> 
> Maybe you could try proxying another app down the tunnel?

Yes, it's working fine, I'm using it for other services. Here is what is shown:

Capture d’écran 2024-11-10 à 02 31 12

and the app runs fine locally:

Capture d’écran 2024-11-10 à 02 33 00

You can check a working app here:

https://monero dot yakitori.capital/

I have noticed that some apps work using the app_host as target, others use umbrel.local:port.

nmfretz:
Hey @Oni-giri. Here is the wiki from the dev of the cloudflare app on the umbrel app store that walks you through what's required:
https://github.com/Radiokot/umbrel-cloudflared/wiki/How-to-set-up-Cloudflare-Tunnel-on-your-Umbrel#advanced-route-to-a-protected-app

Enclosed sits behind our app proxy, which  means it inherits the security properties of our umbrel auth, with the side-effect that you can't tunnel through it. I can't promise a timeline, but there are plans to allow users to optionally remove some of the proxies features on a per app basis.

For now though, what you can do is tunnel to the enclosed Docker container instead of the umbrel proxy container. The link above walks you through how to figure it out, but what you need to use is  `http://enclosed_web_1:8787`

Hope that gets you up and running! Remember to always take care when exposing services to the public internet, particularly those that do not have any authentication of their own.

Oni-giri:
@nmfretz Thanks a lot for the answer, it seems to work for me. I think it would be worth adding it to the short description of the app on the marketplace, as it is the only way to get it to work. Indeed, Enclosed will refuse to accept secrets if it uses an unencrypted http connection, which is the case when you use it on a local network.

Capture d’écran 2024-11-11 à 19 02 31

PR #1564, opened 2024-10-01
Update GitLab to 17.5.1

hugofnm:
This version contains improvements, important bug and security fixes.

> [!WARNING]
> GitLab strongly recommend that all self-managed GitLab installations be upgraded immediately.

https://about.gitlab.com/releases/2024/09/19/gitlab-17-4-released


nmfretz:
Thanks for this update @hugofnm! Gitlab is going to be a finicky one to update and we'll need to be careful not to break current installations.

For example, it looks like this update will break current installs (17.2.1) because Gitlab requires us to first upgrade to 17.3.x before upgrading further: https://docs.gitlab.com/ee/update/#upgrade-paths

```
It seems you are upgrading from 17.2.1 to 17.4.1.
It is required to upgrade to the latest 17.3.x version first before proceeding.
Please follow the upgrade documentation at https://docs.gitlab.com/ee/update/#upgrade-paths
```

We will need to create a `pre-start` bash script that detects and properly walks through the appropriate upgrade path. We'll need to do it in such a way that we are certain that each upgrade step is completed (e.g., database migrations, etc) before the next upgrade step is undertaken.

nmfretz:
For our consideration when updating: https://github.com/getumbrel/umbrel-apps/pull/1543 (cc @DavidGarciaCat)

DavidGarciaCat:
> For our consideration when updating: #1543 (cc @DavidGarciaCat)

Thanks for the ping, @nmfretz. I also got this warning when logging into Gitlab, so I had it on my to-do list to send this upgrade, and I would have probably reached the same conclusion (skip one version). Knowing this PR is already open, I won't send a new one.

nmfretz:
> Thanks for the ping, @nmfretz.

You're welcome. I responded again in the other PR FYI (mentioning it here because it is closed and so we won't get notifications): https://github.com/getumbrel/umbrel-apps/pull/1543#issuecomment-2391115531

DavidGarciaCat:
17.4.2 IS now available.

However, there is another problem: a CVE scored with 10/10 for current versions, so it might be important to get these updates live.

hugofnm:
> For example, it looks like this update will break current installs (17.2.1) because Gitlab requires us to first upgrade to 17.3.x before upgrading further: https://docs.gitlab.com/ee/update/#upgrade-paths

Well, thanks for warning @nmfretz. As I was using my private repo to push updates I didn't really checked if the 17.2.1 to 17.4.1 update would break the GitLab instance...

> We will need to create a `pre-start` bash script that detects and properly walks through the appropriate upgrade path. We'll need to do it in such a way that we are certain that each upgrade step is completed (e.g., database migrations, etc) before the next upgrade step is undertaken.

There is nearly no documentation regarding hooks so I used the NextCloud's pre-start as a base.
I've checked my script with a clean install of GitLab and it manages to upgrade all the way to 17.4.1.
But of course I'm not familiar with umbrelOS hooks so it could have some bugs 😅.

> For our consideration when updating: #1543 (cc @DavidGarciaCat)

Thanks @DavidGarciaCat, I've added the missing detail to the description of the app !

> 17.4.2 IS now available.
> 
> However, there is another problem: a CVE scored with 10/10 for current versions, so it might be important to get these updates live.

Well, according to the [GitLab Upgrade Path tool](https://gitlab-com.gitlab.io/support/toolbox/upgrade-path/) and [the image used here](https://hub.docker.com/r/zengxs/gitlab/tags), version 17.4.2 isn't available for now so I'll stick with the 17.4.1 for now.

nmfretz:
> There is nearly no documentation regarding hooks so I used the NextCloud's pre-start as a base.
I've checked my script with a clean install of GitLab and it manages to upgrade all the way to 17.4.1.
But of course I'm not familiar with umbrelOS hooks so it could have some bugs 😅.

Thanks for diving in and taking a crack at it @hugofnm! Great work. I'll go through tomorrow and test/finalize it.

I've just scanned the hook quickly, and am wondering about this:

```bash
wait_for_upgrade() {
	local -r max_seconds="${1}"
	local -r version="${2}"

	# Wait for the version file to be updated
	for i in $(seq 1 "${max_seconds}"); do
		echo "Attempt: ${i}"
		if [[ "$(cat "${GITLAB_RAILS_VERSION_FILE}")" == "${version}" ]]; then
			echo "Version file updated to: ${version}"

			# Wait for the app to be ready
			sleep 15
			return
		fi
		sleep 1
	done

	echo "Failed after ${max_seconds} seconds"
	exit 1
}
```

Q: In your research, did you find that the `gitlab-rails/VERSION` file is only written out with the newest gitlab version AFTER all migrations/upgrades are 100% complete? If so, then this is a great, simple way for us to be sure that the upgrade process is complete.

Waiting for each upgrade step to complete will be the most critical part of this script, because if we attempt to upgrade to the next version along the upgrade path before migrations/upgrades are 100% complete we'll risk breaking some installs and potentially irreversibly corrupting the database.

I did a quick search and it looks like migrations/upgrades are complete when this command returns a `0`:
```
gitlab-rails runner -e production 'puts Gitlab::BackgroundMigration.remaining'
```

So if we can't rely on the `gitlab-rails/VERSION` file, then we'll need to switch to checking `gitlab-rails runner -e production 'puts Gitlab::BackgroundMigration.remaining'` in the gitlab container.

hugofnm:
> Waiting for each upgrade step to complete will be the most critical part of this script, because if we attempt to upgrade to the next version along the upgrade path before migrations/upgrades are 100% complete we'll risk breaking some installs and potentially irreversibly corrupting the database.

You are right @nmfretz, we should take all precautions when upgrading GitLab. 
I've updated the pre-start hook to include a second check based on the GitLab upgrade guide : [https://docs.gitlab.com/ee/update/background_migrations.html#from-the-database](https://docs.gitlab.com/ee/update/background_migrations.html#from-the-database)

So the script will check GitLab Rails version and check if any batched background migration is pending.
I tested it on my RPI5 and the upgrade went well.

nmfretz:
> I've updated the pre-start hook to include a second check based on the GitLab upgrade guide : https://docs.gitlab.com/ee/update/background_migrations.html#from-the-database

Nice work finding this batched migrations check. Looks like it is the successor to `gitlab-rails runner -e production 'puts Gitlab::BackgroundMigration.remaining'` and what we should be using for GitLab versions > 15:

from https://docs.gitlab.com/ee/update/background_migrations.html#from-the-database
>Non-batched migrations are superseded by batched background migrations. Non-batched migrations were gradually phased out during GitLab 14, with the last one used in GitLab 15.0.

The initial release on the umbrel app store was version 17.2.1, so we are good with just this batched migration check 👍

### Changes to the pre-start script:

> I tested it on my RPI5 and the upgrade went well.

The pre-start script was not set as executable, so it wouldn't actually have run in your test. Interesting that Gitlab still updated without issue.
```
-rw-r--r-- 1 umbrel umbrel 4950 Oct 23 01:35 pre-start
```

I've made some changes to the script and tested updating from 17.2.1. It worked well for me, but based on the GitLab docs and forums I've looked at, it seems like upgrading Gitlab is incredibly finicky, and users may run into issues even if they follow the recommended migration path. In light of this, I'm going to add a warning at the top of the release notes.

Summary of changes:
- made exports.sh executable (`chmod +x`)
 - simplified migration path to `17.2.1 --> 17.3.3 --> 17.5.0`. The [official tool](https://gitlab-com.gitlab.io/support/toolbox/upgrade-path/?current=17.2.1&distro=docker&edition=ce) recommends 17.3.5, but a `zenggxs` image doesn't exist for that tag.
 - simplified the migration loop to only check batched migrations, not a version file in addition
 - allowed the batched migration check to fail without exiting the script. It takes quite a while for `gitlab-psql` to accept commands after startup, and if we don't allow it to fail then the script will exit and users will be left with a broken install.

@sharknoon - could I get you to please give this a check as well and try updating from 17.2.1 to see if migrations work for you?

sharknoon:
I have done the following to test GitLab:
1. Installed the current released GitLab from the official store (17.2.1)
2. Created a test repository with some content
3. Changed some settings in the preferences (dark mode)
4. Removed the gitlab folder `~/umbrel/app-stores/getumbrel-umbrel-apps-github-53f74447/gitlab`
5. Copied over the new gitlab folder from this PR
6. Run `umbreld client apps.update.mutate --appId gitlab`
7. Opened Gitlab

## The results
- Booting went fine
- Theming was still present
- My repo was still there
- Everything seems to be fin until I saw this scary red warning:
grafik
I am on 17.5 (which is the version we are updating to) but it still shows me, I should update (probably from 17.5.0 to 17.5.1)

nmfretz:
Thanks for testing @sharknoon!

> I am on 17.5 (which is the version we are updating to) but it still shows me, I should update (probably from 17.5.0 to 17.5.1)

Ah yes, I see that 17.5.1 was just released earlier today with security updates. I will bump the image (and equivalent pre-start hook logic) and do a final test.

nmfretz:
Just waiting on the maintainer to push the 17.5.1 image: https://github.com/zengxs/gitlab-arm64/pull/141

nmfretz:
Alrighty, the 17.5.1 image is now live on Docker Hub. I have bumped the image, made the corresponding changes to the pre-start script, and re-tested. Going live to the app store. Thanks for your efforts on this @hugofnm and @sharknoon.

### Notes on the pre-start for our records

- When gitlab is updated each time, we need to make sure that the pre-start script's migration logic will trigger properly no matter what version a user is on (that we have had on the app store). This may mean that we will eventually need to update the script to something like I did for Nextcloud recently: https://github.com/getumbrel/umbrel-apps/commit/45115397056814f355b1a2c2a0cae0fd67486f82

- The psql command to check for pending batched migrations cannot be passed in to the legacy `${UMBREL_ROOT}/scripts/app" compose` script due to the way the arguments are parsed. To get around this I used `docker exec` directly in the pre-start script. I have left a comment mentioning that we should change this in the future (likely when the apps framework refactor occurs for umbrelOS).

-The pre-start script has some hacks that we'll want to refine as time goes by and we see how updates go. For example, we loop for a maximum of 1 hour waiting for any batched migrations to finish. This seems pretty conservative to me, but perhaps migrations can take a super long time with a large database.

PR #1565, opened 2024-10-01
App Submission: Formicaio

bochaco:
# App Submission

### App name
Formicaio

### 256x256 SVG icon
https://svgshare.com/i/1B0F.svg
Formicaio icon

### Gallery images
- 
- 
- 

### I have tested my app on:
- [X ] umbrelOS on a Raspberry Pi
- [ ] umbrelOS on Linux VM


nmfretz:
Thanks for submitting Formicaio @bochaco!

nmfretz:
Hi @bochaco, thanks very much for the app submission! And apologies for the extreme delay in reviewing this, that is entirely my fault.

Would you be able to give me a little more info on what this app does? I'm not sure I really understand it based on the description and looking through the github repo.

>Simplify your decentralized experience with this intuitive application,
  designed to streamline your daily tasks when running nodes from home
  on peer-to-peer (P2P) networks. Seamlessly participate in online
  communities using the integrated Nostr client, and manage your
  digital assets with ease through the built-in wallet. Receive, send,
  and store tokens, rewards, and coins earned from running nodes or received
  from third-party sources, all within a single, user-friendly interface.
  
Also, before I do an in-depth review, can you please let me know if the elevated container permissions are necessary for the app to run? For example, Docker-in-Docker is running in privileged mode so the companion container can trivially escape to the host, which is a security consideration that probably prevents us from accepting this as-is.

Sorry again for the delay!

bochaco:
Hi @nmfretz , no worries. Let me elaborate a good response tomorrow, I did imagined DinD could be a problem, I'll explain what this is about, including info about a new approach I already implemented which doesn't require Dind. I'll reply tomorrow.

bochaco:
> Would you be able to give me a little more info on what this app does?

I may be reducing the scope currently described, and keep it simpler to what this is meant to offer in the short/medium term. The main function of the app right now is to allow folks to run nodes on the [Autonomi network](autonomi.com) (I've shared this app with Autonomi community on this post: https://forum.autonomi.community/t/formicaio/40445). This is a P2P network, where people have nodes to share their storage and bandwidth in exchange for ANT tokens as rewards.

I have for the longer-term some ideas around having an integrated wallet which can allow users to move those tokens they receive as rewards for running nodes on Formicaio, plus be able to use an integrated Nostr client, ....; but as I said, for now I have been and will be mostly focused on the aspect of running nodes only, so I'll be updating that in the description for sure.

> Docker-in-Docker is running in privileged mode so the companion container can trivially escape to the host

I originally started with using Portainer, but I didn't like it due to two reasons:
- UX: the user would need to install the Portainer app first.
- Portainer setup: I managed to have first version of Formicaio to setup an account which would interface with Docker through Portainer API but it was way more complicated than accessing a Docker engine API directly.
Thus, looking at Portainer's Umbrel docker-compose file, I basically copied the setup they have for D-in-D, and that was solving the above two aspects.

Recently I implemented another approach which doesn't require Docker, it spawns nodes as native processes (using Rust [process::Command](https://doc.rust-lang.org/std/process/struct.Command.html)), in this way each spawned node becomes a child process of the Formicaio process (this is all implemented in [node_manager](https://github.com/bochaco/formicaio/blob/main/src/node_manager.rs) behind `native` cargo-feature).

Now, I was about to update my Umbrel docker-compose file to use this new approach in order to remove the DinD, but I still have something else to solve: when you run nodes on Formicaio, you decide if their ports will be publicly exposed by the user (e.g. public IP, or port forwarding, ...) or not, this is why I'd need to run Formicaio with `network_mode: host` in case nodes need to expose their ports (which are chosen by the user when spawning those nodes). The problem with this is that I still have to see how to enable the Tor proxy with `app_proxy` service, since it obviously collides with Formicaio port when running in `network_mode: host` mode.

Would this non-DinD approach be better from a security standpoint? if so, I could work on trying to have that setup, I could have a tiny proxy which may solve the issue with Tor/`app_proxy` collision (I can explain more about this if you wish).

At the moment, you can take a look at how current users are running it using my community-app-store: https://github.com/bochaco/formicaio-app-store/blob/master/formicaio-apps-formicaio/docker-compose.yml (you'll see I have a [tiny proxy developed by me](https://github.com/bochaco/metrics_proxy) to overcome the issue with `app_proxy` collision).

Another not so important reason why I set up Formicaio with privileged mode is to be able to access the I2C interface for LCD displays, but this is something not so important perhaps, and I can easily disable for UmbrelOS app deployments if it's an issue: https://github.com/bochaco/formicaio?tab=readme-ov-file#displaying-nodes-stats-on-external-lcd-device .

I'm sorry for such a long response, but just in case you needed all this information, I can explain more details of any of this if you need it, or if something is still confusing, just let me know.

bochaco:
@nmfretz 
I've just pushed a commit where I'm changing it to make use of the other approach I was mentioning before where DinD is removed.
You'll see I'm using a [tiny proxy](https://github.com/bochaco/formicaio_proxy) I've coded myself to have Formicaio working in `network_mode: host` without colliding with Umbrel's `app_proxy`, which is what I was trying to explain before (using nginx may be an option too, I haven't tried it).

If there is a way to configure `app_proxy` so I don't need my own proxy then fine, I just couldn't find a way to do it, I presume you never had to consider an scenario like this app...?....

Please, just let me know if you need more info/details.

formicaio/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
+services:
+  app_proxy:
+    environment:
+      APP_HOST: formicaio_formicaio_proxy_1
```
al-lac, line 4: After adapting the name of the service, change the name to the following.
```suggestion
      APP_HOST: formicaio_proxy_1
```

formicaio/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+      APP_PORT: 52100
+    container_name: formicaio-app_proxy_1
```
al-lac, line 6: container_name is not needed.
```suggestion
      APP_PORT: 52100

```

formicaio/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+  formicaio_proxy:
```
al-lac, line 7: Change name to proxy to make the readability better.
```suggestion
  proxy:
```

formicaio/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    image: bochaco/formicaio_proxy:0.1.0@sha256:de09a722629d5c3f7139eb7a9035ee29d4c6eb804c6031a7af39fe63a4dd298f
+    environment:
+      FORMICAIO_ADDR: host.docker.internal:3000
+      FORMICAIO_PROXY_PORT: 52100
+    extra_hosts:
+      - host.docker.internal:host-gateway
+    container_name: formicaio-formicaio_proxy_1
+  formicaio:
```
al-lac, line 15: Remove container name and rename formicaio service.
```suggestion

  app:
```

formicaio/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    image: bochaco/formicaio:0.3.0-native@sha256:b1cb3beedd399183b0ccf41622832cce67bc22b95edb3b07e24ae1a718632c36
```
al-lac, line 16: Update to latest version.
```suggestion
    image: bochaco/formicaio:0.4.9-native@sha256:2bea87c03cbd545b55627527fbe5ec21f7158f2aaa5ec6f9588b5bff1ec6a3d8
```

formicaio/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    privileged: true
+    restart: on-failure
+    network_mode: host
+    stop_grace_period: 5s
+    environment:
+      DB_PATH: /data
+      NODE_MGR_ROOT_DIR: /data
+      LEPTOS_SITE_ADDR: 0.0.0.0:3000
+    volumes:
+      - ${APP_DATA_DIR}/data/formicaio:/data
```
al-lac, line 26: The folder still needs to be added to the code like this:
`data/formicaio/.gitkeep`

formicaio/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    devices:
+      - /dev:/dev
+    container_name: formicaio-formicaio_1
```
al-lac, line 29: Remove container_name.
```suggestion

```

formicaio/umbrel-app.yml
```diff
@@ -0,0 +1,29 @@
+manifestVersion: 1
+id: formicaio
+name: Formicaio
+tagline: Il potere delle formiche
+icon: ""
+category: social
+version: "0.3.0"
```
al-lac, line 7: Update to latest version.
```suggestion
version: "0.4.9"
```

formicaio/umbrel-app.yml
```diff
@@ -0,0 +1,29 @@
+manifestVersion: 1
+id: formicaio
+name: Formicaio
+tagline: Il potere delle formiche
+icon: ""
+category: social
```
al-lac, line 6: Would this not fit better into the **bitcoin** or **network** section?

formicaio/umbrel-app.yml
```diff
@@ -0,0 +1,29 @@
+manifestVersion: 1
+id: formicaio
+name: Formicaio
+tagline: Il potere delle formiche
```
al-lac, line 4: I like the tagline, but an english one with more context what the app does would probably be better.

formicaio/umbrel-app.yml
```diff
@@ -0,0 +1,29 @@
...
+icon: ""
+category: social
+version: "0.3.0"
+port: 52100
+description: >-
```
al-lac, line 9: Maybe you could add some more details about what users need to do for port-forwarding and so on.

And in general more details would not hurt. Still kind of unclear what the app does exactly.

formicaio/1.png
al-lac, line 1: Please remove the images from the code, only needed in the PR details.

al-lac (changes requested):
Hey @bochaco!

Great that you got it to work without DinD.

I added some more points that would need changes. Please make sure to properly adapt the description as it is still a little unclear about details like port forwarding and other things.

The link for the logo svg seems to be down, maybe you can upload it again and update the link.

I will align with @nmfretz on how we can make progress here.

formicaio/umbrel-app.yml
bochaco: Agreed, I'll set it to bitcoin even that currently it's a bit more related to p2p networks and Ethereum tokens.

bochaco:
Thanks @al-lac , I've just pushed a commit with the requested changes. Please let me know if something else is still required to improve it.

formicaio/umbrel-app.yml
bochaco: ...I actually set it to 'networking' for now, until a category like 'etherum' or 'p2p networks' is available.

al-lac:
Hey @bochaco, thanks for adapting the changes! Lets wait for what @nmfretz has to say about the current state of it.

al-lac:
Hey @bochaco, thanks a lot for your patience on this PR! I had a discussion with @nmfretz and the following topics are still somewhat concerning for us:

- privileged container
- full device passthrough (- /dev:/dev)
- Host networking
- Extra host mapping to the Docker host

Do you think you could make some adaption so the app still works while running as user 1000 and not needed permissions like host networking and full device passthrough?

Let me know your thoughts and if this would be possible at all.

Thanks for your work here, we value it a lot!

bochaco:
> Do you think you could make some adaption so the app still works while running as user 1000 and not needed permissions like host networking and full device passthrough?

Hi @al-lac , thanks for your reply and your time!

I certainly could disable the feature for displaying [stats on LCD display](https://github.com/bochaco/formicaio?tab=readme-ov-file#lcd-display-support) which is the only reason for the need of full device pass-through, so that wouldn't be an issue at all.

However, the Formicaio app would still work in bridge mode, but the spawned nodes won't be able to receive connections from peers in the Autonomi network, thus that'd be a no-go for these nodes I'm afraid. These nodes in Formicaio may be able to connect using UPnP to peers, but I don't think peers would be able to connect back to them if they are behind the docker bridge mode, unless all these ports were somehow docker-exposed of course, which I don't think that is a viable option either....
I've always assumed this could be an issue for having it on Umbrel app-store, so I'm not surprised and I can understand.

al-lac:
Thanks for the quick reply @bochaco!

Got it! What about the privileged permissions and extra host mapping. Could those maybe be addresssed or are they also needed for the app to properly function?

Will further align with the team on how we can handle this.

bochaco:
> Got it! What about the privileged permissions and extra host mapping. Could those maybe be addresssed or are they also needed for the app to properly function?

IIRC the privileged permission is not needed if the use of device pass-through is removed, so that shouldn't be needed. 
The whole `formicaio_proxy` service, which makes use of the `extra_hosts`, is just a hack/workaround I came up with since I didn't know how to have the Formicaio web-UI accessible through the Umbrel's Tor proxy (given it runs in host-network mode), thus I created that dummy proxy that runs in bridge-mode and which works with Umbrel's Tor proxy. In turn, that proxy redirects the requests to the Formicaio app/backend running in host-mode. If there is another way of doing that, then the `formicaio_proxy` service could be removed completely, and I'll be very happy to do so.

al-lac:
Ok good, would it also be possible to run the containers as user 1000?

Regarding the proxy workaround, I also don't know how else this could be solved. Maybe @nmfretz or @sharknoon have an idea on this.

bochaco:
> would it also be possible to run the containers as user 1000?
 
I don't think that could be a problem, I will be giving it a try to test that, and the other changes.

bochaco:
@al-lac I've just pushed a commit which:
- sets `user: 1000:1000` for both services
- removes the `privileged: true`
- removes the device pass-through

With theses changes it's still working fine.

al-lac:
Hey @bochaco,

thanks a lot for your quick responses and adaptions, really appreciate it! 🔝 

For now, we don't have anything to add. We will prepare the gallery assets and let you know once those are ready and if we need some final adaptions before going live! 🚀

al-lac:
Hey @bochaco, thanks a lot for your patience and efforts in getting formicaio ready for the Umbrel App Store! 💯 

We have the gallery assets now and final tests from my side also worked without any issues.

This means we can finally go live! 🚀 

CleanShot 2025-11-14 at 16 39 26@2x

If you want to promote that your app is in the Umbrel App Store now, feel free to use these assets:

badge-dark-logo
badge-light-logo
badge-dark
badge-light

bochaco:
> This means we can finally go live! 🚀

Thanks a lot @al-lac !

al-lac:
Hey @bochaco,

sorry I provided the images without actually linking to the store, those would be the corrected ones:

[badge-dark](https://apps.umbrel.com/app/formicaio)
[badge-dark-logo](https://apps.umbrel.com/app/formicaio)

bochaco:
Thanks @al-lac !

PR #1590, opened 2024-10-03
Update mempool to v3.0.1 and set max heap size to 2gb

knorrium:
This release enables the Rust Get Block Template implementation by default for faster block indexing, there are no other code changes.

We're also taking the opportunity to limit the max heap size of the backend node process to 2gb.

### DIFF
--- a/mempool/docker-compose.yml
+++ b/mempool/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: $APP_MEMPOOL_PORT
       PROXY_AUTH_ADD: "false"
   web:
-    image: mempool/frontend:v3.0.0@sha256:ec84195e6d3050841efef99b9db2ec6f3b8cd64611df0c8dae026a41931ae58c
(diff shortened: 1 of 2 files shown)


wiz (approved):

wiz (approved):

nmfretz:
Thanks @knorrium! I have tested the update and it is working well.

We just need to make two changes to the `umbrel-app.yml` so that users are notified of an available app update in our UI:

Change the `version` to:

```yaml
version: "3.0.1"
```

Change release notes to:

```yaml
releaseNotes: >-
  This update brings faster block indexing and limits memory usage for better performance.


  Full release notes for the previous major update to version 3.0.0 are found at https://github.com/mempool/mempool/releases
  ```
  
Once this is done, we can go live.

knorrium:
@nmfretz thanks for reviewing - I completely forgot about the app metadata 😂

nmfretz:
No worries at all. Everything looks great. Going live.

PR #1605, opened 2024-10-06
Update lnbits to v0.12.11

nmfretz:
Currently untested


nmfretz:
Tested on arm64 and x86 (fresh install and update)

PR #1608, opened 2024-10-08
App Submission: Phantombot

kriakiku:
# App Submission

### App name
Phantombot

### Disclamer

- The service is required to run over HTTPS; otherwise, users won’t be able to connect their Twitch accounts. As a result, the proxy_server is not included in the docker-compose.
- I had to retain the root user, as the application fails to start otherwise.

### 256x256 SVG icon

https://svgshare.com/i/1BJ0.svg

Phantombot

### Gallery images

**1.png:**
Phantombot

**1.png (original):**
umbrel local_25000_panel_

**2.png:**
umbrel local_25000_panel_ (1)

**3.png:**
umbrel local_25000_panel_ (2)

**4.png:**
umbrel local_25000_

**5.png:**
umbrel local_25000_oauth_

### I have tested my app on:
- [x] umbrelOS on Linux VM


kriakiku:
Custom AppStore: https://github.com/kriakiku/umbrel-app-store

screenshot

kriakiku:
@nmfretz Hi! Is there anything else required from me for the approval of this merge request?

kriakiku:
image

phantombot/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+services:
```
nmfretz, line 1: Umbrel's app proxy service should be added here as the first service, which can then point to the `phantombot_server_1` container and default port `25000`. That way you can remove the port mapping in your `server` service and then PhantomBot will running behind our transparent proxy and inherit the security properties of our umbrel auth:

firefox example to follow
https://github.com/getumbrel/umbrel-apps/blob/2dbb41875664e250d34211d395523e46550b5def/firefox/docker-compose.yml#L3-L7

- By default, we only let requests through the proxy if the user already has a valid auth cookie from the Umbrel homescreen. So if the PhantomBot dashboard has no auth of its own, it will still be protected by ours. This gives a good UX because there's zero friction for Umbrel users... they can just access the dashboard without re-entering any credentials if they are already logged in to their Umbrel.
- It also has the benefit of inheriting other security properties of Umbrel auth, such as 2FA if they have it enabled on their Umbrel. They would then get 2FA security for all apps behind the auth proxy too.
- And then if you need to allow external connections to PhantomBot running on port 25000 (e.g., not to the UI but to the api or something), you can whitelist certain routes, so that the web UI is protected by auth, but something like /api/* isn't:

e.g., adding this env var to the proxy container
```yaml
PROXY_AUTH_WHITELIST: "/api/*"
```

- You can also disable the auth portion of the app proxy container entirely with this env var:
```yaml
PROXY_AUTH_ADD: "false"
```

phantombot/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+  server:
+    image: ghcr.io/phantombot/phantombot:3.14.1.0@sha256:dbec9818e40f967ac5aee3abcac5a1857481cbbe6b35400f9a2fa8f1dc638df0
+    user: 0:900
```
nmfretz, line 4: If possible, we want to avoid running services as `root`. I haven't looked at the phantombot Dockerfile to see how they build the image so we might be stuck with this. But we should check if it is possible to run this non-root as the umbrel user (`1000:1000`).

phantombot/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    restart: on-failure
+    volumes:
+      - ${APP_DATA_DIR}/data:/opt/PhantomBot_data
+    ports:
+      - "${APP_PHANTOM_SERVER_PORT}:${APP_PHANTOM_SERVER_PORT}"
```
nmfretz, line 9: This can be removed entirely once the app proxy service is added

phantombot/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    environment:
+      PHANTOMBOT_USEHTTPS: "true"
```
nmfretz, line 11: I haven't tested this submission yet, but I'd imagine this forces user to use https, so they'd click the app and go to https://umbrel.local:25000, is that right? If so they'll be met with the big scary insecure warning that we shouldn't be teaching the average umbrel user to just click through.

We can set this to false, so the user is accessing over http on their local network.

phantombot/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+      PHANTOMBOT_PANELUSER: umbrel
+      PHANTOMBOT_PANELPASSWORD: $APP_PASSWORD
+      PHANTOMBOT_BASEPORT: $APP_PHANTOM_SERVER_PORT
```
nmfretz, line 14: if this is the default port that phantombot runs on, then this can be removed entirely

phantombot/exports.sh
```diff
@@ -0,0 +1 @@
+export APP_PHANTOM_SERVER_PORT="25000"
```
nmfretz, line 1: If this is the default port, then this entire exports.sh should be removed for simplicity

phantombot/umbrel-app.yml
```diff
@@ -0,0 +1,36 @@
+manifestVersion: 1.1
+id: phantombot
+category: social
+name: PhantomBot
+version: "3.14.1.0"
+tagline: PhantomBot is a Twitch chat bot powered by Java
+icon: ""
```
nmfretz, line 7: This entire `icon` line can be removed. We'll host the icon here: https://github.com/getumbrel/umbrel-apps-gallery

And the icon rendering logic in umbrelOS will grab the correct icon.

phantombot/umbrel-app.yml
```diff
@@ -0,0 +1,36 @@
...
+   Required! If you don't perform the initial setup, you will encounter an error when connecting to the control panel!
+
+  1. Connect Your Twitch Account
+     Using the instructions on the page https://umbrel.local:25000/oauth/, create oauth app and connect your Twitch account.
+
+  2. Fill Chanel Configuration Fields
+     Complete the channel and owner fields on the configuration page https://umbrel.local:25000/setup/.
+
+  3. Log In
+     Done! You can log in to the control panel at https://umbrel.local:25000/panel/login using the default credentials.
```
nmfretz, line 24: Awesome work adding these!

If users can navigate to these links by clicking things within the UI, then I would suggest making these instructions more generic. For example, telling the user to "navigate to the oauth settings" instead of providing a specific link.

My reasoning here is that:
1) These links may change in future app updates and it will be difficult to remember to update the app description

2) It's possible that the user accesses their umbrel at a different hostname than `umbrel.local`. For example, they may access their umbrel via local IP address, or by Tailscale, or if they have multiple Umbrel devices they may be using umbrel-2.local

nmfretz (commented):
Thanks for submitting PhantomBot @kriakiku. This looks like a fantastic addition!

Really sorry for the delay in reviewing this. I'll try and be more prompt in follow-ups here, but sometimes I'll have my focus elsewhere.

I've left a review below. Once addressed I can give the app a test. Let me know if anything is confusing or if I've misunderstood something and I can take another look.

nmfretz:
image

Converting to draft for now to keep things organized. @kriakiku ping me as soon as you're ready for me to test!

nmfretz:
Hey @kriakiku, checking in to see if you would still like to get Phantombot working on umbrel?

PR #1606, opened 2024-10-08
App Submission: Baikal - a Calendar + Contacts server 

dennysubke:
# App Submission

### Baikal

h3B1DaG - Imgur

### 

Baikal Server is a lightweight CalDAV and CardDAV server designed for users who want to manage their calendar and contact information securely on their own server. This robust application offers an extensive web interface that simplifies the management of users, address books, and calendars, making it accessible for both beginners and experienced users.

###
InOvJFa - Imgur

hIQ1ZIv - Imgur

nobRqdE - Imgur

### I have tested my app on:
- [ ] umbrelOS on a Raspberry Pi
- [X] umbrelOS on an Umbrel Home
- [ ] umbrelOS on Linux VM

PR #1657, opened 2024-10-16
App Submission: PrivateBin

dennysubke:
# App Submission

### PrivateBin
...

### Icon
Jz4lSRj - Imgur

...

### Gallery images
A

B

C

D

...

### I have tested my app on:
- [ ] umbrelOS on Linux VM


dennysubke:
I would like to request the addition of PrivateBin as an app in the Umbrel App Store. However, since PrivateBin requires HTTPS for proper functionality, its setup would only be possible when used in conjunction with the Umbrel Cloudflare app.

nmfretz:
Thanks for yet another submission @dennysubke!  

Would you be able to take a look through the recent Baikal review https://github.com/getumbrel/umbrel-apps/pull/1606 and make any equivalent changes here? After that I can go through in detail.

Also, from your experience getting this working do you know if we need to supply all of the `.php` files ourselves or will the app create them on first boot?

image

dennysubke:
> Thanks for yet another submission @dennysubke!
> 
> Would you be able to take a look through the recent Baikal review #1606 and make any equivalent changes here? After that I can go through in detail.
> 
> Also, from your experience getting this working do you know if we need to supply all of the `.php` files ourselves or will the app create them on first boot?
> image

I’ve tested it and can confirm that the .php files are not needed. Theme changes are still possible, indicating that these files don’t play a role at all.

nmfretz:
> I’ve tested it and can confirm that the .php files are not needed. Theme changes are still possible, indicating that these files don’t play a role at all.

Nice, that's great to hear 👌

Okay, so if you can go through the Baikal review changes (https://github.com/getumbrel/umbrel-apps/pull/1606) when you have time and make equivalent ones here (e.g., data directory structure, icon line removal, etc etc), make sure it's working fine for you, and then ping me and I will give it a test.

dennysubke:
> > I’ve tested it and can confirm that the .php files are not needed. Theme changes are still possible, indicating that these files don’t play a role at all.
> 
> Nice, that's great to hear 👌
> 
> Okay, so if you can go through the recent [Baikal app review changes](https://github.com/getumbrel/umbrel-apps/pull/1606) when you have time and make equivalent ones here (e.g., data directory structure, icon line removal, etc etc), test that it's working fine for you, and then ping me and I will give it a test.

I’ve made the corresponding changes (updated data directory structure, removed the icon line, etc.) and tested everything - it’s all working smoothly. I also completely deleted the `tpl` folder with the `.php` files, as discussed. 

However, the `conf.php` file in the root directory is still relevant and should remain. 

Let me know when you have a chance to take a look. Thanks! 😎

nmfretz:
Hey @dennysubke, I've made some additional changes to get this finalized. I've kept each commit atomic and descriptive so you can follow along with what I've done.

 >However, the conf.php file in the root directory is still relevant and should remain.
 
This actually works great for me without binding this conf.php. This way a default conf.php file is used when privatebin starts up and we never have to worry about needing to update this file when  updating the app in the app store (e.g., if some conf parameter is deprecated or new ones get added)... new versions of privatebin will always start up with the correct config.

Let's see how users find PrivateBin, and if there is a desire to bind the conf.php file, then we can add that complexity later.
 
 I have tested with a Cloudflare tunnel and it is working well 👌 Thanks again for submitting PrivateBin!
 
image

dennysubke:
> 
>  However, the conf.php file in the root directory is still relevant and should remain.
> 
> This actually works great for me without binding this conf.php. This way a default conf.php file is used when privatebin starts up and we never have to worry about needing to update this file when updating the app in the app store (e.g., if some conf parameter is deprecated or new ones get added)... new versions of privatebin will always start up with the correct config.
> 
> Let's see how users find PrivateBin, and if there is a desire to bind the conf.php file, then we can add that complexity later.
> 

Thank you for publishing the app. 😎 

Unfortunately, I'm not entirely happy with not using `config.php`. Without a custom `config.php` important features will not be available. These include the ability to customize note expiration, adjust encryption settings, or modify the theme (especially dark mode). 

See Screenshots: 


Without `config.php`

without config php



With `config.php`

with config php

nmfretz:
Hey @dennysubke. Thanks for the clarification. We can just send out an update to include the config.php.

> Unfortunately, I'm not entirely happy with not using config.php. Without a custom config.php important features will not be available. These include the ability to customize note expiration, adjust encryption settings, or modify the theme (especially dark mode).

Without the config.php I can still adjust expiration, burn after reading, and password settings. But I don't have dark mode accessible through the UI. This is a bummer, because apps should really expose importing config settings via env vars or ideally through the UI. But we can't really control that here...

As long as you are okay with the maintenance required with adding a config.php then we can add it. To be clear, this means that for app updates you'll need to check if any modifications to the supplied `config.php` file are required, and if so that we properly handle any required changes for _both_ new installs and existing installs.

dennysubke:
> Hey @dennysubke. Thanks for the clarification. We can just send out an update to include the config.php.
> 
> > Unfortunately, I'm not entirely happy with not using config.php. Without a custom config.php important features will not be available. These include the ability to customize note expiration, adjust encryption settings, or modify the theme (especially dark mode).
> 
> Without the config.php I can still adjust expiration, burn after reading, and password settings. But I don't have dark mode accessible through the UI or any other config options. This is a bummer, because apps should really expose importing config settings via env vars or ideally through the UI. But that's out of our control since we're not the devs of the app, and we don't want to unnecessarily reduce functionality.
> 
> As long as you are okay with the maintenance required with adding a config.php then we can add it. To be clear, this means that for app updates you'll need to check if any modifications to the supplied `config.php` file are required, and if so that we properly handle any required changes for _both_ new installs and existing installs.

Thank you for the explanation. Yes, please go ahead and add the `config.php` file to enable the additional settings. I'm aware of the maintenance needed for future updates and am fine with checking and adjusting the `config.php` as required.

Thank you for your efforts!

PR #1661, opened 2024-10-17
App Submission: Datum

Retropex:
# App Submission

### App name
Datum

### 256x256 SVG

soon™

### Gallery images

soon™

### I have tested my app on:


Retropex:
Since Datum haven't been released yet you shouldn't be able to pull the image.

nmfretz:
Really stoked for this one @Retropex! Thanks for working on this.

> Since Datum haven't been released yet you shouldn't be able to pull the image.
 
Sounds good. I've removed the linting bot's earlier comment since that is pretty much what it is complaining about at this stage.

I can leave some initial feedback on packaging to get the ball rolling.
 
Let's do the following for the file structure to remove unneeded files and align app data with the convention used for most other apps:
- remove `.DS_Store` (you can set your git to just globally ignore .DS_Store files)
- Remove the `icon/` dir
- Move `settings/` one level deeper into a new `data/` dir

So it should look something like:
 
 ```
 datum/
├── docker-compose.yml
├── umbrel-app.yml
└── data/
    └── settings/
        └── datum_gateway_config.json
```

File: datum/docker-compose.yml [hunk 1]
```diff
@@ -0,0 +1,11 @@
+version: "3.7"
+
+services:
+  datum:
+    image: ghcr.io/retropex/datum:v1.3
+    volumes:
+      - type: bind
+        source: ${APP_DATA_DIR}/settings/datum_gateway_config.json
+        target: /app/datum_gateway_config.json
```
nmfretz, line 9: Since we're moving `settings` one level deeper inside the `data` directory we can change to:
```suggestion
    volumes:
        - ${APP_DATA_DIR}/data/settings/datum_gateway_config.json:/app/datum_gateway_config.json
```

Do you know if the config file is the only settings file that needs to be persisted? Or is there any other data we need to persist?

datum/umbrel-app.yml
```diff
@@ -0,0 +1,26 @@
+manifestVersion: 1.1
+id: datum
+category: Bitcoin
```
nmfretz, line 3: This just needs to be all lowercase:
```suggestion
category: bitcoin
```

datum/docker-compose.yml
nmfretz, line 9: Since we're moving `settings` one level deeper inside the `data` directory we can change to:
```suggestion
    volumes:
        - ${APP_DATA_DIR}/data/settings/datum_gateway_config.json:/app/datum_gateway_config.json
```

Do you know if the config file is the only settings file that needs to be persisted? Or is there any other data we need to persist?

datum/umbrel-app.yml
```diff
@@ -0,0 +1,26 @@
...
+name: Datum
+version: "1"
+tagline: Make your own mining template
+icon: https://raw.githubusercontent.com/Retropex/Bitcoin-store/refs/heads/datum/btc-datum/icon/datum_logo.svg
```
nmfretz, line 7: This can be removed entirely for the official app store version:
```suggestion
```

datum/umbrel-app.yml
```diff
@@ -0,0 +1,26 @@
...
+  Powered by Bitcoin Knots: https://bitcoinknots.org/
+developer: Ocean
+website: https://ocean.xyz
+dependencies:
+  - bitcoin-knots
+repo: https://github.com/ocean-xyz
+support: https://ocean.xyz
+port: 21000
+gallery:
+  - https://raw.githubusercontent.com/Retropex/Bitcoin-store/refs/heads/datum/btc-datum/icon/preview.png
```
nmfretz, line 21: We can stop the linting bot from complaining for now by just setting this to:
```suggestion
gallery: [ ]
```

When it comes time to ship Datum, we'll add gallery assets here: https://github.com/getumbrel/umbrel-apps-gallery

datum/umbrel-app.yml
```diff
@@ -0,0 +1,26 @@
...
+path: ""
+defaultPassword: ""
+releaseNotes: >-
```
nmfretz, line 24:
```suggestion
releaseNotes: ""
```

datum/umbrel-app.yml
```diff
@@ -0,0 +1,26 @@
...
+submitter: Léo Haf
+submission: https://github.com/Retropex/Bitcoin-store/pull/1
```
nmfretz, line 26: We can change this now to this submission:
```suggestion
submission: https://github.com/getumbrel/umbrel-apps/pull/1661
```

datum/docker-compose.yml
```diff
@@ -0,0 +1,11 @@
...
+    ports:
+      - 21000:21000
```
nmfretz, line 11: Instead of defining a port mapping here, what we can do is add our app proxy container as a service in this compose file. And then point the proxy at the `datum` container.

See your Bitcoin Knots app as an example:
https://github.com/getumbrel/umbrel-apps/blob/d33eef2391a2896194275298b9b884af924f738c/bitcoin-knots/docker-compose.yml#L4-L7

So you'd do this:
```yml
app_proxy:
    environment:
      APP_HOST: datum_datum_1
      APP_PORT: 21000
```
That way Datum is running behind our transparent proxy and inherits the security properties of our umbrel auth:

- By default, we only let requests through the proxy if the user already has a valid auth cookie from the Umbrel homescreen. So if datum has no auth of its own, it is protected by ours.
- It also has the benefit of inheriting other security properties of Umbrel auth, such as 2FA if they have it enabled on their Umbrel. They would then get 2FA security for all apps behind the auth proxy too
- And then if you need to allow external connections to datum running on port 21000 (e.g., from something not accessing the UI), we can whitelist certain routes, so that the web UI is protected by auth, but something like `/connect-to-datum/*` isn't:

e.g., adding this env var to the proxy container
```
PROXY_AUTH_WHITELIST: "/api/*"
```

- You can also disable the auth portion of the app proxy container entirely with this env var:
```yaml
PROXY_AUTH_ADD: "false"
```

Would something like this be compatible with how datum will work?

datum/settings/datum_gateway_config.json
```diff
@@ -0,0 +1,15 @@
...
+    },
+    "api":{
+         "listen_port": 21000
+    },
+    "mining":{
+        "pool_address": "enter your bitcoin address",
+        "coinbase_tag_primary": "enter your block tag",
+        "coinbase_unique_id": 1
+    }
+}
```
nmfretz, line 15: So since the UI is just a read-only dashboard with no ability to adjust configuration, it sounds like we'll be shipping the app with sensible defaults that will make sense for most users. And then we can provide instructions for how to update config via CLI (e.g., naming blocks).

Do you know if any of this is configurable by environment variables in the datum container?

We could take advantage of the existing Knots environment variables and do something really simple so users are configured to Knots on install:
```yaml
datum:
  image: ghcr.io/retropex/datum:v1.3@sha:<digest>
  environment:
    RPC_USER: "${APP_BITCOIN_KNOTS_RPC_USER}"
    RPC_PASSWORD: "${APP_BITCOIN_KNOTS_RPC_PASS}"
    BITCOIN_HOST: "${APP_BITCOIN_KNOTS_NODE_IP}"
    RPC_PORT: "${APP_BITCOIN_KNOTS_INTERNAL_RPC_PORT}"
```

Ideally, we'd pass in these env vars which would tell Datum how to create the initial config file without us having to supply one from the host on install.

datum/docker-compose.yml
Retropex: Only the config file the rest of data is managed by the pool and the node.

Retropex:
Hello @nmfretz thanks for having taking time to review this!

For now the dashboard is the one of datum but @BitcoinMechanic was thinking about forking the bitcoin UI to adapt it to datum, do you think this will be possible?

datum/settings/datum_gateway_config.json
nmfretz: No worries @Retropex, leave that one to me! I'm thinking what we'll do is create a `pre-install` script that writes out the `dataum_gateway_config.json` with the users bitcoind config, so that when they install the app they are fully connected to Knots without having to mess around.

And then because it is a `pre-install`-only script, it won't overwrite this file on app restart or update

datum/docker-compose.yml
Retropex: Good idea, the only port who need to be freely available for ASIC's is the port `23334`, should I expose it with the `docker-compose.yml`

datum/settings/datum_gateway_config.json
nmfretz: Are you okay if I push some commits?

datum/docker-compose.yml
Retropex: also if I add the following block to `docker-compose.yml` should I remove the port under datum section?
```
app_proxy:
    environment:
      APP_HOST: datum_datum_1
      APP_PORT: 21000
```

nmfretz: Yep, that's exactly right.
- add the 23334 port binding to the `docker-compose.yml` if that is the port that the asic will connect on
- remove the 21000 port binding under the `datum` service

datum/settings/datum_gateway_config.json
Retropex: sure!

nmfretz: Added a pre-start script here that configures to Knots RPC: https://github.com/getumbrel/umbrel-apps/pull/1661/commits/92b6554ce7dda21608ecd2440c1515ea4ab1b864

nmfretz:
@Retropex, thanks for making those changes.

I've pushed some commits to finalize the packaging. I've also added a hook that checks whether Datum has been initialized with Knots' RPC credentials. If not, then it updates `datum_gateway_config.json` with the user's credentials. This way when a user installs DATUM they will automatically configure to Knots.

nmfretz:
Once the DATUM Docker image is ready we can properly test and make any necessary changes. Both @lukechilds and I have Knots synced and ready to test this.

Here's a checklist of remaining items for us:
- [ ] icon and gallery assets - (umbrel)
- [ ] add production Docker image (tagged, multi-arch, pinned to digest) - @Retropex/OCEAN
- [ ] test on x86 and arm64 - (umbrel) 
- [ ] Go through `umbrel-app.yml` for final tweaks before launch

nmfretz:
> Should I squash all commits before you merge?

Nope, no need. Thanks for checking. I'll just `Squash and merge` when ready. Just getting gallery assets together and then we'll go live.

nmfretz:
Thanks again for all your efforts with Datum @Retropex. Gallery assets are done and we are going live 🎉

image

Kubaaaaaaaaaa:
Would it be possible to modify the PRE-SART script to check not only the ENV. variables of the App Store version of Knots, but also the variables used in the community version from the Retropex app store (variables compatible with Bitcoin Core)?

APP_BITCOIN_KNOTS_NODE_IP -> BITCOIN_HOST
APP_BITCOIN_KNOTS_INTERNAL_RPC_PORT -> BITCOIN_RPC_PORT
APP_BITCOIN_KNOTS_RPC_USER -> BITCOIN_RPC_USER
APP_BITCOIN_KNOTS_RPC_PASS -> BITCOIN_RPC_PASSWORD

This will ensure compatibility with the community version of Knots and functionality with other apps that require Bitcoin Core.
Thank you,
Jakub

berling2009:
Could you add Best Share Statistics to the Datum Dashboard? It is very important for solo miners. Thank you.

adozenlines:
Adding a widget like public pool will be a nice to have.

Retropex:
Coming soon @adozenlines

PR #1666, opened 2024-10-19
full bitcoin knots policy settings

Retropex:


nmfretz:
Thanks @Retropex.

Noting here that this will need to be merged with or before Datum (https://github.com/getumbrel/umbrel-apps/pull/1661) so that users who install Datum can update Knots to be fully compatible.

There's a couple things to do to finalize this PR:
- [ ] Update the app's version in the `umbrel-app.yml` so that existing installs are notified that an update is available. You could go with something like `version: 27.1.1-datum` if you want to keep the Knots version as 27.1.1.
- [ ] Update release notes in the `umbrel-app.yml`. These could mention that new configuration options have been added for full compatibility with Datum.
- [ ] Add `blocknotify=curl -s -m 5 http://datum_datum_1:21000/NOTIFY` to the bin args in `exports.sh`

nmfretz:
@Retropex, I've made a couple changes to get this working (see below for details). Once we have confirmation from you and the Ocean+Knots teams that the default block template policy options in Knots are fine we can go live!

Based on the [Datum 0.2beta release notes](https://github.com/OCEAN-xyz/datum_gateway/releases/tag/v0.2beta), the recommended Knots config **when actually running Datum is**:
```
blockmaxsize=3985000
blockmaxweight=3985000 
blockreconstructionextratxn=1000000
maxmempool=1000
```

The current default policy based on this PR:
```
blockmaxsize=3985000
blockmaxweight=3985000
blockreconstructionextratxn=1000000
maxmempool=300
```

With this config, Datum users will need to read instructions somewhere to realize they should increase `maxmempool` (which may be what Knots + Datum wants to happen so that default Knots policy remains at 300 and doesn't use additional resources for non-datum users).

### Changes

1) Reverted the `bitcoind` service restart policy back to `unless-stopped` https://github.com/getumbrel/umbrel-apps/pull/1666/commits/6cf67a5e00e323c181cfd775faf8329cdbfaa5c5
The restart policy _must_ be `set to unless-stopped` instead of `on-failure`, because `bitcoind` needs to automatically restart after receiving a shutdown request (not just on error). When users change configurations in Advanced Settings, bitcoind is issued a shutdown request (i.e., rpc stop). Additionally, the Knots app checks on startup if a new configuration file needs to be generated, such as when a new installation is detected or if new default config options are added through an update, as done in this PR. So with an `on-failure` policy everyone would be left with a broken install because bitcoind would receive an rpc stop right as the app is initializing and wouldn't come back online.

2) Changed blocknotify to include datum port 7152 based on Telegram discussion. Sorry, I wrote port 21000 initially in my comment above. Can you + Datum team please confirm that 7152 is the correct port?

nmfretz:
Thanks again for your work on this @Retropex. Going live.

PR #1714, opened 2024-10-30
App Submission: Mattermost

al-lac:
# App Submission

### App name
Mattermost

### 256x256 SVG icon
https://mattermost.com/wp-content/uploads/2024/06/mattermost-logomark.svg

### Gallery images
https://mattermost.com/wp-content/uploads/2024/06/mattermost-home-hero.webp
https://mattermost.com/wp-content/uploads/2021/10/mattermost-platform-channels.webp
https://mattermost.com/wp-content/uploads/2021/10/mattermost-platform-playbooks.webp
https://mattermost.com/wp-content/uploads/2023/01/mattermost-platform-calls.webp

### I have tested my app on:
- [x] umbrelOS on Linux VM

### DIFF
--- a/mattermost/docker-compose.yml
+++ b/mattermost/docker-compose.yml
@@ -1,0 +1,41 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: mattermost_app_1
+      APP_PORT: 8000
+      PROXY_AUTH_ADD: "false"
+  
+  app:
+    image: alexlack/mattermost-app:v10.1.2@sha256:fa306060de89dce0d619537bc8cecc63289008f6ad719d65d5b83404ee8831d0
+    restart: on-failure
+    stop_grace_period: 1m
+    environment:
+      MM_USERNAME: admin
+      MM_PASSWORD: admin
+      MM_DBNAME: mattermost
+    volumes:
+      - ${APP_DATA_DIR}/data/app/mattermost/config:/mattermost/config:rw
+      - ${APP_DATA_DIR}/data/app/mattermost/data:/mattermost/data:rw
+      - ${APP_DATA_DIR}/data/app/mattermost/logs:/mattermost/logs:rw
+      - ${APP_DATA_DIR}/data/app/mattermost/plugins:/mattermost/plugins:rw
+      - ${APP_DATA_DIR}/data/app/mattermost/client-plugins:/mattermost/client/plugins:rw
+      - /etc/localtime:/etc/localtime:ro
+    ports:
+      - "8443:8443/tcp" # WebRTC over tcp
+      - "8443:8443/udp" # WebRTC over udp
+    depends_on:
+      - db
+
+  db:
+    image: alexlack/mattermost-db:v10.1.2@sha256:f75b632696b7d43f85be5c2bd3ac417568c1796819c4444cd3045360ff2bc084
+    restart: on-failure
+    stop_grace_period: 1m
+    environment:
+      MM_USERNAME: admin
+      MM_PASSWORD: admin
+      MM_DBNAME: mattermost
+    volumes:
+      - ${APP_DATA_DIR}/data/db:/var/lib/postgresql/data
+      - /etc/localtime:/etc/localtime:ro
--- a/mattermost/docker-compose.yml
+++ b/mattermost/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   app:
-    image: alexlack/mattermost-app:v10.1.2@sha256:e1d3e97877a153dd184965875828f089ef95e2f672695eab8d09e45b44609f34
+    image: alexlack/mattermost-app:v10.1.2@sha256:9f05d63488b580369bd74a166d40e3f1782842de5261642b3938c758043c850f
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/mattermost/docker-compose.yml
+++ b/mattermost/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   app:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for submitting Mattermost @al-lac! We've got our heads down elsewhere at the moment, but @sharknoon or I will get to this soon. Thanks for your patience.

sharknoon:
I will have a look later today :)

sharknoon (approved):
Thank you so much @al-lac for your app submission. I am blown away by this app submission, as this is one of the best I have ever seen. Great Job! I do not have a single point to improve.
@nmfretz This submission looks good to me

mattermost/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+services:
+  app_proxy:
+    environment:
+      APP_HOST: mattermost_app_1
+      APP_PORT: 8000
+      PROXY_AUTH_ADD: "false"
+  
+  app:
+    image: alexlack/mattermost-app:v10.1.2@sha256:fa306060de89dce0d619537bc8cecc63289008f6ad719d65d5b83404ee8831d0
+    user: "1000:1000"
```
nmfretz, line 12: It looks looks like the `mattermost-app` container must run as root or else the app will fail to start:

```
$ sudo docker logs -f mattermost_app_1
Fixing permissions on /mattermost/data
Fixing permissions on /mattermost/logs
Fixing permissions on /mattermost/config
Fixing permissions on /mattermost/plugins
Fixing permissions on /mattermost/client/plugins
su: must be suid to work properly
```

While not ideal, let's run as root and leave a comment:
```suggestion
    # Needs to run as root or the container fails to run
    # user: "1000:1000"
```

mattermost/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
...
+version: "10.1.2"
+port: 8765
+description: >-
+  🚉 Mattermost is an open-source platform designed for secure collaboration throughout the software development lifecycle.
+
+
+  💬 It serves as a team messaging application that emphasizes developer productivity and the benefits of open-source software.
+
+
+  💾 Mattermost offers a self-hosted alternative to proprietary SaaS messaging services such as Slack and Microsoft Teams.
```
nmfretz, line 15: In order for Mattermost to be useful it needs to be exposed to the internet, is that correct? e.g., in order for other's to use the server.

If so, then you could add a note in the app description that mentions this so that the average user isn't confused when they install the app and can't get it to work out of the box.

For example, you could add something along the lines of:

```yaml
⚠️ Mattermost needs to be exposed to the public internet in order to function as intended. The easiest way to do this is with the Cloudflare Tunnel app or Nginx Proxy Manager app from the Umbrel App Store.
```

nmfretz (commented):
Really great work on this @al-lac! Agree with @sharknoon, this is an impressive submission 🤝

We'll get the gallery assets made up so that we can go live.

In the meantime, there are two small things to do to get this finalized:

mattermost/docker-compose.yml
al-lac: Did it fail to start for you? I had it running with no issues. Also have it running as non-root via podman on a different raspberry pi of mine.

mattermost/umbrel-app.yml
al-lac: It for sure is of benefit if you expose it to the internet, but if you just want to use it as a local home messaging service, it would still work.

mattermost/docker-compose.yml
nmfretz: Yep, it fails to start with `su: must be suid to work properly`.  The app container gets stuck in a reboot loop.

> I had it running with no issues. Also have it running as non-root via podman on a different raspberry pi of mine.

For your podman install, were you using your `alexlack/mattermost-app` image and `user: "1000:1000"` in the compose file? Your initial submission before linting changes didn't have `user: "1000:1000"` https://github.com/getumbrel/umbrel-apps/pull/1714/commits/21fa8503e35bad8cdc599f12fca19d79c2f468c7.
Could you please give it a quick fresh install check to see if you get the same error as me?

nmfretz: It's this entrypoint file that it is erroring on when trying to run as non-root: https://github.com/al-lac/mattermost-docker/blob/main/app/entrypoint.sh

nmfretz: Running as non-root isn't a dealbreaker right now, so if you don't want to dig into the image too much, then we can ship this without `user: "1000: 1000"`

mattermost/umbrel-app.yml
nmfretz: Good point about local home messaging. I have added a note at the bottom of the description that mentions this https://github.com/getumbrel/umbrel-apps/pull/1714/commits/6054479fc1da135f55d7a23ee99ce56902623237

image

nmfretz:
@al-lac, the gallery assets are ready to go https://github.com/getumbrel/umbrel-apps-gallery/tree/master/mattermost.

Once the [user/permissions issue](https://github.com/getumbrel/umbrel-apps/pull/1714#discussion_r1834188042) is resolved we can go live!

image

mattermost/docker-compose.yml
al-lac: True, also failed for me. Really missed to test it after changing the user variable. Sorry about that!

Made some changes so it works also when user is set, see here: https://github.com/al-lac/mattermost-docker/blob/main/app/priv-entrypoint.sh

Published a new build and updated the sha256 in the PR.

Worked nicely on my pi right now.

nmfretz: Excellent, nice work on the fix!

I have tested on arm and x86 and it is working well.

nmfretz:
Great work on the permissions fix @al-lac. And thanks for your efforts bringing Mattermost to the app store. Going live 🎉

PR #1729, opened 2024-11-02
Bump RoboSats to v0.7.1-alpha

Reckless-Satoshi:
Simple version bump. Untested.


nmfretz:
Excellent, thanks for the update @Reckless-Satoshi! I have added release notes for v0.7.1-alpha and have reviewed and tested. Going live.

PR #1739, opened 2024-11-04
Update BTCPay Server to v2.0.1

dennisreimann:
I simply updated the images and could not test this further. Maybe someone can help verify that the upgrade works in production.

### DIFF
--- a/btcpay-server/docker-compose.yml
+++ b/btcpay-server/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   nbxplorer:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this update @dennisreimann! 2.0 is looking great 👌

I have [updated the app manifest](https://github.com/getumbrel/umbrel-apps/pull/1739/commits/94665ed750df4e0a79f0c4df43c597ae0e08b9a8) with version and release notes so that existing installs are notified of an available update like so:

image

I made sure to include a warning at the top regarding integration/plugin/api compatibility.

I have reviewed and tested a fresh install and app update.

dennisreimann:
Looking good! Thanks for including the disclaimer 🙌

PR #1742, opened 2024-11-05
App Submission: WebCheck

dennysubke:
# App Submission

### WebCheck
...

### Icon
GhoYf27 - Imgur

...

### Gallery images
1QWFJhW - Imgur

g9q7xuV - Imgur

3KVJib8 - Imgur

cs0lZzo - Imgur

FGIpxV4 - Imgur

...

### I have tested my app on:
- [ ] umbrelOS on Linux VM


dennysubke:
The Web Check app offers powerful website checks with a modern, user-friendly UI. Loading times, server responses and errors are displayed quickly and clearly. In my opinion this is perfect for integration with Umbrel. 😎

webcheck/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.9'
```
nmfretz, line 1: Let's change this to version 3.7 to ensure compatibility with past OS versions. It should be fine as-is, but I'd like to err on the side of caution.

FYI, this will change in our upcoming apps refactor, but for now let's always use 3.7/

webcheck/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: webcheck_server_1
+      APP_PORT: 3000
+      PROXY_AUTH_ADD: "false"
```
nmfretz, line 8: Shall we remove `PROXY_AUTH_ADD: "false"` entirely so that this app gets umbrel's transparent auth proxy? 

That way the WebCheck UI inherits the security properties of our umbrel auth:

- By default, we only let requests through the proxy if the user already has a valid auth cookie from the Umbrel homescreen. So since the WebCheck dashboard has no auth of its own, it will still be protected by ours. This gives a good UX because there's zero friction for Umbrel users... they can just access the dashboard without re-entering any credentials if they are already logged in to their Umbrel.

- It also has the benefit of inheriting other security properties of Umbrel auth, such as 2FA if they have it enabled on their Umbrel. They would then get 2FA security for all apps behind the auth proxy too.

What do you think?

webcheck/umbrel-app.yml
```diff
@@ -0,0 +1,46 @@
...
+developer: Alicia Sykes
+website: https://web-check.xyz/
+submitter: dennysubke
+submission: https://github.com/getumbrel/umbrel-apps/pull/1742
+repo: https://github.com/Lissy93/web-check
+support: https://github.com/Lissy93/web-check/issues
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
```
nmfretz, line 41: Since you have 4 lovely gallery images, you can add a 4.jpg here

nmfretz (commented):
Excellent submission @dennysubke! We'll do a once-over on the gallery assets before going live.

In the meantime, can you please take a look at these small changes:

webcheck/docker-compose.yml
dennysubke: Understood! :) I will take this into account in the future.

webcheck/umbrel-app.yml
dennysubke: Thanks for the tip! :)

dennysubke:
> Excellent submission @dennysubke! We'll do a once-over on the gallery assets before going live.
> 
> In the meantime, can you please take a look at these small changes:

@nmfretz Short message: All changes have been made. Thank you for your valuable help, and let me know if there's anything else you'd like to comment on. 👍

nmfretz:
Perfect, thanks for those final changes @dennysubke! Tested and going live.
 
image

PR #1743, opened 2024-11-06
Remove (temporarily) Specter Desktop from app store

nmfretz:
This PR removes Specter Desktop from the app store, due to a bug that prevents in from working with Bitcoin Core v28:
https://github.com/cryptoadvance/specter-desktop/issues/2473

It can be re-enabled in the app store once a new version of Specter Desktop has been released by the Specter team.

Behaviour for umbrelOS >= 1.0 users: the app will no longer show up in the app store

Behaviour for umbrelOS 0.5 users: the app will show up with a removal notice in the app description. If a user attempts to install the app they will see this error in the UI:

image

### DIFF
--- a/specter-desktop/umbrel-app.yml
+++ b/specter-desktop/umbrel-app.yml
@@ -1,4 +1,4 @@
-manifestVersion: 1
+manifestVersion: 1.2
 id: specter-desktop
 category: bitcoin
 name: Specter Desktop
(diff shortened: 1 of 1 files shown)


nmfretz:
The linting error shown above is fine in this case. The manifest version has been changed to `1.2` in order to make the app uninstallable on umbrelOS 0.5

PR #1751, opened 2024-11-06
App Submission: Zoraxy - Reverse Proxy Manager

dennysubke:
# App Submission

### Zoraxy - Reverse Proxy Manager
...

### Icon
logo

...

### Gallery images

...

### I have tested my app on:
- [x] umbrelOS on Linux VM


nmfretz:
Thanks for yet ANOTHER submission @dennysubke!

I don't have familiarity with Zoraxy, but do you know if this bind mount is needed:

```yaml
- /var/run/docker.sock:/var/run/docker.sock
``` 

Binding the host's docker socket is, unfortunately, a deal-breaker for security reasons. This would give Zoraxy complete access to the Docker daemon on the host, allowing it to manage other containers and easily break out of the containerized environment. With access to /var/run/docker.sock, the container gains root-level control over the host. Any process in the container could start privileged containers, access sensitive files, or even compromise the entire host system

There's a couple apps in our app store that get around this by running alongside a docker-in-docker (dind) container and binding to the dind's docker socket as a workaround (e.g., portainer and dockge).

dirstel:
docker.sock is "Used for additional functionality with Zoraxy." [Link](https://github.com/tobychui/zoraxy/tree/main/docker)
Maybe an app like zoraxy should be part of Umbrel itself? Thinking about hooks/configs to define a subdomain used by apps...

dennysubke:
> Thanks for yet ANOTHER submission @dennysubke!
> 
> I don't have familiarity with Zoraxy, but do you know if this bind mount is needed:
> 
> ```yaml
> - /var/run/docker.sock:/var/run/docker.sock
> ```
> 
> Binding the host's docker socket is, unfortunately, a deal-breaker for security reasons. This would give Zoraxy complete access to the Docker daemon on the host, allowing it to manage other containers and easily break out of the containerized environment. With access to /var/run/docker.sock, the container gains root-level control over the host. Any process in the container could start privileged containers, access sensitive files, or even compromise the entire host system
> 
> There's a couple apps in our app store that get around this by running alongside a docker-in-docker (dind) container and binding to the dind's docker socket as a workaround (e.g., portainer and dockge).

Good point - binding `/var/run/docker.sock` indeed poses security risks. Could you make the changes to use docker-in-docker (dind) instead?

zoraxy/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+      APP_HOST: zoraxy_server_1
+      APP_PORT: 8000
+      PROXY_AUTH_ADD: "false"
+  server:
+    image: zoraxydocker/zoraxy:v3.1.2@sha256:38f23c5a507542bbd6eac0492bf6456be2004015594ca270a1a5998c8a377b74
+    restart: on-failure
+   #ports: -> The proxy service is listening on port 8400
+      # 8480:80
+      # 8443:443
+      # 8400:8000
```
nmfretz, line 13: Since Zoraxy is a reverse proxy doesn't it need to expose http and https ports for its basic functionality?

User's are going to need to open ports on their router and then port forward to Zoraxy running on their Umbrel. You can bind to free ports that haven't been taken by other apps yet. So something like:

```yml
ports:
  - 41080:80
  - 41443:443
```

Then right at the top of the app description you could note which ports are for which protocol.

zoraxy/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+    volumes:
+      - ${APP_DATA_DIR}/config/:/opt/zoraxy/config/
```
nmfretz, line 15: Let's do the same thing as the other submissions and put this one level deeper on the host inside a parent `data` directory so that it has similar structure to the majority of other apps, and it makes it easy to add more bind mounts inside `data` in the future.

```
- ${APP_DATA_DIR}/data/config/:/opt/zoraxy/config/
```

Reminder, you'll need to make the equivalent change to the committed directories.

zoraxy/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+      # ${APP_DATA_DIR}/zerotier/config/:/var/lib/zerotier-one/ -> If you are not using ZeroTier, this directory is irrelevant.
+      - /var/run/docker.sock:/var/run/docker.sock
```
nmfretz, line 17: We can remove this based on the discussion above.

zoraxy/umbrel-app.yml
```diff
@@ -0,0 +1,52 @@
+manifestVersion: 1
+id: zoraxy
+name: Zoraxy Proxy Server
+tagline: An efficient reverse proxy server with automated SSL management
+category: networking
+version: "3.1.2"
+port: 8400
+description: >-
+  Zoraxy is a powerful, user-friendly reverse proxy server and forwarding tool, designed to provide a straightforward, flexible, and secure method for managing web traffic within home labs and developer environments. Written in Go, Zoraxy caters to both novice and experienced users by simplifying the process of routing HTTP requests to backend servers, enabling efficient and reliable service management without relying on traditional, more complex servers like NGINX or Apache.
```
nmfretz, line 9: I'm thinking we should add port instructions and a warning at the top here, what do you think? You can just steal it directly from the nginx proxy manager app:

https://github.com/getumbrel/umbrel-apps/blob/9432ddfefd4914409f36c7f092b749cc2c1dcacc/nginx-proxy-manager/umbrel-app.yml#L8-L17

zoraxy/umbrel-app.yml
```diff
@@ -0,0 +1,52 @@
...
+developer: Toby Chui
+website: https://zoraxy.aroz.org/
+submitter: dennysubke
+submission: https://github.com/getumbrel/umbrel-apps/pull/1751
+repo: https://github.com/tobychui/zoraxy
+support: https://github.com/tobychui/zoraxy/issues
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
```
nmfretz, line 47: You've got 5 lovely gallery images, so gotta add 4 and 5 😉

nmfretz:
> Maybe an app like zoraxy should be part of Umbrel itself? Thinking about hooks/configs to define a subdomain used by apps...

Thanks for the suggestion @dirstel. When we get a feature like this added (on the roadmap) we'll likely use something lower-level. We've got some ideas 😉

nmfretz (commented):
Apologies for the delay @dennysubke! Thanks again for yet another submission 🫡. 

> Good point - binding /var/run/docker.sock indeed poses security risks. Could you make the changes to use docker-in-docker (dind) instead?

Ah, sorry I hadn't really looked at the app at all when I made that comment. Now that I see what this is we can't use `dind` here because Zoraxy is binding to the Docker socket to get info on what Docker containers are running on the host. If we bind the socket of a `dind` container it would be useless because no containers will be running in it.

I just took a quick look through the [codebase](https://github.com/tobychui/zoraxy/tree/main) and I think there's a path forward for us here:
- the docker socket is being used for some quality of life improvements when running zoraxy via Docker. For example, when creating proxy rules it allows you to select from a list of running containers to get IP addresses, ports, etc.
- that makes things nice, but it isn't strictly required for zoraxy to run
- we can remove the socket bind and also add `DOCKER: "false"` as an env var which I _think_ will make it so the components in the UI that show running docker containers won't show up so users don't think something is broken.

I've gone through and left a review below. Let me know if anything is confusing.

zoraxy/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+services:
```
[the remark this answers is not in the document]
dennysubke: Done! Added version: '3.7' to ensure compatibility with UmbrelOS 0.5. 😊

zoraxy/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+services:
+  app_proxy:
+    environment:
+      APP_HOST: zoraxy_server_1
+      APP_PORT: 8000
+      PROXY_AUTH_ADD: "false"
```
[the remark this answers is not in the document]
dennysubke: Great, already taken care of!

zoraxy/umbrel-app.yml
dennysubke: Thank you! 😃 Added gallery images 4 and 5 to complete the set. 👌

dennysubke:
> Thanks for yet ANOTHER submission @dennysubke!
> 
> I don't have familiarity with Zoraxy, but do you know if this bind mount is needed:
> 
> ```yaml
> - /var/run/docker.sock:/var/run/docker.sock
> ```
> 
> Binding the host's docker socket is, unfortunately, a deal-breaker for security reasons. This would give Zoraxy complete access to the Docker daemon on the host, allowing it to manage other containers and easily break out of the containerized environment. With access to /var/run/docker.sock, the container gains root-level control over the host. Any process in the container could start privileged containers, access sensitive files, or even compromise the entire host system
> 
> There's a couple apps in our app store that get around this by running alongside a docker-in-docker (dind) container and binding to the dind's docker socket as a workaround (e.g., portainer and dockge).

Hey @nmfretz I’ve made the necessary adjustments. Hope I didn’t miss any points. Thank you for your detailed explanations. These will help me for future projects. 👍

nmfretz:
Excellent, thanks @dennysubke! Can you just take a look at these two items from above before we go live?

1) https://github.com/getumbrel/umbrel-apps/pull/1751#pullrequestreview-2450255489: You can add `DOCKER: "false"` as an env var which I think will make it so the components in the UI that are related to showing running docker containers won't show up so users don't think something is broken now that we have removed the docker socket.

For example, without this env var set to false users will still see this in the UI and think the app is broken:
image

2) https://github.com/getumbrel/umbrel-apps/pull/1751#discussion_r1851357597: Now that you have the app sitting behind the app proxy's auth, do you want to remove Zoraxy's built-in auth with `NOAUTH: "true"`? Or do you think its still worth having it?

dennysubke:
> Excellent, thanks @dennysubke! Can you just take a look at these two items from above before we go live?
> 
> 1. [App Submission: Zoraxy - Reverse Proxy Manager #1751 (review)](https://github.com/getumbrel/umbrel-apps/pull/1751#pullrequestreview-2450255489): You can add `DOCKER: "false"` as an env var which I think will make it so the components in the UI that are related to showing running docker containers won't show up so users don't think something is broken now that we have removed the docker socket.
> 
> For example, without this env var set to false users will still see this in the UI and think the app is broken: image
> 
> 2. [App Submission: Zoraxy - Reverse Proxy Manager #1751 (comment)](https://github.com/getumbrel/umbrel-apps/pull/1751#discussion_r1851357597): Now that you have the app sitting behind the app proxy's auth, do you want to remove Zoraxy's built-in auth with `NOAUTH: "true"`? Or do you think its still worth having it?



Thanks for your tips @nmfretz. I made the changes accordingly. 😎 

You're probably right. Since the app_proxy already handles authentication and ensures that only authenticated users have access to the app, I disabled Zoraxy authentication with `NOAUTH: "true"`.

Now everything should be fine. Off with that! 🚀

nmfretz:
Thanks @dennysubke. Tested and working well.

Going live!

image

DERIKKU-GER:
Version 3.3.4 has been released. Please update. Thank you

PR #1767, opened 2024-11-07
Update plex to v1.41.1

nmfretz:


nmfretz:
Reviewed and tested (fresh install and app update)

PR #1770, opened 2024-11-07
Update homebox to v0.15.2

al-lac:
Update HomeBox to an actively maintained fork and version v0.15.2


nmfretz:
Wonderful, thanks for this @al-lac. I've made some small additional tweaks to the app manifest and tested both a fresh install and app update.

PR #1772, opened 2024-11-08
App submission mqttx web

dirstel:
# App Submission

MQTTX Web

### 256x256 SVG icon
https://github.com/emqx/MQTTX/blob/main/web/public/icons/app.png

### Gallery images
https://github.com/emqx/MQTTX/blob/main/assets/mqttx-web-preview.png
https://github.com/emqx/MQTTX/blob/main/assets/mqttx-connection.png
https://github.com/emqx/MQTTX/blob/main/assets/mqttx-muti-topics.png

### I have tested my app on:
- [x] umbrelOS on a Lenovo M720Q

### DIFF
--- a/mqttx-web/docker-compose.yml
+++ b/mqttx-web/docker-compose.yml
@@ -1,0 +1,9 @@
+services:
+  app_proxy:
+    environment:
+      APP_HOST: mqttx-web_app_1
+      APP_PORT: 80
+
+  app:
+    image: emqx/mqttx-web:v1.11.0@sha256:756a5510257309b8921a453b313e0ed4bb5cd812a341b0a10b84518caba4edbe
+    restart: on-failure
--- a/mqttx-web/docker-compose.yml
+++ b/mqttx-web/docker-compose.yml
@@ -7,6 +7,4 @@ services:
   app:
     image: emqx/mqttx-web:v1.11.0@sha256:756a5510257309b8921a453b313e0ed4bb5cd812a341b0a10b84518caba4edbe
     restart: on-failure
-    environment:
-      - PUID=1000
-      - PGID=1000
+    user: "1000:1000"
--- a/mqttx-web/docker-compose.yml
+++ b/mqttx-web/docker-compose.yml
@@ -6,5 +6,5 @@ services:
 
   app:
     image: emqx/mqttx-web:v1.11.0@sha256:756a5510257309b8921a453b313e0ed4bb5cd812a341b0a10b84518caba4edbe
-    restart: on-failure
     user: "1000:1000"
+    restart: on-failure
--- a/mqttx-web/docker-compose.yml
+++ b/mqttx-web/docker-compose.yml
@@ -7,3 +7,6 @@ services:
   app:
     image: emqx/mqttx-web:v1.11.0@sha256:756a5510257309b8921a453b313e0ed4bb5cd812a341b0a10b84518caba4edbe
     restart: on-failure
+    environment:
+      - PUID=1000
+      - PGID=1000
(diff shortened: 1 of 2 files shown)


mqttx-web/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+    environment:
+      APP_HOST: mqttx-web_app_1
+      APP_PORT: 80
+
+  app:
+    image: emqx/mqttx-web:v1.11.0@sha256:756a5510257309b8921a453b313e0ed4bb5cd812a341b0a10b84518caba4edbe
+    restart: on-failure
+    environment:
+      - PUID=1000
+      - PGID=1000
```
nmfretz, line 12: These environment variables don't exist for the emqx/mqttx-web container so won't actually result in the container running as user 1000.

What you can do instead is completely remove the `environment` section and instead add:

```yaml
user: "1000:1000"
```

...to run the container as non-root

mqttx-web/umbrel-app.yml
```diff
@@ -0,0 +1,22 @@
+manifestVersion: 1
+id: mqttx-web
+name: MQTTX Web
+tagline: Your All-in-One MQTT Client Toolbox
+icon: ""
+category: automation
+version: "1.11.0"
+port: 1884
```
nmfretz, line 8: I'm thinking we should probably reserve this port in case an mqtt broker app comes to the app store. 1884 is a common alternate port for this.

And in this case, the port for the UI here can be any free port. For example `9002` is currently free.

mqttx-web/umbrel-app.yml
```diff
@@ -0,0 +1,22 @@
...
+description: >-
+  MQTTX Web is an open-source MQTT 5.0 WebSocket client on the browser to help you debug your MQTT services and applications faster.
+
+  The user interface (UI) of MQTTX adopts a chat-based layout, simplifying operational logic. It enables users to establish multiple M>
```
nmfretz, line 12: It looks like you may have accidentally cut-off the app description. Right now it ends with `...multiple M>`.

Also, to have the whitespace render nicely in the app store, there needs to be 2 blank lines between paragraphs. With just one blank line, it will render like this:

image

mqttx-web/umbrel-app.yml
```diff
@@ -0,0 +1,22 @@
...
+developer: EMQ
+website: https://mqttx.app/
+submitter: dirstel
+submission: https://github.com/getumbrel/umbrel-apps/pull/1772
+repo: https://github.com/emqx/MQTTX
+support: https://github.com/emqx/MQTTX
```
nmfretz, line 18: It looks like their website links here for support:
`https://github.com/emqx/MQTTX/discussions`

nmfretz (commented):
Thanks for submitting MQTTX Web @dirstel! Great work.

We'll get gallery assets prepared. In the meantime, I have left a review below with some small changes to make.

mqttx-web/docker-compose.yml
dirstel: fixed and tested.

mqttx-web/umbrel-app.yml
dirstel: fixed to 9010, as 9001 is used by mqtt via websocket, and 9002 may be likley an alternate port for that.

mqttx-web/umbrel-app.yml
dirstel: fixed: added linefeed and missing text.

mqttx-web/umbrel-app.yml
dirstel: fixed: changed support url

mqttx-web/umbrel-app.yml
nmfretz: Looks like 9010 is sneakily used by MinIO for the API. I have changed to free port 9012.

nmfretz:
Thanks for the changes @dirstel! Really great work on this submission. I have made some final tweaks and we are ready to go live 🎉

image

PR #1789, opened 2024-11-11
App submission mosquitto

dirstel:
# App Submission

### App name
Eclipse Mosquitto

### 256x256 SVG icon
logo

### Gallery images
image
(not really much to screenshot)

### I have tested my app on:
- [x] umbrelOS on Linux VM


nmfretz:
Thanks for another submission @dirstel! We'll try to get to this early next week. Thanks for your patience.

mosquitto/docker-compose.yaml
```diff
@@ -0,0 +1,18 @@
+services:
```
nmfretz, line 1: This docker-compose file and the umbrel-app manifest file need to have the `.yml` suffix instead of `.yaml` ,otherwise they won't render in the app store and you won't be able to install them via the UI or CLI. Were you still able to install it somehow?

mosquitto/docker-compose.yaml
```diff
@@ -0,0 +1,18 @@
...
+  app_proxy:
+    environment:
+      APP_HOST: mosquitto_broker_1
+      APP_PORT: 9001
+    container_name: mosquitto_app_proxy_1
```
nmfretz, line 6: The `container_name` for the app_proxy and broker services can be removed.

mosquitto/docker-compose.yaml
```diff
@@ -0,0 +1,18 @@
...
+  broker:
+    image: eclipse-mosquitto:2.0.20@sha256:8b396cec28cd5e8e1a3aba1d9abdbddd42c454c80f703e77c1bec56e152fa54e
+    restart: on-failure
+    ports:
+      - '1883:1883'
+      - '9001:9001'
```
nmfretz, line 12: Port 1883 is free, but 9001 is used by Penpot. You could change to `9991:9001` and update the instructions accordingly.

mosquitto/docker-compose.yaml
```diff
@@ -0,0 +1,18 @@
...
+    volumes:
+      - ${APP_DATA_DIR}/config:/mosquitto/config:rw
+      - ${APP_DATA_DIR}/data:/mosquitto/data:rw
+      - ${APP_DATA_DIR}/log:/mosquitto/log:rw
+      - ${APP_DATA_DIR}/www:/mosquitto/www:r
```
nmfretz, line 17: To keep everything under one data directory like other apps I'd suggest going with something like this:

```
data
├── mosquitto-data
│   └── .gitkeep
├── config
│   └── mosquitto.conf.default
└── log
    └── .gitkeep
```

And then if serving the UI from a Docker container as described above, just get rid of `- ${APP_DATA_DIR}/www:/mosquitto/www:r`

mosquitto/hooks/pre-start
```diff
@@ -0,0 +1,16 @@
...
+APP_DATA_DIR="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/..")"
+CONFIG_FILE="${APP_DATA_DIR}/config/mosquitto.conf"
+
+if [ -f "${CONFIG_FILE}" ]; then
+  echo "configuration file present, nothing will be changed."
+else
+  echo "configuration file NOT present, defaults will be copied."
+  cp "${CONFIG_FILE}.default" "${CONFIG_FILE}"
+  # set correct owner
+  chown umbrel:umbrel "${CONFIG_FILE}"
```
nmfretz, line 15: Nice hook 🫡

Looking at the [mosquitto Dockerfile](https://github.com/eclipse-mosquitto/mosquitto/blob/master/docker/2.0/Dockerfile), we're actually stuck running the container as user:group `1883:1883` since they have hardcoded this into the image unfortunately. I think this should still work as-is though.

You can see that all data directories for mosquitto end up owned by 1883

```
:~$ ls -l ~/umbrel/app-data/mosquitto/
total 20
drwxr-xr-x 1   1883   1883   72 Nov 28 05:14 config
drwxr-xr-x 1   1883   1883    0 Nov 28 02:04 data
-rw-r--r-- 1 umbrel umbrel  574 Nov 28 04:59 docker-compose.yaml
-rw-r--r-- 1 umbrel umbrel  583 Nov 28 05:14 docker-compose.yml
drwxr-xr-x 1 umbrel umbrel   18 Nov 28 05:14 hooks
drwxr-xr-x 1   1883   1883    0 Nov 28 02:04 log
-rw-r--r-- 1 root   root     17 Nov 28 05:14 settings.yml
-rw-r--r-- 1 umbrel umbrel 1008 Nov 28 04:59 umbrel-app.yaml
-rw-r--r-- 1 umbrel umbrel 1008 Nov 28 04:59 umbrel-app.yml
drwxr-xr-x 1   1883   1883   58 Nov 28 02:04 www
```

mosquitto/umbrel-app.yaml
```diff
@@ -0,0 +1,28 @@
+manifestVersion: 1.1
+id: mosquitto
+name: Mosquitto
+tagline: An open source MQTT broker
+icon: https://raw.githubusercontent.com/eclipse-mosquitto/mosquitto/refs/heads/master/logo/mosquitto-logo-only.svg
```
nmfretz, line 5: This entire icon key-value pair can be deleted. We'll host the icon at https://github.com/getumbrel/umbrel-apps-gallery

mosquitto/www/index.html
```diff
@@ -0,0 +1,306 @@
+<!DOCTYPE html>
+<!-- saved from url=(0024)http://server.home:9445/ -->
+<html><head><meta http-equiv="Content-Type" content="text/html; charset=windows-1252">
+    <title>Samba</title>
```
nmfretz, line 4: Can change to Mosquitto

nmfretz (commented):
Nice work repurposing the Samba/Snowflake index.html to give the app a web UI with instructions @dirstel 👌

Can mosquitto even serve the html page via it's websocket port though? I think that port is just so that mqtt clients can connect using websockets. I tried quickly installing this on my umbrel and nothing shows up.

Regardless, I've got some suggestions below on how to serve the UI and make your life much easier when updating this app in the future.

### Package the UI as a separate Docker image

The way umbrelOS currently works is that on app install, everything from the app's repo is copied over to a users app-data directory. But on app updates only a whitelist of files is copied over so that user data is not overwritten:
https://github.com/getumbrel/umbrel/blob/49d37581c8233f808fc5a79c2cd38810c6cd1b0b/packages/umbreld/source/modules/apps/legacy-compat/app-script#L594-L595

What that means for you here is that if you need to make changes to your UI (e.g., changing the instructions in the html), only new installs would receive the changes. Existing installs that update won't get the changes. You would need to resort to some really hacky things like using a hook that runs on app update and makes modifications to the html/image files. It would end up adding a lot of complexity and overhead for you.

Instead though, you can just package the UI as a super simple Docker image that serves your html and then point the `app_proxy` service to it. That way if you ever need to update the UI, you just build and push a new Docker image to your repo and then change it in the docker-compose.yml when submitting an app update.

I have not tested this at all, but something as simple as this could work for a Dockerfile:
```Dockerfile
# Use the joseluisq/static-web-server as the base image
FROM joseluisq/static-web-server:2.33.1

# Copy the directory with your ui web assets to the /var/public directory which is set by the SERVER_ROOT env var.
# e.g., if your directory is called `public`:
COPY public /var/public
```

You'll then need to make sure to build a multi-architecture image for amd64 and arm64. You'll run something like this:
```
docker buildx build --platform linux/amd64,linux/arm64 -t your-docker-hub-username/static-web-server-mosquitto:version-tag .
```

nmfretz:
Thanks for making those changes @dirstel.

> First, Mosquitto can serve a webpage on its own.
Second the availability of the webpage is a decent indicator of Moswuitto really running.

Thanks for the clarification. With your changes the webpage is now being served correctly for me by mosquitto on app install.

> The MQTT-Service is usuable without permissions. When exposing the service to the internet this should definetly be changed. as there are lots of possibilities (one user, user per client/sensor, acl, aditonal plugins, ...) I think this should not be set in default installation. Maybe a hint or two in the webpage can be added.

Adding a hint to the UI is a great idea. While you're doing that you may want to also change the following:
1) accessinformation --> access information
2) Settings > Advanced Settings > Terminal --> Settings > Advanced settings > Terminal > umbrelOS
3) `nano umbrel/app-data/...` --> `sudo nano ~/umbrel/app-data/...`

> I agree with you in terms of updates. Maybe there should be a detailed description of what is done by an update?

I agree with you. We are redoing the apps framework and will make sure to have detailed descriptions for this new framework.

Under the current framework, an app update copies over the following dirs/files:
https://github.com/getumbrel/umbrel/blob/49d37581c8233f808fc5a79c2cd38810c6cd1b0b/packages/umbreld/source/modules/apps/legacy-compat/app-script#L594-L595
`docker-compose.yml` `*.template` `exports.sh` `torrc` `hooks/` `umbrel-app.yml`

Any other files/dirs are not copied over on update. As an example, the `data` directory in this mosquitto submission won't be copied over meaning that a users data/settings won't get clobbered when they update.

> As far as I can see, the hook used is not needed too.

That's correct. If all you are doing in the hook is creating the config, then you can just delete the entire `hooks/` directory and rename the default config file to `mosquitto.conf.default`.

> In general, headless services should be given the opportunity to display a webpage by umbrel or the app-proxy. In my opinion it is somewhat cumbersome to build a dedicated container for each of this app.

We're redoing the apps framework, so will take your suggestion under advisement 🤝.

The idea with the current framework is that an app should not be tied to umbrelOS. The manifest (`umbrel-app.yml`) is umbrelOS-specific, but the actual app itself... The `app_proxy` is providing transparent proxying and auth (password, 2fa). If the `app_proxy` were to provide a way for headless services to include a webUI then we would be dictating how an app needs to be developed (umbrelOS-specific) instead of just having services be containerized and thus doing anything they want to create a UI.

### App functionality

Can you please do some additional testing to make sure this is working correctly for you? Unfortunately my time is limited for the next while, so I can't spend too much time troubleshooting.

I just quickly tested mosquitto as of your commit https://github.com/getumbrel/umbrel-apps/pull/1789/commits/ac6824c79ac35c6b7322e2f19f44b1f521d92283. The webUI works for me, but I cannot connect via MQTTX Web using port 9991. This is because your compose file is exposing 9991 on the host and binding to internal port 9001 in the container. So when you try to connect to something like `http://umbrel.local:9991` you'll hit the internal container port 9001 which isn't associated with a listener in your mosquitto.conf.

What you can do instead is simplify the exposed ports since what the `app_proxy` container does is proxy requests directly to a port INSIDE the `broker` container. There is no need to expose 9991. The app proxy will upgrade the request to ws automatically. See my screen-recording below for the changes to make:

https://github.com/user-attachments/assets/99a432b4-6b53-4212-8a87-1f4ff9fb91f1

We may need to whitelist certain routes in the app proxy like `/mqtt` if external (to umbrelOS) applications need to access.

Before considering implementing this though, can you please first test using your commit https://github.com/getumbrel/umbrel-apps/pull/1789/commits/ac6824c79ac35c6b7322e2f19f44b1f521d92283 to make sure that I have not made a mistake while briefly playing around.

dirstel:
Sorry for the confusion about the port(mapping)s - I'm glad, the app_proxy is able to serve the web page AND the mqtt-via-webservice as well at the same time. This takes away some confusing entries.

* removed hook as it is not needed + downgraded manifestVersion to 1 as no hooks needed
* removed port mapping in docker container: webpage and mqtt via webservice is handled and exposed by app_proxy
* added information about configuring authentication + corrected typos in web page

* successfully tested access to web page
* successfully tested access via Port 1883 and mqtt-protocol
* successfully tested access via Port 9021 and mqtt-via-webservice

nmfretz:
Excellent, thanks for making those changes and thoroughly testing @dirstel 🤝. I have tested as well and everything is working nicely.

The gallery assets are complete and we're going live. Thanks for all your efforts bringing this second app to the store!

image

PR #1809, opened 2024-11-15
Update mattermost to v10.2.0

al-lac:
Tested on Raspberry Pi 5.


nmfretz:
Wonderful, thanks for this update @al-lac, it is working great! Sorry for the delay in reviewing while we had our heads down on the recent beta launch. 

FYI, I added an extra newline in the release notes so they render like this:
image

PR #1812, opened 2024-11-17
App Submission: Super Productivity

dennysubke:
# App Submission

### Super Productivity
...

### Icon
bNDXJHt - Imgur

...

### Gallery images
a

b

c

d

...

### I have tested my app on:
- [x] umbrelOS on Linux VM


dennysubke:
I unfortunately noticed too late that the Docker image `johannesjo/super-productivity:v10.1.1` does not support `arm64` architectures. It might make sense to wait for a future version that offers multi-architecture support.

dennysubke:
FYI: 

I have opened an issue on the developer site: 

https://github.com/johannesjo/super-productivity/issues/3694

dennysubke:
The PR for the multi-arch image has been created and is ready—just waiting for it to be merged now.

https://github.com/johannesjo/super-productivity/pull/3699

nmfretz:
> The PR for the multi-arch image has been created and is ready - just waiting for it to be merged now.

Nice! Awesome work reaching out on their GitHub @dennysubke and coordinating 👑.

Ping me when it's ready to go.

dennysubke:
> > The PR for the multi-arch image has been created and is ready - just waiting for it to be merged now.
> 
> Nice! Awesome work reaching out on their GitHub @dennysubke and coordinating 👑.
> 
> Ping me when it's ready to go.

Ping! 🎉

nmfretz (commented):
Thanks for pinging me @dennysubke! I have left a review below. When you dig in to the webdav backend can you please do some tests to make sure data is persisted on the backend and not just in the browser's local storage or indexdb or wherever Super Productivity stores data by default? I think you'll be able to easily check by just opening the app in an incognito window to see if your data has persisted.

dennysubke:
Hey @nmfretz!

Thank you for your detailed review and feedback! 👍 

I will test the WebDAV backend to ensure data persistence as you suggested. Opening the app in an incognito window and checking if data is stored on the backend (rather than just locally in the browser) is a great idea. I'll report back with my findings. 💪 

The exposed port was included for debugging purposes and to allow potential direct access to the WebDAV backend. However, if this isn't necessary for the intended use case (like syncing across devices), I can remove the exposed port.

By the way, this is from their documentation:


grafik

https://github.com/johannesjo/super-productivity/tree/master?tab=readme-ov-file#run-as-docker-container

Can you take a look at it in the meantime?

al-lac:
Hey @dennysubke! I tested it with webdav and could not properly get it to work because of CORS and other errors. Will be hard to enable this without HTTPS.

I would actually skip the included webdav and let the users configure it by themself. They can sync to one of available backends like nextcloud or dropbox.

FYI: @nmfretz

dennysubke:
> Hey @dennysubke! I tested it with webdav and could not properly get it to work because of CORS and other errors. Will be hard to enable this without HTTPS.
> 
> I would actually skip the included webdav and let the users configure it by themself. They can sync to one of available backends like nextcloud or dropbox.
> 
> FYI: @nmfretz

@al-lac Thanks for your feedback! 

You're right, without HTTPS, WebDAV is tricky. We can close the pull request if it’s too complicated for users. Is not a problem for me.

If anyone really wants the app, they can pull it themselves via Portainer or install it from my community store:

https://github.com/dennysubke/dennys-umbrel-app-store

al-lac:
I think it would still make sense to add it, just need to inform users that data is stored in the local browser storage until they configure a backend like nextcloud or dropbox. 

But lets hear what @nmfretz has to say about this.

nmfretz:
Thanks for taking a look @al-lac! I agree, let's get this to the app-store without WebDAV and tweak the description so that users are aware of the configuration.

I'll get this done today.

nmfretz:
Alrighty, we are going live. Thanks for this submission @dennysubke!

I have removed the WebDAV service and bind mounts and added the following info to the release notes:

image

Also, @dennysubke, my review comments on the bind mounts would apply to your community app store as well if they are configured the same as your original submission here. Might be worth checking.

image

dennysubke:
> Alrighty, we are going live. Thanks for this submission @dennysubke!
> 
> I have removed the WebDAV service and bind mounts and added the following info to the release notes:
> 
> image
> 
> Also, @dennysubke, my review comments on the bind mounts would apply to your community app store as well if they are configured the same as your original submission here. Might be worth checking.
> 
> image
> 

@nmfretz Thanks for merging! I appreciate the improvements you made. 

I'll review the bind mount configurations in my community app store and adjust them accordingly. 👍🏻

PR #1814, opened 2024-11-19
Update Firefox to 132.0.2

al-lac:
Update Firefox to the latest version.

Not tested yet.


nmfretz:
Excellent, thanks for this update as well @al-lac. Tested and working well 👌

PR #1835, opened 2024-11-20
Update btcpay-server to v2.0.3

nmfretz:


nmfretz:
Tested fresh install and app update

PR #1839, opened 2024-11-22
Update Grafana to version 11.3.1

al-lac:
Tested on Raspberry Pi 5

### DIFF
--- a/grafana/docker-compose.yml
+++ b/grafana/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api/*"
       
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Wonderful, thanks for this update @al-lac! I have tested fresh install and app update and it is working well.

I made the following changes to the release notes:
- Added a warning at the top about breaking changes since grafana kinda heavily promote this in there update articles/release notes/etc.
- Used their intro paragraph from their [what's new article](https://grafana.com/docs/grafana/latest/whatsnew/whats-new-in-v11-3/). Seems like they do a pretty good summary for each release so it will make it easy for us to just copy-paste this each time.
- Changed the release notes link to not use markdown syntax. This app store still runs on umbrelOS 0.5 which won't render it nicely, so I've changed it to make it compatible. In a later version of the app store we should be able to start including them.

PR #1863, opened 2024-11-27
Update Affine to v0.18.0

al-lac:
Tried to update to v0.18.1, but there does not seem to be a proper container for it.

Tested on Raspberry Pi 5.

### DIFF
--- a/affine/docker-compose.yml
+++ b/affine/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       APP_PORT: 3010
 
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @al-lac! 

> Tried to update to v0.18.1, but there does not seem to be a proper container for it.

It was sneakily hiding on you
https://github.com/toeverything/AFFiNE/pkgs/container/affine-graphql/310200263?tag=stable-8b066a4

I was too slow, and 0.18.2 got released as I was reviewing this so I've bumped and tested.

PR #1861, opened 2024-11-27
Update stirling-pdf to version v0.34.0

nmfretz:
🤖 This is an automated pull request to update stirling-pdf to version v0.34.0. This PR must be reviewed and tested before merging.

PR #1864, opened 2024-11-27
Add pre-start migration script for Stirling PDF updates

sharknoon:
This PR is a follow-up for this PR: https://github.com/getumbrel/umbrel-apps/pull/1861

I have added a pre-start hook to update to `0.28.3` first, if the installed version is < `0.28.0`.

A couple of notes from my side:
- I have tried reproducing the upgrade error (upgrading straigt from `0.25.0` to `0.34.0` without hook), but to my suprise that worked wonderfully
- Stirling PDF seems to not store their version in any persisted files, so I get the version from the image label `org.opencontainers.image.version`
- I don't know exactly how to test this hook. I have copied it into the appstore directory and ran `umbreld client apps.update.mutate --appId stirling-pdf`, which worked, but I am not sure, if the script execution was successful. I am also not sure where to find the `echo` logs

@nmfretz If you have time, can you help me test this hook? That would be a great help :) Thank you!

### DIFF
--- a/stirling-pdf/hooks/pre-start
+++ b/stirling-pdf/hooks/pre-start
@@ -1,0 +1,131 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# This script updates Stirling PDF by first migrating older versions (< 0.28.0) to 0.28.3 first.
+# https://github.com/Stirling-Tools/Stirling-PDF/releases
+
+APP_DATA_DIR="$(readlink -f $(dirname "${BASH_SOURCE[0]}")/..)"
+
+STIRLING_PDF_SETTINGS_FILE="${APP_DATA_DIR}/data/configs/settings.yml"
+
+APP_COMPOSE_FILE="${APP_DATA_DIR}/docker-compose.yml"
+APP_COMPOSE_BACKUP_FILE="${APP_DATA_DIR}/docker-compose.yml.bak"
+
+# List of versions on migration path
+# stirling pdf on the umbrel app store was initially released with version 0.25.0
+VERSIONS=()
+VERSIONS+=("0.25.0")
+VERSIONS+=("0.26.1")
+VERSIONS+=("0.27.0")
+
+# Image to migrate to from older versions before migrating to the latest version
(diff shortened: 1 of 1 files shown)


sharknoon:
Regarding your Question: No, they don't store any important data, just some preferences
grafik
I think maybe it is easier, as you suggested, to just add this note, because 1. probably most of the users have installed the updates for Stirling PDF and 2. even if they didn't, they would not loose any important data

nmfretz:
> I think maybe it is easier, as you suggested, to just add this note, because 1. probably most of the users have installed the updates for Stirling PDF and 2. even if they didn't, they would not loose any important data

Ya, I'm really leaning this way now especially since it's just some minor preferences that would be lost. I think the tradeoff is worth it to keep this super clean... I'm pretty sure no one is going to be impacted at all. But if they are the consequences are pretty inconsequential (just re-install with no data loss). Let's go with a warning in the release notes as discussed here https://github.com/getumbrel/umbrel-apps/pull/1861

> I don't know exactly how to test this hook. I have copied it into the appstore directory and ran umbreld client apps.update.mutate --appId stirling-pdf, which worked, but I am not sure, if the script execution was successful. I am also not sure where to find the echo logs

Nice job on the hook with grabbing the version and waiting for stirling to be up and running 👌.

You can stream the logs while updating with:
```
sudo journalctl -u umbrel -f
```

Or if you want to just have one terminal open you can do:
```
umbreld client apps.update.mutate --appId stirling-pdf & sudo journalctl -u umbrel -f
```

Those will contain the echo logs from any app lifecycle hooks that run. But one gotcha is that the hook file has to be executable. Right now the one in this repo isn't executable so it won't actually run:
```
$ ls -l stirling-pdf/hooks/pre-start | awk '{print $1}'
-rw-r--r--@
```

You can fix with `chmod +x stirling-pdf/hooks/pre-start`

Which means that your tests are showing that we probably don't even have to run the script haha 🎉

sharknoon:
Yes, I agree that not having the hook at all is the best solution. Thank you for explaining how to stream the logs :)

PR #1883, opened 2024-12-02
Update GitLab to version 17.6.1

al-lac:
Updating GitLab to the latest version. Could not test it yet, not enough free memory on my Pi 5.


nmfretz:
Thanks for the gitlab update @al-lac! I've update the pre-start hook so that users on older versions make the required stops when updating. I also added a warning to the release notes about GitLab taking a while to start back up after the update, just so users don't panic when first trying to access the UI after the update.

Tested and working well.

PR #1886, opened 2024-12-03
Update vikunja to version 0.24.5

al-lac:
Tested on Raspberry Pi 5


nmfretz:
Thanks for all these perfect updates @al-lac! I really appreciate the help.

PR #1900, opened 2024-12-06
Update grafana to version 11.4.0

al-lac:
Tested on Raspberry Pi 5.


nmfretz:
Thanks very much @al-lac, this is perfect. Reviewed and tested on arm64 and x86.

PR #1903, opened 2024-12-06
App Update: Snort v0.3.0

v0l:


nmfretz:
Excellent, thanks for this update @v0l. Tested fresh install and update.

PR #1905, opened 2024-12-08
new debian based image

Retropex:
Forgot to push the Debian based image with `curl` for the datum release. 🤦‍♂️

### DIFF
--- a/bitcoin-knots/docker-compose.yml
+++ b/bitcoin-knots/docker-compose.yml
@@ -44,7 +44,7 @@ services:
         ipv4_address: $APP_BITCOIN_KNOTS_IP
 
   bitcoind:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for the fix @Retropex. Tested app update and that curl is available in the container.

```
$ sudo docker exec -it bitcoin-knots_bitcoind_1 curl -I https://example.com
HTTP/2 200
accept-ranges: bytes
age: 449672
cache-control: max-age=604800
content-type: text/html; charset=UTF-8
date: Mon, 09 Dec 2024 11:37:25 GMT
etag: "3147526947"
expires: Mon, 16 Dec 2024 11:37:25 GMT
last-modified: Thu, 17 Oct 2019 07:18:26 GMT
server: ECAcc (lac/55BE)
x-cache: HIT
content-length: 1256
```

PR #1917, opened 2024-12-10
Wireguard: Add note about using nginx for external access

### LINKED ISSUE #1642: Potential issues with Wireguard?
Hey, guys,

I have noticed two points when trying to use WireGuard with my Umbrel Home:

1. The configuration file points to `umbrel.local:51820`, meaning any external connection will never occur because that's an inaccessible endpoint if I'm not at home. It would be convenient to explain how we can create an out-home accessible endpoint so our devices can connect when we're in a hotel or anywhere else. Can the Cloudflare Tunnel App help here? But I saw we can only define TCP connections, not UDP.

2. WireGuard usually works with TCP/UDP protocols and port 51820. To prevent unexpected issues, maybe adding a note in the App description informing the user that they will need to open this port in their router/firewall would be convenient.

Cheers,

al-lac:
Just a small note to make the use of wireguard easier for users.

Fixes #1642

### DIFF
--- a/wireguard/umbrel-app.yml
+++ b/wireguard/umbrel-app.yml
@@ -38,6 +38,11 @@ description: >-
   📐 Well Defined & Thoroughly Considered
   
   WireGuard is the product of careful academic consideration, with a clearly defined protocol and well-considered design decisions.
+
+
+  ℹ️ Access from outside the local network
+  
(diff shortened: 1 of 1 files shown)


nmfretz:
Excellent, thanks for this @al-lac 👌

PR #1923, opened 2024-12-13
Update tailscale to v1.78.3

al-lac:
Tested on Raspberry Pi 5.


nmfretz:
Thanks for this update as well @al-lac!
Reviewed and tested.

PR #1925, opened 2024-12-14
Update n8n to 1.72.1

MajorTom327:


nmfretz:
Thanks for the effort of updating n8n @MajorTom327, I really appreciate the help 🤝.

It looks like 1.72.1 is a `pre-release` version for n8n, and we have a policy of using `latest`/`stable` releases when they available.

https://github.com/n8n-io/n8n/releases/tag/n8n%401.72.1
image


I've update n8n to the latest stable release v1.71.3 here: https://github.com/getumbrel/umbrel-apps/pull/1945, so you can see the changes we make to the umbrel-app.yml for future updates.

MajorTom327:
Thanks, I didn’t know that.
It makes sense!

PR #1926, opened 2024-12-14
Update Elements Core to v23.2.4 & PeerSwap to v4rc2

Impa10r:
Enables discounted Confidential Transactions on mainnet


nmfretz:
Excellent, thanks @Impa10r. I'm just syncing elements so I can test properly

nmfretz:
Thanks again for these updates @Impa10r. I have tested and they are working well.

For our records: the `user` changes in https://github.com/getumbrel/umbrel-apps/pull/1926/commits/9aa416470855ff1abb047d796f03c1d8a7b9f349 to satisfy the linter for the elements app are okay to do in this case. The `ghcr.io/vulpemventures/elements` image was already running as user 1000 so there are no permissions issues to handle for users who update:
https://github.com/vulpemventures/docker-elements/blob/170f26627eb51ed79c3790a58a3a053245015bc2/23.2.4/Dockerfile#L19-L31

And for the `ghcr.io/tiero/elements-web` that was previously running as root, there appears to be no data dir that it requires root permissions for.

PR #1927, opened 2024-12-15
[METUBE] delete files from disk

MontejoJorge:
Add variable to delete files from disk when deleting from web ui

https://github.com/alexta69/metube?tab=readme-ov-file#configuration-via-environment-variables

### DIFF
--- a/metube/docker-compose.yml
+++ b/metube/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8081
 
   metube:
-    image: ghcr.io/alexta69/metube:2024-11-05@sha256:46a478214738d251cfe8dd40c1ca628149e014a6e041a261af04fd483015fd7d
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this @MontejoJorge. I've been on-the-fence about whether or not to include this env var in the past.

On one hand, the `Completed` section is kind of just a "downloads history" for videos. The videos are stored in a user's main storage directory on their Umbrel and not in an app directory: `/home/umbrel/umbrel/data/storage/downloads/metube`. So originally I thought it might be a bit weird to delete the videos permanently with MeTube since you can use the File Browser app to handle all files in Umbrel's main storage... kind of like how most torrent clients work.

But on the other hand, I've seen a few requests now from users who want this functionality within MeTube so let's change it!

I've update MeTube and added a note about the change in the release notes. I've reviewed and tested.

PR #1947, opened 2024-12-18
App Submission: WordPress

al-lac:
# App Submission

WordPress

### 256x256 SVG icon
https://en.m.wikipedia.org/wiki/File:WordPress_blue_logo.svg

### Gallery images
wordpress-1
wordpress-2
wordpress-3

### I have tested my app on:
- [ ] umbrelOS on Linux VM

### Notes
Port mapping needed as the app_proxy seems to remove headers set by nginx or other proxies.

### DIFF
--- a/wordpress/docker-compose.yml
+++ b/wordpress/docker-compose.yml
@@ -1,0 +1,29 @@
+version: "3.7"
+services:      
+  app:
+    image: wordpress:6.7.1@sha256:2f3572d5cd722489fe47d59ed45d947dc9507f5a019eb0567ddf24aedf9257ed
+    user: "1000:1000"
+    environment:
+      - WORDPRESS_DB_HOST=wordpress_db_1
+      - WORDPRESS_DB_NAME=wordpress
+      - WORDPRESS_DB_USER=umbrel
+      - WORDPRESS_DB_PASSWORD=umbrelwordpress
+    volumes:
+      - ${APP_DATA_DIR}/data/wordpress:/var/www/html
+    ports:
+      - "8567:80"
+    restart: on-failure
+    depends_on:
+      - db
+
+  db:
+    image: mariadb:11.5.2@sha256:2d50fe0f77dac919396091e527e5e148a9de690e58f32875f113bef6506a17f5
+    user: "1000:1000"
+    environment:
+      - MYSQL_ROOT_PASSWORD=umbrelwordpressroot
+      - MYSQL_DATABASE=wordpress
+      - MYSQL_USER=umbrel
+      - MYSQL_PASSWORD=umbrelwordpress
+    volumes:
+      - ${APP_DATA_DIR}/data/db:/var/lib/mysql
+    restart: on-failure
--- a/wordpress/docker-compose.yml
+++ b/wordpress/docker-compose.yml
@@ -1,5 +1,11 @@
 version: "3.7"
-services:      
+services:
+  app_proxy:
+    environment:
+      APP_HOST: wordpress_app_1
+      APP_PORT: 80
+      PROXY_AUTH_ADD: "false"
+
   app:
     image: wordpress:6.7.1@sha256:2f3572d5cd722489fe47d59ed45d947dc9507f5a019eb0567ddf24aedf9257ed
     user: "1000:1000"
@@ -10,8 +16,6 @@ services:
       - WORDPRESS_DB_PASSWORD=umbrelwordpress
     volumes:
       - ${APP_DATA_DIR}/data/wordpress:/var/www/html
-    ports:
-      - "8567:80"
     restart: on-failure
     depends_on:
       - db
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks for yet another app submission @al-lac! We'll start on gallery assets.

This looks really good. Here's one consideration:
 
> Port mapping needed as the app_proxy seems to remove headers set by nginx or other proxies.

Ah yes, here's some extra info on that:
https://github.com/getumbrel/umbrel-apps/pull/1037
https://github.com/getumbrel/umbrel/issues/1428

One option here that could be worth testing is:
- Include the `app_proxy` in the compose file
- This would mean that the wordpress admin dashboard would sit behind the proxy
- Then when exposing the public site (not the admin dashboard) to the internet, you could proxy/tunnel to the app container directly, bypassing the proxy.
- As an example, using a Cloudflare Tunnel you would set the service to `http://wordpress_app_1` (port 80 is default).

It's possible though that this doesn't do everything we need.

nmfretz:
Gallery assets done:

image

nmfretz:
@al-lac, ah yes you're right. Re-reading my comment I'm not sure what I was talking about regarding exposing the public site but not the dashboard... sorry about that.

So essentially our two options are:

1) Don't put it behind the app_proxy.

In this configuration a user could tunnel or point a reverse proxy to the host at port 8567 or they could also choose to point to the `wordpress_app_1` container directly (internal port 80).

2) Put it behind the app_proxy.

In this configuration, you wouldn't be able to point to the host at port 8567 because of the `app_proxy` container, but you could still use the Cloudflare Tunnel app or Nginx Proxy Manager app and point directly to the `wordpress_app_1` container.

Right now, the precedent is that we are always using the app proxy container unless an app needs to run in host networking mode. Since WordPress does not need host networking mode, I don't think that we should break that precedent. 

Let's place this behind the app proxy container for the initial release. We can then gauge the reaction to it and make a decision on whether the proxy should be removed.

I'll make the required changes and include some instructions in the app descriptions and then we can test.

nmfretz:
@al-lac - could you please give this a test with the recent changes and let me know what you think? It's important that we don't reduce the functionality of the app needlessly so please let me know if I have accidentally done so and we can revert.

I have added the app proxy and have also added `/wp-admin` as the `path` in the manifest. This makes it so that when a user clicks on the app from the umbrelOS homescreen they are taken to the sign-in page if they are currently logged out. Without this path, if a user signs out of WordPress and then clicks on the app they'll be taken to the site page that has no option for signing in and they may be confused as to how to sign back in.

al-lac:
Hey @nmfretz!

Everything seems to work fine, also really nice that you direct users towards the admin page!

Was able to get the reverse proxy working by entering `wordpress_app_1` as the nginx forwarded hostname, together with the linked nginx config from the wordpress documentation.

The only thing that the app_proxy limits, is that you cannot use a proxy that is running on a different umbrel or other system. But I am not sure how many users would take advantage of this.

sharknoon:
Hey @al-lac and @nmfretz,

I have some experience with WordPress (have created a custom theme and layout for an intranet page at my previous job). 

Regarding the proxy, WordPress is basically just a PHP Webserver with a Database, therefore I think you have made the right choice by adding a proxy infront of it.

Also WordPress has its own auth, so you have already disabled the auth in the proxy.

Me recommendation is to keep it as it is 😀

nmfretz:
Excellent, thanks for that info and testing @sharknoon!

And thanks again for getting WordPress to the app store @al-lac 🚀  


image

amtexugo:
Hi gang!

Somebody here had the experience with cloudflare instalation proxy with wordpress app in umbrel? 

Here had a headcake about that. The local domain umbrel.local:8567 ok but the subdomain.domain.com dont open: i used http://wordpress_app_1:80 in Cloudflare config. umbrel CF do the tunnel ok but the page dont open.  the browser change the port to 8567 when is opening but dont open nothing

al-lac:
Hey @amtexugo!

I think you are already really close here. I used nginx proxy manager to make my site available on the net. I think you also need to change the values here in wordpress:
CleanShot 2025-09-18 at 21 16 45@2x

Beware though, could break the site.

amtexugo:
Yes!

I have sucess
Maked what you Send
Locallity is not possible
Just by domain.com

Sent from [Proton Mail](https://proton.me/mail/home) for iOS

18, Alexander L. ***@***.***(mailto:18, Alexander L. <<a href=)> escreveu:

amtexugo:
Thanks for replay

Sent from [Proton Mail](https://proton.me/mail/home) for iOS

18, Alexander L. ***@***.***(mailto:18, Alexander L. <<a href=)> escreveu:

amtexugo:
Wp in umbrel have much chalengers

Sent from [Proton Mail](https://proton.me/mail/home) for iOS

18, Alexander L. ***@***.***(mailto:18, Alexander L. <<a href=)> escreveu:

PR #1948, opened 2024-12-18
Update Mattermost to v10.3.1

al-lac:
Tested update and install on Raspberry Pi 5 and Umbrel Home.

### DIFF
--- a/mattermost/docker-compose.yml
+++ b/mattermost/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   app:
-    image: alexlack/mattermost-app:v10.2.1@sha256:8caaad949601b961d44d4f57d387c5c8a392db992082dbd0916ff0a891aa7c0e
+    image: alexlack/mattermost-app:v10.3.1@sha256:bfc5ffe1362b16f635e4963862b0d9b649f78b7616994c640793cb4d494a99ae
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
@@ -30,7 +30,7 @@ services:
       - db
 
   db:
-    image: alexlack/mattermost-db:v10.2.1@sha256:9cc76651b07fb8b4f7bacdf7adec2195ecf9fc4d790ef411140921c972628df0
+    image: alexlack/mattermost-db:v10.3.1@sha256:73580144209dc1402de7c85695664be020625355d799e7c4c9cbd22c6f54360f
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/mattermost/umbrel-app.yml
+++ b/mattermost/umbrel-app.yml
@@ -3,7 +3,7 @@ id: mattermost
 name: Mattermost
 tagline: Team Chat, Open Source, Self-Hosted
 category: social
-version: "10.2.1"
+version: "10.3.1"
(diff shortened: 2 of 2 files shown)


nmfretz:
Excellent. Thanks for this @al-lac. Tested and working great.

image

Shall I add Mattermost to the list of apps that our auto-update script will update? Or would you rather keep doing them manually?

al-lac:
Great, thanks for testing as well! :-) 

Feel free to add it to the auto-update script. Is the script / action also open source? Would be really interesting to see how you automatically deploy and screenshot an app.

nmfretz:
> Feel free to add it to the auto-update script. 

Great, I've added it!

> Is the script / action also open source? Would be really interesting to see how you automatically deploy and screenshot an app.

Ah, it's just code I have locally on my computer at the moment. It's in a bit of a hacky state as I tweak/modify it to get it to be even more automated. I haven't touched it in quite a long time, so will need to improve and clean things up.

But at a high level this is what it does:

- Checks for a new version of the app by fetching latest version from GitHub and comparing against current version in the umbrel-app.yml. I've got some logic to deal with tags and versions... for example, we want to grab the `latest` version if available and not use a `pre-release`. Also fetches release notes.

- Uses AI to analyze the release notes and compose files and identify which Docker images need updating.

- Checks that the identified images we're updating to support both linux/amd64 and linux/arm64 architectures (I'm using `mplatform/mquery` as a super easy way to do this) and fetches the verified digests.

- Uses AI to update the docker-compose.yml and umbrel-app.yml files. This is the trickiest part and relies on good prompting. You'll see by the completed updates in this repo that I now usually don't need to make edits, but occasionally do when it can't grab the right image for some reason or it adds some irrelevant release note point, etc.

- Creates a new local branch, commits changes, creates a remote PR.

- Installs the app in an `umbrel-dev` instance, verifies that the app was installed, tests the UI accessibility, and then uses a headless browser to take a screenshot of the UI. Then it adds these testing results to the PR.

Overall, this allows me to focus on testing and checking the release notes for breaking changes instead of getting bogged down in manually making the update. The auto-tester gives me the "fresh install" test that I can quickly verify because it will be freshly installed in my umbrel-dev instance. And then I can focus on testing that an app update persists data, etc.

I haven't updated this table in a bit, but here's a rough look at how many apps the updater-script is "responsible" for:

```
 * Update Process for Maintained Apps (141 total):
 * +------------------------------------------------+-------+-------+
 * | Update Method                                  | Count |     % |
 * +------------------------------------------------+-------+-------+
 * | Maintained by app developer                    |    35 |  25%  |
 * | Updateable via script without manual input     |    76 |  54%  |
 * | Updateable via script with manual input        |    22 |  16%  |
 * | Maintained by Umbrel (development work needed) |     8 |   6%  |
 * +------------------------------------------------+-------+-------+
 *
 ```
 
 For the update script to work without intervention the app's github repo must have (1) tags and (2) release notes. If they don't have both of these then they fall under the `Updateable via script with manual input` category where I pass in version and raw release notes via the command line when running the update script.

al-lac:
Wow, really nice workflow! 🚀 

I thought that those release notes maybe are refined by AI 😁 

I guess with some effort most of it could be done via GitHub Actions, so it is not dependent on a local machine. But not sure how the umbrel-dev tests and so on would work there.

PR #1949, opened 2024-12-18
Remove Nitter

nmfretz:
The Nitter project is unmaintained and is now non-functional due to changes in X's API: 
https://github.com/zedeus/nitter/issues/919

This PR:
- adds `disabled: true` to the app manifest which results in Nitter not being shown in the app store on umbrelOS >= 1.0
- adds a removal notice to the app description so that umbrelOS 0.5 users are aware that it is no longer functional

This does not remove existing installs of Nitter.


realtobman:
..so how do I remove an existing install on umbrel?
The Icon has a progress bar on it, and I can't click it or remove it.
When I do a right-click -> view in appstore, this happens:
[console.json](https://github.com/user-attachments/files/18449358/console.json)

PR #1952, opened 2024-12-19
Update Firefox to version 133.0.3

al-lac:
Tested update and fresh install on Raspberry Pi 5 and Umbrel Home.


nmfretz:
Excellent, thanks @al-lac. This is working great.

PR #1977, opened 2024-12-21
App Submission: Open WebUI

al-lac:
# App Submission

Open WebUI

@nmfretz It could make sense to split ollama into a separate app. But it has no UI other than showing that it is running when you access the port.

Both parts do not work properly without the root user.

### 256x256 SVG icon

https://filetransfer.io/data-package/N7NbYURd#link

### Gallery images

open-webui-1
open-webui-2
open-webui-3
open-webui-4

### I have tested my app on:
- [x] umbrelOS on Linux VM

### DIFF
--- a/open-webui/docker-compose.yml
+++ b/open-webui/docker-compose.yml
@@ -22,5 +22,5 @@ services:
     depends_on:
       - ollama
     environment:
-      - 'OLLAMA_BASE_URL=http://ollama:11434'
+      OLLAMA_BASE_URL: "http://open-webui_ollama_1:11434"
     restart: on-failure
--- a/open-webui/docker-compose.yml
+++ b/open-webui/docker-compose.yml
@@ -1,6 +1,12 @@
 version: '3.6'
 
 services:
+  app_proxy:
+    environment:
+      APP_HOST: open-webui_web_1
+      APP_PORT: 8080
+      PROXY_AUTH_ADD: "false"
+
   ollama:
     image: ollama/ollama:0.5.4@sha256:18bfb1d605604fd53dcad20d0556df4c781e560ebebcd923454d627c994a0e37
     volumes:
@@ -9,14 +15,12 @@ services:
       - 11434:11434
     restart: on-failure
 
-  open-webui:
+  web:
     image: ghcr.io/open-webui/open-webui:v0.4.8@sha256:c1e4f0927fb0acd53bcbd8bbc92ecaf7ca36a23ee4cb8f25ce90c541012a473a
     volumes:
       - ${APP_DATA_DIR}/data/open-webui:/app/backend/data
     depends_on:
       - ollama
-    ports:
-      - 2876:8080
     environment:
       - 'OLLAMA_BASE_URL=http://ollama:11434'
     restart: on-failure
--- a/open-webui/docker-compose.yml
+++ b/open-webui/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   ollama:
-    image: ollama/ollama:0.5.4@sha256:18bfb1d605604fd53dcad20d0556df4c781e560ebebcd923454d627c994a0e37
+    image: ollama/ollama:0.5.7@sha256:7e672211886f8bd4448a98ed577e26c816b9e8b052112860564afaa2c105800e
     volumes:
       - ${APP_DATA_DIR}/data/ollama:/root/.ollama
     # We could uncomment this if we want to expose the ollama port to the host for other external apps/programs to use. 
@@ -17,7 +17,7 @@ services:
     restart: on-failure
 
   web:
-    image: ghcr.io/open-webui/open-webui:v0.5.4@sha256:42e8fa544facc38d731e3d516fbf478abe435bb4b80798e0934930afea6c5bab
+    image: ghcr.io/open-webui/open-webui:v0.5.7@sha256:b9a3425659236186df16ccf4432a247a353e54dec9549fb475d8b57f0c29a93d
     volumes:
       - ${APP_DATA_DIR}/data/open-webui:/app/backend/data
     depends_on:
--- a/open-webui/docker-compose.yml
+++ b/open-webui/docker-compose.yml
@@ -12,5 +12,7 @@ services:
     volumes:
       - ${APP_DATA_DIR}/data/open-webui:/app/backend/data
     environment:
+      # Exported from ollama app, which is currently a required dependency.
+      # This will need to change once optional dependencies are supported.
       OLLAMA_BASE_URL: $APP_OLLAMA_URL
     restart: on-failure
--- a/open-webui/docker-compose.yml
+++ b/open-webui/docker-compose.yml
@@ -11,8 +11,9 @@ services:
     image: ollama/ollama:0.5.4@sha256:18bfb1d605604fd53dcad20d0556df4c781e560ebebcd923454d627c994a0e37
     volumes:
       - ${APP_DATA_DIR}/data/ollama:/root/.ollama
-    ports:
-      - 11434:11434
+    # We could uncomment this if we want to expose the ollama port to the host for other external apps/programs to use. 
+    # ports:
+    #   - 11434:11434
     restart: on-failure
 
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for yet another submission @al-lac!

> @nmfretz It could make sense to split ollama into a separate app. But it has no UI other than showing that it is running when you access the port.

I think what you've done here makes sense. Users get easy access to an ollama instance + a web UI that can manage model downloads and use them.

I've just taken a quick look at the compose file and I think we could stick the web-ui behind the app proxy (with or without umbrelOS authentication) and then even remove the host-bind port for Ollama so that Docker uses its internal DNS to connect containers. Let me give it a quick test and let you know my findings so we can decide how to move forward.

Also, we've gone ahead and created the gallery assets in the meantime:
https://github.com/getumbrel/umbrel-apps-gallery/tree/master/open-webui

open-webui/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
+version: '3.7'
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: open-webui_web_1
+      APP_PORT: 8080
+      PROXY_AUTH_ADD: "false"
```
nmfretz, line 8: I think it makes sense to disable the proxy authentication here because Open WebUI has its own built-in authentication that includes an initial set-up for an admin user.

I checked and it is possible to disable open-webui's in-built authentication by adding ` WEBUI_AUTH: "False"` as an env var to the open-webui container if we'd rather have umbrelOS proxy auth instead, but then we lose the ability in OPen WebUI to have multiple users with the admin user behind a separate password.

open-webui/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+
+  ollama:
+    image: ollama/ollama:0.5.4@sha256:18bfb1d605604fd53dcad20d0556df4c781e560ebebcd923454d627c994a0e37
+    volumes:
+      - ${APP_DATA_DIR}/data/ollama:/root/.ollama
+    # We could uncomment this if we want to expose the ollama port to the host for other external apps/programs to use. 
+    # ports:
+    #   - 11434:11434
```
nmfretz, line 16: The `open-webui` container connects to the `ollama` container using Docker's internal DNS via the `OLLAMA_BASE_URL` so we don't need to bind this port to the host if Open WebUI is the only application connecting to Ollama.

However, if users want to connect to their Ollama instance via some other external application then we'd need to have this exposed on the host. But I think that for that use-case it probably makes sense to have ollama running as it's own standalone service like you mention:

> @nmfretz It could make sense to split ollama into a separate app.

What are your thoughts on removing this port binding for now? That way this app is a pre-packaged Web UI + Ollama instance that is easy for the average user to use. In the future we can consider either exposing ollama's port on the host or including a standalone ollama instance as another app.

open-webui/docker-compose.yml
```diff
@@ -0,0 +1,27 @@
...
+    restart: on-failure
+
+  web:
+    image: ghcr.io/open-webui/open-webui:v0.4.8@sha256:c1e4f0927fb0acd53bcbd8bbc92ecaf7ca36a23ee4cb8f25ce90c541012a473a
+    volumes:
+      - ${APP_DATA_DIR}/data/open-webui:/app/backend/data
+    depends_on:
+      - ollama
+    environment:
+      OLLAMA_BASE_URL: "http://open-webui_ollama_1:11434"
```
nmfretz, line 26: I've changed this to use the full container name of the ollama instance: `open-webui_ollama_1`.

The reason is that if we kept it as just `ollama` and then a user has another container named `ollama` running (e.g., as a standalone app or as part of another app's compose file) then Docker will "helpfully" try and resolve the correct container by just guessing haha. So sometimes it would connect and other times it wouldn't.

open-webui/umbrel-app.yml
```diff
@@ -0,0 +1,33 @@
...
+  🌐 Open WebUI is an extensible, feature-rich, and user-friendly self-hosted WebUI designed to operate entirely offline.
+
+
+  🤖 It supports various LLM runners, including Ollama and OpenAI-compatible APIs.
+
+
+  🦙 This app comes packaged with [Ollama](https://ollama.com/) pre-configured to work out of the box. 
+  
+  To start downloading models via Ollama, simply type in a model name in the model search bar and click "Pull from Ollama.com".
+  A list of models can be found at https://ollama.com/search.
```
nmfretz, line 18: I've added some instructions for downloading models. We could also consider not doing this and instead linking to something like https://docs.openwebui.com/getting-started/quick-start/starting-with-ollama#a-quick-and-efficient-way-to-download-models

nmfretz (commented):
@al-lac I've committed some atomic changes that you can follow. What do you think of this set up for Open Web UI? Feel free to revert anything that you feel doesn't make sense. I've explained some of my changes below.

nmfretz:
image

open-webui/docker-compose.yml
al-lac: Makes sense! 

I am actually leaning towards giving ollama into its own app or at least keeping the port binding. This would enable the users to connect from their Workstations or use it with other services such as Lobe Chat, Home Assistant, Mattermost and more.

Also we could update ollama better if it was a separate app.

Sadly the ollama website is just an indication that it is running, without any guide on how to connect. So the downside would for sure be that it is not as easy to use as it is now.

But I am fine with whatever you think makes the most sense here!

nmfretz: Ya, I hear you. It would be nice to have Ollama as a standalone service that other services could connect to.

Then Open WebUI could just be shipped as the `open-webui` container, and if users optionally want Ollama functionality they would download the separate Ollama app and connect through Open WebUI's UI. Another option would be to have ollama be `required` in Open WebUI's manifest so that it must first be installed by the user and then when they install Open WebUI it will be pre-configured with Ollama... but it might actually be nice Ollama to be entirely optional for Open WebUI. For example, a user may only want to use Open WebUI with an OpenAI API.

I see what you mean though about the less-than-ideal status page for Ollama:
image

Ideally we'd be able to ship Ollama without having to include some custom frontend like we do for things like Back That Mac Up, Tor Snowflake Proxy, etc, since that just adds overhead when updating the app (e.g., needing to check if the frontend needs updating).

Let me run this by the team and see what we think about having Ollama run separately with just its simple status html. Potentially we could allow this and then include some general instructions in the app description on how to connect to other apps.

nmfretz:
@al-lac, this branch is a proof-of-concept for having Ollama as a standalone app: https://github.com/getumbrel/umbrel-apps/tree/standalone-ollama.

In that branch I've removed ollama from Open WebUI's compose file (https://github.com/getumbrel/umbrel-apps/blob/standalone-ollama/open-webui/docker-compose.yml) and added ollama as a standalone app (https://github.com/getumbrel/umbrel-apps/blob/standalone-ollama/ollama/docker-compose.yml).

Feel free to give it a test and see what you think. Here's a screen recording of:
- connecting to ollama from Open WebUI using the docker container name `ollama_ollama_1` (resolved by Docker's DNS)
- connecting to ollama from Open WebUI using umbrel's Docker's gateway IP `10.21.0.1`
- connecting to ollama from another computer on the same local network using `umbrel.local`

https://github.com/user-attachments/assets/44830765-073e-45b0-9dbb-7deab8022b9e

open-webui/docker-compose.yml
nmfretz: Proof of concept for ollama as a standalone app: https://github.com/getumbrel/umbrel-apps/pull/1977#issuecomment-2588500389

al-lac:
Hey @nmfretz, this is great! I really prefer it this way.

Tested pretty much the same as you did, also tried it in mattermost (copilot) and Lobe Chat.

For Lobe Chat to be able to connect, this environment variable needs to be set on ollama:
```yaml
environment:
      OLLAMA_ORIGINS: "*"
```

nmfretz:
> For Lobe Chat to be able to connect, this environment variable needs to be set on ollama:

Nice, thanks for testing that.

> Hey @nmfretz, this is great! I really prefer it this way.

Ya, having it standalone and obviously available for other apps seems the way to go to me as well. It's just a shame that packaging it alone leaves it without a UI for downloading/managing models. You'd need to either use the terminal (totally fine for technical users) or download a UI app like this Open Web UI submission.

Let's finalize our decision after Feb 1 😉

nmfretz:
Thanks for all your efforts getting this to the app store @al-lac! We're live: 

image

PR #1980, opened 2024-12-23
Update Helipad to v0.2.1

ericpp:
Updates Helipad from v0.2.0 to v0.2.1


nmfretz:
Thanks for this update @ericpp! I have reviewed and tested. Going live.

PR #1969, opened 2024-12-21
Update pocketbase to version v0.23.11

nmfretz:
🤖 This is an automated pull request to update pocketbase to version v0.23.11. This PR must be reviewed and tested before merging.

PR #1983, opened 2024-12-24
Update pocketbase to version 0.23.12 and create initial superuser

al-lac:
This PR updates pocketbase to version 0.23.12 and creates the initial superuser with a migration script.

Due to changes made to pocketbase starting with version 0.23, the intial superuser creation has to happen via the cli, the URL that is shown on the terminal or a migration script. See: https://github.com/pocketbase/pocketbase/discussions/5814

Tested both an upgrade and a fresh install on the Raspberry Pi 5 and Umbrel Home.

Existing users / admins will not be touched.

Closes #1969

### DIFF
--- a/pocketbase/data/migrations/initial_superuser.js
+++ b/pocketbase/data/migrations/initial_superuser.js
@@ -1,0 +1,22 @@
+// https://pocketbase.io/docs/js-migrations/#creating-initial-superuser
+// pb_migrations/1687801090_initial_superuser.js
+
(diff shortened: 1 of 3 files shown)


nmfretz:
Fantastic work @al-lac! I have reviewed and tested, including checking that existing admin user's credentials are not nuked.

I added a sentence to the release notes about the default credentials so that existing users aren't confused when they go to open the app after updating and are met with this popup:

image

PR #1987, opened 2024-12-29
Added bitcoind container 'user' parameter

rstmsn:
Added 'user' parameter to bitcoind docker-compose service, to ensure container is launched as 'umbrel' user. This in turn ensures all files created by the container have the expected user:group (umbrel:umbrel).

This update resolves permission related errors encountered by third party apps (e.g. Electrs), when attempting to access bitcoin-knots .cookie file.


Retropex (approved):

nmfretz:
Thanks for this @rstmsn! We need to make sure that we don't accidentally break existing user's installations. For example, just adding this change will work for new app installs, but existing users would then have `bitcoind` running with user 1000 permissions and potentially not be able to access files previously created/overwritten with root permissions. We can do this with a lifecycle hook that chowns the bitcoind data directory. I can add this to the existing `pre-start` hook.

Before I do this though, @Retropex is the issue with knots `bitcoind` running as root an accidental change from a recent update? Like a change to the Docker image build that resulted in the default user being root?

rstmsn:
Thanks @nmfretz. @Retropex may provide more clarity, however my understanding is that the 'bitcoin-knots' app used the 'bitcoin' app as a starting template, which was then customised to run the 'bitcoin-knots' client, as opposed to the 'bitcoin-core' client.
The 'user' parameter for the 'bitcoind' container, in the 'bitcoin-knots' 'docker-compose.yml' file was unintentionally omitted, possibly as far back as the birth of the 'bitcoin-knots' app.

As many other apps in the Umbrel eco system typically run as user '1000' (umbrel), so too should the 'bitcoin-knots' app/containers, to ensure these third party apps have sufficient priviledge to access/read files created by the 'bitcoin-knots' services. 

In terms of supporting historic installs, presumably something along the lines of 'chown -R umbrel:umbrel ~/umbrel/app-data/bitcoin-knots' would take care of this, if that could be included in the pre-start hook? 

Many thanks for your guidance,
R.

nmfretz:
Thanks for the details @rstmsn! I just want to clarify the image build characteristics with @Retropex before we make the change.

> The 'user' parameter for the 'bitcoind' container, in the 'bitcoin-knots' 'docker-compose.yml' file was unintentionally omitted, possibly as far back as the birth of the 'bitcoin-knots' app.

The 'user' parameter was very likely not included in the original submission because the knots `bitcoind` image hardcoded user 1000:1000 (similar to what the old [lncm/bitciond image](https://github.com/lncm/docker-bitcoind) did before we migrated to our own Docker file https://github.com/getumbrel/docker-bitcoind). I suspect that there has been a recent change in the knots `bitcoind` Docker image build because we extensively tested the Bitcoin Knots app compatibility with Electrs (and other apps) when we released the umbrelOS 1.3 with swappable apps: https://x.com/umbrel/status/1861030142321254785

@Retropex is this maybe a new debian-based image build based on https://github.com/getumbrel/docker-bitcoind? I just want to make sure before we push a fix in case other changes are needed.

Retropex:
Hello guys, Bitcoin Knots effectively seems to run in root, I have not hardcoded any users in the Dockerfile. I have effectively used your [template](https://github.com/Retropex/docker-bitcoind-prebuilt/blob/master/Dockerfile) with Bitcoin Core.

nmfretz:
> Hello guys, Bitcoin Knots effectively seems to run in root, I have not hardcoded any users in the Dockerfile. I have effectively used your [template](https://github.com/Retropex/docker-bitcoind-prebuilt/blob/master/Dockerfile) with Bitcoin Core.

Excellent, thanks for the confirmation @Retropex! What happened then is when the switch was made to using the debian-based template, user 1000:1000 was no longer defined in the Dockerfile like in the old lncm template.

I will create a life-cycle hook to make sure we reset permissions for impacted users.

nmfretz:
I have added logic to the existing pre-start hook that recursively `chown`'s the bitcoind data directory. We write out a flag file so that we don't `chown` unnecessarily on every app start (i.e, for fresh installs where it is unnecessary, and after we have completed the `chown` action for the first time). 

Tested on fresh install and app update. Electrs connects successfully.

Thanks again @rstmsn! Really appreciate it.

PR #2018, opened 2025-01-07
Update homebox to version 0.17.1

al-lac:
The split between architectures was finally fixed in sysadminsmedia/homebox#365 🎉 

Tested update and fresh install on Raspberry Pi 5 and Umbrel Home.

### DIFF
--- a/homebox/docker-compose.yml
+++ b/homebox/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       APP_PORT: 7745
     
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks for staying on top of this one @al-lac! Tested and working great.

> The split between architectures was finally fixed in https://github.com/sysadminsmedia/homebox/pull/365 🎉

With this fix, I've added homebox back into the list of apps that the auto-updater will attempt to update.

PR #2019, opened 2025-01-07
Update zabbix to version 7.2.2

al-lac:
Tested fresh install and update on Umbrel Home.


nmfretz:
Thanks for this update as well @al-lac. Tested and working great.

PR #2038, opened 2025-01-08
Update btcpay-server to v2.0.5

nmfretz:
Updates both btcpayserver and nbxplorer containers. Requires testing before merge.


nmfretz:
Reviewed and tested (app update and fresh install)

PR #2039, opened 2025-01-08
Update duplicati to v2.1.0.2_beta_2024-11-29

nmfretz:
Requires testing before merge

### DIFF
--- a/duplicati/docker-compose.yml
+++ b/duplicati/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   duplicati:
-    image: duplicati/duplicati:2.0.8.1_beta_2024-05-07@sha256:97b8ef7bcaa095793ef9f39049851f987860a8c832aaf67ef977ff398f9e996e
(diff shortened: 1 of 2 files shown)


nmfretz:
@sharknoon can I please get a double check on this duplicati update?

I first tested the update by only changing the Duplicati image to the [latest beta release](https://github.com/duplicati/duplicati/releases/tag/v2.1.0.2_beta_2024-11-29), not canary release.

I was met with this error, indicating that the old `entrypoint` command was failing

```
Jan 09 00:30:25 umbrel-dev umbreld[939609]: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "/usr/bin/duplicati-server": stat /usr/bin/duplicati-server: no such file or directory: unknown
```

I've removed the entrypoint, and replaced the server options with equivalent environment variables. I have tested a fresh install and app update, but would appreciate some extra testing.

sharknoon (approved):
I have tested it on my umbrel and it works like a charm. I am glad that duplicati has added environment variables, because modifying the entrypoint was a last resort anyway. This way it is easier to update in the future as well.
Thank you for your update 😀

nmfretz:
Wonderful, thanks @sharknoon!

>  I am glad that duplicati has added environment variables

Ya me too! We no longer have to worry about the entrypoint command changing with an update.

PR #2040, opened 2025-01-09
feat(jellyfin): Add hardware-accelerated video decoding support for Jellyfin

Knufle:
… server by mapping required GPU device nodes (/dev/dri, /dev/vcsm, /dev/vchiq, and /dev/video*) in the Docker Compose configuration.

#### **Context:**  
The current Docker Compose configuration for the Jellyfin server does not include hardware-accelerated video decoding capabilities. This feature is essential for improving media playback performance, especially for high-definition video streams, by leveraging GPU-based transcoding.  

#### **Proposed Changes:**  
Modify the `docker-compose.yml` to include the following device mappings:  
- `/dev/dri`: Enables Direct Rendering Infrastructure (DRI) for GPU access.  
- `/dev/vcsm` and `/dev/vchiq`: Facilitates GPU memory and communication, particularly for Raspberry Pi systems.  
- `/dev/video10`, `/dev/video11`, `/dev/video12`: Grants access to video device nodes for hardware-based video encoding/decoding.  

#### **Benefits:**  
- Enables hardware-based transcoding for Jellyfin, reducing CPU usage.  
- Enhances performance during media playback, particularly for high-definition and 4K content.  
- Provides compatibility with GPUs and Raspberry Pi-specific hardware.  

#### **Steps to Validate:**  
1. Update the `docker-compose.yml` file with the proposed changes.  
2. Deploy the updated Jellyfin container.  
3. Run `docker exec <container_name> ls -l /dev/` to verify the container has access to the mapped devices.  
4. Check Jellyfin's playback settings and logs to confirm that hardware acceleration is enabled and functional.  

#### **Notes:**  
- Ensure the host system has the appropriate GPU drivers installed and configured.  
- Additional Jellyfin configuration may be required to enable hardware acceleration in the application itself.

### DIFF
--- a/jellyfin/docker-compose.yml
+++ b/jellyfin/docker-compose.yml
@@ -20,10 +20,3 @@ services:
     ports:
       # Service auto-discovery
       - 7359:7359/udp
-    devices:
-      - /dev/dri:/dev/dri
-      - /dev/vcsm:/dev/vcsm
-      - /dev/vchiq:/dev/vchiq
-      - /dev/video10:/dev/video10
-      - /dev/video11:/dev/video11
-      - /dev/video12:/dev/video12
--- a/jellyfin/docker-compose.yml
+++ b/jellyfin/docker-compose.yml
@@ -20,3 +20,10 @@ services:
     ports:
       # Service auto-discovery
       - 7359:7359/udp
+    devices:
+      - /dev/dri:/dev/dri
+      - /dev/vcsm:/dev/vcsm
+      - /dev/vchiq:/dev/vchiq
+      - /dev/video10:/dev/video10
+      - /dev/video11:/dev/video11
+      - /dev/video12:/dev/video12
--- a/jellyfin/hooks/pre-install
+++ b/jellyfin/hooks/pre-install
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Path to docker-compose.yml one directory up from the hooks folder
+COMPOSE_FILE="$(cd "$(dirname "$0")" && pwd)/../docker-compose.yml"
+
+# Dynamic device mapping for hardware acceleration
+DEVICE_MAPPINGS=""
+
+# Check for DRI devices
+if [ -e /dev/dri ]; then
+    DEVICE_MAPPINGS+="      - /dev/dri:/dev/dri\n"
+fi
+
+# Check for Raspberry Pi specific devices
+if [ -e /dev/vcsm ]; then
+    DEVICE_MAPPINGS+="      - /dev/vcsm:/dev/vcsm\n"
+fi
+
+if [ -e /dev/vchiq ]; then
+    DEVICE_MAPPINGS+="      - /dev/vchiq:/dev/vchiq\n"
+fi
+
+# Check for video devices
+for i in {10..12}; do
+    if [ -e "/dev/video$i" ]; then
+        DEVICE_MAPPINGS+="      - /dev/video$i:/dev/video$i\n"
+    fi
+done
+
+# Update docker-compose.yml with dynamic device mappings
+if [ -n "$DEVICE_MAPPINGS" ]; then
+    {
+        # Use printf instead of echo to prevent unwanted newlines
+        [ "$(tail -c1 "$COMPOSE_FILE")" != "" ] && printf "\n"
+        printf "    devices:\n"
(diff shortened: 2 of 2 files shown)


nmfretz:
Hey @knufle, thanks for this PR! Hardware acceleration for Jellyfin would be a fantastic addition, but unfortunately with our current apps framework this implementation won't work.

This is because umbrelOS users could be using any device (e.g., raspberry pi, umbrel home, mini-pc, laptop etc), but will all be installing the jellyfin app from the same `docker-compose.yml` file.

When you list device bindings in the compose file, the user's device needs to have all those devices actually exist on their host machine. If they don't exist, then Docker will error and the app will fail to install.

```yaml
devices:
  - /dev/dri:/dev/dri
  - /dev/vcsm:/dev/vcsm
  - /dev/vchiq:/dev/vchiq
  - /dev/video10:/dev/video10
  - /dev/video11:/dev/video11
  - /dev/video12:/dev/video12
```

For example, say your machine doesn't have `/dev/vcsm`, Docker will fail with something like:
```
Error response from daemon: error gathering device information while adding custom device "/dev/vcsm": no such file or directory
```

Looking at [Jellyfin's docs for hardware acceleration](https://jellyfin.org/docs/general/administration/hardware-acceleration/), there are different device config options depending on the vendor of the installed GPU (if any). So with our current framework we have no way of knowing this in advance and Jellyfin would be broken for the majority of users.

We have hardware acceleration for various apps/use-cases on our radar. Hopefully this will be something available in the future.

Knufle:
@nmfretz Thanks for checking that, I really thought it would not be a problem as I used another home server application in the past that did something similar in its docker-compose.yml for jellyfin and it didn't have any issues as far as I'm aware, I thought docker would just ignore the not found devices.

Anyways, with that in mind, do you think this could be a solution? https://github.com/getumbrel/umbrel-apps/pull/2040/commits/7687a5c8f32a1593f9322fdb62b83a6587ab8524

nmfretz:
@Knufle - really clever thinking of using a life-cycle hook. We've thought about this before as well but came to the conclusion that handling this with a bash script will not be robust enough. If something goes wrong, the user will have no feedback as to what the issue is. And there are so many variations of hardware that would require gpu support (e.g., all x86 variants running umbrelOS, home, RPi's with different driver/hardware versions), and potentially certain hardware needs to install drivers for certain devices to show up.

Our plan is to handle this at the `umbreld` level https://github.com/getumbrel/umbrel/tree/master/packages/umbreld
This will be safer and easier. We could potentially also allow users to  manually change `devices` and environment variables related to hardware acceleration in apps so that there is an easy way to manually adjust and roll back when necessary.

For now, the best way to enable hardware acceleration on Jellyfin would probably be to either:

1) Create your own [Community App Store](https://github.com/getumbrel/umbrel-community-app-store) with a Jellyfin app that is configured for hardware acceleration for your specific device. Then others can use your app store if they are interested in hardware acceleration, and you can easily update the jellyfin version on your app store, which when you go to update the app on your umbrel the hardware acceleration configuration will persist.

2) Use the Portainer app to run Jellyfin with your specific hardware acceleration config.

> I really thought it would not be a problem as I used another home server application in the past that did something similar in its docker-compose.yml for jellyfin and it didn't have any issues as far as I'm aware, I thought docker would just ignore the not found devices.

Ya, it would be great if we could just include a list of all devices, but unfortunately Docker will fail to bring up the container when they are missing. You can test this by adding a device to the compose file that your specific device doesn't have and trying to install the app.


Closing this PR, but thanks again for opening it. We've bumped up hardware acceleration higher on our priority list now.

only1thor:
I have tried to get this working on a system with an Intel N150, but for the life of me i can't get any devices to load in `/dev/dri/`. is this a limitation on the kernel Umbrel uses?

only1thor:
another thought on loading drivers based on system;
could umbrel provide a set of environmentvariables that could be used for stuff like this? 
say set `$AVAILABLE_HARDWARE_RENDER_PATHS` to `"/dev/dri/renderD128", "/dev/video10"`
to then include in the docker compose like this:

```yaml
devices: [${AVAILABLE_HARDWARE_RENDER_PATHS}]
```

I think this should work since json is valid yaml. 
and variables like `${DEVICE_HOSTNAME}` are already in use, so some passing of variables to the compose files is already implemented.

Edit:
Just realized that an empty list is still an issue, and would break startup of the app.

PR #2029, opened 2025-01-08
[ON HOLD] - Update syncthing to version v1.29.0

nmfretz:
🤖 This is an automated pull request to update syncthing to version v1.29.0. This PR must be reviewed and tested before merging.

PR #2043, opened 2025-01-10
Update syncthing to version 1.29.1

al-lac:
Tested on Umbrel Home.

Closes #2029


nmfretz:
Excellent, thanks for this @al-lac! I have reviewed and tested.

PR #2065, opened 2025-01-17
Update Lobe Chat to version 1.47.5

al-lac:
Tested update on Umbrel Home and fresh install on Raspberry Pi 5.


nmfretz:
Excellent, thanks for this update as well @al-lac! I have reviewed and tested.

One small thing that is just personal preference and is not a blocker at all for updates like this:
I removed `- Upgraded to Next.js 15 and React 19` from the release notes since development/dependency updates like this won't really mean much to the average user. Totally just personal preference though.

al-lac:
Totally makes sense, thanks! 🎉

PR #2069, opened 2025-01-19
Update Firefox to version 134.0.1

al-lac:
Tested update on Raspberry Pi 5 and fresh install on Umbrel Home.


nmfretz:
Another perfect update 👑. Thanks @al-lac.

PR #2094, opened 2025-01-21
Update Linkwarden to version 2.9.3 and run as root

al-lac:
Tested update on Umbrel Home and fresh install on Raspberry Pi 5.

I noticed some bugs like Linkwarden not being able to create snapshots of websites. This is due to the fact that official support for non-root is still missing. 
There is already a PR open for non-root support: linkwarden/linkwarden#779 

Also changed the database url as I have seem some errors in the logs. Changing it to the full name seems to fix it.

Would run it as root for the time being if that is ok @nmfretz


nmfretz:
Excellent update @al-lac. Great find and fixes re the non-root user issue and postgres url.

Tested and working great.

PR #2113, opened 2025-01-27
Update immich to version 1.125.6 and disable umbrel app proxy

al-lac:
Tested update and fresh install on Umbrel Home. Fresh install on Raspberry Pi 5. (Took a long time to start on the Pi)

I disabled the app proxy authentication as immich comes with authentication built in. 
This way users are able to setup a reverse proxy in front of immich for external access. Please let me know if it is ok like this @nmfretz.

### DIFF
--- a/immich/docker-compose.yml
+++ b/immich/docker-compose.yml
@@ -22,7 +22,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   server:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @al-lac! Immich seems to have decent auth, so this sounds good to me considering it allows easier proxying. The one thing people will lose is umbrel's build in 2fa if they have that enabled (good idea adding the warning in the release notes).

I bumped to 1.125.6 since immich issued a couple updates before I got to this.

PR #2153, opened 2025-02-04
Update tailscale to v1.80.0

al-lac:
🤖 This is an automated summary of installation tests for tailscale:

- ✅ App successfully installed in umbrel-dev instance
- ✅ App successfully listed in umbrel.yaml
- ✅ App UI successfully returns a 200 status code
- 📸 Screenshot of the app:
  App Screenshot

**This PR must still be reviewed and tested before merging.**


nmfretz:
Reviewed and tested

wh0am1-dev:
Hi @al-lac, I've just updated to 1.80.0 and after that I had to log in again. When clicking in the login button, this message shows:

image

Searching through tailscale repo, I've found there's currently an issue when accessing tailscale 1.80.0 web interface through HTTP: https://github.com/tailscale/tailscale/issues/14872

Since accessing the web interface through HTTP is the only way of using tailscale on umbrel (unless you tinker with something like nginx proxy manager or anything similar, which is not trivial), could we revert this until we have a fix merged into the tailscale repo?

al-lac:
Hey @wh0am1-dev! Good find. I think it makes sense to roll it back. What do you say @nmfretz?

al-lac:
@wh0am1-dev thanks for the hint. The version has been rolled back to 1.78.3.

PR #2169, opened 2025-02-07
App Submission: RustDesk Server

Knufle:
# App Submission
RustDesk Server

This PR adds RustDesk server to Umbrel. RustDesk is an open-source remote desktop software that allows secure remote connections.

The server consists of two components:
- hbbs (ID/Rendezvous server)
- hbbr (Relay server)

Both components are required for proper functionality and are included in this deployment.

### 256x256 SVG icon

https://limewire.com/d/ceee5ca3-4cfd-4d82-ba71-a0af37693c14#yLGrXJICd91f3YfXMUzWZgJR8tmMIhcXtc0DZn75uLk

### Gallery images

There's no image to be shown here since it's all server

### I have tested my app on:
- [ ] umbrelOS on Linux VM

### DIFF
--- a/rustdesk/docker-compose.yml
+++ b/rustdesk/docker-compose.yml
@@ -1,0 +1,23 @@
+version: '3.7'
+
+services:
+  hbbs:
+    container_name: hbbs
+    image: rustdesk/rustdesk-server@sha256:680f8ba5accafc264d15076f33a6fdb9cb6f4d963a0fc92e01023ca0e919cc83
+    command: hbbs
+    volumes:
+      - ./data:/root
+    network_mode: "host"
+
+    depends_on:
+      - hbbr
+    restart: unless-stopped
+
+  hbbr:
+    container_name: hbbr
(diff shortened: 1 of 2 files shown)


al-lac:
Hey @Knufle! Nice submission!

Sadly only the Pro version comes with a Web Console. For this to make it to the app store, there should be some Web UI that would at least explain users what they need to do to connect to the server in the client apps.

FYI: @nmfretz

nmfretz:
Thanks for the submission @Knufle!

@al-lac is correct that we need some sort of UI so that when a user clicks on the app icon they are directed to something. This could be as simple as a container that simply shows whether or not the services are running. For example, the mosquitto app (https://github.com/getumbrel/umbrel-apps/tree/master/mosquitto) uses html/css from other apps we've added in the past to do this:

nmfretz:
Whoops, sorry I accidentally closed this when navigating around

al-lac:
Closing this as #2169 fixes the missing Web UI part.

PR #2174, opened 2025-02-09
App Submission: cobalt

dennysubke:
# App Submission

### cobalt
...

### Icon
logo
...

### Gallery images

...

### I have tested my app on:
- [x] umbrelOS on Linux VM


dennysubke:
Hey @nmfretz 

Just for your information: The developers of Cobalt didn't provide a Docker image with UI support, so I built one myself. The app is truly amazing, and I’d love to see this PR merged. 🚀

Thanks for reviewing!

cobalt/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+    restart: on-failure
+    user: "1000:1000"
+    ports:
+      - 0.0.0.0:9000:9000
+    environment:
+      API_URL: "http://umbrel.local:9000/"
+      API_AUTH_REQUIRED: 0
+      DURATION_LIMIT: 86400
+    volumes:
+      - ${APP_DATA_DIR}/data/cookies:/cookies.json
```
al-lac, line 22: This would mount to a directory called `cookies.json' which would not work. From what I have seen in the documentation, this file would need to be pre-filled with tokens for various sites. I would leave this out for now.
```suggestion
```

cobalt/umbrel-app.yml
```diff
@@ -0,0 +1,36 @@
...
+  cobalt is a powerful and intuitive media downloader built to simplify the process of saving your favorite content from the web. Designed for users who want to enjoy their media without the interruptions of ads, trackers, or paywalls, cobalt offers a streamlined experience. Whether you're downloading videos, music, images, or other types of media, cobalt acts as a proxy that fetches content directly from its source, ensuring a fast and seamless process.
+
+
+  What sets cobalt apart is its commitment to user privacy and freedom. The tool doesn’t cache content, so you’re always downloading directly from the original source. It also avoids unnecessary distractions, like ads or third-party interference, letting you enjoy a cleaner, faster, and more reliable media retrieval process.
+
+
+  cobalt is ideal for anyone who values simplicity and control over their media downloads. With its easy-to-use interface and efficient functionality, you can quickly retrieve and store content without the usual hassle. Whether you’re an enthusiast collecting media or someone simply trying to access content without restrictions, cobalt makes it easy to download what you love, whenever you want.
```
al-lac, line 15: Add single quotes instead.
```suggestion
  What sets cobalt apart is its commitment to user privacy and freedom. The tool doesn't cache content, so you're always downloading directly from the original source. It also avoids unnecessary distractions, like ads or third-party interference, letting you enjoy a cleaner, faster, and more reliable media retrieval process.


  cobalt is ideal for anyone who values simplicity and control over their media downloads. With its easy-to-use interface and efficient functionality, you can quickly retrieve and store content without the usual hassle. Whether you're an enthusiast collecting media or someone simply trying to access content without restrictions, cobalt makes it easy to download what you love, whenever you want.
```

cobalt/umbrel-app.yml
```diff
@@ -0,0 +1,36 @@
+manifestVersion: 1
+id: cobalt
+name: cobalt
```
al-lac, line 3: Would name it with an uppercase `C` for consistency.
```suggestion
name: Cobalt
```

cobalt/umbrel-app.yml
```diff
@@ -0,0 +1,36 @@
...
+tagline: A media downloader that doesn't piss you off
```
al-lac, line 4: Use the tagline from their repo.
```suggestion
tagline: Best way to save what you love
```

cobalt/docker-compose.yml
```diff
@@ -0,0 +1,22 @@
...
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: cobalt_web_1
+      APP_PORT: 9001
+      PROXY_AUTH_ADD: "false"
+  web:
+    image: dennysubke/cobalt:10.6@sha256:870110e7002b7080adc81f9cee5357ab784adba81385a7b51663a7c1e92030cd
+    init: true
```
al-lac, line 11: I don't think `init` is needed here.
```suggestion

```

al-lac (changes requested):
Hey @dennysubke! Another great submission! :octocat: 

Added some points that should be changed. Also I think you can remove the directory as the cookies.json would not make a lot of sense now.

Tested it without it and everything worked just fine.

cobalt/umbrel-app.yml
dennysubke: Hey, thanks for the suggestion! I kept it lowercase because cobalt officially uses that styling. 

Umbrel already has several apps that follow their own naming conventions, like autobrr, code-server, nolooking, ntfy, noStrudel, tautulli und usocial. So I think it makes sense to stay consistent with the app's original branding. 

Let me know what you think! 😊

cobalt/umbrel-app.yml
dennysubke: I knew it wouldn't go through! 😂

dennysubke:
> Hey @dennysubke! Another great submission! :octocat:
> 
> Added some points that should be changed. Also I think you can remove the directory as the cookies.json would not make a lot of sense now.
> 
> Tested it without it and everything worked just fine.

All changes have been made now! Thanks again for your help and feedback! 😊

cobalt/umbrel-app.yml
al-lac: True that, lets leave it like this then!

cobalt/umbrel-app.yml
nmfretz: ahaha

nmfretz:
Excellent, thanks @dennysubke. Great initiative creating the image 👌. And thanks for reviewing @al-lac. 

I've changed the API to the nearest free port (`9013`). We'll make the equivalent change in the gallery image. Great idea including the important config instruction as a gallery asset.

Something to consider: how do you feel about moving the note about the processing server config higher in the app description? This way it's not hidden until the user clicks "read more" in the app store. If the user doesn't set this then they'll be using cobalt's server instead which may not be clear to the average user.

If we don't want a ⚠️ symbol to be the first thing shown in the app description we could just use a settings cog ⚙️ or tool icon 🛠️ instead. Something like:

> ⚙️ In order use your own cobalt instance, go to settings > instances > and toggle "use a custom processing server". Add http://umbrel.local:9013 as your server.

dennysubke:
> Excellent, thanks @dennysubke. Great initiative creating the image 👌. And thanks for reviewing @al-lac.
> 
> I've changed the API to the nearest free port (`9013`). We'll make the equivalent change in the gallery image. Great idea including the important config instruction as a gallery asset.
> 
> Something to consider: how do you feel about moving the note about the processing server config higher in the app description? This way it's not hidden until the user clicks "read more" in the app store. If the user doesn't set this then they'll be using cobalt's server instead which may not be clear to the average user.
> 
> If we don't want a ⚠️ symbol to be the first thing shown in the app description we could just use a settings cog ⚙️ or tool icon 🛠️ instead. Something like:
> 
> > ⚙️ In order use your own cobalt instance, go to settings > instances > and toggle "use a custom processing server". Add http://umbrel.local:9013 as your server.

Thanks for the feedback and suggestions @nmfretz! I agree with moving the note higher in the app description to make it more visible. Also, I think using a settings icon ⚙️ instead of the ⚠️ symbol is a great idea for a cleaner look. I've made the necessary adjustments.

nmfretz:
Thanks again @dennysubke. I'm sending this live now with 4 of the gallery assets and then will add the 5th one with port 9013 instead of 9000 soon.

image

dennysubke:
> Thanks again @dennysubke. I'm sending this live now with 4 of the gallery assets and then will add the 5th one with port 9013 instead of 9000 soon.
> 
> image
> 

Thanks @nmfretz! Do you want me to take care of adding the 5th gallery asset with port 9013, or do you already have it covered?

al-lac:
@dennysubke can you also look into updating the container of this app? :-)

al-lac:
@dennysubke pinging once again, latest version would be `10.9.3`

Could not find a changelog though, would probably need to be created by looking at the recent commits.

dennysubke:
Moin @al-lac!

Thanks for the reminder! I've gone ahead and updated the container to version 10.9.4 and opened a PR for it. Let me know if anything needs to be adjusted! 💪🏻

dennysubke:
Hey @johnstrong1 

That’s a good idea! 👍🏻 

I’ll look into it and see how best to implement it. Maybe I can get it done with the next update.

PR #2218, opened 2025-02-17
App Submission: Kimai

al-lac:
# App Submission

Kimai

### 256x256 SVG icon
https://github.com/kimai/images/blob/main/kimai_logo_transparent.svg

### Gallery images
https://www.kimai.org/images/screenshots/screenshot-dashboard.webp
https://www.kimai.org/images/screenshots/screenshot-reporting.webp

### I have tested my app on:
- [ ] umbrelOS on Linux VM

### DIFF
--- a/kimai/docker-compose.yml
+++ b/kimai/docker-compose.yml
@@ -17,7 +17,7 @@ services:
     environment:
       ADMINMAIL: umbrel@umbrel.local
       # cannot use $APP_PASSWORD here, as the maximum allowed length is 60 characters
-      ADMINPASS: kimaiumbrel
+      ADMINPASS: changeme
       DATABASE_URL: "mysql://kimaiuser:${APP_PASSWORD}@kimai_db_1/kimai?charset=utf8mb4&serverVersion=8.3.0"
     restart: on-failure
     depends_on:
--- a/kimai/docker-compose.yml
+++ b/kimai/docker-compose.yml
@@ -1,0 +1,42 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: kimai_app_1
+      APP_PORT: 8001
+      PROXY_AUTH_ADD: "false"
+
+  app:
+    # currently does not work properly rootless
+    #user: "1000:1000"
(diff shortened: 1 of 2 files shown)


nmfretz:
Perfect submission @al-lac 🫡. Let's get gallery assets prepared.

### A thought on the default credentials

I like your solution here of warning the user right in the app description to change the password.

```yaml
 # cannot use $APP_PASSWORD here, as the maximum allowed length is 60 characters
 ADMINPASS: kimaiumbrel
```

> ⚠️ To get started, log in with the provided default credentials. Make sure to change the default admin password before exposing your instance to the internet.

We could also consider these, but totally not necessary:

1) Keep your same description but just change the password to `changeme` so it is even more obvious to someone that they should change it.

2) If instead we want a unique, secure password by default where we don't rely on the user changing it we could export one from an `export.sh` file. We could take advantage of [derive_entropy](https://github.com/getumbrel/umbrel/blob/86f452c8e4abc97b18b8787ea5302565a382399c/packages/umbreld/source/modules/apps/legacy-compat/app-script#L128-L147) and do something like:

```sh
export APP_KIMAI_PASSWORD=$(derive_entropy "${app_entropy_identifier}-KIMAI_PASSWORD" | head -c32)
```

What do you feel is best to go with?

al-lac:
@nmfretz ah yeah I really like the second option! 

How would we be able to show this value in the default credentials then?

nmfretz:
> @nmfretz ah yeah I really like the second option!
>
>How would we be able to show this value in the default credentials then?

Ah shit, we wouldn't be able to! I didn't think that through properly. Sorry about that.

Shall we change it to `changeme` then so it's very obvious that it should be changed?

al-lac:
@nmfretz damn what a shame. A feature for generating secrets and other values that can easily be used in the compose files and shown on the web ui would be really nice to have!

Done. Should we also put the password notice higher on the description?

nmfretz:
@al-lac going live 🎉

image

> @nmfretz damn what a shame. A feature for generating secrets and other values that can easily be used in the compose files and shown on the web ui would be really nice to have!

Totally agree.

> Done. Should we also put the password notice higher on the description?

I think what you've done here is great, since (at least for now) this will primarily be used without exposing to the the internet. It's an easy change to make later if you want to place the warning higher in the description.

PR #2225, opened 2025-02-19
App Submission: Gitingest

dennysubke:
# App Submission

### Gitingest
...

### Icon
klXVPov - Imgur

...

### Gallery images
XDnMRVm - Imgur

jpoRiVD - Imgur

NlddoBz - Imgur

...

### I have tested my app on:
- [x] umbrelOS on Linux VM


dennysubke:
Gitingest is a real hidden gem for anyone looking to turn Git repositories into LLM-friendly prompts quickly and effortlessly! 🚀

gitingest/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+      APP_HOST: gitingest_app_1
+      APP_PORT: 8000
+
+  app:
+    image: elestio/gitingest:v0.1.3@sha256:c1f0d290be9e49141ecc500c0002ee2401fe11b1b391e3f375a66dd99f5e2c08
+    restart: on-failure
+    environment:
+      - ALLOWED_HOSTS=*
+      - UID=1000
+      - PUID=1000
```
al-lac, line 15: I would use the `user:` option instead of the environment variables.
```suggestion
    image: elestio/gitingest:v0.1.3@sha256:c1f0d290be9e49141ecc500c0002ee2401fe11b1b391e3f375a66dd99f5e2c08
    user: "1000:1000"
    restart: on-failure
    environment:
      - ALLOWED_HOSTS=*

```

al-lac (dismissed):
Really nice submission @dennysubke!

Only have a minor change to add, the rest is flawless! 🔝

dennysubke:
@al-lac Completed! 😎

al-lac:
@nmfretz this app would be good to go from my side!

nmfretz:
Thanks guys. Really great work @dennysubke. We'll send the gallery assets through design.

@al-lac @sharknoon - I'm wondering whether or not we should tighten this up:

```yaml
environment:
      - ALLOWED_HOSTS=*
``` 

It's a Django variable used to prevent host header attacks: https://docs.djangoproject.com/en/5.1/ref/settings/#allowed-hosts. 

We could change to this:

```
- ALLOWED_HOSTS=${DEVICE_DOMAIN_NAME}
``` 

But then people with multiple umbrels may run in to issues when accessing via something like umbrel-2.local.

We could also include the device IP and just `${DEVICE_HOSTNAME}` so people can access via IP and tailscale.

What do you guys think?

sharknoon:
That is a very tricky question. Host header attacks are a form of cross-site-scripting attacks. While for many platforms, this is a major concern, but I think umbrel devices are rarely public. It is a matter of trust, do I trust all my friends and family memebers, who use my umbrel? Is my umbrel device publicly accessible or just via VPN (Tailscale, Cloudflare Tunnels, ...)?

On the other hand how important is it to allow multiple umbrel devices?

I personally think, the best solution would be to add a setting to umbrelOS, allowing the hostname (and maybe domains too) to be specified, under which the umbrel is reachable. Until then, I would make this more strict by setting `- ALLOWED_HOSTS=${DEVICE_DOMAIN_NAME}` with a note in the Readme, on how to change it manually to a different hostname, if needed.

al-lac:
Yeah I would also just add the `${DEVICE_DOMAIN_NAME}` and the local + tailscale IPs would be nice.

@nmfretz there is no tailscale ip in the exports file for the app, guess we would need to add that right?

nmfretz:
Thanks @sharknoon!

What I've done is allowed umbrel.local, the device's IP, and tailscale magicDNS name since these are very simple to do. Let's keep our ears out for anyone having issues with this and then we can take another look if needed.

Going live @dennysubke!  

image

dennysubke:
Ok. I don't think the app was needed. 😂

Changing github.com to **uithub.com** gives you the entire code repo and structure, so you can pass it to an LLM.

Example: https://uithub.com/getumbrel/umbrel-apps

Who of you knew this? 😂 @al-lac @nmfretz @sharknoon

sharknoon:
There are many proprietary non-selfhostable alternatives available. I think it is always good to be able to choose, weather you want to go with self-hosted or SaaS solution

nmfretz:
Ha, ya no worries at all @dennysubke. I agree with @sharknoon here. For example, Gitingest hosts their SaaS solution at gitingest.com and you can just do the exact same thing as uithub and just change hub to ingest in the url.

dennysubke:
Absolutely correct @sharknoon @nmfretz I just felt like being a smartass... 😜Thought that uithub.com was still unknown.

mamolas:
Copy buttons not working on latest Firefox Version 151.0.4 (64-bit) Snap Ubuntu.
Also it would be better to record the personal token for private repos at .env or something, I think that's the only reason to prefer the self hosted app

PR #2226, opened 2025-02-20
App Submission: Threema Web

dennysubke:
# App Submission

### Threema Web
...

### Icon
logo

...

### Gallery images
a

b

c

...

### I have tested my app on:
- [x] umbrelOS on Linux VM


dennysubke:
Threema is a privacy-focused messaging app that is end-to-end encrypted and widely used in German-speaking countries (Germany, Austria, and Switzerland). While its popularity is mainly limited to these regions, it remains a strong alternative to mainstream messengers.

I understand that Threema might not be relevant for all Umbrel users worldwide. If this PR is not a good fit for the Umbrel App Store, that's totally fine. I just wanted to contribute and see if there’s interest in adding it.

threema/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: threema_web_1
+      APP_PORT: 80
+
+  web:
+    image: dennysubke/threema-web:v2.6.1@sha256:963e3bf1268213308e5a1856a71306a9600a9f61faa04a4cbae2055be0318f8d
+    user: "1000:1000"
```
al-lac, line 11: @dennysubke the app does not seem to work when it is not running as root:
```
threema_web_1        | /usr/local/bin/entrypoint.sh: line 32: userconfig.overrides.js: Permission denied
threema_web_1        | Patching userconfig file...
```

threema/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
+manifestVersion: 1
+id: threema
+name: Threema Web
+tagline: An intuitive and secure web client for Threema users
+category: files
+version: "2.6.1"
+port: 88
```
al-lac, line 7: Please use a port higher than 1023. Ports below that are called system ports and are reserved for other purposes. See: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports

al-lac (changes requested):
Hey @dennysubke,

Thanks for another nice submission!

Only have some things to add. For me the app is not working when it is not running as root.

al-lac:
Hey @dennysubke, yeah give it a try to change the permissions and see if that works out.

dennysubke:
> Hey @dennysubke, yeah give it a try to change the permissions and see if that works out.

@al-lac 

I have now created the image several times with all possible “permission variables”. No chance. Always the same error message as soon as I set `user: 1000:1000`. 

I also watched the documentary from Threema: https://github.com/threema-ch/threema-web/blob/master/docs/docker.md 

Unfortunately I can't figure it out. It can't be my image either, as the original Threema image has exactly the same problem.

What next? Should I close the PR?

al-lac:
I would say we wait for what @nmfretz has to say about it. 

It running as root is not a deal breaker from my point of view. But lets see.

nmfretz:
Excellent, thanks @dennysubke for the submission, and @al-lac for reviewing.

Looking at the Dockerfile they're using nginx as a base image: https://github.com/threema-ch/threema-web/blob/master/Dockerfile, which will be running as root. 

> I hope the line in the Dockerfile RUN chown -R 1000:1000 /usr/share/nginx/html /usr/local/bin/entrypoint.sh sets the correct permissions for the copied files, ensuring the container can access them properly. I’ve adjusted the image accordingly and will push an update.

You'd need to create a non-root user/group in addition to this, but I'd honestly probably not dig too far into trying to change the user since the project's Dockerfile doesn't support it and we'd need to do pretty extensive testing.

dennysubke:
> Excellent, thanks @dennysubke for the submission, and @al-lac for reviewing.
> 
> Looking at the Dockerfile they're using nginx as a base image: https://github.com/threema-ch/threema-web/blob/master/Dockerfile, which will be running as root.
> 
> > I hope the line in the Dockerfile RUN chown -R 1000:1000 /usr/share/nginx/html /usr/local/bin/entrypoint.sh sets the correct permissions for the copied files, ensuring the container can access them properly. I’ve adjusted the image accordingly and will push an update.
> 
> You'd need to create a non-root user/group in addition to this, but I'd honestly probably not dig too far into trying to change the user since the project's Dockerfile doesn't support it and we'd need to do pretty extensive testing.

Hey @nmfretz!  Thanks for the feedback! I get the issue with the root user in the Dockerfile. Totally makes sense that changing it would be too much work. Do you think this will cause any issues with merging, or can we go ahead and merge the app? Appreciate the heads-up!

al-lac:
@dennysubke I think it will not block merging. After the gallery assets have been reviewed we can get it live once @nmfretz gives the final go.

nmfretz:
> Do you think this will cause any issues with merging, or can we go ahead and merge the app? Appreciate the heads-up!

We can go ahead with this.

Going live 🎉

image

dennysubke:
> > Do you think this will cause any issues with merging, or can we go ahead and merge the app? Appreciate the heads-up!
> 
> We can go ahead with this.
> 
> Going live 🎉
> 
> image
> 

@nmfretz 

Thank you so much for merging! I’m especially excited about it because I use the app every day. 👍🏻

al-lac:
@dennysubke can you update the container to version 2.6.2? :-)

dennysubke:
> @dennysubke can you update the container to version 2.6.2? :-)

@al-lac Gladly! 😎 

```threema-web2.6.2
dennysubke/threema-web:v2.6.2@sha256:3445db391dda9fa74f8cedab43edc20376c1f0dd510156dfd01c0c12d5f53fcf

PR #2243, opened 2025-02-21
Update wordpress to version 6.7.2

al-lac:
Tested the update on Umbrel Home and an Umbrel Dev instance.


nmfretz:
Reviewed and tested as well.

PR #2254, opened 2025-02-25
Update synapse to v1.125.0

al-lac:
Tested on Umbrel Dev instance.


nmfretz:
Reviewed and tested on my main synapse instance. Tested update, compatibility with element update https://github.com/getumbrel/umbrel-apps/pull/2257, and compatibility with latest element iOS app

PR #2262, opened 2025-02-26
Update affine to version 0.20.2

al-lac:
Tested fresh install on Umbrel Dev instance and update on Umbrel Home.


nmfretz:
Reviewed and tested

PR #2263, opened 2025-02-26
Update Lobe Chat to version 1.65.1

al-lac:
Tested fresh install on Umbrel Dev instance and update on Umbrel Home.


nmfretz:
Reviewed and tested with open ai key

PR #2264, opened 2025-02-26
Update firefox to version 135.0.1

al-lac:
Tested fresh install on Umbrel dev instance and update on Umbrel Home.


nmfretz:
Reviewed and tested

PR #2267, opened 2025-02-26
Update immich to version 1.127.0

al-lac:
Tested fresh install on dev instance and update on Umbrel Home.


nmfretz:
Reviewed and tested on main daily-use Immich instance. Tested update and compatibility with latest iOS 1.126.1 iOS client

PR #2268, opened 2025-02-26
Update pi-hole to version 2025.02.6

al-lac:
Tested fresh install on dev instance and update on Umbrel Home. 

Some more testing would be needed still.

### DIFF
--- a/pi-hole/docker-compose.yml
+++ b/pi-hole/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.7"
 
 services:
   server:
-    image: pihole/pihole:2024.07.0@sha256:0def896a596e8d45780b6359dbf82fc8c75ef05b97e095452e67a0a4ccc95377
+    image: pihole/pihole:2025.02.6@sha256:bcf79fe65348067d17b7b14fd7e2cb8177bbb4e972e1880a905169334e69c1a2
     # Pi-hole doesn't currently support running as non-root
     # https://github.com/pi-hole/docker-pi-hole/issues/685
     # user: "1000:1000"
--- a/pi-hole/docker-compose.yml
+++ b/pi-hole/docker-compose.yml
@@ -12,10 +12,9 @@ services:
       - ${APP_DATA_DIR}/data/pihole:/etc/pihole/
       - ${APP_DATA_DIR}/data/dnsmasq:/etc/dnsmasq.d/
     environment:
-      - VIRTUAL_HOST=${APP_DOMAIN}
-      - WEBPASSWORD=${APP_PASSWORD}
-      - WEB_PORT=8082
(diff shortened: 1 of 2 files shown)


nmfretz:
Nice work on the new env vars @al-lac. I have tested a fresh install and app update and both work well when pointing devices at pihole.

dirstel:
As I did not configure a backup DNS, I ran into some trouble here and the update did not complete. reinstalling using ip-adresses and a backup uf my pihole-data-dir worked. So better give a warning to the users in the releasenotes?!

al-lac:
@nmfretz should we add something here? Guess it always makes sense to have a backup DNS in case Pi-hole is down.

al-lac:
Yeah I see what you mean @dirstel, opened up a PR here #2274

nmfretz:
Thanks for bringing this up @dirstel. A warning is a really good idea. We can have @al-lac's warning in https://github.com/getumbrel/umbrel-apps/pull/2274 be a permanent one to the top of the release notes section so that users are warned for every update.

> The upgrade to v6 did not work without having a secondary DNS. The other updates before did work. so maybe this is just a one-time-thing

Out of curiosity, did you make any changes to your network between the previous pi-hole updates and this one that would account for the different behaviour you experienced? There should be no difference in behaviour between the updates from just a pi-hole app perspective, because the update process is always stop the app (so pihole is offline) > download any new images > start the app. So say you have your entire router pointed at pihole for DNS, with your umbrel pointed at the same router, then when pihole goes down your dns will go down as well unless there is a backup DNS set at the router level.

Also, what is this advanced setting feature set at for you? You can find it at Settings > Advanced settings

image

nmfretz:
Tweaked the warning to:

> ⚠️ If you have set up your router to use Pi-hole as its only DNS server, umbrelOS may be unable to resolve domain names during the update, preventing it from downloading the update files. To avoid issues, temporarily set a backup DNS server (e.g., 1.1.1.1) on your router before updating— or configure a permanent fallback DNS to prevent future issues.

dirstel:
> Out of curiosity, did you make any changes to your network between the previous pi-hole updates and this one that would account for the different behaviour you experienced? 
No, I did not make any changes. I'm using pihole for my entire networks, so even with umbrel. The DNS config is distributed via DHCP, so I had no trouble in changing this to a secondary DNS. My router itself does NOT use pihole, and my secondary DNS is my router (now :)). I had no problems in upgrading pihole so far.

> Also, what is this advanced setting feature set at for you? You can find it at Settings > Advanced settings
nope, this one isn't set.

nmfretz:
Thanks for the explanation @dirstel. Super helpful to understand use-cases and setups.

al-lac:
@kroese good find! Thanks for pointing that out here.

Will take a look at the PR, we could probably also combine this with the next update.

PR #2271, opened 2025-02-26
Remove docker compose options that were not compatible with older umbrel versions

### LINKED ISSUE #2228: RomM app don't install
i'm on Umbrel 0.5.4 and when i try to install RomM in the app store, it stays installing for long a long time. In the end, it won't being installed at all.

Something in the umbrel logs said that the docker-compose.yml is wrong, i don't remember exactly the error

al-lac:
This update removes newer docker compose options that prevent the app from being installed on older umbrel versions like 0.5.4. 

Fixes #2228

### DIFF
--- a/romm/docker-compose.yml
+++ b/romm/docker-compose.yml
@@ -20,7 +20,6 @@ services:
     depends_on:
       db:
         condition: service_healthy
-        restart: true
     environment:
       - DB_HOST=romm_db_1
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks @al-lac. I have tested fresh install and app update on umbrelOS 1.0+ and fresh install on umbrelOS pre-1.0.

PR #2272, opened 2025-02-26
Add .gitignore file

al-lac:
Adding a .gitignore file so development files are not commited by accident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,43 @@
+# Mac OS specific
+.DS_Store
+.AppleDouble
+.LSOverride
+._*
+.Spotlight-V100
+.Trashes
+
+# Docker
+.docker/
+docker-compose.override.yml
+
+# Environment variables and secrets
+.env
+.env.*
+!.env.example
+*.pem
+*.key
+
+# Editor specific files
+.idea/
+.vscode/
+*.swp
+*.swo
+*~
+
+# Logs
+logs/
+*.log
+
+# Build artifacts
+dist/
+build/
+out/
+
+# Dependency directories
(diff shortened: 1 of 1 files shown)


nmfretz:
Good idea @al-lac.  The macOS-specific section in particular would be helpful as we see those all the time in PRs.

We'll just need to be careful not to gitignore important files needed by some existing apps.

For example the env variables and secrets section clashes with some existing apps:

```bash
$ find . -name ".env" -o -name ".env.*" -o -name "*.pem" -o -name "*.key"
./blockstream-blind-oracle/data/server_private_key.key
./lndboss/.bosgui/.env
```

And same with the logs section:

```bash
$ find . -path "./logs" -o -name "*.log"
./lndg/data/logs/lndg-controller.log
```

And if we're too aggressive then I could also see us accidentally gitignoring files needed by future app submissions and causing headaches trying to troubleshoot what's going wrong (when we inevitably forget about the gitignore!)

What's your thoughts on pairing this down for now to only ignore OS- and editor-specific files to handle the most common issues we currently see with PRs?

al-lac:
Good points @nmfretz! Yeah I think just having it for OS- and editor-specific files would be enough for now!

nmfretz:
Excellent, thanks for the changes @al-lac! I have done the following to finalize:

- added `.cursorrules` to the .gitignore
- double checked that the .gitignore isn't ignoring currently tracked files that apps rely on:
  ```bash
  $ git ls-files -ci --exclude-standard
  ``` 
- tested the repo on both umbrelOS 0.5 and 1.0+

PR #2273, opened 2025-02-26
Wikijs repo fix

### LINKED ISSUE #2176: WikiJS Git Permissions Error
WikiJS will not sync to Git by default due to the container not allowing write permissions to the requisite directory. Adding a volume in the docker-compose.yml file fixes this issue. 

- ${APP_DATA_DIR}/repo:/app/wiki/data/repo

al-lac:
This update adds the repo folder to wikijs to fix permissions errors when using the git sync feature.

Fixes #2176.

Maybe you can test and take a look at the new hook @nmfretz.

### DIFF
--- a/wikijs/docker-compose.yml
+++ b/wikijs/docker-compose.yml
@@ -8,11 +8,10 @@ services:
       PROXY_AUTH_ADD: "false"
   
   server:
+    user: "1000:1000"
     image: linuxserver/wikijs:2.5.306@sha256:4868253594b2511b7c6054aa83695332bf3e77257b667c84e276e62668e36b62
     restart: on-failure
-    environment:
-      - PUID=1000
-      - PGID=1000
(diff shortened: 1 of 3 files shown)


nmfretz:
Reviewed and tested. Great work remembering to update the umbrel.yaml manifest version to 1.1 with the inclusion of the hook. And also good find on the container not running as user 1000:1000 even though it is a linuxserver image which usually abide by the PUID/PGID environment variables.

I have tweaked the hook very slightly to take into account that all previous installs of wikijs on umbrel will already have the `data/` and `config/` bind mounts created. I also updated the comments slightly to remind us why we do specific things.

wikijs now runs as `1000:1000`:
```
$ sudo docker exec wikijs_server_1 id
uid=1000 gid=1000(users) groups=1000(users)
```

PR #2274, opened 2025-02-27
Add upgrading warning to pi-hole release notes

al-lac:
Making sure that users don't break their internet connection if they have setup pi-hole as their DNS server.

Please review @nmfretz.


nmfretz:
Excellent, thanks @al-lac! I have added some additional wording.

PR #2279, opened 2025-03-04
Update maybe to version v0.4.3

al-lac:
Tested the update on a dev instance and on Umbrel Home.


nmfretz:
Reviewed and tested (including app update).

PR #2280, opened 2025-03-04
Update immich to version v1.128.0

al-lac:
Tested update on dev instance and Umbrel Home.


nmfretz:
Bumped to 1.129.0 and added a release note about the memories bug fix. Tested, including latest iOS client.

PR #2288, opened 2025-03-05
fix: Pihole over IPv6

kroese:
Previously the web GUI was reachable over IPv4 and IPv6.

But recently someone updated this app to v6 and forgot to include the IPv6 port. So this corrects that issue.


nmfretz:
Thanks for this @kroese. Tested and working.

I don't have write permissions to your fork, so I will merge this as-is and then push another commit to master that modifies the manifest file so that existing installs can update.

PR #2291, opened 2025-03-06
Update Open-webUI to v0.5.20

beecho01:
Tested by updating app-store files on Umbrel OS 1.4, and causing update to trigger:

  - `open-webui/docker-compose.yml`
  - `open-webui/umbrel-app.yml`

Update was successful without issues and existing data persists.


nmfretz:
Thanks @beecho01! Tested and working well. I added in release notes from 0.5.19 as well since we missed that update.

PR #2303, opened 2025-03-10
Update home assistant 2025.3.1

al-lac:
Tested fresh install on Umbrel Dev instance and update on Umbrel Home. All seemed fine after the update.


nmfretz:
Reviewed and tested with existing integrations that I use

PR #2306, opened 2025-03-11
Alby Hub v1.15.0

bumi:
Updates to Alby Hub v1.15.0

Which contains an important bug fix for a relay connection issue. 
In some cases a relay connection fails for some app connections which is then not recovered which causes the app to fail.


nmfretz:
Excellent, thanks for the update @bumi! Tested and working well. 

image

PR #2309, opened 2025-03-11
Update Jellyseer to version 2.5.0

johnpc:


nmfretz:
Thanks for this update @johnpc 🤝. Tested fresh install and app update.

PR #2327, opened 2025-03-12
Update electrs to v0.10.9

nmfretz:
>   Highlights:
      - Separates blocks reading & index writing into scoped threads
      - Sets HTTP Content-Type header for Prometheus response
      - Doesn't deserialize transactions if not needed
      - Allows concurrent DB background operations for SSDs


nmfretz:
- Tested on arm64 and x86
- Tested fresh install and app update

PR #2353, opened 2025-03-14
Fix Lightning Node compatibility with testnet3

Impa10r:
Bitcoin Core now exports `APP_BITCOIN_NETWORK` as "testnet3", but LND still works with `/home/umbrel/umbrel/app-data/lightning/data/lnd/data/chain/bitcoin/testnet` folder.

For Lightning Node app to still be compatible with testnet3, it is necessary to export another variable `APP_LIGHTNING_NODE_CHAIN_DIR` that is equal "testnet" when `APP_BITCOIN_NETWORK` == "testnet3", and use it in macaroon and backup path instead of `APP_BITCOIN_NETWORK`:

```
MACAROON_DIR: "/data/.lnd/data/chain/bitcoin/${APP_LIGHTNING_NODE_CHAIN_DIR}/"
CHANNEL_BACKUP_FILE: "/data/.lnd/data/chain/bitcoin/${APP_LIGHTNING_NODE_CHAIN_DIR}/channel.backup"
```


jkuchar:
testnet3 is deprecated, testnet4 should be used. ref: https://mempool.space/testnet

Impa10r:
> testnet3 is deprecated, testnet4 should be used. ref: https://mempool.space/testnet

LND and CLN do not support testnet4 yet. I use them with testnet3 to test things, it still works for my purposes.

nmfretz:
Thanks @Impa10r. Taking a look at this now.  Looks like LND v0.19.0 (currently an RC)[ will have testnet4 support](https://github.com/lightningnetwork/lnd/blob/master/docs/release-notes/release-notes-0.19.0.md), so we can look at adding logic to exports.sh for that when it is released.

nmfretz:
@Impa10r I was just toying with the idea of waiting until LND 0.19.0-beta is released with testnet4 compatibility so that this problem would go away, but it looks like we're out of luck because LND's `/data/chain/bitcoin/<network>` directory name for testnet3 will still differ from bitcoind's network name:

https://github.com/guggero/lnd/blob/d757bb51ee992976d3b4dc625ed46d0751909069/lncfg/config.go#L108-L124
```go
// NormalizeNetwork returns the common name of a network type used to create
// file paths. This allows differently versioned networks to use the same path.
func NormalizeNetwork(network string) string {
	// The 4th testnet isn't the "default" yet, so we'll want to explicitly
	// point that to a "testnet4" directory.
	if network == "testnet4" {
		return network
	}


	// We want to collapse "testnet3" and "testnet" to the same "testnet"
	// directory.
	if strings.HasPrefix(network, "testnet") {
		return "testnet"
	}


	return network
}
```

^ the code above is showing that the testnet3 dir will remain as `testnet` in 0.19.0-beta

That being said, I think we can actually just solve this upstream in the Bitcoin Node app very simply and it will ensure that all dependent apps work correctly. We can just export `APP_BITCOIN_NETWORK` as `testnet` instead of `testnet3` and everything will just work without needing any of the changes in this PR.

I will test and get back to you.

nmfretz:
@Impa10r can you please test this PR https://github.com/getumbrel/umbrel-apps/pull/2596?

Impa10r:
This PR https://github.com/getumbrel/umbrel-apps/pull/2596 solves it better, closing.

PR #2354, opened 2025-03-15
Mount Umbrel's home directory in the docker container in Portainer.

ArcaEge:
The `${UMBREL_ROOT}/home` directory is now mounted at `/umbrel` in the container.

### DIFF
--- a/portainer/docker-compose.yml
+++ b/portainer/docker-compose.yml
@@ -25,6 +25,7 @@ services:
     volumes:
       - ${APP_DATA_DIR}/entrypoint.sh:/entrypoint.sh
       - ${APP_DATA_DIR}/data/docker:/data
+      - ${UMBREL_ROOT}/home:/umbrel
 
   portainer:
     image: portainer/portainer-ce:2.27.1@sha256:99c3047d44991af08f2a34df16e69ae2654bee43444b2e9857aa6b5864c4f602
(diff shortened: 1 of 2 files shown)


al-lac:
Hey @ArcaEge,

thanks for the contribution!

It is on purpose though, that the home directory is not mounted for portainer. We do not want portainer to mess with the data of other umbrel apps.

FYI @nmfretz.

ArcaEge:
Hi @al-lac,
I should've explained this a little bit better, but I think this might be beneficial when you need containers in Portainer to interact with Umbrel apps, especially if the Umbrel app isn't supported yet - for example, running Calibre in Portainer and using Calibre-Web's book library as Calibre's own library so that they're synchronised (the SMB share workaround doesn't work in this case due to Calibre). This could also be useful for backups too (Duplicati can't access the data of other apps right now).
Also, I don't think Portainer messing with the data of other apps is a particularly big issue as the target users of Portainer are developers anyway.
Thanks

al-lac:
Hey @ArcaEge,

Thanks for the detailed explanation.

Even though Portainer may be more suitable for developers, giving the app too much access can cause a lot of other issues. And every app should run in its own sandbox if possible.

I would recommend you change it yourself in the docker-compose file, or you open up a community app store and use that for your own version of portainer.

PR #2380, opened 2025-03-18
Update photoprism to version 250228-43447fa38

al-lac:
Tested update on Umbrel Home.


nmfretz:
- moved user action to top of release notes so it's super visible
- Tested fresh install and app update

PR #2381, opened 2025-03-19
update to datum `v0.3.1beta`

Retropex:
I'm hitting a problem with this release its no longer possible to use `"pool_address": "enter your bitcoin address"` we need to put a valid bitcoin address, this can be now done in the UI but there is a permission error (I will fix that).

The second thing is the the autoconfig isn't working when upgrading from v0.2.2

### DIFF
--- a/datum/data/settings/datum_gateway_config.json
+++ b/datum/data/settings/datum_gateway_config.json
@@ -2,40 +2,28 @@
   "bitcoind": {
     "rpcuser": "auto-configure-on-install",
     "rpcpassword": "auto-configure-on-install",
-    "rpcurl": "auto-configure-on-install",
-    "work_update_seconds": 40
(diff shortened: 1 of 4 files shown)


Retropex:
hum, I haven't been able to fix the permission problem. is the `pre-start` script executed at each start or just at the first install?

Retropex:
Fixed the permissions problem.

Retropex:
All problem fixed, and I have added some migration to be sure address don't cause datum to stop.

al-lac:
@Retropex as of now there is no way to influence the length or other parameters of the `deterministicPassword`.

al-lac:
@Retropex is this update ready to be merged or are there still some changes needed?

Retropex:
yes, just pushed a small change and it's now ready.

al-lac (approved):

PR #2441, opened 2025-03-28
Update GitLab to version 17.10.1

al-lac:
Tested update on Umbrel Home.


al-lac:
Tested update on Umbrel Home and fresh install on dev instance.

PR #2444, opened 2025-03-28
Update seafile to version 12.0.11

al-lac:
Tested update on Umbrel Home.

Maybe you can also take a look @nmfretz as this is quite a big update.

### DIFF
--- a/seafile/docker-compose.yml
+++ b/seafile/docker-compose.yml
@@ -22,13 +22,14 @@ services:
     entrypoint: memcached -m 256
 
   seafile:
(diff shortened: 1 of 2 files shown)


al-lac:
Tested update on Umbrel Home and fresh install on dev instance.

hannpet:
Bug/ Big omission:
This new version requires Seadoc in the compose-file. If Seadoc is missing, the Wiki-function of Seafile is not working (Internal Server Error when trying to add Wiki pages)

PR #2465, opened 2025-04-01
Update btcpay-server to version 2.0.8

al-lac:
Tested on Umbrel Home, maybe you can also take a look @nmfretz @sharknoon.


sharknoon:
I have installed the live version of `btcpay-server`, set it up and then updated to app to this new version (2.0.8). It has kept the data and it works fine for me. But I haven't intensively tested the app because I am not that familiar with crypto payment providers.

nmfretz:
Reviewed and tested. Working great @al-lac, we're good to go live.

PR #1922, opened 2024-12-12
App Submission: Ente Photos

starfreck:
# App Submission

### Ente Photos

ente-photos-icon-white
ente-photos-icon-transparent
ente-photos-icon-green

This PR is for #1064 

### Gallery images
https://community.ente.io/media-kit/

### I have tested my app on:
- [ ] umbrelOS on a Raspberry Pi
- [ ] umbrelOS on an Umbrel Home
- [x] umbrelOS on Linux VM

Issue #1064, opened 2024-04-24
Ente Auth and Photos


Lenni-builder:
It would be really nice to have [Ente](https://ente.io) Photos and especially Auth available.

highghlow:
Well, they don't seem to have a Dockerfile in their repo, but if they don't have permission issues it should be pretty easy to make

Lenni-builder:
Interesting. You could maybe contact them if you have questions/concerns

FlyinPancake:
I looked trough the repo and man it's a s***show of tags... the last tagged version seems to be ente auth(?), there is no version tags for the containers they build, and the web frontend has no packaging whatsoever oh and it's split up into multiple apps so packaging would be a great challenge. the webapp from the main branch crashes immediately too... let's just wait until someone cleans up that mess so umbrel can have a complete version of the app. For now it seems to properly host ente you would need a broader understanding of the codebase

Lenni-builder:
I think they have a guide on how to selfhost it somehwere, that might help.

FlyinPancake:
Yeah the backend can be set up quite easily, however the web interfaces are not packaged.

Another issue is the versioning which is unclear to me so I don't know which versions I should package if I went with creating images myself.

Lenni-builder:
They could probably answer most of your questions if you contact them for example on Discord.

starfreck:
Maybe this would be helpful
https://youtu.be/Gu-zAxAOn1E?si=CBw31P6jSyg12Sv5

shrimpwtf:
bump, this would be awesome

shrimpwtf:
> Since there haven't been any updates in the past nine months, I decided to work on this for **Ente Photos**. If anyone is interested in helping with the testing or keeping track of progress, please check https://github.com/starfreck/umbrel-apps/tree/ente-photos.
> 
> I have tested it in Docker, and it seems to be functioning well, except for the issue with **admin access privileges**. Without these privileges, it only provides **5 GB** of storage. I am currently working on resolving this issue.

amazing work trying to get this to function! ill see if i can find any solutions to this

starfreck:
The PR has been created for `Ente Photos` #1922 I have wrote instructions on how to setup. Linting checks are failing at the moment but I will try to fix them ASAP.

PR #2482, opened 2025-04-02
App Submission: Ente Photos

starfreck:
# App Submission

### Ente Photos

ente-photos-icon-white
ente-photos-icon-transparent
ente-photos-icon-green

This PR is for #1064 

### Gallery images
https://community.ente.io/media-kit/

### I have tested my app on:
- [x] umbrelOS on Linux VM

Previous PR #1922, which I closed.

### DIFF
--- a/ente-photos/README.md
+++ b/ente-photos/README.md
@@ -1,0 +1,340 @@
+# Ente Photos Self-Hosting Guide
+
+## Table of Contents
+1. [Introduction](#introduction)
+2. [Initial Setup After Installation](#initial-setup-after-installation)
+3. [How to Use Predefined OTPs](#how-to-use-predefined-otps)
+4. [How to Find Verification Codes in Logs](#how-to-find-verification-codes-in-logs)
+5. [How to Connect Desktop and Mobile Apps](#how-to-connect-desktop-and-mobile-apps)
+6. [How to Update Subscription to Unlimited](#how-to-update-subscription-to-unlimited)
+7. [How to Configure Custom Email Senders](#how-to-configure-custom-email-senders)
+8. [Additional Resources](#additional-resources)
+
+## Introduction
+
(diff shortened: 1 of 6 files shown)


ente-photos/README.md
starfreck, line 1: @al-lac Thank you for reviewing this PR. I've documented the entire setup process step-by-step in this file to make it as clear as possible. I acknowledge there are two aspects that remain somewhat cumbersome:

1. Finding verification codes in logs (though users can set up predefined OTPs as documented)
2. Updating subscription to unlimited storage

We do not know if Ente will ever fix these things or not, so would it be acceptable to release this with a clear warning that some manual configuration is required on the user's end? The documentation provides detailed instructions for each step, which should help users navigate through the setup process.

starfreck:
Hey @al-lac,

Thank you for your feedback. I'll keep this PR open as a reference in case Ente enhances their process in the future, at which point I'll update this accordingly.

nmfretz:
Thanks @starfreck, we really appreciate your work on this! Hopefully Ente can simplify 🤞

nmfretz:
Hey @starfreck, sorry this has been sitting open for so long. Thanks again for all the work you put into this.

I took another look, and there is actually some good news: Ente's self-hosting setup has improved quite a bit since this was opened, and the current upstream images are multi-arch now. So Ente may be a lot more realistic for the App Store than it was before.

That said, this PR is pretty stale against current Ente, and I think the right path now is a fresh package based on the current upstream setup rather than trying to keep iterating on this one. There are still a few Umbrel-specific things that would need to be handled cleanly, especially storage endpoints/CORS, generated secrets, first-run config, and how mobile/desktop clients connect.

I'm going to close this one out for now, but Ente is definitely worth revisiting with a fresh PR. Really appreciate you getting this started and documenting the rough edges here.

PR #2493, opened 2025-04-03
Update fulcrum to version 1.12.0

al-lac:
Update works, but could not test a lot so far as it still seems to be syncing up.

Maybe @nmfretz or @sharknoon can test as well.


nmfretz:
Tested fresh install and app update @al-lac . Both working great!

nmfretz (approved):

PR #2522, opened 2025-04-08
Update Monero daemon to v0.18.4.0

colder-is-better:
This is the v0.18.4.0 release of the Monero software. It is a recommended release that fixes multiple daemon-related network vulnerabilities.

### DIFF
--- a/monero/docker-compose.yml
+++ b/monero/docker-compose.yml
@@ -43,7 +43,7 @@ services:
     restart: unless-stopped
     stop_grace_period: 15m
     command: "${APP_MONERO_COMMAND}"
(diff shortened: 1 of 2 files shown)


al-lac:
Tested, everything working fine on Umbrel Home. @nmfretz should we change the restart policy on the monerod container?

nmfretz:
Thanks for another monerod update @colder-is-better!

 > Tested, everything working fine on Umbrel Home. @nmfretz should we change the restart policy on the monerod container?
 
 Excellent. Then this is good to go @al-lac. The reason that the `monerod` container is set with a restart policy of `unless-stopped` is so that when the user restarts the daemon from the app's UI (i.e., when changing config through advanced settings), Docker will bring the `monerod` container back up. With a `on-failure` policy, Docker would only bring up the monerod container again if monerod exited with a non-zero status.

al-lac:
@nmfretz added a comment so this does not come up again on future updates.

PR #2524, opened 2025-04-08
App Submission: Neko

dennysubke:
# App Submission

### Neko
...

### Icon
logo

...

### Gallery images
a

b

c

### I have tested my app on:
- [X] umbrelOS on Linux VM


dennysubke:
Neko is a self-hosted virtual browser running in Docker containers, utilizing WebRTC to deliver low-latency, real-time streaming directly to the user's browser. It allows users to access a fully functional web browser in a secure, isolated environment on any device, without exposing their own system or compromising privacy.

neko/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+      PROXY_AUTH_ADD: "false"
+
+  web:
+    image: ghcr.io/m1k1o/neko/firefox:3.0.3@sha256:9b5d3e04af1024a44c2aa8938c78f155d64b0ab9459feb424c34b461155a3f98
+    environment:
+      PUID: 1000
+      PGID: 1000
+      NEKO_DESKTOP_SCREEN: 1920x1080@30
+      NEKO_MEMBER_MULTIUSER_USER_PASSWORD: neko
+      NEKO_MEMBER_MULTIUSER_ADMIN_PASSWORD: umbrel
```
al-lac, line 17: Maybe use the deterministic password here.
```suggestion
      NEKO_MEMBER_MULTIUSER_ADMIN_PASSWORD: $APP_PASSWORD

```

neko/umbrel-app.yml
```diff
@@ -0,0 +1,37 @@
...
+support: https://github.com/m1k1o/neko/issues
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
+releaseNotes: ""
+dependencies: []
+path: ""
+defaultUsername: ""
+defaultPassword: "umbrel"
```
al-lac, line 37: Use of the deterministic password.
```suggestion
deterministicPassword: true
```

al-lac:
Tested the app again, works fine. Using it via tailscale seems to fail however, it is just not connecting after logging in.

One more thing, could we have an issue with a port clash @nmfretz @sharknoon? 

Neko is using the range 52000-52100 (udp) and snort is using 52027 (tcp).

nmfretz:
Nice work guys!

> One more thing, could we have an issue with a port clash @nmfretz @sharknoon?
>
> Neko is using the range 52000-52100 (udp) and snort is using 52027 (tcp).

Nice work catching and thinking about this. In this case, there will be no clash on the host because they are using separate protocols (udp vs tcp).

al-lac:
@nmfretz thanks! Then we can go live with this one @dennysubke  🚀
CleanShot 2025-04-30 at 11 23 25

al-lac (approved):

PR #2528, opened 2025-04-10
Remove libretranslate from 1.x app store

al-lac:
This PR removes the LibreTranslate from the app store for now over trademark issues.


arguser:
Hopefully this an UI bug, since this merge I'm seeing this on my dashboard. I'm using Libre Translate via API (disabled UI) and I want to keep using it.

image

al-lac:
Hey @arguser, a reboot should probably fix this.

PR #2553, opened 2025-04-11
Remove app-proxy from umami

al-lac:
Remove the app-proxy of umami due to the issues described in #2390

Also closes #2390.

Please take a look if it is ok like this @nmfretz

### DIFF
--- a/umami/docker-compose.yml
+++ b/umami/docker-compose.yml
@@ -1,17 +1,13 @@
 version: "3.7"
 
 services:
-  app_proxy:
-    environment:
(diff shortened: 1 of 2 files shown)


sharknoon:
Good work @al-lac. When I was creating the umami app, I hadn't in mind that users want to see the original IP address. Thanks for fixing this.

This PR looks good to me

al-lac:
Tested it again on my Raspberry Pi and Umbrel Home.

Updated the release notes to better reflect why the app proxy was removed.

PR #2554, opened 2025-04-12
Disable n8n telemetry, fix config permissions, activate task runners

skibitsky:
This PR disables telemetry in the n8n app

https://docs.n8n.io/privacy-security/privacy/#opting-out-of-telemetry


al-lac:
Hey @skibitsky!

Thanks for bringing this up. When testing the app I also saw some issues in the logs regarding the config file and the task runners. Added two more options to also fix those, now there are no more warnings to be found in the logs.

PR #2555, opened 2025-04-13
App Submission: Ungoogled Chromium

dennysubke:
# App Submission

### Ungoogled Chromium
...

### Icon
logo

...

### I have tested my app on:
- [X] umbrelOS on Linux VM


dennysubke:
Unfortunately I noticed too late that the image used in this PR is only available for amd64 and not as a multiarch image. Since Umbrel also runs on ARM devices like the Raspberry Pi this would cause issues. I am closing this PR for now and might open a new one later with a suitable image. 💪

PR #2591, opened 2025-04-18
Update bitcoin-knots description spacing

copy2018:
Looking at https://apps.umbrel.com/app/bitcoin-knots it looks like double spacing is needed for correct paragraph spacing.


al-lac:
@copy2018 good catch, thanks!

PR #2595, opened 2025-04-19
Update shopstr to version 0.1.2

calvadev:
- Minor bug fixes and an update to the latest version of Shopstr


calvadev:
@al-lac got some bug fixes pushed and also bumped up the digest hash for the image. Realized I hadn't updated the hash since I first started testing so the current app is a bit outdated, not up to even the previous version.

al-lac:
Tested update on Umbrel Home, everything working nicely.

PR #2625, opened 2025-04-22
Add note on OAuth setup for email and calendar in Twenty (closes #2594)

### LINKED ISSUE #2594: Twenty email and calendar providers list is empty
As you can see in the below screenshot, the email providers list is empty and no pathway to enter one in custom. I am wondering if there is some more technical path I need to follow to integrate my email like setting the auth details in env vars or something.

Note: I am on Umbrel version 1.3 so a tad out of date.
And I am on the latest available version of Twenty in the Umbrel app store (0.50.0)

Image

dennysubke:
This adds a clarification to the app description of the Twenty app. Since email and calendar providers (like Gmail or Outlook) are not preconfigured and require custom OAuth credentials, users must manually add these credentials as environment variables in the `docker-compose.yml` file.

The added note explains this behavior and links to the official setup guide:
https://twenty.com/developers/section/self-hosting/setup#for-gmail-and-google-calendar

This should help users understand why the provider list may appear empty and how to proceed with configuration.
This resolves issue #2594.


nmfretz:
> One thing that should be noted though is that those settings are lost on an update.

Ah ya, the classic ha! Would be so nice if these were configurable via the app's UI. But alas, we're stuck with workarounds for now until we offer a way to configure optional env vars for apps.

There actually is a way we can get around this to get these env vars to persist across updates. We could do something like this if you guys think having the email/calendar provider functionality is super important (I haven't actually tried the app to know):

We could include a `.env` file at the same level as the `docker-compose.yml` file. Then users can just add their own env vars to that file. We'd then set this in the `server` services block in the compose file:

```yaml
env_file:
  - .env
```

This way the `server` container will bring in anything that users place in there.

If Twenty gracefully handles blank strings for env vars then we could even pre-populate the .env with the variable names like:

```.env
AUTH_GOOGLE_CLIENT_ID=
AUTH_GOOGLE_CLIENT_SECRET=
etc...
```

to make it easy for users to just fill in with their values.

What do you guys think?

al-lac:
@nmfretz good idea! I think the email/calendar integration is kind of important for the app, but also did not use it too much so far.

Could for sure make sense to implement it like this, guess we can always switch over once we have a way for the users to configure env vars via the UI.

@dennysubke what do you think about this?

dennysubke:
Hey @al-lac @nmfretz 

Sounds good, that approach makes sense. Including a .env file at the same level as the docker-compose.yml would definitely make it easier for users to customize without touching the compose file.

I'm on board with that. 👍

al-lac:
Hey @dennysubke @nmfretz,

I just tested this out, and it works fine. Still quite a lot of effort as you need to create a lot in either the google or microsoft cloud console. Also we would need to put the `SERVER_URL` in the `.env` file as all of this only works if you provide a valid redirect URL to the provider (no .local domains). So I had to to expose the app via the nginx proxy manager and change the `SERVER_URL` to an external domain to get this to work.

So, should we still do it? I mean it would at least give users the option to do if they would want to.

dennysubke:
Thanks for testing it out, @al-lac! I agree, it’s quite a bit of effort, but having the option to configure email and calendar providers could still be valuable for those who are willing to go through the setup.

Regarding the `SERVER_URL`, we could mention in the description that a public domain is required for OAuth to work properly and that .local domains won’t suffice. That way, users are aware of the limitation upfront.

nmfretz:
> I just tested this out, and it works fine. Still quite a lot of effort as you need to create a lot in either the google or microsoft cloud console. Also we would need to put the SERVER_URL in the .env file as all of this only works if you provide a valid redirect URL to the provider (no .local domains). So I had to to expose the app via the nginx proxy manager and change the SERVER_URL to an external domain to get this to work.

Damn, thanks for diving in to this @al-lac, great work. But man what a pain that is! I really don't think we can expect the average user to set this up (not average self-hoster, but average person). Just think about the level of background knowledge you need to understand what to do.

> So, should we still do it? I mean it would at least give users the option to do if they would want to.

But I think if we can gracefully add the functionality without making app updates or upgrades needlessly difficult then we should do it in this case.

How about we set it up so that we ship an .env like this:

```
# Uncomment these lines and add your custom values to configure Twenty
# SERVER_URL=
# AUTH_GOOGLE_CLIENT_ID=
# AUTH_GOOGLE_CLIENT_SECRET=
```

Or potentially we just don't include the .env and just tell Docker not to fail if it doesn't exist like this in the `server` container's service block:

```
env_file:
  - path: ${APP_DATA_DIR}/.env
    required: false
 ```
 
 And then all we do in the `environment` section of the `server` is change `SERVER_URL` to:
 ```
 SERVER_URL: ${SERVER_URL:-http://${DEVICE_DOMAIN_NAME}:2020}
 ```
 
 So that it defaults to `http://${DEVICE_DOMAIN_NAME}:2020}`, but can be overridden by the .env

al-lac:
Hey @dennysubke, together with @nmfretz I worked on a different way to tackle the current limitation.

We added an environment file where users can set a custom server url and required values for the google / microsoft integration.

Please test if possible and provide some feedback.

@sharknoon maybe you also have time to test this.

dennysubke:
@al-lac I’ll test it out over the next few days and see how well the .env file workarounds work. Unfortunately, I don’t have a Google or Microsoft account myself, so I’ll need to create those first. 🙄

In the meantime, I’ve opened an [issue](https://github.com/twentyhq/twenty/issues/12529) on their page.

dennysubke:
@al-lac 

Quick update: 🎉

The devs just added a way to configure all those environment variables directly through the admin panel in the app. So no need to mess with .env anymore.

https://github.com/twentyhq/twenty/issues/12529

al-lac:
Hey @dennysubke, thanks for opening the ticket and looking into this!

I now updated twenty to the latest version and added a modified version of your guide to the description. 🚀

PR #2633, opened 2025-04-24
App Submission: Domain Locker

dennysubke:
# App Submission

### Domain Locker
...

### Icon
logo

...

### I have tested my app on:
- [X] umbrelOS on Linux VM


domain-locker/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+  app:
+    image: lissy93/domain-locker:0.0.6@sha256:bd8b494b2a8b02e9fb612859f64c336e6eb1b21745199253b130705bd4ac8853
+    user: "1000:1000"
+    restart: on-failure
+    environment:
+      DL_ENV_TYPE: ${DL_ENV_TYPE:-selfHosted}
+      DL_PG_HOST: ${DL_PG_HOST:-postgres}
+      DL_PG_PORT: ${DL_PG_PORT:-5432}
+      DL_PG_USER: ${DL_PG_USER:-postgres}
+      DL_PG_PASSWORD: ${DL_PG_PASSWORD:-changeme2420}
```
al-lac, line 18: I think you can remove the variable names here as those will not be used anyways. And maybe change the value to something else besides `changeme`.
```suggestion
      DL_PG_PASSWORD: domain-locker

```

domain-locker/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+      DL_PG_NAME: ${DL_PG_NAME:-domain_locker}
+    depends_on:
+      - postgres
+
+  postgres:
+    image: postgres:15-alpine@sha256:ef9d1517df69c4d27dbb9ddcec14f431a2442628603f4e9daa429b92ae6c3cd1
```
al-lac, line 24: Maybe we could try a more up-to-date version of postgres here like 17.4, just so we don't have to update this soonish.

domain-locker/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    user: "1000:1000"
+    restart: on-failure
+    environment:
+      POSTGRES_DB: ${DL_PG_NAME:-domain_locker}
+      POSTGRES_USER: ${DL_PG_USER:-postgres}
+      POSTGRES_PASSWORD: ${DL_PG_PASSWORD:-changeme2420}
```
al-lac, line 30: Would also remove the variable names here and adapt the password.

al-lac (changes requested):
Hey @dennysubke!

Another great app submission, you are on a roll! 🚀 

Only have some minor points, tested the app and it works. Sometimes I had to reload pages for them to show something, this could also be due to the app still being in pretty fresh and being more like a beta version.

dennysubke:
Moin @al-lac! 

Thanks again for the feedback! I’ve addressed all your suggestions. Everything should be good to go now. 😊 

Let me know if there’s anything else!

al-lac (approved):

al-lac:
Thanks again for the submission @dennysubke! Going live now! 🚀 

image

PR #2631, opened 2025-04-23
Update Urbit to 3.2

yapishu:
Updates urbit app with the [vere 3.2](https://github.com/urbit/vere/releases/tag/vere-v3.2) runtime

PR #2636, opened 2025-04-25
Fix Urbit 3.2 app release

yapishu:
- Fixed [broken download URL](https://github.com/mopfel-winrux/urbit-umbrel/pull/33/files) for new vere version; didn't notice this in [previous PR](https://github.com/getumbrel/umbrel-apps/pull/2631) because the old binary would remain in place


al-lac (approved):
Tested on Umbrel Home.

PR #2641, opened 2025-04-27
update bitwatch to 1.5

zapomatic:
This release adds support for extended public key address watching, cleans up UI and fixes several minor bugs.

- Add support for xpub/ypub/zpub addresses as a sub-collection
- extended keys can be configured with an initial number of addresses to watch, a gap limit of empty addresses to keep updating on the list, and a skip count to jump over already spent addresses.
- adds better public mempool.space API rate limiting and a new apiDelay config for being nicer to the API
- new defaults in config for public mempool.space API usage to better align with rate limits
- UI impprovements to unify styles, cleanup modules and make buttons cleaner

New gallery images (if you want to update them):

### DIFF
--- a/bitwatch/docker-compose.yml
+++ b/bitwatch/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3117
 
   web:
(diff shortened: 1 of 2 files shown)


al-lac:
Hey @zapomatic,

thanks for the update! 🚀 

We are quite full with work, so that does not leave a lot of time for updating the assets, will keep it for when there are bigger changes to the UI.

Tested successfully on my Umbrel dev instance.

PR #2662, opened 2025-04-28
Urbit v3.2-2 -- use permanent storage for urbit data instead of app data directory

yapishu:
[This PR](https://github.com/mopfel-winrux/urbit-umbrel/pull/40/files) modifies the Urbit app to use the `data/storage` directory to store piers so that they aren't lost if the ship is uninstalled. It also migrates existing piers, and fixes a minor UI bug that preventing logs from displaying.


yapishu:
Originally this was just a feature update to move app data to storage but there was an urbit release this morning so I threw that in here

al-lac:
Hey @yapishu, thanks for the update! 🎉 

It is a good idea to store the data so it does not deleted on uninstall. But most users expect the data to be cleaned up anyways, as this is the default for all non-media apps now. So please remove the new volume mount for now.

Best to inform the users via the description about taking a backup before uninstalling for now.

yapishu:
@al-lac This isn't backing up the pier data, it's migrating it from app data to storage to avoid accidental deletion. A bit of background about why accidentally deleting is of particular concern:

An urbit instance is a pure function of its event history; everything that has happened to it since it was first turned on, like packets and software updates and keystrokes, gets added to an event log. The p2p communication channels are also stateful, and each channel with a peer has a counter. This means that if you try to boot a non-current backup copy, or you lose your copy of your instance and boot a new one with the same keys, it will be unable to communicate with outside instances since the counters will not align; worse, it could also create 'double booted' instances, which may initiate communications with new peers and start incrementing message counters with them, leaving the original copy unable to communicate with them. This can be remedied by an external PKI public key reset, which is an out-of-band way of announcing a reset to peers, but this forces you to reset your entire instance and generally lose previous data.

For this reason, in urbit land we are very cautious about avoiding accidental data loss. What I changed in this update was not backing up the instance to storage, but migrating it (it attempts to pick it up and move it, though abandons if it fails). Since PKI resets can carry a monetary cost (if they are L1 identities) and are almost always associated with data loss, I thought a better default was persisting the instance's data by default so that a user does not accidentally delete it.

I can understand if you'd prefer to not make this default behavior; would changing it to a configurable option be acceptable?

arguser:
Hello @al-lac, I was in conversation with @yapishu about this topic.

I'm a long time Umbrel and Urbit user and I must say **it's not obvious** that if you uninstall the Umbrel application you would also lose your Urbit data.

You could think of this Umbrel app as a Launcher. Data created by launching Urbit (via Key) is not `app-data` is user data. If I'm uploading my Urbit Instance (via Zip) to run in this app it's not `app-data` what I'm uploading, it's mine and wouldn't think it will be deleted when uninstalling the Umbrel app.

In short: Urbit data is User data.

al-lac:
Hey @yapishu and @arguser, I totally get what you are trying to achieve here, and it makes sense. Guess it could be clarified a bit more that uninstalling apps also deletes user data. This applies to most apps in the store right now, may it be databases or image uploaders.

Lets see what @nmfretz has to say to this topic.

arguser:
> Hey @yapishu and @arguser, I totally get what you are trying to achieve here, and it makes sense. Guess it could be clarified a bit more that uninstalling apps also deletes user data. This applies to most apps in the store right now, may it be databases or image uploaders.
> 
> Lets see what @nmfretz has to say to this topic.

Hey, is this topic being discussed on any channel? If so, I'd like to take part on it. 
At least we should get an option to **preserve app-data** when uninstalling an application on Umbrel.

nmfretz:
Hi All- 

Sorry for the delay in responding. @yapishu would you be okay if we issue a v3.3.0 update without the data directory change just so we don't hold up the release further? And then keep a PR open to keep discussing the data directory change?

### Data directory persistence

Totally understand your points about being able to persist the data, as well as the risks of booting up an instance from an out-of-date backup. The main issue is that in our current OS, storing Urbit's data outside the app's data directory would break the current apps model (e.g., bypasses filesystem sandboxing, makes it so that the OS won't show proper disk usage for the app, and the OS doesn't inform the user that data is persisted even when they go to delete the app).

We have a couple things on the horizon that should allow us to implement this nicely for Urbit:

1) Near-term, a new backups feature will allow users to backup their data and restore. FYI, even with umbrelOS today users can manually backup Urbit data, which can be found in the UI at Files > Apps > urbit. I understand there are risks with out-of-date backups as you mention so this doesn't solve accidental app deletion, but this would currently allow a user to easily persist the data in the event they needed to re-install the app or change devices, etc.
2) After that, we plan on implementing the ability for apps (and users) to request to store certain data outside of the app data dir and it will persist across installs. The OS would be aware of the configuration and there would then be clear communication to the user as to what is happening.


@arguser for this:
> I'm a long time Umbrel and Urbit user and I must say it's not obvious that if you uninstall the Umbrel application you would also lose your Urbit data.

are you saying that this is the case for all apps? Or just Urbit in particular because of what it is? Just to be clear, right now if a user goes to uninstall Urbit (or any app) they are met with this confirmation dialog that warns them that all app data will be permanently deleted.

Screenshot 2025-07-02 at 3 34 19 PM

@yapishu for this:
> I thought a better default was persisting the instance's data so that a user does not accidentally delete it. I can understand if you'd prefer to not make this default behavior; would changing it to a configurable option be acceptable?

That's a cool idea. Let me run it by the team. Are you saying that within the Urbit app the user could select to store piers data at `${UMBREL_ROOT}/data/storage/urbit`? And then internally the app would save it's piers data to `/storage` in the container instead of `/data/piers`?

arguser:
@nmfretz 
> are you saying that this is the case for all apps? Or just Urbit in particular because of what it is? Just to be clear, right now if a user goes to uninstall Urbit (or any app) they are met with this confirmation dialog that warns them that all app data will be permanently deleted.

From how I think about my instance data and in the same sense that if I see:
image
I wouldn't think that means all my downloaded content would be deleted.

Being able of opting-out from data deletion on uninstall would be nice.

nmfretz:
> I wouldn't think that means all my downloaded content would be deleted.

@arguser hmmm ya that's a fair point. It's not super clear in that case. But in the case of all the download clients (transmission, qbittorrent, etc) you aren't downloading data for those apps to use. You're not trying to download a movie and then watch it with Transmission. You're torrenting files to store on your Umbrel. Whereas for Urbit, it's data is for use within the app (sorry if I'm missing nuance, I/m not an Urbit user myself).

> Being able of opting-out from data deletion on uninstall would be nice.

Yes, agreed. That's what my second point here would allow https://github.com/getumbrel/umbrel-apps/pull/2662#issuecomment-3030559060

yapishu:
@nmfretz with urbit, your instance's state is fully contained within the data directory, so if you accidentally delete it, you lose all of your chat history etc. Urbit IDs are a decentralized identity, so  the downside to accidentally deleting your app state (losing your chat history and whatever else, plus potential monetary cost of resetting your identity's PKI keys in order to be able to use it again) far outweighs the cost of possibly leaving some stray data on the disk after an intentional uninstall, IMO. If one is an infrequent user (most likely case if you are deleting it and don't care about the data), the size of the data is likely on the order of 200MB. This might sound surprisingly large, but it's a fully self-contained virtual machine's state, and a pittance on modern disks.

nmfretz:
Hey @yapishu. Thanks again for the clarification. We've discussed as a team again and decided we cannot allow binding outside the app directory with the current umbrelOS implementation, since 
1) It sets the wrong precedence for other app developers (eg. immich and photoprism devs could make the same argument) and doesn't comply with user expectations (deleting an app should delete the app's data).
2) It will make cleaning things up when we implement user-specificed storage locations* quite complex if apps are mounting outside their data dir right now.

*We're going to elegantly solve the issue of apps like Urbit wanting to store and persist data outside their app directory in the future by allowing apps to configure their own storage destination. As soon as we do this, all Urbit users can take advantage of it and it will be super clear what is happening.

al-lac:
Hey @arguser and @yapishu, do you think we can go ahead here with updating to the latest version (3.5) and skipping the storage mount until the new solution is in place?

yapishu:
Sure; there is a actually a new runtime due out this week(?) so we can strip some stuff out here in the meantime

arguser:
@al-lac I think we are good to go now, could you give it a try so we can merge this?

arguser:
ping to @nmfretz

al-lac:
Hey @arguser, thanks updating the app version and making the requested changes. Really appreciate it! 🤝 

Will test it and then we can go live here.

al-lac (approved):
Tested on my Umbrel dev instance and everything was working flawless after the update.

Thanks again for your patience on this update @arguser and @yapishu.

Going live! 🚀

PR #2692, opened 2025-05-01
Change name of bitcoin app

copy2018:
Given the current issues surrounding OP_Return etc users should be clearly able to choose between core and Knots.

If the name remains 'Bitcoin Node' this would seem more ''supported'' and would be umbrel choosing sides and favouring Core.

To keep umbrel neutral the name should be changed to Bitcoin Core

### DIFF
--- a/bitcoin/umbrel-app.yml
+++ b/bitcoin/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1.1
 id: bitcoin
 category: bitcoin
 name: Bitcoin Core
-version: "28.1"
+version: "29.0"
 tagline: Run your personal node powered by Bitcoin Core
 description: >-
   Run your Bitcoin node and independently store and validate
--- a/bitcoin/umbrel-app.yml
+++ b/bitcoin/umbrel-app.yml
@@ -1,7 +1,7 @@
 manifestVersion: 1.1
 id: bitcoin
 category: bitcoin
-name: Bitcoin Node
+name: Bitcoin Core
 version: "28.1"
 tagline: Run your personal node powered by Bitcoin Core
 description: >-


Retropex:
I agree

Retropex (approved):

copy2018:
Time is of the essence.

copy2018:
Whats holding this back from being merged?

rstmsn:
Concept ACK.

nmfretz:
Hey @copy2018, thanks for the suggestion, and to others for chiming in!

Bitcoin Node is our flagship app that we maintain ourselves and is what we are best known for. And while the `bitcoind` container that sits behind the app is [currently Core](https://github.com/getumbrel/umbrel-apps/blob/75a1a6093ca68dfeb7984905809c464faaf3f506/bitcoin/docker-compose.yml#L43), the app itself is more than just "Bitcoin Core". It is our app that we developed to make operating a Bitcoin node both beautiful and easy.

The other node apps on the app store have been forked from our frontend+server app and are not maintained by us. In the case of Knots, it is @Retropex who is doing an excellent job of the maintenance. If our official umbrel app changes its name to `Bitcoin Core` isn't this more "choosing sides" than if we left it as `Bitcoin Node`?

> If the name remains 'Bitcoin Node' this would seem more ''supported'' and would be umbrel choosing sides and favouring Core.

One of our main tenets at Umbrel is that **YOU** should run whatever you want on umbrelOS and that we should not force it on you. And I think our actions reflect that, regardless of whether we create some of the apps ourselves or choose to name them in certain ways.

Anyone can run a[ Community app store](https://github.com/getumbrel/umbrel-community-app-store) and distribute it. [Portainer](https://apps.umbrel.com/app/portainer) is in the app store to allow you to run any container/compose setup. And most importantly for this discussion, we've made it so that all Bitcoin node apps are the same first-class citizens in umbrelOS. You can swap them out easily and connect to any* dependent app in the ecosystem.

When you go to install an app that is dependent on having a Bitcoin node installed you'll see this:

image

*actually, Knots has the only exception to this right now with Datum, where it benefits from having the only dependent app in the ecosystem that will only allow connection to Knots and no other implementation:

image

-----------------

We're also super clear about the app being backed by Core, since we mention it in:

1) The app store tagline:
All the node apps have the same style tagline so it is easy for people to compare
image


2) In the app store gallery:

image

3) In the app itself:
image

copy2018:
Hi @nmfretz, If a new user were to pick up an umbrel to run a ''Bitcoin Node'' and they see two apps, on named ''Bitcoin Node'' and one named ''Bitcoin Knots'' I would think that this new user would just run the ''Node'' rather than potentially taking a second to think and read about what they actually want to run if they saw core and knots in the store for example.

I appreciate your GUI is also a part of the ''Bitcoin Node'' app however at the end of the day It is still core, respectfully the name should reflect that in my opinion. One could ask the question since @Retropex has taken your GUI and ''improved'' it with various new settings toggles that are not exclusive to knots (constatsindex will be coming in the next version for example) should that mean that Bitcoin KNots can also be called bitcoin node. It just muddies the waters, lets unify everything under the same logic one way or another.

>If our official umbrel app changes its name to Bitcoin Core isn't this more "choosing sides" than if we left it as Bitcoin Node?
I don't see how this is the case, Umbrel maintains their own ''Bitcoin Core'' node and they also allow the community to maintain a ''Bitcoin Knots Node''. That seems more than fair.

To conclude, I think that Umbrel is _inadvertently_ funneling users to bitcoin core by presenting it as the ''Normal Node'', to make the market truly free all 3 Node implementations should be presented under their real name under the hood or they all should be called ''Bitcoin Node'' with **SECONDARY** text revealing the truth of what you are running. Lets be consistent here.

Im not being accusatory in saying umbrel is biased and im sorry if I somehow can across in that way, I simply think that calling it Bitcoin node is a fragment of when umbrel only supported one implementation which now must be deprecated in the interest of creating one LEVEL playing field. (_Or allow all implementations to be called Bitcoin Node to make it fair_ (_which would be stupid, but fair_))

rstmsn:
> To conclude, I think that Umbrel is _inadvertently_ funneling users to bitcoin core by presenting it as the ''Normal Node''

I agree with this conclusion. 
If I may, here is a link to a recent example, where the user is excited to finally "Have my own node now": 
https://x.com/BachleTanja/status/1919830771373617159

In the process of setting up the "new node", the user has installed Umbrel's "Bitcoin Node", which as we are all aware is actually a proxy for Bitcoin Core. 

Perhaps the user made a conscious decision to install Bitcoin Core, or perhaps they did so inadvertently, and without being aware that there were multiple reference implementations to choose from. 

If it is the result of the latter, then it is clear that Bitcoin Core is benefiting from being ambiguously labled "Bitcoin Node", where less experienced users will simply choose to install it, by default.

On the basis that it would probably be unreasonable to use Knots as the reference client for Umbrel's default "Bitcoin Node", would it not therefore be fairer, across the board, to simply make clear what reference implementation the user is chosing, in all cases?

PR #2700, opened 2025-05-04
Update Robosats to v0.7.7-alpha

KoalaSat:


al-lac:
Hey @KoalaSat!

Tested the update and it works fine for me, just noticed that it switches over to `nostr` as the default, is this to be expected?
CleanShot 2025-05-06 at 17 51 34

KoalaSat:
> Hey @KoalaSat!
> 
> Tested the update and it works fine for me, just noticed that it switches over to `nostr` as the default, is this to be expected? CleanShot 2025-05-06 at 17 51 34

Yes! We are slowly switching some functionalities to nostr 🙂

al-lac:
Great! Then everything is working as it should. 

Do you maybe want to mention this change in the release notes or should I merge it like this?

KoalaSat:
> Great! Then everything is working as it should.
> 
> Do you maybe want to mention this change in the release notes or should I merge it like this?

I think its fine as it is, that will only affect newly installed apps. An user who actively set that configuration in the past will stay with its choise.

PR #2723, opened 2025-05-06
App submission: sat.watch

jpcummins:
# App Submission

### sat.watch

sat.watch monitors your Bitcoin and sends real-time alerts for mempool and block confirmation activity.

### Icon
[SVG Icon](https://sat.watch/static/satwatch.svg)

### Gallery images

Screenshots
1. "Monitor your Bitcoin" https://sat.watch/static/screenshots/1.1.5-1.png 
2. "Add your addresses, xpubs, or descriptors" https://sat.watch/static/screenshots/1.1.5-3.png 
3. "Peace of mind with real-time alerts" https://sat.watch/static/screenshots/1.1.5-4.png 

I could really use some help on the gallery images. If you need additional screenshots, please let me know. 

...

### I have tested my app on:
- [x] umbrelOS on Linux VM

### DIFF
--- a/satwatch/docker-compose.yml
+++ b/satwatch/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       APP_HOST: satwatch_web_1
       APP_PORT: 8080
   web:
-    image: ghcr.io/jpcummins/sat.watch:1.1.2@sha256:c7afa6021549cb040ccf1608f1810ee98150c3485f0f03a47b5e4c4089a4496e
+    image: ghcr.io/jpcummins/sat.watch:1.1.5@sha256:8fbdd04b01d1ab8fa5626a908622a1dc36fea1c93c24eb24b8213e0bb2a94b5a
     restart: on-failure
     stop_grace_period: 1m
     environment:
--- a/satwatch/docker-compose.yml
+++ b/satwatch/docker-compose.yml
@@ -1,0 +1,39 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: satwatch_web_1
+      APP_PORT: 8080
+  web:
+    image: ghcr.io/jpcummins/sat.watch:1.1.2@sha256:c7afa6021549cb040ccf1608f1810ee98150c3485f0f03a47b5e4c4089a4496e
+    restart: on-failure
+    stop_grace_period: 1m
+    environment:
+      DATABASE_URL: "postgresql://satwatch:satwatch@db:5432/satwatch?sslmode=disable"
+      ELECTRUM_HOST: $APP_ELECTRS_NODE_IP
+      ELECTRUM_PORT: $APP_ELECTRS_NODE_PORT
+      URL: "http://$DEVICE_DOMAIN_NAME:3883"
+      ZMQ_HOST: $APP_BITCOIN_NODE_IP
+      ZMQ_PORT: $APP_BITCOIN_ZMQ_RAWTX_PORT
+      RPCHOST: "http://$APP_BITCOIN_NODE_IP:$APP_BITCOIN_RPC_PORT"
+      RPCUSER: $APP_BITCOIN_RPC_USER
+      RPCPASSWORD: $APP_BITCOIN_RPC_PASS
+    ports:
+      - "8080:8080"
+    depends_on:
+      - db
+  db:
+    image: docker.io/library/postgres:17@sha256:fe3f571d128e8efadcd8b2fde0e2b73ebab6dbec33f6bfe69d98c682c7d8f7bd
+    restart: on-failure
+    environment:
+      POSTGRES_USER: satwatch
+      POSTGRES_PASSWORD: satwatch
+      POSTGRES_DB: satwatch
+    volumes:
+      - ${APP_DATA_DIR}/data/pgdata:/var/lib/postgresql/data
+    healthcheck:
+      test: pg_isready -U satwatch -h localhost -d satwatch
+      interval: 5s
+      timeout: 5s
+      retries: 10
--- a/satwatch/hooks/post-install
+++ b/satwatch/hooks/post-install
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Create the default username and password.
+for attempt in $(seq 1 3); do
+  if "${UMBREL_ROOT}/scripts/app" compose "${APP_ID}" run --rm web './create-user -admin -username satwatch -password satwatch'; then
+    echo "Successfully created default user"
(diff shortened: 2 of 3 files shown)


satwatch/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+  app_proxy:
+    environment:
+      APP_HOST: satwatch_web_1
+      APP_PORT: 8080
+  web:
+    image: ghcr.io/jpcummins/sat.watch:1.1.5@sha256:8fbdd04b01d1ab8fa5626a908622a1dc36fea1c93c24eb24b8213e0bb2a94b5a
+    restart: on-failure
+    stop_grace_period: 1m
+    environment:
+      DATABASE_URL: "postgresql://satwatch:satwatch@db:5432/satwatch?sslmode=disable"
```
al-lac, line 13: Use of the full database service name.
```suggestion
      DATABASE_URL: "postgresql://satwatch:satwatch@satwatch_db_1:5432/satwatch?sslmode=disable"
```

satwatch/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+      ELECTRUM_PORT: $APP_ELECTRS_NODE_PORT
+      URL: "http://$DEVICE_DOMAIN_NAME:3883"
+      ZMQ_HOST: $APP_BITCOIN_NODE_IP
+      ZMQ_PORT: $APP_BITCOIN_ZMQ_RAWTX_PORT
+      RPCHOST: "http://$APP_BITCOIN_NODE_IP:$APP_BITCOIN_RPC_PORT"
+      RPCUSER: $APP_BITCOIN_RPC_USER
+      RPCPASSWORD: $APP_BITCOIN_RPC_PASS
+    user: "1000:1000"
+    depends_on:
+      - db
```
al-lac, line 24: Wait for db to be healthy.
```suggestion
    depends_on:
      db:
        condition: service_healthy
```

satwatch/hooks/post-install
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Create the default username and password.
+for attempt in $(seq 1 3); do
+  if "${UMBREL_ROOT}/scripts/app" compose "${APP_ID}" run --rm web './create-user -admin -username satwatch -password satwatch'; then
```
al-lac, line 6: Use of deterministic password.
```suggestion
  if "${UMBREL_ROOT}/scripts/app" compose "${APP_ID}" run --rm web "./create-user -admin -username satwatch -password ${APP_PASSWORD}"; then
```

satwatch/umbrel-app.yml
```diff
@@ -0,0 +1,31 @@
...
+  - bitcoin
+  - electrs
+website: https://sat.watch
+gallery: []
+repo: https://github.com/jpcummins/sat.watch
+support: https://github.com/jpcummins/sat.watch/issues
+port: 3883
+path: ""
+defaultUsername: "satwatch"
+defaultPassword: "satwatch"
```
al-lac, line 29: Show deterministic password on app page.
```suggestion
deterministicPassword: true
```

al-lac (changes requested):
Hey @jpcummins!

Thanks for the app submission! 🚀 

I added some comments on what should be adapted to get the app into the store.

One thing I noticed is that I was not able to add any addresses without providing a valid smtp config. Even after adding a webhook I still got an error about missing smtp. Is this expected?

About the admin user, as I did not see a way to create another user via the UI, it would make sense to adapt the hook from a post-install to a pre-start script to make sure the user is created again if it is deleted by mistake. Otherwise a reinstall would be needed.

al-lac:
@jpcummins, thanks for making the adaptions.

It seems to be possible to delete the satwatch user from the Web UI.

Adding a pre-start script just makes sure that the user is added back in case that happens.

A pre-start script is preferred because if the post-install fails, it will never run again. With a pre-start script you can just restart the app to run it again.

Could do something similar to `solidtime` where a file is generated after the user has successfully been created:
https://github.com/getumbrel/umbrel-apps/blob/master/solidtime/hooks/pre-start#L50-L74

jpcummins:
@al-lac  Thanks again for the helpful feedback—and apologies for the confusion on my end regarding user management. You’re absolutely right; there’s a delete account button, and I completely blanked on that. 

I’ve now implemented the changes and moved user creation to a pre-start script. Instead of using a file, I updated the user creation script to be idempotent, which I think makes the process a bit more robust. You can see the changes here:
https://github.com/jpcummins/sat.watch/commit/770f4ba97b1750a69b6190e2928cc22aef4a9bf1

I’ve tested this on both umbrel-dev and a Raspberry Pi. I confirmed that deleting the admin user and restarting correctly recreates the user, and restarting without deleting the user doesn’t result in duplicates.

I’ve also added new screenshots for the latest version of the app:

[Screenshot 1](https://sat.watch/static/screenshots/1.1.8-1.png)
[Screenshot 2](https://sat.watch/static/screenshots/1.1.8-2.png)
[Screenshot 3](https://sat.watch/static/screenshots/1.1.8-3.png)
[Screenshot 4](https://sat.watch/static/screenshots/1.1.8-4.png)
[Screenshot 5](https://sat.watch/static/screenshots/1.1.8-5.png)

Let me know if there’s anything else you’d like me to adjust.

al-lac:
Really great job @jpcummins! 👏 

Tested it and it works well! Also re-added the user after I deleted it after a restart of the app.

One thing that would still be nice is if you could let the **DB** run as user `1000:1000`, maybe you could even manage to let your **app** run rootless.

Will start to get the assets made now!

jpcummins:
@al-lac Database permissions updated. :saluting_face: 

It turned out to be a bit tricky—initially, the container failed to start due to permission errors. After some debugging, I realized the issue was that the volume mount point didn’t exist and was being created at startup by the container with root permissions. As a result, user 1000 couldn’t access the data directory. I fixed this by adding a /data/db/.gitkeep file to ensure the directory exists ahead of time with the correct permissions. You might add this as a warning to your linter -- make sure volume mount directories exist if referenced.

Also, I’ve updated the app display name from "satwatch" to "sat.watch" to keep things consistent with the documentation and website.

al-lac:
@jpcummins great job on the adaptions! 🚀 

Actually the linter already has a check for that, guess both of us just missed that, see here for a previous job result: https://github.com/getumbrel/umbrel-apps/actions/runs/14980416434

al-lac (approved):

al-lac:
Hey @jpcummins,

thanks for your patience! We are going live now! 🚀 

image

PR #2764, opened 2025-05-13
App Submission: Endurain

al-lac:
# App Submission

### App name
Endurain

### 256x256 SVG icon
https://cdn.jsdelivr.net/gh/selfhst/icons/svg/endurain.svg

### Gallery images
https://docs.endurain.com/assets/gallery/screenshot_home.png
https://docs.endurain.com/assets/gallery/screenshot_activity_page_1.png
https://docs.endurain.com/assets/gallery/screenshot_gears_page.png
https://docs.endurain.com/assets/gallery/screenshot_health_weight.png
https://docs.endurain.com/assets/gallery/screenshot_profile.png

### I have tested my app on:
- [x] umbrelOS on Linux VM

### DIFF
--- a/endurain/docker-compose.yml
+++ b/endurain/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   app:
-    image: ghcr.io/joaovitoriasilva/endurain:v0.10.7@sha256:e7a8e976ae5b3b87bff8f5fa463bda75ae91644976eaf67d26d824da10f5f628
+    image: ghcr.io/joaovitoriasilva/endurain:v0.11.2@sha256:2bd4c73e564a152ac17e336c7923e36cc6a28a3be6e67664db7ba11659455033
     # not working rootless right now
     # user: "1000:1000"
     environment:
@@ -19,6 +19,8 @@ services:
       SECRET_KEY: ${APP_PASSWORD}
       ENDURAIN_HOST: http://${DEVICE_DOMAIN_NAME}:8865
       BEHIND_PROXY: "true"
+      FERNET_KEY: changeme
+      GEOCODES_MAPS_API: changeme
     volumes:
       - ${APP_DATA_DIR}/data/app/user_images:/app/backend/user_images
       - ${APP_DATA_DIR}/data/app/files/bulk_import:/app/backend/files/bulk_import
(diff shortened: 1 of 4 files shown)


endurain/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+      APP_HOST: endurain_app_1
+      APP_PORT: 8080
+      PROXY_AUTH_ADD: "false"
+
+  app:
+    image: ghcr.io/joaovitoriasilva/endurain:v0.11.2@sha256:2bd4c73e564a152ac17e336c7923e36cc6a28a3be6e67664db7ba11659455033
+    # not working rootless right now
+    # user: "1000:1000"
+    environment:
+      DB_TYPE: postgres
```
nmfretz, line 15: I think we can run as non-root using the `UID` and `GID` env variables:
https://docs.endurain.com/getting-started/

endurain/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+      DB_HOST: endurain_db_1
+      DB_PORT: 5432
+      DB_PASSWORD: endurain
+      SECRET_KEY: ${APP_PASSWORD}
+      ENDURAIN_HOST: http://${DEVICE_DOMAIN_NAME}:8865
```
nmfretz, line 20: Darn, it's a shame that we get locked in to a domain like this for the actual frontend to communicate with the backend, because it means that users trying to connect via something like tailscale or even just by local IP, or if they have multiple umbrels that are `umbrel-<#>` they won't have a working app. The only thing that will work is `http://umbrel.local`.

Not much we can do about it though.

endurain/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+      BEHIND_PROXY: "true"
+      FERNET_KEY: changeme
+      GEOCODES_MAPS_API: changeme
```
nmfretz, line 23: Should we include a .env for these do you think? Or is the usecase not high enough?

nmfretz (commented):
Nice one @al-lac! This is a really cool usec-ase for a selfhosted app.

I've just left some minor comments below that we should address before going live.

endurain/docker-compose.yml
al-lac: Yup, really a shame. I will open a ticket, maybe this can be adapted by the developer.

al-lac: I also used an environment file for this now, can you test it out?

endurain/docker-compose.yml
al-lac: Added a .env file, the `FERNET_KEY` can be generated on device.

nmfretz (approved):
Exellent work @al-lac 👑. It's working perfectly for me.

al-lac:
Going live with this now! 🏃 

image

PR #2779, opened 2025-05-14
App Submission: OpenResty Manager

Safe3:
# App Submission

### App name
OpenResty Manager

### 256x256 SVG icon
logo

### Gallery images
om
cfg

### I have tested my app on:
- [x] umbrelOS on Linux VM


openresty-manager/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
...
+  web:
+    image: uusec/openresty-manager:1.1.3@sha256:5dfc233feb1f222927303fce93b0f63193cbe470cfec45f1b9c30b25ba9ef493
+    hostname: openresty-manager_web_1
+    restart: on-failure
+    ports:
+      - 3080:80
+      - 3443:443
+    volumes:
+      - /etc/localtime:/etc/localtime:ro
+      - /etc/resolv.conf:/etc/resolv.conf:ro
```
al-lac, line 18: This would override the docker dns, so I would remove this.
```suggestion
```

al-lac (changes requested):
Hey @Safe3!

Thanks for the submission, really don't have a lot to add. Great work!

Just added a minor thing that should be changed.

I installed the app on my Raspberry Pi 5. Could not figure out how to properly add upstreams, I tried some service names that I also used for nginx-proxy-manager like `romm_server_1`, but it would always just give me an error like this:
`nginx: [emerg] host not found in upstream "romm_server_1:8080" in /opt/om/nginx/conf/upstreams/14.conf:4 nginx: configuration file /opt/om/nginx/conf/nginx.conf test failed`
Can you help out here and maybe add an example to the description?

openresty-manager/docker-compose.yml
Safe3: Or there is another way just as the offical docker-compose.yml that use network_mode: host, than use the server ip to visit it.

﻿services:
  openresty-manager:
   image: uusec/openresty-manager:latest
   container_name: openresty-manager
   restart: always
   network_mode: host
   volumes:
     - /etc/localtime:/etc/localtime:ro
     - /etc/resolv.conf:/etc/resolv.conf:ro
     - om_acme:/opt/om/acme
     - om_data:/opt/om/data
     - om_conf:/opt/om/nginx/conf

volumes:
  om_acme:
  om_data:
  om_conf:

Safe3: The network_mode: host benefit is that it can listen any ports on host without any limits

Safe3: Hey @al-lac , the best is use the OpenResty Manager to proxy all apps, change the umbrel default port 80 to others, so it     
 can renewal free ssl certificates with both HTTP-01 and DNS-01 challenges

al-lac: Is it possible to change the default dns resolver on install?

Managed to add an app like romm_server_1 now, but still got a 502 when trying to open the page.

Safe3: Now support NGINX_RESOLVER environment to init nginx resolver config

Safe3: OK, resolv.conf was removed

al-lac (approved):
Hey @Safe3,

works really great now after the changes. I can now use it pretty much the same way as the `nginx proxy manager` and it is really easy to add other apps by using the container name like `mattermost_app_1:8000`.

We will start to prepare the assets now.

Thanks again for your contribution! 🚀

Safe3:
Hey @al-lac ,

The `umbrel` is really a great product too, I love it and be happy to contribue this.

openresty-manager/docker-compose.yml
```diff
@@ -0,0 +1,34 @@
...
+      - PUID=1000
+      - PGID=1000
+  
+  docker-host:
+    image: qoomon/docker-host:3.3.1@sha256:ec9d214d7fa88932e9186ca2f4419d44661376929931248525b08e7fa9e270f7
+    hostname: $DEVICE_DOMAIN_NAME
+    cap_add:
+      - NET_ADMIN
+      - NET_RAW
+    restart: on-failure
```
al-lac: Hey @Safe3, thanks for your patience. We would now be ready to launch. However I am questioning if this service is even needed. I tried the app without it and everything worked fine. What is the main use case for the docker-host service in that case?

Safe3: Sure, you can remove docker-host service

Safe3:
I clean the docker-compose.yml file

al-lac:
Thanks @Safe3!

Then we are ready to launch! 🚀 

CleanShot 2025-07-03 at 16 24 48@2x

PR #2807, opened 2025-05-19
App Submission: IT-Tools

dennysubke:
# App Submission

### IT-Tools
...

### Icon
mzXqNcH - Imgur

...

### I have tested my app on:
- [X] umbrelOS on Linux VM

### DIFF
--- a/ittools/docker-compose.yml
+++ b/ittools/docker-compose.yml
@@ -9,7 +9,8 @@ services:
   app:
     image: corentinth/it-tools:2024.10.22-7ca5933@sha256:8b8128748339583ca951af03dfe02a9a4d7363f61a216226fc28030731a5a61f
     restart: on-failure
-    user: 1000:1000
+    # not working rootless yet
+    #user: 1000:1000
     mem_limit: 2g
     cpu_shares: 768
     security_opt:


ittools/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+services:
+  app_proxy:
+    environment:
+      APP_HOST: ittools_app_1
+      APP_PORT: 80
+
+  app:
+    image: corentinth/it-tools:2024.10.22-7ca5933@sha256:8b8128748339583ca951af03dfe02a9a4d7363f61a216226fc28030731a5a61f
+    restart: on-failure
+    user: 1000:1000
```
al-lac, line 12: I get an error if the user is set like this. Without it, it works just fine. Can you confirm @dennysubke?
```suggestion

```

al-lac (changes requested):
Hey @dennysubke! 

Thanks for yet another wonderful submission! 💻 

Really like all this swiss army knife of an app!

I only have one change to suggest as the app does not seem to run if the user is set to `1000:1000`

ittools/docker-compose.yml
dennysubke: ✅ Confirmed. I get the same error when user: 1000:1000 is set. Without that line, everything works fine.

dennysubke: Would it be okay if we just leave out the user line?

al-lac: Yeah I think so, maybe open up an issue about it in the upstream repo if one does not exist yet about this topic.

al-lac:
Hey @dennysubke, 

thanks for this contribution and opening the ticket upstream! 💯 

Will align with @nmfretz for the final go! 🚀 

CleanShot 2025-07-03 at 17 12 19@2x

al-lac (approved):

PR #2827, opened 2025-05-21
App Submission: Actual

abhirajshourya:
# App Submission

### 256x256 SVG icon
https://avatars.githubusercontent.com/u/37879538?s=200&v=4

### Gallery images
https://github.com/actualbudget/actual/blob/master/demo.png
https://actualbudget.org/img/homepage/actual-report-dashboard-light.png
https://actualbudget.org/img/homepage/actual-main-transaction.png
https://actualbudget.org/img/homepage/actual-main-budget.png

### I have tested my app on:
- [x] umbrelOS on Linux VM


actual/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+services:
+  app_proxy:
+    environment:
+      APP_HOST: actual_web_1
+      APP_PORT: 5006
+      PROXY_AUTH_ADD: "false"
+
+  web:
+    user: 1000:1000
+    image: ghcr.io/actualbudget/actual:sha-4dba958@sha256:a22cafc52d51d7f84b1f1d30a3fbd6ad53b792edc04f627067216e9f0517d714
```
al-lac, line 12: Use the tag for the version that is being installed.
```suggestion
    image: ghcr.io/actualbudget/actual:25.5.0@sha256:3b486fbc0db02839917f65ed277aa509980f710b6e96dde966f5219983978179
```

actual/umbrel-app.yml
```diff
@@ -0,0 +1,33 @@
...
+support: https://github.com/actualbudget/actual/issues
+port: 5006
+submitter: Abhiraj Shourya
+submission: https://github.com/getumbrel/umbrel-apps/pull/2827
+gallery: []
+dependencies: []
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+releaseNotes: ""
```
al-lac, line 33: Add a final newline.
```suggestion
releaseNotes: ""

```

actual/umbrel-app.yml
```diff
@@ -0,0 +1,33 @@
...
+
+  - A beautifully designed interface is fine-tuned to get out of your way and make it as fast as possible to explore your finances.
+
+  - Actual is a local app, plain and simple. Your data is synced in the background so all devices have access, but the app totally works regardless of your network connection. This also allows end-to-end encryption to keep your data private.
+
+  - Save hundreds of dollars a year (at least!) by tracking your spending. Based on tried and true methods, our budgeting system is based off of your real income instead of made up numbers. This makes you face your real spending, and clearly shows how much you are saving each month. We make this process as simple as possible. 
+
+  - Breeze through your transactions and update them easily with a streamlined, minimal interface. Categorizing your transactions correctly is important and we’ve optimized this process. Manage split transactions and transfers all in the same editor.
+  
+  - Intuitive reports give you a quick way to learn about your finances. By default, we include net worth and cash flow reports. Actual also includes a powerful custom report engine to design your own reports to fit your needs.
```
al-lac, line 20: Maybe add a note and explain what to do about the error that is shown after opening up the app for the first time. Otherwise users might think that the app is broken.
CleanShot 2025-05-22 at 22 41 03

al-lac (changes requested):
Hey @abhirajshourya, 

Thanks for the submission! 💸 

Really great job on preparing the app, only have some minor things to add. Please adapt the requested changes.

actual/umbrel-app.yml
abhirajshourya: Hey @al-lac I've added a note regarding the issue, let me know if you need any other changes. Thanks!

al-lac:
Hey @abhirajshourya,

Thanks again for your contribution!

After aligning with the team I was informed that there were already submissions for Actual already. But they were declined due to the error one gets when starting the app (SharedArrayBuffer).

So we will put this submission on hold until umbrelOS supports HTTPS out of the box.

Thanks for your patience.

abhirajshourya:
Hi @nmfretz I did update so the description for users to go around the error and start using the app. I've been using it in http mode locally, works great for me. I haven't faced any issues yet.

abhirajshourya:
Re: the SharedArrayBuffer "Fatal Error" some reviewers may hit when opening Actual over plain HTTP on the LAN — wanted to explain the reasoning behind how this PR handles it, since it's not something fixable in `docker-compose.yml`.

**Root cause:** Actual's server already sends the correct `Cross-Origin-Opener-Policy`/`Cross-Origin-Embedder-Policy` headers on its own. The only missing piece is a secure context — browsers only allow `SharedArrayBuffer` over HTTPS or `localhost`, and Umbrel serves every app over plain HTTP on the LAN by default via `app_proxy`. There's no manifest field or dependency mechanism that lets one app auto-terminate TLS in front of another, so this can't be resolved from inside the app's own compose file.

Given that, I followed the pattern already established elsewhere in this store for apps with the same hard requirement:
- [`privatebin/umbrel-app.yml`](https://github.com/getumbrel/umbrel-apps/blob/master/privatebin/umbrel-app.yml) — *"PrivateBin requires HTTPS... the easiest way is with the Cloudflare Tunnel app or Nginx Proxy Manager app on the Umbrel App Store"*
- [`mattermost/umbrel-app.yml`](https://github.com/getumbrel/umbrel-apps/blob/master/mattermost/umbrel-app.yml) and [`wordpress/umbrel-app.yml`](https://github.com/getumbrel/umbrel-apps/blob/master/wordpress/umbrel-app.yml) — same two apps (Cloudflare Tunnel / Nginx Proxy Manager) recommended for HTTPS access

This PR's description note follows the same shape, but is a bit more specific: it explains *why* the error happens (secure-context requirement, not a bug), and calls out Tailscale's "Serve" feature by name as a low-friction way to get HTTPS on the LAN, alongside a generic reverse-proxy mention and Actual's own built-in "unsupported fallback mode" workaround for single-device use. I tested all of this locally — reproduced the exact Fatal Error over a non-localhost origin, confirmed the headers Actual sends are already correct, and verified fallback mode works — before writing the note.

Also bumped the pinned image from `25.5.0` to the current latest `26.9.0` while updating this.

abhirajshourya:
Hi @al-lac can I request a re-review please? Thanks!

al-lac:
Hey @abhirajshourya!

umbrelOS 2.0 will also support Apps that require HTTPS, so maybe a workaround is not needed.

@nmfretz or @kaichappel will probably get to reviewing this soon.

PR #2844, opened 2025-05-23
Update DATUM setup instructions

nmfretz:
Hey @Retropex, could you please take a look at these updated instructions for the DATUM app description and let me know if they look okay?

We've gotten some support queries recently with some confusion on how to set up Datum, and I realized that things have changed since you guys added config in the UI (which is a fantastic addition by the way!).

Users won't see an available app update with this PR, the app store description will just change.


Retropex (approved):
It’s ok for me. 👌

nmfretz:
Awesome, thanks @Retropex!

PR #2871, opened 2025-05-27
add widget for datum

Retropex:


al-lac:
Hey @Retropex, thanks for adding widget support here! 🚀 

Seems to work fine for me, not sure though about the `three-stats` type and `two-stats` example. Thought that `two-stats-with-guage`could also make sense here. What do you think @nmfretz?

Retropex:
The `two-stats` exemple is a leftover of my test I will fix this to `three-stats` since `two-stats` seems to [not exist](https://github.com/getumbrel/umbrel/blob/93a4b1801d1141252e73ff759a4ef52d038fff92/packages/ui/src/modules/widgets/index.tsx#L83C2-L125C3).

Retropex:
@al-lac change pushed and tested again on my side.

al-lac (approved):

al-lac:
Looks good to me! Will merge it now. FYI @nmfretz.

PR #2952, opened 2025-06-13
Update shopstr to version 0.7.5

calvadev:
This update of Shopstr brings the following improvements and fixes:
    - Fixed race condition when minting new proofs at checkout
    - Revised checkout page design
    - Minor code styling fixes


al-lac:
Hey @calvadev, thanks for the update! Tested successfully on my Umbrel dev instance.

PR #2985, opened 2025-06-16
Update twenty to v0.55.6

### LINKED ISSUE #2594: Twenty email and calendar providers list is empty

al-lac:
This updates twenty to version 0.55.6.

This release comes with a new feature that gives users the option to configure environment variables via the twenty admin panel in the settings. This fixes #2594.

Also closes #2625 as the workaround with the environment files is not needed anymore then.

Tested on Umbrel dev instance.

### DIFF
--- a/twenty/docker-compose.yml
+++ b/twenty/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
   
   server:
-    image: twentycrm/twenty:v0.54.4@sha256:6a6e5a34426867e746f43ef000989cae3afc44f2e1cedf8bcefa447f6988187a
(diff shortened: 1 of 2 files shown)


al-lac:
A fresh install works fine, but the update fails for me, more details on the issue i opened here: https://github.com/twentyhq/twenty/issues/12651

al-lac:
The upstream issue has been solved. Tested the update again and it needs a hook to clear the cache. This hook can be removed in future versions as it was really needed for updates from 0.54 to 0.55 and newer.

Also tested a fresh install.

PR #3045, opened 2025-06-24
 Update cobalt to version 11.1

dennysubke:
I tried the update on my Umbrel Home and everything ran smoothly without any problems.


al-lac:
Hey @dennysubke, for me the local media processing was enabled per default and it did not start downloads for me. After disabling it everything worked just fine. Does it work both ways for you?

dennysubke:
Hey @al-lac 

I gotta admit I was a bit lazy testing that part 🤷‍♂️ I am running into the same issue local media processing is enabled by default and downloads just do not start. 

There are two options either open an issue in the official Cobalt repo or add a note in the app description telling folks to disable local processing if it causes trouble. 

Since this feature is still in beta hopefully it will get fixed in a future update.

Thanks for bringing it up. ✌🏻

al-lac:
Yeah an issue on their repo might be a good idea. We are in no urgency to upgrade before this is fixed.

al-lac:
@dennysubke can you update to version `11.2.2`? This could maybe solve the issue. If it does not, I would open up an issue on their repo.

https://cobalt.tools/updates#11.2
https://github.com/imputnet/cobalt/pkgs/container/cobalt/454426263?tag=11.2.2

dennysubke:
https://github.com/getumbrel/umbrel-apps/pull/3152

al-lac:
Newer version was pushed via #3152.

PR #3075, opened 2025-06-25
App Submission: Brave

dennysubke:
# App Submission

### Brave
...

### Icon
logo

...

### I have tested my app on:
- [X] umbrelOS on Linux VM


brave/umbrel-app.yml
```diff
@@ -0,0 +1,43 @@
...
+repo: https://github.com/linuxserver/docker-brave
+support: https://github.com/linuxserver/docker-brave/issues
+gallery:
+  - 1.jpg
+  - 2.jpg
+  - 3.jpg
+  - 4.jpg
+releaseNotes: ""
+dependencies: []
+path: "https://umbrel.local:3935"
```
al-lac, line 41: While this is a nice workaround to get self-signed https to work, it will break the app if it is not running on `umbrel.local,` if you have multiple instances the url could also be `umbrel-2.local` or `umbrel-dev.local`

dennysubke: Yes, I was aware of that limitation and used umbrel.local as a quick workaround to make HTTPS work. I hope someone on your end has a better idea for a proper fix. 💪🏻

dennysubke:
Hey @al-lac  

Thanks for the review. 💪🏻 

We will wait for @nmfretz and hope he has an idea.💡

nmfretz:
Hey @dennysubke thanks for your tireless efforts submitting apps!

Unfortunately we can't accept the Brave submission right now for the same reason the Omada Controller PR is on hold (see #1196).

The main issue is that we don't want to have apps on the official app store that will inadvertently train non-technical users to click through cert warnings, since this is identical UX to a real man-in-the-middle attack.

Once umbrelOS has inbuilt HTTPS support, then we can revisit.

One tip for your community app store version: you should look into removing the `seccomp:unconfined` security profile, since this makes it possible for the container to run high-risk syscalls that could be exploited to break out of the container. I don't think it's actually necessary to have for the Docker and debian versions used in umbrelOS currently.

dennysubke:
Hey @nmfretz! 👋

Thanks a lot for the thoughtful feedback and the clear explanation. I completely understand the concern and agree we should not train users to ignore cert warnings. It totally makes sense to keep it on hold until Umbrel supports HTTPS out of the box.

Also appreciate the tip about seccomp:unconfined. I will remove it from my Community App Store version.

Thanks again for your time and your review! 💪🏻

nmfretz:
Thanks for being so awesome @dennysubke! Would you like to keep this PR open and moved to draft? Or closed for now?

dennysubke:
@nmfretz 

I’ll close the PR now. Thanks a lot for your time and support! 😊

PR #3078, opened 2025-06-26
New knots UI + datum fixs

Retropex:
Hey, I have rebased your new UI for Knots, but I'm experiencing some issues with the other apps. They can’t connect to Knots. I don’t know why so far, but I'm investigating. I hope you can help me with this. Thanks!


Retropex:
Port fixed at least for datum, I can't test applications that require a non pruned node for now.

What is the logic behind the "swap" for apps? what env variable are used to switch app from core to knots?

nmfretz:
Hey @Retropex, I'm back in town and will make sure to get to this today!

nmfretz:
> Port fixed at least for datum, I can't test applications that require a non pruned node for now.

I also only have a pruned Knots node right now. I'll do some datadir copying and get a non-pruned node spun up so we can properly test apps that require `txindex` etc.

nmfretz:
> What is the logic behind the "swap" for apps? what env variable are used to switch app from core to knots?

There are two main things required for umbrelOS to show multiple implementations of an app and then for those apps to work with other apps in the ecosystem.

### 1) umbrel-app.yml manifest

First, in the manifest the app tells umbrelOS that it `implements` bitcoin.

```yml
manifestVersion: 1.1
id: bitcoin-knots
implements:
  - bitcoin
```

With this in the manifest, umbrelOS shows all impementations of an app when a user goes to install an app that has `bitcoin` as a dependency. For example, here's Electrs in the app store:

image

This isn't just for Bitcoin, although it is what prompted us to add the feature. It's an OS level feature we've added that lets any app do this. For example, say someone wanted to submit another Matrix server besides Synapse which is currently the only one in the app store, they could now do it and then when a user goes to install the Element client they would see different Matrix server options to connect it to.

Say a user has Bitcoin Node installed and wants to install Bitcoin Knots instead, they would install Bitcoin Knots and then to switch any apps over they would right click on the app and select `Settings` which will show them the UI for connecting to Knots:

image

image

### 2) exports.sh

Then the thing that actually makes the apps in the ecosystem work with each other is that implementations of apps need to export the exact environment variable names that other connecting apps use. For example, your app must export a variable named exactly `APP_BITCOIN_DATA_DIR` in order for apps like Electrs and Ride The Lightning to function since that is the variable name they use in their compose files.

This is what these lines are doing in the Knots version currently on the app store:

https://github.com/getumbrel/umbrel-apps/blob/181fb8aa23ecc192dc6300496d7b85d36b30dd04/bitcoin-knots/exports.sh#L139-L170

This loop iterates through a list of variable name suffixes and creates corresponding APP_BITCOIN_* environment variables based on existing APP_BITCOIN_KNOTS_* variables. When a dependent app starts up (like Electrs) it runs the connected Knots app's exports.sh and therefore gets the correct variable names it needs.

I'll update the exports.sh file @Retropex. Is there anything else that you're still working on? We tried to make it super easy to maintain the new codebase, so adding new settings should be as simple as just updating the single `settings.meta.ts` file and everything should just work (backend config, backend + frontend validation, new UI  elements + responsiveness).

Edit: updated exports.sh in https://github.com/getumbrel/umbrel-apps/pull/3078/commits/a2fd87d05874c4eb23bc2b7195fb284ad1458bf4 and https://github.com/getumbrel/umbrel-apps/pull/3078/commits/44886656e9fed565c47a77347c6c2c10986a2bfc

nmfretz:
Also @Retropex, there's a couple things we ran into in the early hours of our launch that you should be aware of and check for:

1) It is somehow possible that some users can have a pre-update `~/umbrel/app-data/bitcoin/data/apps` data directory that is owned by the root user (so will likely be the same for Knots). This is despite the fact that the app gets installed by umbrelOS with this data dir owned by user 1000. The pre-start script currently handles this, but just so you're aware.

2) We didn't initially account for all ways that a user may have manually gone in via the command line and edited the `bitcoin.conf`. This is handled correctly in the latest `1.0.2` release so make sure you have pulled in those changes.

Retropex:
Thanks a lot for the help! For now, I don't plan to add any policy, and the app has been rebased on v1.0.2, so I think it's ready if you have been able to run other apps with Knots.

With the new app, does the user have to restart all knots dependent apps?

nmfretz:
> Thanks a lot for the help! For now, I don't plan to add any policy, and the app has been rebased on v1.0.2, so I think it's ready if you have been able to run other apps with Knots.

I'm just waiting on full sync of a non-pruned node to test app connections and then we can go live!

> With the new app, does the user have to restart all knots dependent apps?

For this specific update I think we should recommend that users do this. The reason is that I have just made the internal container and external host port mappings the same https://github.com/getumbrel/umbrel-apps/pull/3078/commits/9d1d4c3062dc3434bbc3b5916a47319022a55265. This means that any apps using the env vars `APP_BITCOIN_KNOTS_INTERNAL_RPC_PORT` and `APP_BITCOIN_KNOTS_INTERNAL_P2P_PORT` will need to restart to source the change properly.

You had actually previously suggested we do this (https://github.com/getumbrel/umbrel-apps/pull/2450#issuecomment-2796309770). At the time I advised not to do this since it would require changes to the old app's source code (rpc and p2p ports were hardcoded). And it would mean that user's who wanted to use the bitcoin cli would need to include `rpcport=9332` in their commands... Welp, now things are the exact opposite with the new app 😅. The `P2P_PORT` and `RPC_PORT` variables in the compose file for the new app control both the conf settings and what the UI shows in the connection modal. So the easiest thing now is to set the internal ports to the same as the external ones instead of adding complexity to the source code.

Also, the rpc user and password are no longer passed in to the daemon as startup flags. Instead rpcauth is set in the conf file so if advanced users go to use the bitcoin cli from the command line they will need to include rpc username, password, and rpcport anyways which can be found in the connection modal.


In the meantime @Retropex here's a couple things to consider, but feel free to ignore since you're maintaining the app:

Did you want the in-app header to say Bitcoin Node or Bitcoin Knots?

image


In the QR codes within the Connect modal, did you want to swap out the Umbrel logo for the Knots logo?

In the "Advanced" tab of the settings page I see you added some conf options. Did you want these here or under a different tab? The reason I ask is because the Advanced tab currently states that changes in this tab are not validated... but the options you added are in fact validated because they come from the settings.meta.ts file, which is used to automatically generate zod validation schema used in both the UI and the backend.

image

Retropex:
> I'm just waiting on full sync of a non-pruned node to test app connections and then we can go live!

I'm also doing this. It would be nice to wait for the release until I have also been able to test it.

> Did you want the in-app header to say Bitcoin Node or Bitcoin Knots?

Since it's in app, I think I will let this as "Bitcoin Node"

> In the QR codes within the Connect modal, did you want to swap out the Umbrel logo for the Knots logo?

Yes I will change that!

> In the "Advanced" tab of the settings page I see you added some conf options. Did you want these here or under a different tab?

No it's an error on my side, I will move the to policy or optimization.

Retropex:
@nmfretz I have been able to fully sync knots and test the migration from v0.8.0 to v1.0.2. Policy settings are correctly migrated, and the app connects correctly to knots (tested LND, Core Lightning, Bitfeed, Electrs, and Fulcrum).

Unless you get problems on your side I think we are good to go!

Also don't forget the PR to update the knots logo: https://github.com/getumbrel/umbrel-apps-gallery/pull/50

nmfretz:
Excellent @Retropex. I've tested compatibility and all appears to be working well.

I haven't tested DATUM yet though. Are you passing in a `blocknotify` flag in your fork? If not, we need to add the flag here in the compose file. We have added `BITCOIND_EXTRA_ARGS` to the compose file for just such a purpose. You can add cli flags here and bitcoind will start with them. That way, if you need to make changes you can so so simply by updating the compose file and not needing to build a new app image.

**Edit** - actually @Retropex is `blocknotify` even required anymore? Just tested now and I'm getting stratum updates fine:

```
datum_datum_1      | 2025-07-07 04:06:46.825 [               datum_gateway_template_thread]  INFO: Updating priority stratum job for block 904401: 3.12509021 BTC, 15 txns, 5968 bytes (Sent to 0 stratum clients)
datum_datum_1      | 2025-07-07 04:07:28.613 [               datum_gateway_template_thread]  INFO: Updating standard stratum job for block 904401: 3.12626215 BTC, 102 txns, 60763 bytes (Sent to 0 stratum clients)
datum_datum_1      | 2025-07-07 04:08:10.833 [               datum_gateway_template_thread]  INFO: Updating standard stratum job for block 904401: 3.12766500 BTC, 184 txns, 184818 bytes (Sent to 0 stratum clients)
datum_datum_1      | 2025-07-07 04:08:53.034 [               datum_gateway_template_thread]  INFO: Updating standard stratum job for block 904401: 3.13046604 BTC, 297 txns, 317671 bytes (Sent to 0 stratum clients)
datum_datum_1      | 2025-07-07 04:09:35.255 [               datum_gateway_template_thread]  INFO: Updating standard stratum job for block 904401: 3.13096654 BTC, 364 txns, 347260 bytes (Sent to 0 stratum clients)
```

I've run across a migration issue that can impact some users, so I need to issue an update for https://github.com/getumbrel/umbrel-bitcoin. It will be 1.0.3. Once I do that, you can pull in the changes to your fork and then we can go live!

nmfretz:
@Retropex, 1.0.3 changes have been pushed to https://github.com/getumbrel/umbrel-bitcoin.

nmfretz:
@Retropex I've just seen your commit come in for blocknotify https://github.com/getumbrel/umbrel-apps/pull/3078/commits/9d544294f88e1b98dde97ed619ffef90607bdfc1

Seeing the actual command with whitespace reminds me that this probably doesn't work right now. The new app splits the extra args on whitespace so that command won't get parsed correctly.

I think the easiest fix for you will be to split on commas instead of whitespace in `manager.ts`

nmfretz:
Something like this:

```ts
// Grab extra flags from env, if present
// This allows us to add extra flags in the app store compose file without changing this codebase
const envArgs = (process.env['BITCOIND_EXTRA_ARGS'] ?? '')
  .trim()
  .split(',') // splits on commas to allow spaces in arguments
  .map(arg => arg.trim()) // trim whitespace from each argument
  .filter(Boolean) // removes empty strings
```

Retropex:
Yeah, sorry, I shouldn't have pushed. I'm making the datum command a switch on the UI so that users who don't want to use datum can disable it so they don't waste resources for nothing.

Retropex:
I'm experiencing some issues with datum after the update. I'm investigating. Otherwise, the new datum block notification setting works!

bitcoin-knots/exports.sh
```diff
@@ -11,113 +10,55 @@ export APP_BITCOIN_KNOTS_ZMQ_RAWBLOCK_PORT="48332"
...
 export APP_BITCOIN_KNOTS_ZMQ_HASHBLOCK_PORT="48334"
 export APP_BITCOIN_KNOTS_ZMQ_SEQUENCE_PORT="48335"
-export APP_BITCOIN_KNOTS_INTERNAL_RPC_PORT="8332"
-export APP_BITCOIN_KNOTS_INTERNAL_P2P_PORT="8333"
+export APP_BITCOIN_KNOTS_ZMQ_HASHTX_PORT="48336"
 
-BITCOIN_CHAIN="main"
-BITCOIN_ENV_FILE="${EXPORTS_APP_DIR}/.env"
+# These are legacy env vars we need to keep around for DATUM compatibility
+export APP_BITCOIN_KNOTS_INTERNAL_RPC_PORT="9332"
```
Retropex, line 16: For some reason, these variables are not detected by DATUM even after restarting DATUM, but it works if I reinstall DATUM completely. Any ideas on what could go wrong?

Retropex: I found the problem, DATUM configure the RPC url only once https://github.com/getumbrel/umbrel-apps/blob/5f92fcaa4cd8066fc0d84f6b616e2db2cd087da2/datum/hooks/pre-start#L20

What do you think of just removing this condition so that datum check for credentials at each start?

Retropex:
Oh and sorry I didn't answered that:

> actually @Retropex is blocknotify even required anymore? Just tested now and I'm getting stratum updates fine:

Yes, it's still required otherwise your Stratum jobs will indeed actualize, but you will mine stale work which is very bad.

bitcoin-knots/exports.sh
nmfretz: Ah ya, the best-practice here would be to have the DATUM container read these values on startup. The easiest way is to have it so that DATUM accepts the RPC config as environment variables so they can be added to the compose file. That way updates are incredibly easy and you never have to worry about .

> What do you think of just removing this condition so that datum check for credentials at each start?

But yes, this is a good option to get this out the door quickly. You can just have it so that on every start of DATUM the pre-start hook writes out the rpc credentials.

nmfretz: One consideration here though is that in the DATUM UI, you have the option for users to edit these values:

image

This means if you rewrite these values every start (via env vars or pre-start script) then a user wouldn't be able to persist their own changes. This shouldn't be a huge issue though because the DATUM app has the Knots app as a dependency so I can't see anyone really needing to change these. But it could be confusing if there are some users who do this.

Retropex: I decided to change the script because modifying datum to recognize an environment variable requires changing its C source code and necessitates review to ensure I don’t break anything. I don’t want to add more work to the OCEAN devs for the review.

Since, as you said, it's very unlikely that people change the RPC in the interface, I have disabled these UI elements.

image

Retropex:
Ok changes have been tested and I no longer encounter any issues, if it's also OK on your side then we are good to go!

Again don't forget the PR to update Knots logo on the App Store. https://github.com/getumbrel/umbrel-apps-gallery/pull/50

nmfretz:
Thanks @Retropex. I have tested:
- arm64 and x86
- fresh install and update
- external wallet connection via electrum server and rpc
- port forwarding for incoming clearnet peer connections to make sure internal port change works.
- Datum compatibility

```
2025-07-09 02:32:13.882 [               datum_gateway_template_thread]  INFO: Updating priority stratum job for block 904677: 3.14795778 BTC, 2008 txns, 1353799 bytes (Sent to 0 stratum clients)
2025-07-09 02:32:13.891 [               datum_gateway_template_thread]  INFO: NEW NETWORK BLOCK NOTIFICATION RECEIVED
2025-07-09 02:32:55.937 [               datum_gateway_template_thread]  INFO: Updating standard stratum job for block 904677: 3.14795778 BTC, 2008 txns, 1353799 bytes (Sent to 0 stratum clients)
```

> Again don't forget the PR to update Knots logo on the App Store. https://github.com/getumbrel/umbrel-apps-gallery/pull/50

Yep, we'll update both the logo and the images. They are separate from this apps repo, so as soon as they are updated in umbrel-apps-gallery everyone will see the new logo/images no matter what version of Knots they are on. We'll tweak your logo to be non-rounded (umbrelOS will responsively round it to the correct degree throughout the homescreen and app store UI).

Going live 🎉 Good luck.

Retropex:
@nmfretz thank you for your precious help! 🙏

PR #3094, opened 2025-06-30
Update btc-rpc-explorer to v3.5.0

nmfretz:
~~DO NOT MERGE YET~~

Tested and working on x86 and arm64. It looks like there is a small issue where the project forgot to update the version number in their UI:

image

Let's wait a couple days to see if they release a fix, just so we don't release 2 updates close together with only a cosmetic version change.


nmfretz:
No fix for in-app version number yet. Going live with the following note:

image

PR #3121, opened 2025-07-03
Update category of monero

al-lac:
This updates the category of monero to the new `crypto` one.


nmfretz:
It turns out we'll need to wait for an umbrelOS update to allow new categories on the official app store UI

PR #3153, opened 2025-07-08
Bump rtl and boltz-client

jackstar12:


al-lac:
Thanks for the update @jackstar12! 

I tested both apps successfully on my Umbrel dev instance. Also updated the formatting of the release notes.

PR #3240, opened 2025-07-17
update datum to v0.4.0beta

Retropex:
tested on umbrel 1.4.2


al-lac:

PR #3289, opened 2025-07-26
Update Monero daemon to v0.18.4.1

colder-is-better:
Monero release v0.18.4.1 is out, introducing various bug fixes.


al-lac:
Tested update on my Umbrel dev instance, everything seems to work just fine!

Thanks for the update @colder-is-better!

PR #3332, opened 2025-07-29
Update openresty-manager to version v1.7.2

Safe3:


al-lac:
Hey @Safe3, thanks for the update!

I made some changes so the new `logs` folder also works for updates to the app. Could you please update the pre-start hook with 755 permissions? I don't have access to push to your repository: `chmod +x hooks/pre-start`

Also had to remove access to the docker socket, that somehow slipped though on the initial submission. If the app should continue to work with apps, we would need to add dind (docket-in-docker), similar to how it is solved for apps like portainer here: https://github.com/getumbrel/umbrel-apps/blob/4d686f5a5e48fdbeb74618b5a304be54eed895af/portainer/docker-compose.yml#L9-L27

Safe3:
It doesn't matter, only affect the use of openresty-manager appstore function. I added you to the member of my repository, now you can modify any file as you wish!

al-lac:

Safe3:
Great !

Issue #3325, opened 2025-07-28
[Feature Request] Add ffmpeg in Sabnzbd container


ernesto75:
Would be nice to have ffmpeg already included in the container. I found it so useful running ffmpeg in scripts to automate the conversion of downloaded files! Thanks a lot if you manage to add it!

PR #3339, opened 2025-07-31
ffmpeg support for Sabnzbd

ernesto75:
I'm editing sabnzbd docker-compose.yml by adding environment variables to make use of docker-mods to install ffmpeg package. See https://github.com/linuxserver/docker-mods and https://github.com/linuxserver/docker-mods/tree/universal-package-install

This will also close issue #3325


al-lac:
Hey @ernesto75,

thanks for the update! Tested it and all seems good.

PR #3373, opened 2025-08-06
Dynamic port allocation for Bitcoin Knots

Retropex:
This feature has been largely requested by Knots users. I really think it should be added to Umbrel.

### DIFF
--- a/bitcoin-knots/exports.sh
+++ b/bitcoin-knots/exports.sh
@@ -2,19 +2,30 @@ export APP_BITCOIN_KNOTS_NODE_IP="10.21.21.7"
 export APP_BITCOIN_KNOTS_TOR_PROXY_IP="10.21.22.12"
 export APP_BITCOIN_KNOTS_I2P_DAEMON_IP="10.21.22.13"
 
+installed_apps=$(ls /home/umbrel/umbrel/app-data)
+
+if echo "$installed_apps" | grep --quiet -x 'bitcoin'; then
+	export APP_BITCOIN_KNOTS_RPC_PORT="9332"
+	export APP_BITCOIN_KNOTS_P2P_PORT="9333"
+	# These are legacy env vars we need to keep around for DATUM compatibility
+	export APP_BITCOIN_KNOTS_INTERNAL_RPC_PORT="9332"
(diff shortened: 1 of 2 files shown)


al-lac:
Hey @Retropex, 

did you test what happens if a user installs `Bitcoin Core` afterwards?

From looking at the code this would probably end up in a failed install.

Waiting on @nmfretz to provide more input here.

Retropex:
Hey

> did you test what happens if a user installs Bitcoin Core afterwards?

That's the only downside, it will fail. That's said the user just have to stop Knots to be able to install Core, Knots will take port 9333 upon restart.

al-lac:
Yeah we probably need some handling on the bitcoin core side as well to avoid it failing on install. But lets wait for what @nmfretz has to say.

Retropex:
Hey @nmfretz, would it be possible to have your input here?
Thanks!

nmfretz:
Hey @Retropex, thanks for the ping. And sorry for the delay in responding.

>  I really think it should be added to Umbrel.

Totally agree with you here. We're going to need to wait until we implement this at the OS level though unfortunately. We plan to implement this feature but we're currently working on some other high priority stuff at the moment (backups etc). In the future, it will look something like this:

Apps will request preferred ports and then umbrelOS can assign the ports if they are free, or fallback to other ports (e.g., port +1) if they are already in use. So apps will never have port conflicts and won't break each other on install. So Knots would just set 8332 and 8333 as it's preferred ports and the OS would handle any conflicts with clear info presented to the user.

Right now we'd have to hackily add lifecycle hooks to both Bitcoin Knots and Bitcoin Node apps to restart each other on install and it would be super brittle. We also wouldn't have a clear way to indicate to users what is happening.

Retropex:
Ok, thanks for the feedback!

Please let me know the day the feature is ready.

nmfretz:
@Retropex will do! Do you want this PR reopened though? If so, please feel free to do so and we've tagged it as "pending umbrelOS feature" to track it.

nmfretz:
Hey @Retropex I'm just doing some cleaning of PRs here and I'm going to close this one out here and track the dynamic port work in the main umbrel repo instead

Still completely agree this is the right feature. Once the umbrelOS side is in place, we can come back to Knots and wire it up properly.

PR #3375, opened 2025-08-07
App Submission: Bassin

duckaxe:
A zero-fee Bitcoin solo mining pool for umbrellOS. Run your own CKPOOL at home.

# App Submission

### Bassin
...

### 256x256 SVG icon

app-icon

...

### Gallery images
gallery_1
gallery_2
gallery_3

...

### I have tested my app on:
- [ ] umbrelOS on Linux VM

### DIFF
--- a/bassin/docker-compose.yml
+++ b/bassin/docker-compose.yml
@@ -58,7 +58,6 @@ services:
     depends_on:
       - ui
     restart: on-failure
-    container_name: bassin_www_1
 
   widget-server:
     image: ghcr.io/duckaxe/umbrel-bassin-widget:main@sha256:293d4a17579bbef339c3f4595ed2dc36bef70699d34909bfa3ac20bba8facb7c


bassin/umbrel-app.yml
```diff
@@ -0,0 +1,55 @@
...
+          text: "2.10"
+          subtext: "Th/s"
+        - title: "Workers"
+          text: "2"
+        - title: "Best Share"
+          text: "21"
+          subtext: "G"
+        - title: "Total Shares"
+          text: "210"
+          subtext: "M"
```
al-lac: Add final newline.
```suggestion
          subtext: "M"

```

bassin/docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+      - ${APP_DATA_DIR}/data/www:/www:ro
+    restart: on-failure
+    container_name: bassin_www_1
+  
+  widget-server:
+    image: ghcr.io/duckaxe/umbrel-bassin-widget:main@sha256:293d4a17579bbef339c3f4595ed2dc36bef70699d34909bfa3ac20bba8facb7c
+    user: "1000:1000"
+    environment:
+      - BASSIN_API_POOL_URL=http://bassin_www_1/pool/pool.status
+    restart: on-failure
```
al-lac: Add final newline.
```suggestion
    restart: on-failure

```

duckaxe: Done

bassin/umbrel-app.yml
duckaxe: Done

duckaxe:
@al-lac I have already tried it. It's complicated because the CKPOOL (log & json) files and front-end assets must be in the same folder (`data/www`) and accessible via the same web server.

I noticed that some Umbrel apps include front-end assets. That's why I did it this way. It is and will remain a single-page application. Only the JS and CSS files will change, if at all. Everything else stays the same. 

This makes things much easier for me.

al-lac:
Hey @duckaxe, ok got it. Lets see what @nmfretz has to say about it.

Usually it is fine to just have some simple html files in there, but once we go into javascript and other logic we have to think about what makes sense here.

duckaxe:
Thank you for response, @al-lac 
Then waiting for @nmfretz

duckaxe:
@al-lac I found a way to extract the user interface into a docker image. It's much cleaner now ;)
I also updated the gallery images.

al-lac:
Awesome work @duckaxe! 🚀 

I made some adaptions to the description formatting and removed the leftover directory.

duckaxe:
@al-lac Thanks for that. You added `awaiting gallery assets` flag - I already provided all ready2go gallery assets. Do we need more?

al-lac:
Hey @duckaxe, really great work on those assets! 🔥 

I added the label as there is usually a final design check and possible adaptions before apps make it to the store.

But don't worry, does not take too long usually!

duckaxe:
❤️

duckaxe:
> But don't worry, does not take too long usually!

Hope it happens soon.

bassin/umbrel-app.yml
```diff
@@ -0,0 +1,58 @@
+manifestVersion: 1
+version: "2.0.0"
+id: bassin
+name: Bassin
+tagline: Solo Bitcoin Mining Pool
+description: >-
+  A zero-fee Bitcoin solo mining pool for umbrellOS. Run your own CKPOOL at home.
```
al-lac: Fix wording.
```suggestion
  A zero-fee Bitcoin solo mining pool for Umbrel. Run your own CKPOOL at home.
```

bassin/umbrel-app.yml
```diff
@@ -0,0 +1,58 @@
+manifestVersion: 1
+version: "2.0.0"
+id: bassin
+name: Bassin
+tagline: Solo Bitcoin Mining Pool
```
al-lac: Make it clear that it is based on CKPOOL
```suggestion
tagline: Solo Bitcoin Mining Pool (CKPOOL)
```

bassin/umbrel-app.yml
```diff
@@ -0,0 +1,58 @@
...
+description: >-
+  A zero-fee Bitcoin solo mining pool for umbrellOS. Run your own CKPOOL at home.
+
+
+  No setup, no hassle. 
+    - Point your miner to **umbrel.local:3456** or **<umbrelIP>:3456**
```
al-lac: Fix umbrelIP string.
```suggestion
    - Point your miner to **umbrel.local:3456** or **`<umbrelIP>:3456`**
```

duckaxe:
@al-lac Thank you for your comments. 

1. My knowledge in this area is limited. Unfortunately, I have no idea how to solve this problem.
2. The loading indicator is a kind of fallback for all cases where something goes wrong. A generic error message would not help the user. The note “Blockchain must be synchronized” is included in the app description & FAQ. 

I would deploy the current version and wait for user feedback to see how many users have a problem with it.

bassin/docker-compose.yml
```diff
@@ -0,0 +1,69 @@
+version: '3.7'
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: bassin_www_1
+      APP_PORT: 80
+      PROXY_AUTH_ADD: "false"
+    container_name: bassin_app_proxy_1
```
al-lac: Is generated automatically by umbrelOS.
```suggestion
```

bassin/docker-compose.yml
```diff
@@ -0,0 +1,69 @@
...
+    depends_on:
+      - init
+    restart: on-failure
+    stop_grace_period: 30s
+    ports:
+      - "3456:3333/tcp"
+    volumes:
+      - ${APP_DATA_DIR}/data/config:/config:ro
+      - ${APP_DATA_DIR}/data/www:/www
+    container_name: bassin_ckpool_1
```
al-lac: Is generated automatically by umbrelOS.
```suggestion
```

bassin/docker-compose.yml
```diff
@@ -0,0 +1,69 @@
...
+    user: "1000:1000"
+    environment:
+      - FOLDER=/www
+      - PORT=80
+    volumes:
+      - ${APP_DATA_DIR}/data/www:/www:ro
+    depends_on:
+      - ui
+    restart: on-failure
+    container_name: bassin_www_1
```
al-lac: Is generated automatically by umbrelOS.
```suggestion
```

bassin/docker-compose.yml
```diff
@@ -0,0 +1,69 @@
...
+      - APP_BITCOIN_RPC_USER=${APP_BITCOIN_RPC_USER}
+      - APP_BITCOIN_RPC_PASS=${APP_BITCOIN_RPC_PASS}
+      - APP_BITCOIN_ZMQ_RAWBLOCK_PORT=${APP_BITCOIN_ZMQ_RAWBLOCK_PORT}
+
+  ckpool:
+    image: ghcr.io/duckaxe/ckpool-solo:main@sha256:2c8b700e6b3b03537155199bb7de319539a62adbfeb87227eef9c886f06444f7
+    entrypoint: /bin/sh -c '/opt/ckpool/src/ckpool -k -B -c /config/ckpool.conf'
+    user: "1000:1000"
+    depends_on:
+      - init
```
al-lac: This makes sure that `ckpool` only gets started after `init` is done.
```suggestion
      init:
        condition: service_completed_successfully
```

duckaxe:
@al-lac Thank you very much for improvements. I didn't remove the `container_name: bassin_www_1` because the dashboard widget doesn't work after removing this line. I have adopted other suggestions.

al-lac:
Hey @duckaxe, thanks a lot for making the adaptions! 👍 

Really weird that it does not work without the `container_name: bassin_www_1`, maybe something going wrong with DNS.

Did you also check the issue with the ckpool image? Did not see a change here, so the app would still be broken on some amd64 devices.

duckaxe:
@al-lac the ckpool image was created by using the following command where _amd64_ is included:
`docker buildx build --platform linux/amd64,linux/arm64 -t ghcr.io/duckaxe/ckpool-solo:main --push .`

https://github.com/duckaxe/ckpool-solo

al-lac:
Hey @duckaxe,

Ok got it!

I think you would need to do something like this to make it compatible with processors that do not support AVX2.

Edit Dockerfile:
```
ARG CFLAGS=""
ARG CXXFLAGS=""

# Later:
RUN ./autogen.sh && \
    ./configure CFLAGS="${CFLAGS}" CXXFLAGS="${CXXFLAGS}" && \
    make
```

Build command:
```
docker buildx build --platform linux/amd64,linux/arm64 \   --build-arg CFLAGS="-O2 -march=x86-64 -mtune=generic" \   --build-arg CXXFLAGS="-O2 -march=x86-64 -mtune=generic" \   -t ghcr.io/…:main --push .
```

But really not sure if that would work out, don't have a lot of experience with that either. Maybe this is also something that has already been solved upstream or should be noted there.

duckaxe:
@al-lac No chance to create an image if the flag contains “-march=x86-64”; Docker kept crashing. However, I tried something else and created a new image. On [GitHub](https://github.com/duckaxe/ckpool-solo/pkgs/container/ckpool-solo), you can see that it is definitely a multi-OS for both platforms.

al-lac:
Hey @duckaxe, thanks for putting in the effort to try to fix this! 🙌 

We will try it with the new image and let you know if it works now.

nmfretz:
Hey @duckaxe thanks for the hard work to bring Bassin on the app store! This looks like a really great addition.

I've given this a test as well, and the avx2 instruction set issue that @al-lac mentioned is a deal-breaker which needs to be resolved before we go live.

The current `ckpool` image that you're building is indeed multi-architecture (amd64 + arm64), however it looks like the amd64 image is somehow built to REQUIRE the avx2 instruction set. This means that for any users running umbrelOS on an amd64 machine whose CPU doesn't support the avx2 instruction set, the app will be broken. This will be the case for thousands of umbrelOS users, not just an edge-case.

As an example, when I go to install the app on an Umbrel Home with a N5105 CPU the `ckpool` container errors out with:

```
bassin_ckpool_1     | Illegal instruction
bassin_ckpool_1     | Terminated
```

It may be worth you reaching out the the ckpool devs to see if they already have a solution for this. It should be possible to only use avx2 when it's detected. I think that's the recommended way for most applications to use it (I think that's what Ollama does as an example).

nmfretz:
Also @duckaxe, for this:
> I didn't remove the container_name: bassin_www_1 because the dashboard widget doesn't work after removing this line. I have adopted other suggestions.

Removing the hardcoded `container_name: bassin_www_1` from the compose file in this repo works fine for me. The widget server still communicates correctly. As @al-lac mentioned, umbrelOS will automatically name the www container to `bassin_www_1` :
https://github.com/getumbrel/umbrel/blob/0f4b0c306749b5307f8fc10cfee29e0f88092625/packages/umbreld/source/modules/apps/app.ts#L115-L117

i.e., umbrelOS itself will add the `container_name: bassin_www_1` line to the compose file when the app is installed.

Would you mind testing again to confirm on your end?

duckaxe:
@nmfretz Thank you for your reply and for being willing to help me.

Give me another chance. I have now created both images manually and combined them into a multi-image. I have referenced the new image in the compose file.

If that doesn't work, I will close the PR until I have found a solution.

Regarding the `container_name`: I have just tried it. If I remove the container name, neither the widget nor the UI are accessible.

duckaxe:
@nmfretz @al-lac I installed umbrelOS on my **Intel** Mac (amd64). Bassin runs flawlessly with the latest CKpool image.

nmfretz:
Hey @duckaxe thanks for giving it another shot! The same issue persists unfortunately. Here's a screen recording with sound for you to watch to hopefully help clear things up. Let me know if anything is confusing:

https://github.com/user-attachments/assets/cc4212dd-2af9-4763-a09e-23a9a9126339

duckaxe:
@nmfretz I have found a solution. I am using the ckPool image, which I have previously used in the community app and which is compatible with all architectures. This solves the problem.

Sorry for the inconvenience. I hope that all the issues have now been resolved.

nmfretz:
> @nmfretz I have found a solution. I am using the ckPool image, which I have previously used in the community app and which is compatible with all architectures. This solves the problem.

Excellent, working great for me! It's very likely that image is built to optionally use avx2 when it is available.

Let's get Bassin to the official app store 🚀. @al-lac this is good to go after a final look through and gallery assets are finalized.

duckaxe:
@nmfretz Awesome, thank you!
@al-lac Regarding gallery assets: Yesterday I made a few adjustments (just replaced the bitcoin address by a dummy address to avoid any misunderstandings). If you have time, feel free to update them. Thanks!

duckaxe:
@al-lac The assets update isn't important. I prefer to merge Bassin now instead of waiting for assets update. Thank you.

al-lac:
Hey @duckaxe, thanks for all the effort you put into this app, we appreciate it a lot! 🥇 

Did a final test, everything works well. We can finally go live now! 🚀 

CleanShot 2025-11-14 at 15 24 35@2x

duckaxe:
@al-lac @nmfretz Thank you for your patience, time, and fantastic support!

nmfretz:
Hey @duckaxe, great work again on Bassin! I've been having fun with it recently. Love that you have the best share shown in the widget which I have an addiction to checking.

Would you be okay to change the coinbase tag to something like `mined by Bassin on Umbrel`? Currently Public Pool does `Public Pool on Umbrel` and DATUM does `DATUM on Umbrel`. It's a pretty great feeling for the community when it's clear that it was a little guy running an Umbrel that mined a block.

duckaxe:
@nmfretz Good idea. I'm planning a new release for the next days, so I can easily adjust the btcsig.

nmfretz:
Amazing @duckaxe. Looking forward to the new release!

PR #3376, opened 2025-08-07
Rusty Kaspad update

al-lac:
Disable rusty kaspad for now as apps are not allowed access to the docker socket.


al-lac:
Hey @elldeeone,

had to disable the app again as I missed the fact that it had access to the docker socket. Could you please update the frontend to be able to get the logs without requiring the docker socket?

elldeeone:
@al-lac sorry, i've now updated and pushed to remove the usage of docker socket. Could you please check?

PR #3407, opened 2025-08-07
App Submission: Ziit

0PandaDEV:
### App Name 

Ziit

### SVG Logo

https://ziit.app/logo.svg

### Gallery Images

Dashboard
Login
Settings

### Tested on

- MacOS using VM


ziit/hooks/pre-start
```diff
@@ -0,0 +1,25 @@
...
+    NEW_PASETO_KEY=$(openssl rand -base64 32)
+    sed -i "s|^\( *NUXT_PASETO_KEY: *\).*$|\1\"k4.local.${NEW_PASETO_KEY}\"|" "$ENV_FILE"
+fi
+
+if [[ "$CURRENT_PASSWORD" == "CHANGEME" ]] || [[ "$CURRENT_PASSWORD" == "OUTPUT_OF_OPENSSL_COMMAND" ]]; then
+    NEW_PASSWORD=$(openssl rand -hex 16)
+    sed -i "s|^\( *POSTGRES_PASSWORD: *\).*$|\1${NEW_PASSWORD}|" "$ENV_FILE"
+    
+    sed -i "s|^\( *NUXT_DATABASE_URL: *\).*\(#.*\)\?$|\1\"postgresql://postgres:${NEW_PASSWORD}@postgres:5432/ziit\"\2|" "$ENV_FILE"
+fi
```
al-lac: Adapt pre-start script.
```suggestion
#!/usr/bin/env bash
set -euo pipefail

# This scripts checks the environment file for placeholder values and replaces them with generated keys or system variables.

APP_DIR="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/..")"
ENV_FILE="$APP_DIR/settings.env"

# Check if the file exists
if [ ! -f "$ENV_FILE" ]; then
    echo "Error: Environment file not found at $ENV_FILE"
    exit 1
fi

# Check if the NUXT_HOST value is set to "changeme"
if grep -Eq '^NUXT_HOST= *changeme' "$ENV_FILE"; then
    echo "Placeholder value for NUXT_HOST detected. Adding proper hostname..."

    # Replace the placeholder with the hostname
    sed -i -E 's|^NUXT_HOST= *changeme( *#.*)?$|NUXT_HOST="'"http://${DEVICE_DOMAIN_NAME}:8865"'" \1|' "$ENV_FILE"

    echo "Updated NUXT_HOST in $ENV_FILE"
else
    echo "NUXT_HOST is already set. No changes made."
fi

```

ziit/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+  app:
+    image: ghcr.io/0pandadev/ziit:v1.0.2@sha256:ad8d23ea1a5927f85b0ab36f77c0891f1490794e1baf1c719d8cdf18ca8594f5
+    environment:
+      NUXT_DATABASE_URL: "postgresql://postgres:CHANGEME@postgres:5432/ziit" # change the password for production use
+      NUXT_PASETO_KEY: "k4.local.OUTPUT_OF_OPENSSL_COMMAND" # -> openssl rand -base64 32
+      NUXT_GITHUB_CLIENT_ID: "" # Github client id
+      NUXT_GITHUB_CLIENT_SECRET: "" # Github client secret
+      NUXT_GITHUB_REDIRECT_URI: "" # Your redirect url for the github auth e.g. https://ziit.example.com/api/auth/github/callback
+      NUXT_HOST: "http://${DEVICE_DOMAIN_NAME}:3009" # The domain where ziit will be hosted e.g. ziit.example.com
+      NUXT_DISABLE_REGISTRATION: "false" # By default users are allowed to register a new account
```
al-lac: Move some env vars to a file and use generated ones.
```suggestion
      NUXT_DATABASE_URL: "postgresql://postgres:umbrelziit@ziit_postgres_1:5432/ziit" # change the password for production use
      NUXT_PASETO_KEY: ${APP_ZIIT_PASTEO_KEY}
    env_file:
      - ${APP_DATA_DIR}/settings.env
```

ziit/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+    restart: on-failure
+    depends_on:
+      - postgres
```
al-lac: Use healthcheck.
```suggestion
    depends_on:
      postgres:
        condition: service_healthy
```

ziit/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+
+  postgres:
+    image: timescale/timescaledb:latest-pg17@sha256:c9a5107ed594445a5d1ab581711c22de5fd8ca42b1a0d6844e85507b18166bf8
+    restart: on-failure
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: CHANGEME # change this for production use
```
al-lac: You can use a fixed password here.
```suggestion
      POSTGRES_PASSWORD: umbrelziit
```

ziit/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+      POSTGRES_DB: ziit
+    volumes:
+      - ${APP_DATA_DIR}/data/postgres:/var/lib/postgresql/data
```
al-lac, line 28: Add healthcheck.
```suggestion
    volumes:
      - ${APP_DATA_DIR}/data/postgres:/var/lib/postgresql/data
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U postgres"]
      interval: 5s
      timeout: 5s
      retries: 5
```

ziit/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+
+volumes:
+  postgres:
```
al-lac: Remove volumes.
```suggestion
```

ziit/umbrel-app.yml
```diff
@@ -0,0 +1,21 @@
+manifestVersion: 1
```
al-lac: Use manifest version 1.1 to enable hooks.
```suggestion
manifestVersion: 1.1
```

ziit/umbrel-app.yml
```diff
@@ -0,0 +1,21 @@
...
+repo: https://github.com/0PandaDEV/Ziit
+support: https://github.com/0PandaDEV/Ziit/issues
+port: 3009
+gallery: []
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+dependencies: []
+submitter: PandaDEV
+submission: https://github.com/getumbrel/umbrel-apps/pull/3407
```
al-lac: Add final newline.
```suggestion
submission: https://github.com/getumbrel/umbrel-apps/pull/3407

```

al-lac (changes requested):
Hey @0PandaDEV,

thanks for the submission! 

I added some things that should be changed.

A new file called `exports.sh` needs to be added with this content as well:
```bash
export APP_ZIIT_PASTEO_KEY=$(derive_entropy "${app_entropy_identifier}-ZIIT_PASTEO_KEY" | head -c32)
```

Also a `settings.env` file with this content:
```bash
NUXT_GITHUB_CLIENT_ID="" # Github client id
NUXT_GITHUB_CLIENT_SECRET="" # Github client secret
NUXT_GITHUB_REDIRECT_URI="" # Your redirect url for the github auth e.g. https://ziit.example.com/api/auth/github/callback
NUXT_HOST=changeme # The domain where ziit will be hosted e.g. ziit.example.com
NUXT_DISABLE_REGISTRATION="false" # By default users are allowed to register a new account
```

It would be nice if some of those settings like the github client id and secret could be set via the web ui as well, this would make the setup here way easier.

The app seems to fail at the first startup, only after a restart, it shows the login. But when trying to register, I get the following error:
```bash
2025-08-12T12:03:39.800Z Error 500: Registration failed: undefined is not an object (evaluating 'passwordValidation.error.errors[0]')
2025-08-12T12:03:52.611Z Error 500: Registration failed: undefined is not an object (evaluating 'passwordValidation.error.errors[0]')
```

al-lac:
@0PandaDEV thanks for adapting the changes! Still fails though when trying to register with the error I posted above.

0PandaDEV:
@al-lac Yes I know the problem is the token used for auth which was generated using the pre-start script before and now in the export.sh. It has the wrong format it should be k4.local.BASE64 string but the k4.local is missing now.

al-lac:
@0PandaDEV I added it back in again, however, I still get the same error after applying the fix.

0PandaDEV:
@al-lac  I get the following error from the docker container `Key must be 32 bytes long.`

al-lac:
Well, kind of makes sense, does the PASTEO_KEY have to be 32-characters including the `k4.local.`?

0PandaDEV:
No I think just the part after k4.local has to be 32 bytes but isn't the app_entropy_identifier the same as the $APP_SEED which is 256 bits / 8 = 32 bytes so it should technically work. Maybe because it's hex instead of base64, but that shouldn't matter anyways.

al-lac:
Hey @0PandaDEV, good catch! Adapted it to base64. Seems to work better now for me, but after logging in, it just seems to log me out instantly. Any ideas on what could be wrong here?

al-lac:
Hey @0PandaDEV, checking in. Did you manage to find out what my issue could be related to?

al-lac:
Hey @0PandaDEV, just checking in here, any update on the app?

0PandaDEV:
Hi sorry this PR completely drowned in my notifications I'm currently not at home, so I can't look into this right now. Do you still have the issue where it logs you out immediately?

al-lac:
Hey @0PandaDEV, no worries! Yup, still got the same issue.

al-lac:
Hey @0PandaDEV, I took another look at this, and the login seems to be failing as Ziit sets the login cookie with `secure: true`.

This will not work via a http URL like `http://umbrel.local`. 

So there would need to be an upstream fix from your side that enables the app to work on http (maybe via an env var), or we would need to wait until HTTPS support for Umbrel is here.

PR #3417, opened 2025-08-12
Update Robosats to v0.8.1-alpha

KoalaSat:


al-lac:
Thanks for the update @KoalaSat! 

Tested the update on my Umbrel dev instance.

PR #3446, opened 2025-08-14
Remove svgur.com from PR description example

duckaxe:
Since GitHub now allows SVG files to be attached to a PR, there is no longer a need for external services. In particular, https://svgur.com has been hacked. It should not be recommended.

Any other places like PR templates that recommend svgur.com should also be updated.


al-lac:
Hey @duckaxe, nice! 

I did not know that GitHub finally added SVG support, good catch! Thanks! 🔥

PR #3452, opened 2025-08-16
Alby Hub: lnd dependency replaced by dynamic backend detection getalby#1349

NodeDiver:
This PR allows users to install Alby Hub with an LDK backend directly without having the dependency of the LND (lightning) app. In case the LND is indeed installed, then the Alby Hub will install automatically using the LND as the backend.

Dynamic backend detection implemented in the `exports.sh` file, similar to mempool's app implementation.
- If the LND app (`lightning`) is installed, Alby Hub will be configured to use **LND**.
- If not, it will fallback to **LDK** as default.

Addresses [issue#1349](https://github.com/getAlby/hub/issues/1349)

### DIFF
--- a/albyhub/docker-compose.yml
+++ b/albyhub/docker-compose.yml
@@ -10,12 +10,13 @@ services:
     user: 1000:1000
     volumes:
       - ${APP_DATA_DIR}/data:/data
-      - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
+      # Using UMBREL_ROOT here to ensure exports.sh can read mounted LND certs/macaroons properly
+      - ${UMBREL_ROOT}/app-data/lightning/data/lnd:/lnd:ro
     environment:
-      LN_BACKEND_TYPE: "LND"
-      LND_ADDRESS: $APP_LIGHTNING_NODE_IP:$APP_LIGHTNING_NODE_GRPC_PORT
-      LND_CERT_FILE: "/lnd/tls.cert"
-      LND_MACAROON_FILE: "/lnd/data/chain/bitcoin/$APP_BITCOIN_NETWORK/admin.macaroon"
+      LN_BACKEND_TYPE: ${APP_ALBYHUB_LN_BACKEND:-LDK}
+      LND_ADDRESS: ${APP_ALBYHUB_LND_ADDRESS:-}
+      LND_CERT_FILE: ${APP_ALBYHUB_LND_CERT_FILE:-}
+      LND_MACAROON_FILE: ${APP_ALBYHUB_LND_MACAROON_FILE:-}
       WORK_DIR: "/data/albyhub"
       COOKIE_SECRET: ${APP_SEED}
       LOG_EVENTS: "true"
--- a/albyhub/exports.sh
+++ b/albyhub/exports.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+
+export APP_ALBYHUB_LN_BACKEND="LDK"
+export APP_ALBYHUB_LND_ADDRESS=""
+
+# Check if Lightning Node app is installed and export required variables if so
+installed_apps=$("${UMBREL_ROOT}/scripts/app" ls-installed)
+
+if echo "$installed_apps" | grep --quiet 'lightning'; then
+  export APP_ALBYHUB_LN_BACKEND="LND"
+  export APP_ALBYHUB_LND_ADDRESS="10.21.21.9:10009"
+  export APP_ALBYHUB_LND_CERT_FILE="/lnd/tls.cert"
+  export APP_ALBYHUB_LND_MACAROON_FILE="/lnd/data/chain/bitcoin/mainnet/admin.macaroon"
+fi
--- a/albyhub/umbrel-app.yml
+++ b/albyhub/umbrel-app.yml
@@ -39,11 +39,11 @@ releaseNotes: >-
     - Addressed password change authorization bug
     - Added migration button for isolated connections to sub-wallets
     - Included Wave.space in the app store
+    - Removed lightning(LND) dependency: LND + LDK backend auto-detection.
 
 
   Full release notes are found at https://github.com/getAlby/hub/releases
-dependencies:
-  - lightning
+
 path: ""
 defaultUsername: ""
 defaultPassword: ""


albyhub/docker-compose.yml
```diff
@@ -10,12 +10,13 @@ services:
     user: 1000:1000
     volumes:
       - ${APP_DATA_DIR}/data:/data
-      - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
+      # Using UMBREL_ROOT here to ensure exports.sh can read mounted LND certs/macaroons properly
```
bumi: I don't think exports.sh needs to or does read the certs/macaroon file. 

as `lightning/data/lnd` will likely not be existent if `lightning` is not installed will this work? or will it create the directory because it is mounted?

albyhub/exports.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+
+export APP_ALBYHUB_LN_BACKEND="LDK"
+export APP_ALBYHUB_LND_ADDRESS=""
```
bumi: why an empty export on this but not on the cert/macaroon variable? 
I think this should not be needed.

albyhub/docker-compose.yml
```diff
@@ -10,12 +10,13 @@ services:
...
+      - ${UMBREL_ROOT}/app-data/lightning/data/lnd:/lnd:ro
     environment:
-      LN_BACKEND_TYPE: "LND"
-      LND_ADDRESS: $APP_LIGHTNING_NODE_IP:$APP_LIGHTNING_NODE_GRPC_PORT
-      LND_CERT_FILE: "/lnd/tls.cert"
-      LND_MACAROON_FILE: "/lnd/data/chain/bitcoin/$APP_BITCOIN_NETWORK/admin.macaroon"
+      LN_BACKEND_TYPE: ${APP_ALBYHUB_LN_BACKEND:-LDK}
+      LND_ADDRESS: ${APP_ALBYHUB_LND_ADDRESS:-}
```
bumi: why not just `${APP_ALBYHUB_LND_ADDRESS}` which is probably the same as you don't set a default? or does it fail when the env variable is not existing at all?

albyhub/umbrel-app.yml
```diff
@@ -39,11 +39,11 @@ releaseNotes: >-
     - Addressed password change authorization bug
     - Added migration button for isolated connections to sub-wallets
     - Included Wave.space in the app store
+    - Removed lightning(LND) dependency: LND + LDK backend auto-detection.
```
rolznz: I'm not sure this should be included here?

albyhub/docker-compose.yml
NodeDiver: **About the volume for LND:**
I belive docker does not supports conditional volume mounts, and I have not found a way to bypase this and make some sort of contitional volume mount using the aid of the export file, so I left this volume to be mounted. When in backend type: LDK, the lnd folder will be created and will be empty if no lightning (lnd) app was installed ever.

The `APP_LIGHTNING_NODE_DATA_DIR` will not exist since we removed the lightning (lnd) dependency. We have to use another way to define the path, I used UMBREL_ROOT because it is already defined in the exports.sh

----

**About the certs/macaroon file:**
If we remove the cert/macaroon file from the exports and the compose file, the Hub will install and launch, but after clicking in "Get Started(LND)" we will have a "Failed to start: one or more required LND configurations are missing". We need those variables for the Hub with backend: LND. 

If you remove the certs/macaroon values in the exports.sh, and you place them back into the docker-compose file. The hub will install but will not launch; we will have a "This site can’t be reached: ERR_CONNECTION_REFUSED" error in the browser. I believe this is because the values for these variables can't really be hardcoded, they must have a conditional because when the backend is LDK then these values will not exist, and the variables will be Null
 
When we are at the LDK case and the  values of the certs/macaroon variables will not exist, to disregard those variables, In the docker compose file we need the :- at the end, so that if that value is not existant then the variable will be disregarded and its absence will not trigger errors. 

However, I believe we can't write:
`LND_CERT_FILE: ${"/lnd/tls.cert":-}`
`LND_MACAROON_FILE: ${"/lnd/data/chain/bitcoin/mainnet/admin.macaroon":-}`
Because, if I understand correctly, the conditional :- is only valid for variable names but not file paths. That means I have to wrap the file path of the cert and macaroon in a variable name, and I did that in the exports.sh  

Perhaps there is another way to to disregard the cert and macaroon variables when their values are absent directly from the compose file. I have not found another way.

albyhub/exports.sh
NodeDiver: Those lines are indeed not needed and are now removed. thanks 🙏

albyhub/docker-compose.yml
NodeDiver: It fails in my testing environment when the variable does not exist at all. The Hub will install but when we launch it we will have a "This site can’t be reached: ERR_CONNECTION_REFUSED" error in the browser.

albyhub/umbrel-app.yml
NodeDiver: (removed, 👌)

albyhub/exports.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Check if Lightning Node app is installed and export required variables if so
+installed_apps=$("${UMBREL_ROOT}/scripts/app" ls-installed)
+
+if echo "$installed_apps" | grep --quiet 'lightning'; then
+  export APP_ALBYHUB_LN_BACKEND="LND"
+  export APP_ALBYHUB_LND_ADDRESS="10.21.21.9:10009"
```
bumi, line 8: for the umbrel reviewer: Is this reliable? or is there anyway to get the IP:PORT dynamically?

NodeDiver: The standard dynamic variables: `APP_LIGHTNING_NODE_IP` & `APP_LIGHTNING_NODE_GRPC_PORT `are used in many other apps. But I interpreted that they can only be used if lightning (LND) is a dependency..

The lightning (LND) app always uses 10.21.21.9 in Umbrel and port 10009 is LND's standard gRPC port, so this should be safe to use. This IP and port are also used  similarly by the mempool app in their own export.sh script.

albyhub/docker-compose.yml
nmfretz: What you guys have done here is fine:

```yml
- ${UMBREL_ROOT}/app-data/lightning/data/lnd:/lnd:ro
```

If the user doesn't have the Lightning Node app installed then Docker will create the bind mount at `${UMBREL_ROOT}/app-data/lightning/data/lnd` on the host with root priveledges. It's not super ideal because then a user without Lightning Node installed will have that path on their umbrel (although it will be empty). But it doesn't have other negative consequences, because if a user goes to install Lightning Node later on, umbrelOS will properly fix the permissions for Lightning Node's data directory.

albyhub/exports.sh
nmfretz: > The standard dynamic variables: APP_LIGHTNING_NODE_IP & APP_LIGHTNING_NODE_GRPC_PORT are used in many other apps. But I interpreted that they can only be used if lightning (LND) is a dependency.

Yep, that is correct. An app will only be able to source another app's variables if that app is listed as a dependency in its umbrel-app.yml file. So even though there is this guard...

```bash
if echo "$installed_apps" | grep --quiet 'lightning'; then
```

...any reference to `APP_LIGHTNING_NODE_IP` or `APP_LIGHTNING_NODE_GRPC_PORT ` within that block would still fail even though the Lightning Node app is installed, because umbrelOS would not allow albyhub to source the lightning node's exports.sh file (where those variables are defined).

So what you have here is good.

nmfretz:
Hey @NodeDiver and @bumi, thanks for this PR to get Alby Hub to conditionally work with LND or LDK.

And @NodeDiver, really great work digging in and understanding the docker bind mount and exports.sh variable nuances 👌. I've added some comments to your discussions above with some additional details.

In the future we plan to have more expressive dependencies on umbrelOS, to do things like allow apps to list "optional dependencies". For example, if LND exists then ask the user if they want to integrate with it, but if it doesn't exist then the app can function without it.

For now though, the workaround you're using is fine because we already have precendence for it with the Mempool app.

There's a couple things we need to think about before going live with this:

### 1) Test Networks

Currently, Alby Hub has `lightning` as a dependency:

https://github.com/getumbrel/umbrel-apps/blob/e8241d1fcd7bbd2143d7b4c654b85d4e0d8d26bf/albyhub/umbrel-app.yml#L45-L46

And lightning in turn has `bitcoin` as a dependency:

https://github.com/getumbrel/umbrel-apps/blob/e8241d1fcd7bbd2143d7b4c654b85d4e0d8d26bf/lightning/umbrel-app.yml#L34-L35

Which means that the current version of Alby Hub on the app store always has access to the `APP_BITCOIN_NETWORK` variable from any bitcoind app.

This means that Alby Hub ends up automatically working with any bitcoin network that the user is currently running their Bitcoin node + Lightning node on because the compose file currently uses this:

https://github.com/getumbrel/umbrel-apps/blob/e8241d1fcd7bbd2143d7b4c654b85d4e0d8d26bf/albyhub/docker-compose.yml#L18

But with this PR removing the `lightning` dependency (and by association the `bitcoin` dependency as well) we'll no longer get access to `APP_BITCOIN_NETWORK`, which is why this line you have now hardcodes "mainnet":

```sh
export APP_ALBYHUB_LND_MACAROON_FILE="/lnd/data/chain/bitcoin/mainnet/admin.macaroon"
```

But this will break any current user's setups who are running on test chains like signet, testnet4, etc. When they update to this latest version, Alby Hub would be broken for them because the macaroon filepath would be incorrect.

 It would also result in a broken install for any new installs of Alby Hub for users who are running their bitcoin/lnd setups on test networks.


### 2) Behaviour when a user installs LND after AlbyHub

How should Alby Hub handle this case:

1) user does not have LND installed
2) user installs Alby Hub and clicks to configure the LDK backend
3) Later, the user installs Lightning Node (LND). Now when alby hub restarts (manually or via update) it will source the variables from LND.

bumi:
> 1) Test Networks

Any thoughts what should be done here? should it check the `/lnd/data/chain/bitcoin/mainnet` directory? 
I hoped people don't really do that ;)

> 2) Behaviour when a user installs LND after AlbyHub

a change of LN_BACKEND_TYPE env variable will be ignored. so once it's setup with either LDK or LND and then this env variable gets changed it doe not have any effect.

albyhub/exports.sh
NodeDiver: Thanks for the clarification!

nmfretz:
> a change of LN_BACKEND_TYPE env variable will be ignored. so once it's setup with either LDK or LND and then this env variable gets changed it doe not have any effect.

Sweet, that's great since the worst-case scenario we want to avoid is a user setting up LDK on purpose, but then thinking they've lost access to their funds if they later install LND on their machine and it auto-configures their Alby Hub.

Let's update the app description in the `umbrel-app.yml` to mention LDK vs LND to help users out a bit, since now the app won't be as fool-proof for install. Essentially, we want to avoid scenarios like the following as much as possible so Umbrel and Alby don't get increased support requests:

- User installs Alby Hub **BEFORE** Lightning Node, which is now possible without `lightning` as a dependency.
- User then might realize they want to connect to LND so they install Lightning Node.
- But Alby Hub won't auto-configure to LND until Alby Hub is restarted so it can source the required env vars.

We might also get confused users clicking the LDK option, then realizing they messed up and then needing to know to re-install Alby Hub to configure it how they want.

I'll update the manifest for you guys to review.

> Any thoughts what should be done here? should it check the /lnd/data/chain/bitcoin/mainnet directory?
I hoped people don't really do that ;)

Ya I think it's pretty low probability that anyone is running Alby Hub on a test network on umbrelOS right now, since they'd need to be running a Bitcoin node and Lightning node on a test network as well with test network lighting channels which seems quite unlikely to me. Do you guys have any stats on whether there are users running on test networks?

I think the path forward here is just to hardcode to `/lnd/data/chain/bitcoin/mainnet` since there isn't a very reliable way for us to determine the network without `lightning` set as a dependency. We can include a warning at the top of these release notes that out-of-the-box compatibility with test networks is no longer supported. This will likely be solved in the future when umbrelOS has "optional dependencies".

nmfretz:
@NodeDiver @bumi do these setup instructions look okay to you?

https://github.com/getumbrel/umbrel-apps/pull/3452/commits/8b0e3113e38e357165a699e0882c091a794e3fb8
image

nmfretz:
@NodeDiver @bumi - This isn't a blocker for release at all, but just a consideration for you guys:

It could be good to always show the LDK option during the configuration screen regardless of the value of `LN_BACKEND_TYPE`. Again, I think "optional dependencies" on umbrelOS will solve this for you eventually when we implement it, because a user would be prompted by umbrelOS on install of Alby Hub to connect to LND if they want. But the way this works now is that if a user has Lightning Node installed on their Umbrel then they **ONLY** have the option of connecting Alby Hub to it, since Alby Hub won't show them the LDK option. So this combo is not currently possible:

`User has LND installed + User wants to use Alby Hub with built in LDK`

What do you guys think?

NodeDiver:
Thanks, @nmfretz the app description text looks good to me. A few confirmations and tweaks:

**Confirmaitons:**
• On the first run right after the Hub was installed, the start button shows “Get Started (LDK)” or “Get Started (LND)” based on what’s detected. The user will see which node is being installed. A user wanting to install LND when LND was not installed, might see that something is not as it should be, seeing "Get Started (LDK)" should be an indicator for the user that something is not as the user desires. We can only hope after seeing this, the user will go back to the app store and read the description and understand lnd needs to be installed first, exactly as point nº2 you wrote instructs it.

• Installing LND later: Alby Hub won’t auto-switch once configured (LN_BACKEND_TYPE is effectively locked). If a user installs Lightning Node afterward and wants LND, they should reinstall Alby Hub; to change backends, This was also correctly explained on the point nº2 of the setup instructions.

**Things to add:**
" `User has LND installed + User wants to use Alby Hub with built in LDK
` "
• Installing LDK when LND is present: It is already possible with the current built. The use just have to enter the “Advanced mode” in the setup, this allows choosing other backends even if LND is installed, including choosing LDK. 
_Note: I believe both LND and LDK uses the same port 9735 by default but for different uses cases, I'm not sure if or how this could be a conflict CC @bumi ._

• I tried to think of ways to detect the mainnet/testnet/signet routes, but I think anything can overcomplicate things. I guess we'll state that testnet/signet aren’t supported on Umbrel out of the box. Hopefully, users wanting to use testnet/signets might possibly be savvy enough to run the Docker stack of Alby Hub locally for testing/developing. We could add a phrase at the bottom of the app description stating Alby Hub only works with mainnet in Umbrel and devs that used Alby Hub with signet or testnet are invited to use the Hub with signet/testnet only outside umbrel, in dev environment. Alby's github repository can be checked for instructions.⚠️ If someone happens to contact us complaining they lost the ability to use the Hub on testnet, I will personally help them mount a local environment for testing & experimenting.

Thank you all, I appreciate the time you've put into this.

zmjohnso:
I was just looking for a way to run LDK with Alby Hub on Umbrel. This will be awesome to have. Thanks for working on it all!

NodeDiver:
Hey @nmfretz , is there anything else that needs to be discussed before this gets merged? Thanks!

nmfretz:
Thanks for the ping @NodeDiver! And thanks for the additional info.

> Note: I believe both LND and LDK uses the same port 9735 by default but for different uses cases, I'm not sure if or how this could be a conflict CC @bumi .

This should not be an issue because your LDK port 9735 would be inside the container only and not exposed on the host so there would be no conflict with LND (LIghtning Nodea app) also running on a user's Umbrel at the same time if they had it installed.

> I tried to think of ways to detect the mainnet/testnet/signet routes, but I think anything can overcomplicate things. I guess we'll state that testnet/signet aren’t supported on Umbrel out of the box.

Okay, I can add a warning to both the release notes and app description noting that only mainnet is supported when connected to the Lightning Node app specifically. Later down the line when umbrelOS supports optional dependencies we can reliably support test networks for Alby Hub.

**Edit:** on reflection I don't think it is worth a warning here as running alby hub on a test network is likely not being done on umbrelOS, does not have monetary value, and extra warnings would just add confusion for users.

Does this all sound okay to you @NodeDiver and @bumi?

NodeDiver:
I'm good! It all sounds okay for me!
Later down the line, when Umbrel supports optional dependencies, we could adjust things to support test networks again. 🙌

nmfretz:
Thanks again @NodeDiver! I have:
- added release notes, as well as a comment to the exports file noting that test network compatibility can be revisited once optional dependencies are added to umbrelOS.
- merged in changes from master so we are on the latest Alby Hub version 1.20.0
- re-tested

Thanks again for all your efforts getting this done 🤝

PR #3504, opened 2025-08-22
Update wger CSRF list to enable access via IPs

al-lac:
This enables wger to be accessed via the local ip of umbrel or also via tailscale magic dns if the hostname is set to `umbrel` there.


al-lac:

PR #3525, opened 2025-08-26
Update Monero daemon to v0.18.4.2

colder-is-better:
Monero 0.18.4.2 (Fluorine Fermi) has been released, so we upgrade the daemon in the corresponding Umbrel app accordingly.


al-lac:
Hey @colder-is-better,

thanks for this monero update! 💸 

Tested it on my Umbrel dev instance and everything is working as it should. Waiting on a final go from @nmfretz or @sharknoon before merging.

nmfretz:
Awesome, thanks for these regular updates @colder-is-better. Tested and working great. Going live.

PR #3561, opened 2025-09-04
update Bitcoin Knots to v29.1

Retropex:


nmfretz:
Thanks for the update @Retropex! Tested fresh install and app update. Going live.

PR #3567, opened 2025-09-09
Bump ThunderHub

apotdevin:


nmfretz:
Awesome @apotdevin, thanks for always keeping ThunderHub up to date on the app store 🤝

Reviewed and tested. Going live.

PR #3571, opened 2025-09-10
Update public-pool-web to version 0.1.0

martinbarilik:
Hi,

more fixes.

Thanks in advance


al-lac:
Thanks @martinbarilik 🙌 Appreciate you keeping public-pool-web updated in the app store.

Reviewed and tested. Going live now 🚀

al-lac (approved):

PR #3572, opened 2025-09-10
Update toshi-moto to v1.31.1 - support for z-pub and y-pub

toshimoto821:
fixes https://github.com/toshimoto821/toshi-moto/issues/306


al-lac:
Awesome @toshimoto821  🎉 Thanks for making sure toshi-moto stays current in the app store.

Everything’s reviewed and tested. Rolling it out 🚀

PR #3658, opened 2025-09-23
Allow Jellyfin to access Videos folder

Zac0511:
Jellyfin being a an app made to host video files, it would make sense to be able to put my video files directly inside UmbrelOS's video folder, instead of somewhere in the app data files.

Note: i don't know if `${UMBREL_ROOT}/data/storage/videos:/videos` is the correct thing to point to the videos folder, so please correct it if its wrong.

### DIFF
--- a/jellyfin/docker-compose.yml
+++ b/jellyfin/docker-compose.yml
@@ -17,7 +17,7 @@ services:
     volumes:
       - ${APP_DATA_DIR}/data/config:/config
       - ${UMBREL_ROOT}/data/storage/downloads:/downloads
-      - ${UMBREL_ROOT}/data/storage/videos:/videos
+      - ${UMBREL_ROOT}/home/Videos:/videos
     ports:
       # Service auto-discovery
       - 7359:7359/udp


Zac0511:
I modified the `docker-compose.yml` file locally, and it seems like `${UMBREL_ROOT}/data/storage/videos:/videos` does **not** lead to the Video folder.
I'l try with `${UMBREL_ROOT}/home/umbrel/umbrel/home/Videos:/videos`

Zac0511:
Okay it seems to work with `${UMBREL_ROOT}/home/Videos:/videos`, i'l make a new commit.

Zac0511:
Should work now, tho i dont understand while the `${UMBREL_ROOT}/data/storage/downloads:/downloads` in the App-Store version of the file is a `${UMBREL_ROOT}/home/Downloads:/downloads` in my Umbrel's file, even tho i installed Jellyfin litterally 30 minutes ago and never edited the file.

al-lac:
Hey @Zac0511, thanks for coming up with this great idea for letting Jellyfin access the video folder. Appreciate you contributing here. 🏅 

After some discussion with the team, we decided not to merge this right now as it would cause issues if the folder does not exist in addition to other things that would need to be considered.

Be assured, another concept on how this can be handled in the future is already in the works! 👀

PR #3699, opened 2025-09-29
Update fulcrum to v2.0

nmfretz:
https://github.com/cculianu/Fulcrum/releases/tag/v2.0.0

From Fulcrum's official release notes:
>#### Upgrade Instructions
>Fulcrum 2.0 has a different database format than the 1.x series. As such, you have two options for upgrading:
>
>Create a blank new datadir and just re-synch from block 0 (slow, but 100% reliable)
Upgrade your existing 1.x datadir (faster, but irreversible and destructice)
To upgade your existing database, be sure to pass the one-time --db-upgrade flag to Fulcrum. It will refuse to start up if it detects that your datadir is in the old format and you did not pass this flag. This ensures that admins know what they are getting into.
>
>Note: The upgrade process takes seconds to minutes on BCH and BTC testnet(s), maybe ~15 minutes on BCH mainnet, and around an hour or more (depending on hardware) on BTC mainnet.

>#### Critical Note On Upgrading the DB from 1.x -> 2.0
>Let it run to completion!
>
>While Fulcrum 2.0 is 100% reliable and may be killed at any time during its operation -- there is one exception to this rule: the upgrade process that converts 1.x databases -> 2.0. As such please let the upgrade process run to completion and do not kill, stop, molest, or otherwise abuse the Fulcrum process while it's upgrading the DB!!. If you do, the old 1.x database will be lost and your 2.0 database will also be corrupted.
>
>You have been warned!

I don't think there is a simple and reliable way for us to go the database upgrade route for users. This would involve a one-time container startup with `--db-upgrade`, then allow migration to run fully with no user feedback other than the app stuck in "starting...", and then bringing up the app normally after migration is complete. The migration on mainnet can take hours, AND during that time if the user restarted the app or their Umbrel for any reason they would fully brick their Fulcrum app (would lose 1.x database and corrupt 2.0 database) requiring a re-install and full re-sync.

So the simplest, bomb-proof route for us here is to delete index data for pre-2.0 installs which will automatically force a re-index which apparently won't be corrupted during start/stop even during index anymore due to 2.0 db changes.

I've done this super simply with a flag file we set for new installs (empty fulcrum data dir) and for existing installs one-time after deleting the contents of the fulcrum data dir.

Other changes:
- reverted https://github.com/getumbrel/umbrel-apps/pull/2894 changes that resulted in a broken UI (no fulcrum.log)
- removed legacy post-update script
- updated the backupIgnore filepaths to take into account Fulcrum 2.0 structure:

image

### DIFF
--- a/fulcrum/docker-compose.yml
+++ b/fulcrum/docker-compose.yml
@@ -29,7 +29,9 @@ services:
   fulcrum:
     image: cculianu/fulcrum:v2.0.0@sha256:cb1c006d0cff104696f4791d0f1516699b2c163120165461385e4de206271943
     init: true
-    stop_grace_period: 1m
+    # Rely on Docker's default SIGTERM; Fulcrum treats SIGTERM/SIGINT/SIGQUIT as graceful
+    # Generous grace period prevents SIGKILL during checkpoint/flush on exit
+    stop_grace_period: 5m
     restart: on-failure
     user: "1000:1000"
     environment:
@@ -41,7 +43,12 @@ services:
       PEERING: "false"
       ANNOUNCE: "false"
       DATA_DIR: /data
-    command: sh -c 'Fulcrum -D /data _ENV_ 2>&1 | tee /logs/fulcrum.log'
+    # Exec Fulcrum as PID1 so it receives SIGTERM directly; FIFO+tee mirrors output to docker logs and /logs/fulcrum.log
+    # rm -f makes FIFO creation idempotent across restarts/crash loops
+    command: >
+      sh -c 'rm -f /tmp/fulcrum.pipe; mkfifo /tmp/fulcrum.pipe;
+      tee -a /logs/fulcrum.log < /tmp/fulcrum.pipe &
+      exec Fulcrum -D /data _ENV_ > /tmp/fulcrum.pipe 2>&1'
     volumes:
       - "${APP_DATA_DIR}/data/fulcrum:/data"
       - "${APP_DATA_DIR}/data/fulcrum-logs:/logs"
--- a/fulcrum/docker-compose.yml
+++ b/fulcrum/docker-compose.yml
@@ -29,8 +29,7 @@ services:
   fulcrum:
     image: cculianu/fulcrum:v1.12.0.1@sha256:ee6a67224a766448507eddc0efba8649b5598f1ea3def39673e2a08210ae4921
     init: true
-    stop_signal: SIGINT
-    stop_grace_period: 5m
+    stop_grace_period: 1m
     restart: on-failure
     user: "1000:1000"
     environment:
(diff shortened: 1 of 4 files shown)


al-lac:
Tested by doing a fresh install of Fulcrum `1.12.0` and updating to `2.0.0`. Everything works fine afterwards:
```
Sep 29 09:43:42 umbrel-dev umbreld[126614]: Executing hook: /home/umbrel/umbrel/app-data/fulcrum/hooks/pre-start
Sep 29 09:43:42 umbrel-dev umbreld[126690]: App: fulcrum - Removing '/home/umbrel/umbrel/app-data/fulcrum/hooks/post-update'
Sep 29 09:43:42 umbrel-dev umbreld[126690]: App: fulcrum - Pre-2.0 data detected, clearing '/home/umbrel/umbrel/app-data/fulcrum/data/fulcrum' once
```

nmfretz:
Thanks for testing @al-lac! 

I've also fixed the fulcrum service command so that Fulcrum actually receives the default SIGTERM from compose when shutting down https://github.com/getumbrel/umbrel-apps/pull/3699/commits/a82eb54b762da925c836699b683be4547f585a31

Previously I had the container running `sh -c 'Fulcrum ... | tee ...'`, so Docker’s SIGTERM hit PID1→sh, not Fulcrum so it was probably getting SIGKILLed. And we can't use the last PRs command (https://github.com/getumbrel/umbrel-apps/pull/2894) because it breaks the logging we need for the UI to be reliable.

IIIspaceIII:
@nmfretz 
with the update to v2.0.0 there is a flag that will migrate the db 
from the release description -> https://github.com/cculianu/Fulcrum/releases/tag/v2.0.0

"Added --db-upgrade CLI arg (required to upgrade an existing DB from Fulcrum 1.x to 2.0 format)"

nmfretz:
> with the update to v2.0.0 there is a flag that will migrate the db

Hey @IIIspaceIII, ya an easy data migration would have been nice, but the risks of users ending up in a completely broken state were quite high. See my comments on it above https://github.com/getumbrel/umbrel-apps/pull/3699#issue-3463185606.

Hopefully this is just a one-time painful re-sync 🤞

lkroll:
Thanks for the update.
Because I didn't want to wait 1-2 weeks for a full resync on my RP4 I created the `POST_2_0_VERSION` file manually before updating and then added the `--db-upgrade` flag to the Fulcrum command in the docker-compose file.
After a restart it is now updating and should take less time :)
A bit cumbersome but doable for advanced users.
Maybe it is possible to add instructions for that in the Changelog?
Or maybe it is too late now and most people are fine with a full resync.

UPDATE: The container got killed by oom-kill so apparently 4GB on the RP4 is not enough for the db-upgrade command. Will open an issue on the Fulcrum repo for that.

PR #3700, opened 2025-09-29
Update Lightning Node v0.19.3

nmfretz:


nmfretz:
Tested on amd64 and arm64, both fresh install and app update.

PR #3770, opened 2025-10-10
Bump monerod ver

colder-is-better:
This is a highly recommended release that enhances protection against spy nodes.


al-lac:
Hey @colder-is-better,

thanks a lot for keeping the monero app up-to-date in the Umbrel App Store! 🥇 

Tested the update on my Umbrel dev instance and everything worked perfectly. Going live!

PR #3778, opened 2025-10-12
Proposed fix for redirect to BTCPay login/admin page

jimbrend:
Heya friends,

I noticed after a recent reinstall navigating to http://umbrel.local:3003/ on various browsers taking me to app page
and not the admin page from the dashboard,

umbrel.local:3003/apps/btcpay-server/login
umbrel.local:3003/apps/btcpay-server/admin

Also those don't resolve and show the 404 server page, but umbrel.local:3003/login works just fine every time and loads the admin page.

Proposing the change to compose file for upstream true if this may resolve upstream routing for those pages, and umbrel-app file for redirect

The /login page works fine each time when instead entered manually to get to admin page once already logged in each time (http://umbrel.local:3003/admin 404s as well currently), 

Hopefully useful will share anything I come across to help promote as seamless experience as possible :), i.e. users clicking through on dashboard so it takes to admin page every time, mine currently defaults to http://umbrel.local:3003/ which loads the default app/storefront that was created instead of admin page...

Thank you!

### DIFF
--- a/btcpay-server/umbrel-app.yml
+++ b/btcpay-server/umbrel-app.yml
@@ -25,6 +25,7 @@ dependencies:
   - lightning
 repo: https://github.com/btcpayserver/btcpayserver
 support: https://chat.btcpayserver.org
+port: 3003
 
 proxy:
   path: "/apps/btcpay-server"


al-lac:
Hey @jimbrend,

I tried to replicate this issue, but I always get sent to either the login or the admin dashboard. Maybe @nmfretz or @sharknoon can replicate this on their instances.

jimsbr:
hmm, yeah no rush!
It does it still on my home network loads up app page first on Chrome, Safari, and mobile so i have to manually add the /login

I threw in my current Umbrel and BTCPayServer logs below here if it helps for review at all later, cheers!

[umbrel_btcpay-server_2025-10-13_12-56.log](https://github.com/user-attachments/files/22889456/umbrel_btcpay-server_2025-10-13_12-56.log)

[umbrel-1760374607032.log.gz](https://github.com/user-attachments/files/22889460/umbrel-1760374607032.log.gz)

jimsbr:
So, for this issue specifically, it was a non-default setting I changed recently for display website on root causing it: 
E71DF9BA-8B3C-453B-9C3D-BE5DBFE8FE9C

Keeping that not selected on your storefront/app page and on the default (None) undos that and accomplishes the intended behavior I was suggesting here and defaults to admin dashboard successfully when clicking from Umbrel dashboard, and should be default behavior unless someone creates a storefront and selects that (maybe adding a blurb somewhere on the app page how to get to login page without reinstalling or for that setting could be enough if someone doesn't know how to do it manually to consider)... 

I guess disregard, from what I'm reading in their [docs here](https://docs.btcpayserver.org/FAQ/Apps/?utm_source=chatgpt.com) it's not too definitive after you set other things up
"If for any reason, you want an app to be on the same domain as your BTCPay Server homepage, you can select to display it on the root. In that case, no DNS configuration is needed … Using an app on the root domain of BTCPay Server means you’ll have to access the login and other pages manually."

I seem to be reading that "this by design: BTCPay assumes the root is for customer-facing use." If you go to your domain root (like https:// yourdomain.com), BTCPay Server usually redirects you to the public store page — not the admin dashboard. The admin page lives at /admin..."
So it's inferred you just go to /admin by default manually, but note that 404s on that suggestion, adding /login is more reliable manually for me..

I can confirm I rooted and added the lines I proposed here, adding the upstream true line didn't seem to change anything, I also SSH'd and wrote in the redirect line that didn't seem to change anything either after restart, until I changed the selection for Display app on website root back to default shared in first screenshot above, which works with those lines still edited in after a restart...

For anyone stumbling across this I have helpful steps here I added  https://community.umbrel.com/t/setting-up-cloudflare-tunnels-on-umbrel/10239/8?u=usernameisjim in addition to [Radiokot's helpful set up of Cloudflare](https://github.com/Radiokot/umbrel-cloudflared/wiki/How-to-set-up-Cloudflare-Tunnel-on-your-Umbrel) steps

On default settings now, it seems the public-facing root domain defaulting to a login page with red error "You cannot login over an insecure connection. Please use HTTPS or Tor."

To resolve this instead of the other setting that caused that behavior I made sure to only select this one where you can map your store to a specific domain like this:


2770A9B3-D721-451B-88CC-B515683AE473

Once I made sure that was right it's all working again for a public-facing URL and with https:// just fine again with Cloudflare, and with (None) and mapping to specific domain as shown in last screenshot

My written in lines are still there proposed in this PR which isn't seemingly causing an issue right now but will probably remove them, which I guess is not impacting anything...

I will maybe add to Radiokot's setup guide a few blurbs, to try and document things and make it easier for people who run into anything confusing and want to run local stores and stores that are public-facing on the internet with a high uptime :)

Noting my next thing I want to try and figure out is if I can map other domains on Cloudflare tunnels to other websites, as Cloudflare seems to default to just the one App and domain (like if someone wants to embed a crowdfund, a couple other store pages, or a shoutbox on one Umbrel Home)...

but that could add more complexity and start breaking things, anyways ty!

jimsbr:
closing, general login/UX redirect stuff for defaulting to admin page as described in last comment moreso domain settings related

jimsbr:
mental note.

PR #3782, opened 2025-10-14
Bump rtl to 0.15.6 and boltz client to 2.8.9

kilrau:


nmfretz:
Thanks for these 2 updates @kilrau! Really appreciate you keeping these up-to-date in our app store 🤝

Reviewed and tested.

PR #3834, opened 2025-10-17
fix[linter]: Mounted directory uptime-kuma does not exists

Mr-Sunglasses:
ref to linter error: https://github.com/getumbrel/umbrel-apps/pull/3833#issuecomment-3413312254

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,13 @@ submitter: Umbrel
 submission: https://github.com/getumbrel/umbrel/pull/334
 ```
 
+When submitting a new app, leave the `gallery` and `releaseNotes` fields empty. Use the following values:
+
+```yml
+gallery: []
+releaseNotes: ""
+```
+
 The `dependencies` section within the app manifest gives Umbrel a list of app IDs that must be already installed in order for the user to install BTC RPC Explorer and also function.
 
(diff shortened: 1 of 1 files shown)


al-lac:
Hey @Mr-Sunglasses,

thanks a lot for the proposed fix for this linting issue. 🤝 

Solving it like this would however switch the directory that is being used to a different one, caused a reset of the app data.

The right way to fix this would be to add an `uptime-kuma` folder in the `data` one.

Do you want to tackle this or should I take over?

Mr-Sunglasses:
> Hey @Mr-Sunglasses,
> 
> 
> 
> thanks a lot for the proposed fix for this linting issue. 🤝 
> 
> 
> 
> Solving it like this would however switch the directory that is being used to a different one, caused a reset of the app data.
> 
> 
> 
> The right way to fix this would be to add an `uptime-kuma` folder in the `data` one.
> 
> 
> 
> Do you want to tackle this or should I take over?

Thanks @al-lac , I'll do the suggested changes.

Mr-Sunglasses:
Thanks @al-lac for the suggestions on the PR, I've done the necessary changes PTAL.

al-lac:
Looks good, thanks for the contribution @Mr-Sunglasses! 🚀

PR #3837, opened 2025-10-17
Knots: add version selection

Retropex:
Hey,

Due to some users that want to run an alternative version of knots, I would like to add the support of an alternative version of knots directly inside the package in order to avoid the user having to manually move the data from one package to another.

Such change should be merged only if it's very stable since bitcoind is upstream to a lot of services. that why code review would be very appreciated, thanks in advance!

PR of the package change:
https://github.com/Retropex/umbrel-bitcoin/pull/7

### DIFF
--- a/bitcoin-knots/docker-compose.yml
+++ b/bitcoin-knots/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
   
   app:
-    image: ghcr.io/retropex/umbrel-bitcoin-knots:1.0.5-knots.v29.2@sha256:d75df53e0fc8b15ace68d8b9edbaca475d3e71735b63b2061fadbc9c92e2e193
+    image: ghcr.io/retropex/umbrel-bitcoin-knots:1.1.0@sha256:f455204f1f8b4f7a2b6f0c70966f77b50015a171cdaf218353f7d4f251301425
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 15m30s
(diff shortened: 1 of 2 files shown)


nmfretz:
Awesome idea @Retropex! We've actually got an imminent Bitcoin Node app update about to go live that you can pull in changes from that will actually do this and have full version-aware validation of the config. It should make it really easy to add new knots versions to your app each time.

> Such change should be merged only if it's very stable since bitcoind is upstream to a lot of services. that why code review would be very appreciated, thanks in advance!

Ya, totally agree. If you pull in the upstream changes and then tweak the Dockerfile to include the binaries you want, and tweak any settings as well it will make testing super simple for me and then keeps things easy moving forward for you to keep merging in upstream app changes.

nmfretz:
@Retropex changes are good to merge in now 🫡
https://github.com/getumbrel/umbrel-bitcoin

Retropex:
Thanks @nmfretz, changes has been rebased on Knots package!

nmfretz:
Excellent, thanks @Retropex! I've tested fresh install, app update, and switching versions. I've also taken a general look through https://github.com/Retropex/umbrel-bitcoin/pull/7/files and the changes look good (including still installing curl 🤝). The only thing I haven't checked is the Knots-specific setting additions, and the specific values (default, min, max, etc) for settings that differ from Core, since those are an implementation detail of the app and not related to packaging for umbrelOS.

Thanks again! Going live.

PR #3889, opened 2025-10-25
App Submission: copyparty

dennysubke:
# App Submission

### copyparty
...

### Icon
logo

...

### I have tested my app on:
- [X] umbrelOS on Linux VM

### DIFF
--- a/copyparty/docker-compose.yml
+++ b/copyparty/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3923
 
   web:
-    image: ghcr.io/9001/copyparty-ac:1.19.19@sha256:0799b3a495e4ba97304d8eeb052aafec744d871bdcbb0ad996a04c573d3ba696
+    image: ghcr.io/9001/copyparty-ac:1.19.20@sha256:81bae1a42e0ad879231f84eeb88f6a8b7dc943e84c1fb838dd44ccfac0fffc5a
     restart: on-failure
     user: "1000:1000"
     volumes:
--- a/copyparty/umbrel-app.yml
+++ b/copyparty/umbrel-app.yml
@@ -3,7 +3,7 @@ id: copyparty
 name: copyparty
 tagline: A versatile all-in-one file server for everything
 category: files
-version: "1.19.19"
+version: "1.19.20"
 port: 3923
 description: >-
   📼 copyparty is a lightweight yet incredibly powerful file sharing platform designed to give you complete control over your data. It transforms any environment into a personal and fully featured file hub where you can upload, download, edit, and organize files directly through an intuitive web interface. Everything happens in the browser without the need for additional software, making it fast, convenient, and accessible from virtually anywhere.


dennysubke:
@Mr-Sunglasses @al-lac 

copyparty is the hottest thing in self hosting right now. 🔥

It comes with everything: a smooth web interface, FTP, WebDAV, a music player and even a code viewer, all in one container.


Watch the video: 

[Watch the video](https://youtu.be/15_-hgsX2V0)



It is blowing up in the self hosting scene for a reason. 😎

copyparty/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+services:
+  app_proxy:
+    environment:
+      APP_HOST: copyparty_web_1
+      APP_PORT: 3923
+
+  web:
+    image: ghcr.io/9001/copyparty-ac:1.19.17@sha256:4fed30b24b217378314e10ebfbc67ad6c02be89734e8735c0b65f0dd8e524e0b
+    volumes:
+      - ${APP_DATA_DIR}/config/config.conf:/cfg/config.conf:ro
```
Mr-Sunglasses: We can store all types of data whether configuration files or media within a single `data` directory. Please also change the directory path accordingly
```suggestion
      - ${APP_DATA_DIR}/data/config/config.conf:/cfg/config.conf:ro
```

copyparty/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+      - ${APP_DATA_DIR}/data/cp_data:/cfg/hists
+      - ${APP_DATA_DIR}/data/shared:/shr
+      - ${APP_DATA_DIR}/media:/media
```
Mr-Sunglasses: We can store all types of data whether configuration files or media within a single data directory. Please also change the directory path accordingly
```suggestion
      - ${APP_DATA_DIR}/data/media:/media
```

copyparty/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: copyparty_web_1
+      APP_PORT: 3923
+
+  web:
+    image: ghcr.io/9001/copyparty-ac:1.19.17@sha256:4fed30b24b217378314e10ebfbc67ad6c02be89734e8735c0b65f0dd8e524e0b
```
Mr-Sunglasses: A new version is available, we can update to that
```suggestion
    image: ghcr.io/9001/copyparty-ac:1.19.19 @sha256:0799b3a495e4ba97304d8eeb052aafec744d871bdcbb0ad996a04c573d3ba696
```

copyparty/umbrel-app.yml
```diff
@@ -0,0 +1,38 @@
+manifestVersion: 1
+id: copyparty
+name: copyparty
+tagline: A versatile all-in-one file server for everything
+category: files
+version: "1.19.17"
```
Mr-Sunglasses: Matches with the updated version of docker image.
```suggestion
version: "1.19.19"
```

copyparty/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+    volumes:
```
Mr-Sunglasses, line 13: I've tested the app, and it works perfectly in rootless mode. It’s a good practice to run it this way for better security and isolation.
```suggestion
	user: "1000:1000"
    volumes:
```

Mr-Sunglasses (commented):
Hey @dennysubke

Thanks a lot for packaging copyparty , It's a very famous app on self-hosted forums and we are glad that you package it for Umbrel App Store 🌟

I truly appreciate your continued contributions and the effort you put into growing the app ecosystem. Amazing work as always! 🚀👏

There are some small suggestions which I commented, apart from that everything is working flawlessly, I've tested on my umbrel home.

copyparty/docker-compose.yml
Mr-Sunglasses, line 13: Missing `restart` policy.
```suggestion
	restart: on-failure
    volumes:
```

dennysubke:
@Mr-Sunglasses Thanks for the review! I’ve handled all your comments. Mind double checking if the config.conf looks good now?

copyparty/data/config/config.conf
```diff
@@ -0,0 +1,18 @@
...
+  xff-src: 10.21.0.0/16
+  xff-hdr: x-forwarded-for
+
+[accounts]
+  admin: umbrel
+
+[/]
+  /media
+  accs:
+    rwda: admin
```
Mr-Sunglasses: Hey @dennysubke 👋🏻, here are some suggestions and improvements we can make:

- removed `qr` as it will never visible on the user side and we don't need it.
- removed:
```
  rproxy: -1
  xff-src: 10.21.0.0/16
  xff-hdr: x-forwarded-for
```
as it is suggested that that explicitly setting a wrong value can have unfortunate consequences. ref.: 

> "Enable reverse proxy support with auto mode" is also weird, and by weird I guess I mean it doesn't make sense -- this option (with that specific value) tells copyparty to trust the rightmost ("nearest/closest") value in the xff-hdr, which is the value that your local reverseproxy specifies. Depending on the rest of the stack which copyparty is running inside (reverseproxy, other middlewares, cdn/waf), this may be the wrong value. The best way to find the correct value is running copyparty without this option specified at all, and see what it says in the log. If Umbrel is rigid enough in those aspects, then it might be possible to set a static predefined value, but beware that explicitly setting a wrong value can have unfortunate consequences (see "thank you for playing" in the readme)

I tested after removing these options and the app works as expected. Logs now show the proxy IP for clients, which is acceptable for this setup. For now, keeping these unset avoids misconfiguration risk.

- changes permission from `rwda` to `rwmda` as we need to give all the permission to the admin.


```suggestion
[global]
  p: 3923                          # Listen on port 3923
  e2dsa                            # scans all mounted volumes (including readonly ones)
  e2ts                             # scans for tags in all files that don't have tags yet
  z                                # Enable Zeroconf discovery
  shr: /shr                        # base URL where shares will be accessible from
  hist: /cfg/hists/                # database (fs+media index) and thumbnail cache
  
[accounts]
  admin: umbrel                    # Create a user named admin with password

[/]
  /media               # Expose /media as the root volume
  accs:
   rwmda: admin       # the user "admin" gets read-write-move-delete-admin
```

9001: some quick comments on the config -- specifically the comments, which I find very surprising and confusing;

I'm curious where those explanations for e2dsa and e2ts came from, because they are entirely incorrect -- those options are not at all related to encryption or authentication; searching for them [in the readme](https://github.com/9001/copyparty/#file-indexing) shows that `e2dsa` can be explained as "scan and index all volumes on startup", and `e2ts` is "also scan and index multimedia metadata". The `e2dsa` option is very expensive since it will create checksums of all your files, but it is necessary for file deduplication (another default-disabled feature). A lighter alternative to `e2dsa`, which does not actively index existing files but only files uploaded into copyparty, is `e2d`

and regarding the "qr" option, this only produces a qr-code in the server console on startup, and is not login-related -- it is mainly intended for quick access when the server is temporarily launched on a LAN. Perhaps "show  a qr-code in the console on startup", or maybe remove the option if this is never visible / useful in the first place?

"shr" is perhaps better explained as "base URL where shares will be accessible from"; the current comment could indicate that the value is a filesystem path, which is not the case (or maybe that's just me),

and a better explanation for the "hist" option is that this is where the database (the filesystem and music index) is stored, along with the thumbnail cache -- but note that logfiles are NOT stored there! (another curiosity)

"Enable reverse proxy support with auto mode" is also weird, and by weird I guess I mean it doesn't make sense -- this option (with that specific value) tells copyparty to trust the rightmost ("nearest/closest") value in the xff-hdr, which is the value that your local reverseproxy specifies. Depending on the rest of the stack which copyparty is running inside (reverseproxy, other middlewares, cdn/waf), this may be the wrong value. The best way to find the correct value is running copyparty without this option specified at all, and see what it says in the log.  If Umbrel is rigid enough in those aspects, then it might be possible to set a static predefined value, but beware that explicitly setting a wrong value can have unfortunate consequences (see "thank you for playing" in the readme)

EDIT: oh, and is there a reason we are giving the user read/write/delete access, but not `rwmda` so they can also move/rename files?

Mr-Sunglasses: Thanks a lot @9001  for the detailed clarification! I really appreciate you taking the time to explain everything so clearly. I now realize that a few of my comments were way off. I was referencing this article: https://noted.lol/copyparty/ looks like some of the explanations there aren’t accurate, and I should have confirmed things against the official docs first.

Sorry for the confusion! I’ll update the suggestion to match the correct behavior for e2dsa, e2ts, qr, hist, the reverse proxy settings, and also adjust the permissions as you suggested. Thanks for catching all of that.

Thanks again for the guidance and for being so helpful. and for being so helpful ❤️

dennysubke: Hey @Mr-Sunglasses, I went through all your feedback and updated the config accordingly.
Really appreciate how clearly @9001 explained everything, that breakdown was incredibly helpful and detailed.

Thanks again to both of you for the input and guidance 🙌

Everything should be good to go now 🚀

Mr-Sunglasses (approved):
Thanks a lot @dennysubke making changes, I'm excited to see copyparty on umbrel app store, awesome work as always 🥇 

We will prepare the assets now 🧑‍🍳 .

al-lac:
Hey @dennysubke,

thanks once again for providing another amazing app to the Umbrel App Store! 🥇 ⏯️ 

Saw a lot about copyparty on YouTube, really cool that we finally have it here.

Thanks @Mr-Sunglasses for the great review 👏 

We can go live now!

CleanShot 2025-11-21 at 14 09 27@2x

And thanks for developing such a cool app @9001, always happy when developers of the app reach out here, we appreciate it a lot! 🔝 

Feel free to use any of those assets:
[badge-light](https://apps.umbrel.com/app/copyparty)
[badge-dark](https://apps.umbrel.com/app/copyparty)
[badge-light-logo](https://apps.umbrel.com/app/copyparty)
[badge-dark-logo](https://apps.umbrel.com/app/copyparty)

PR #3966, opened 2025-11-06
fix: remove email field and fix support link

Mr-Sunglasses:
- Remove email field as it does not have any use
- Fix support link and change it to github issue as discussion is not enabled in the repository.


al-lac:
Thanks a lot @Mr-Sunglasses for finding and fixing these small issues 🥇

PR #3967, opened 2025-11-06
fix: change support link to github issues

Mr-Sunglasses:


al-lac:
Another great catch, thanks @Mr-Sunglasses!

PR #3969, opened 2025-11-06
App Submission: Twenty Eight

duckaxe:
# App Submission

### Twenty Eight

Bitcoin Node versions 29 and 30 are great apps, but they use a lot of RAM (6-8 GB) on my RasPi4. This is problematic for devices with limited RAM (4-8 GB). _Twenty Eight_ attempts to solve this problem.

_Twenty Eight_ is a lightweight Bitcoin node pointing to Bitcoin Core v28.1. This version is renowned for its low RAM usage of approximately 1 GB. The Bitcoin node is optimized to operate efficiently on devices with limited memory, making it an ideal choice for users with older or less powerful hardware.

This leads to greater decentralization, as a higher number of devices can act as active Bitcoin nodes.

### 256x256 SVG icon
icon

...

### Gallery images
gallery_1

gallery_2

...

### I have tested my app on:
- [x] umbrelOS on Linux VM

### DIFF
--- a/bitcoin-28/docker-compose.yml
+++ b/bitcoin-28/docker-compose.yml
@@ -43,7 +43,7 @@ services:
     image: getumbrel/bitcoind:v28.1@sha256:c565266ea302c9ab2fc490f04ff14e584210cde3d0d991b8309157e5dfae9e8d
     user: "1000:1000"
     command: "${APP_BITCOIN_28_COMMAND}"
-    restart: unless-stopped
+    restart: on-failure
     stop_grace_period: 15m30s
     volumes:
       - "${APP_BITCOIN_28_DATA_DIR}:/data/.bitcoin"


nmfretz:
Hey @duckaxe, sorry for the delay in getting to this! Really dig the app name of `Twenty Eight` and the idea behind lower RAM usage.

Unfortunately we can't accept this one to the official app store purely for security reasons. Obviously though, it could still be installed by users via a Community App Store.

**Security:**
The issue is that this app is locked to Bitcoin Core v28.1, with the idea being to never upgrade to 28.2, 28.3, or anything later for RAM reasons. Even though the 28.x branch is currently still eligible for critical security backports (that window ends when v31 ships), those fixes only land in newer point releases like 28.2 / 28.3 etc. Because your app pins users to 28.1 forever, users will never get those critical patches even while the branch is technically still supported. And then after v31 is released the 28 branch wouldn't get any critical fixes at all. Here's the maintenance schedule for Core releases: https://bitcoincore.org/en/lifecycle/. You can see that 28.x is already out of maintenance phase and will soon be end-of-life.

On top of that, Bitcoin Core’s disclosure policy (Antoine Poinsot’s announcement: https://groups.google.com/g/bitcoindev/c/Q2ZGit2wF7w) keeps medium- and high-severity bugs secret until 2 weeks after the last affected version reaches EOL. So imagine that a medium/high bug was fixed in v30 (October 2025) that is present in v28. v29 is in maintenance so likely gets the fix as a point release, but v28 may not. That makes v28.x the last supported version still affected, so the bug details only become public 2 weeks after v28.x goes EOL  when v31 ships (~late April 2026). Anyone permanently locked on v28.1 right now therefore runs that undisclosed, remotely exploitable bug without an upgrade path.

**RAM:**
A recent update to umbrelOS in v1.5 should help with RAM a lot. There's now a zram swap for ram compression: https://github.com/getumbrel/umbrel/commit/dade727827bb738a3fca65c5818843aca4f2ca99

So 8GB Pi's on umbrelOS 1.5 should get about 12GBa ddressable memory.

Currently, the Bitcoin Nodes I'm running are consuming about 2-3 GB:

image

duckaxe:
Thank you for your detailed response. I agree with you regarding security. Will try latest Bitcoin Node on my Raspi4. 

PR closed.

PR #3972, opened 2025-11-06
BTC-Tracker v0.6.2: Reintroduce Breaking Change Notice to Prevent Data Loss Reports

dennysubke:
I’d suggest keeping the Breaking Change notice in place for a few more updates, as some users have reported data loss after updating. The reminder helps ensure everyone exports their data before upgrading and avoids further issues.

@wilqq-the FYI 💪🏻

### DIFF
--- a/activepieces/docker-compose.yml
+++ b/activepieces/docker-compose.yml
@@ -9,7 +9,7 @@ services:
   app:
     ## does not work rootless as of yet
     #user: "1000:1000"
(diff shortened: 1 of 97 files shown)


al-lac:
@dennysubke, good idea! Lets keep this notice in until most users have updated. Thanks for adding this again! 💯

wilqq-the:
Thanks guys, I asked @dennysubke to re-add that message because I got like 2-3 messages how to rollback data - good that we managed to get the .json stuff from the container.

Unfortunately due to privacy I don't have any telemetry either inside app or on the image repo so it's not easy to track current version distribution - I guess it is a decision to make in the near future just to get rid of that message :)

PR #3973, opened 2025-11-07
Update obsidian to version 1.9.14

dennysubke:
Update Obsidian to version 1.9.14

- Improved performance when opening quick switcher and command palette  
- Fixed inaccurate result counts in Global Search  
- Word count hidden in status bar when a base file is active  
- Fixed lag when resizing table columns  
- Tag detection now case-insensitive  
- Nested tags now recognized correctly  
- Formulas work when a file is embedded in Canvas  
- Formulas return null when file not in sidebar or embedded  
- Fixed pop-out windows not reopening after quitting the app

### DIFF
--- a/obsidian/umbrel-app.yml
+++ b/obsidian/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: obsidian
 category: files
 name: Obsidian
-version: "1.8.10"
+version: "1.9.14"
 tagline: The private and flexible writing app that adapts to the way you think
 description: >-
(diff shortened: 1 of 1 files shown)


al-lac:
Hey @dennysubke, me and @Mr-Sunglasses tested the app again and found out now that it does not work as intended.

A lot of newer linuxserver images require HTTPS, and this blocks us from publishing the update as of now.

Thanks for your efforts in trying to keep obsidian up-to-date though, hopefully we will have a solution for this soon.

PR #4018, opened 2025-11-08
fix: repo and support link for homarr

Mr-Sunglasses:
homarr is migrated to a new org and repo and the old one is archived.


al-lac:
Good catch @Mr-Sunglasses! 🎣 

Thanks for keeping these links up-to-date! 🚀

PR #4028, opened 2025-11-12
fix: enable albyhub app proxy auth by default

NodeDiver:
- Turned on PROXY_AUTH_ADD so Alby Hub now sits behind Umbrel’s app proxy.

This forces anyone visiting the app to authenticate with their Umbrel account even if the Alby Hub unlock password is known. Adds a safety layer when users mismanage credentials, requiring both the Umbrel password and the Alby Hub unlock password to gain access.

### DIFF
--- a/albyhub/umbrel-app.yml
+++ b/albyhub/umbrel-app.yml
@@ -3,7 +3,7 @@ id: albyhub
 name: Alby Hub ✨
 tagline: Self-custodial Lightning wallet with integrated node and app connections
 category: bitcoin
-version: "1.20.0-ldk"
+version: "1.20.0-auth"
 port: 59000
 description: >-
   Alby Hub is an open-source, self-custodial Bitcoin Lightning wallet, with the easiest-to-use Lightning Network node for everyone.
@@ -39,11 +39,10 @@ gallery:
   - 3.jpg
   - 4.jpg
 releaseNotes: >-
-  This release adds the option to run Alby Hub as a standalone app on umbrelOS with Alby Hub's embedded LDK based Lightning node. This means there are now two ways to run Alby Hub on umbrelOS:
-    - As a standalone app without Lightning Node (LND) installed
-    - Connected to the Lightning Node app (LND) when installed on your Umbrel
   
+  This release adds additional security measures to your Alby Hub app on umbrelOS by having it be protected by Umbrel's authentication system in addition to Alby's. Now, if you are logged out of your Umbrel when you access your Alby Hub app, you will be prompted to first login to your Umbrel account, which protects you even in the event that your Alby account is compromised. 
   
+
   No other changes to Alby Hub are included in this release.
 
 


albyhub/docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
     environment:
       APP_HOST: albyhub_server_1
       APP_PORT: 8080
-      PROXY_AUTH_ADD: "false"
+      PROXY_AUTH_ADD: "true"
```
reneaaron: Isn't _true_ the default here so you can omit the whole line? 

From the [docs](https://github.com/getumbrel/umbrel-apps/blob/840a1ef8ebb30c3de8378442b21e05108708762c/README.md?plain=1#L455C1-L455C105): 
> The Umbrel App Proxy automatically protects an app by requiring the user to enter their Umbrel password


```suggestion
```

NodeDiver: It is the default. Also I just test it locally and indeed the whole line is unnecessary, it's removed now.

NodeDiver:
@nmfretz I thank you in advance for reviewing and merging this 🙏

nmfretz:
Awesome @NodeDiver, really glad to see Alby Hub utilizing the auth proxy.

> This forces anyone visiting the app to authenticate with their Umbrel account even if the Alby Hub unlock password is known. Adds a safety layer when users mismanage credentials, requiring both the Umbrel password and the Alby Hub unlock password to gain access.

And in addition to needing to enter the umbrel password if not signed in, if a user has 2FA set up on their Umbrel then Alby also sits behind that! 

In order for existing users to get an app update notification the `version` field in the `umbrel-app.yml` needs to be updated. As-is, only fresh installs of Alby Hub would site behind the auth proxy.

What's better for you guys:

1) We wait until the next version of Alby Hub is released, and then merge this PR with updated image, version, and release notes. This means users would continue to not have the auth proxy enabled until then.

2) Merge this PR now with only the auth proxy added. We change the version to something like
```yml
version: 1.20.0-auth
```

or 
```yml
version: 1.20.0-hotfix.1
```

And then add release notes like:
```yml
releaseNotes: >-
  This release adds additional security to your Alby Hub app on umbrelOS by ...
  

  No other changes to Alby Hub are included in this release.


  Full release notes are found at https://github.com/getAlby/hub/releases
```

NodeDiver:
Hey, @nmfretz ,
Since this change is independent of the next version of Alby Hub and it is an enhancement on security, you are welcome to merge it now.

You can use `version: 1.20.0-auth`

```
releaseNotes: >-
  This release adds additional security to your Alby Hub app on UmbrelOS by adding existing UmbrelOS security measures
  

  No other changes to Alby Hub are included in this release.


  Full release notes are found at https://github.com/getAlby/hub/releases
```

Thanks!!!

nmfretz:
Excellent, thanks again @NodeDiver! Here's what I've done for the release notes: 

image

Reviewed and tested connections. If a user is not logged in to their Umbrel they would be met with this screen if trying to access the Alby Hub app directly:

image

PR #4061, opened 2025-11-17
Update Tor Browser to version v15.01

dennysubke:
**PR Note on Architecture Support:**

- amd64 builds use the official Tor Browser release: 15.0.1 (from torproject.org)

- arm64 builds use an unofficial community port: 13.0.9 (from tor-browser-ports on SourceForge)

The Tor Project does not provide an official ARM64 build, so the ARM variant is based on the widely used community-maintained port. Both versions are handled automatically via `TARGETARCH` during build.

If the Tor Project ever publishes an official ARM64 release, I will update the image accordingly. 💪🏻


al-lac:
Hey @dennysubke, thanks a lot for keeping the tor browser up to date! Really appreciate it a lot! 🤝 

Great work on making sure the image supports both amd64 and arm as well 🔝 

Made some adaptions to the release notes and description. Tested on my Umbrel Home and dev instance (arm64). Everything worked out well. Going live!

PR #4118, opened 2025-11-27
App Submission: DumbWare Suite

dennysubke:
# App Submission

### DumbWare Suite
...

### Icon
logo

...

### I have tested my app on:
- [X] umbrelOS on Linux VM


al-lac:
Hey @dennysubke,

thanks a lot for this submission! So many tools in one app, wow. 🤯 

I need to check with @nmfretz and @Mr-Sunglasses if this is a way to go here. For example, [dumbpad](https://apps.umbrel.com/app/dumbpad) is already on the store.

Having multiple tools in one app is great, but it makes updating them somewhat harder.

Will let you know once I know more!

dennysubke:
Hey @al-lac,

thanks a lot for taking a look and for looping in @nmfretz and @Mr-Sunglasses 😎!

I totally understand the concern.

My idea here was simply to provide a small, clean Dumbware landing page where all these tiny tools live together with a consistent look and Umbrel friendly defaults. More of a mini suite than a single app.

About the images:
I intentionally kept the latest tags instead of version pinned tags because the Dumbware apps all use different versioning styles. Some follow semantic versioning, some use date based versions, and some do not tag at all. Using latest was the only reliable way to keep the whole suite consistent.

And yes, I promise I really put a lot of love into this one. Layout, landing page, icons, the full experience. I tried hard to make it feel like a small Umbrel native toolbox. 😄

If the overlap with DumbPad becomes a blocker, I am happy to remove it or adjust the suite for an official store version.

And if it simply does not fit Umbrels overall concept, no worries at all. I can close the PR and keep it exclusive to my community store. 💪🏻

al-lac:
Hey @dennysubke,

We appreciate all the hard work that you put into this. 🥇 
I noticed already how much care you took to make everything look good and it really fits in nicely into the Umbrel design language. 🔝 

Lets just wait until we hear back from @nmfretz and @Mr-Sunglasses.

Have a great weekend!

dennysubke:
Thanks a lot, really appreciate the kind words! 😃

Happy to wait for feedback from @nmfretz and @Mr-Sunglasses 😎. 

Hope you also have a great weekend! ☀️

nmfretz:
Hey Denny, thanks for yet another submission. You've been absolutely crushing it. I don't know if you've seen yet, but our apps preview site has a cool new feature where you can see all the apps that have been submitted by someone. In the "submitted by" field of an app you'll see a little pill showing how many apps someone has submitted, and you can click it so see all the apps they've submitted. Check out yourself here 🥇:
https://apps.umbrel.com/submitter/dennysubke

For DumbWare, I'm really thinking the apps should be kept separate in the official app store. Please feel free to push back, I can see how much effort you've put into this. Here's my reasoning:
- The developer of the actual apps doesn't bundle them together. They are all separate. So if people come across one of the apps elsewhere (say a reddit post about the kanban board app) and want to check for it on umbrelOS they'd look for the individual app.
- The actual apps are unrelated except for the fact that they are all the same developer. It feels a little weird that if you wanted the TODO app (DumbDo) you'd also end up with 7 other apps for things like kanban boards, ip tools, terminal emulator, etc. And then if you just want the TODO app, you'd click on the DumbWare bundle app from the umbrelOS homescreen > which will open a new browser window where you'll be in a new dashboard > then you click on the TODO app > which will open yet another browser window.
- It becomes confusing for people to understand what all the apps do or to even find them in the app store just by browsing for cool apps because they are all behind an all-in-one DumbWare bundle. Imagine you're filtering the appstore by networking... you'd never see DumbWhoIs.
- It won't be immediately clear that a particular app is getting an update unless you read through the release notes (believe it or not, most users don't read release notes!).
- the current dashboard code makes it a bit difficult for you to do app updates that need changes to the dashboard. Like imagine a new DumbWare app is released. You'd need to create a forever-existing pre-start hook that does some finicky sed command to update the index.html and also download and save a new app icon asset so that existing installs get the new frontend code. We could get around this though by just packaging the UI as a Docker image so that updates are easy.
- I don't even want to think about this one haha, but imagine you decide to stop contributing to the app ecosystem at some point (please don't! 🫂). Like you end up having less time you can devote to it... then someone else would need to step in to maintain the custom UI.

Are you okay with this just being a Community App Store app?

dennysubke:
Hey @nmfretz, thanks a lot for your detailed feedback and the kind words.

And wow… that submitter page really made my day. I had no idea this feature even existed. 🥇🔥

To be honest, I already had a feeling when submitting that the DumbWare apps might be better kept separate in the official store. Your arguments about discoverability, updates and the fact that the original developer does not bundle them either are completely understandable.

My idea with DumbWare was simply to offer a clean and unified interface as an optional experience for users who like having everything in one place. But I fully understand that the official store needs a more modular structure.

So if the best decision is to keep it in the Community App Store, I am totally fine with that. No problem and no hard feelings. 😊

Thanks again for your thoughtful feedback and for the great work the team is doing. I am always happy to help wherever I can. 🙌

And by the way… you might want to avoid looking at my Pull Request for NoteDiscovery. 💥😄

Have a great week. 💪🏻

PR #4126, opened 2025-12-01
App Submission: Plane

al-lac:
# App Submission

Plane

### 256x256 SVG icon
Could only find a PNG as of now:
plane-icon

### Gallery images
CleanShot 2025-12-01 at 15 53 31@2x
CleanShot 2025-12-01 at 15 22 33@2x
CleanShot 2025-12-01 at 15 24 26@2x
CleanShot 2025-12-01 at 15 26 34@2x
CleanShot 2025-12-01 at 15 30 14@2x

### I have tested my app on:
- [x] umbrelOS on Linux VM


al-lac:
Hey @nmfretz and @Mr-Sunglasses, would be great if you could review this app!

Tested it on my Umbrel Home, dev instance and Raspberry Pi 5 already.

I am still thinking of moving some env vars over to a .env file as this would make it easier for users to adapt the base url. If it is static like it is now, the app will only work when accessed via umbrel.local.

nmfretz:
Really nice work @al-lac this is quite the complex docker setup!

Reviewed and tested. Working well 👌

> I am still thinking of moving some env vars over to a .env file as this would make it easier for users to adapt the base url. If it is static like it is now, the app will only work when accessed via umbrel.local.

Ya it's probably a good idea. We can then simplify once we have the ability to change env vars within umbrelOS

Local access with `${DEVICE_DOMAIN_NAME}` will probably work for the majority of users unless they have multiple umbrels (e.g., `umbrel-2.local`), but sharing it for remote access would be an issue. I tested tunneling to the app proxy with cloudflare (`http://umbrel.local:8762`) and changing the `WEB_URL` and `API_BASE_URL` env vars to my domain.

al-lac:
Hey @nmfretz and @Mr-Sunglasses, I added a `settings.env` that contains some environment variables like `WEB_URL`. Changing those to an external URL enables Plane to work behind an external Proxy like nginx. Tested it via nginx proxy manager on Umbrel Home.

Maybe you can do a final test as well.

Mr-Sunglasses (approved):
Tested on my umbrel dev instance working like a charm 🌠 , great work @al-lac 💜

al-lac:
Tested app on Umbrel dev instance and Umbrel Home. Going live! 🎉

Issue #4116, opened 2025-11-27
Regression: Cloudflare Tunnel + Public-Pool proxy no longer works due to forced :2019 port in API URLs after PR #4016


daredevilbear:
# Regression: Cloudflare Tunnel + Public Pool proxy no longer works due to forced `:2019` port in API URLs after PR #4016

### What happened

Before PR **#4016**, Umbrel’s Cloudflare Tunnel setup with the **Public Pool** proxy worked perfectly by pointing the reverse proxy to:

```text
http://public-pool_proxy_1:80
```

After the update introduced by PR #4016, the Cloudflare Tunnel still successfully routes to the app **but the Public Pool UI now fails to load its API endpoints.**

The frontend (or proxy config) is now effectively forcing the port **`:2019`** into the API requests, for example:

- `https://<tunnel-domain>:2019/api/info`
- `https://<tunnel-domain>:2019/api/info/chart`

Because Cloudflare Tunnels only expose **HTTPS 443** and don’t map that internal port externally, these URLs fail to resolve.

If the exact same endpoints are accessed **without the `:2019` port**, they work as expected:

- `https://<tunnel-domain>/api/info`
- `https://<tunnel-domain>/api/info/chart`

So: **before PR #4016 everything worked; after it, the forced `:2019` breaks Public Pool when accessed over a Cloudflare Tunnel.**

---

### What I expected to happen

The Public Pool frontend should continue using **relative API paths** (or otherwise not hard‑coding the internal port) so that reverse proxies and Cloudflare Tunnels can route traffic correctly.

Expected, working behavior (pre‑PR #4016):

```text
/api/info
/api/info/chart
```

These resolve correctly through the tunnel when the app is reverse‑proxied to `http://public-pool_proxy_1:80`.

---

### Steps to reproduce

1. Install Public Pool on Umbrel (latest version after PR #4016).  
2. Expose Umbrel via a Cloudflare Tunnel using `cloudflared`.  
3. Open the Public Pool UI via the tunnel domain, e.g.:  
   ```text
   https://<tunnel-domain>/public-pool
   ```
4. Observe that the main UI loads, but network calls to the API are made with `:2019` in the URL and fail to resolve over the tunnel.  
5. Manually calling the same API endpoints **without** `:2019` works.

---

### Environment

- Umbrel version: *(your version here)*  
- Public Pool app version: Latest (post‑PR #4016)  
- Exposure method: **Cloudflare Tunnel** (`cloudflared`)  
[...]

PR #4128, opened 2025-12-01
Update public-pool-ui to 2.2.1

duckaxe:
This small update resolves an issue that prevented the Public Pool UI from being accessed when the connection was tunneled.

Successfully tested with Cloudflare DNS enabled.

Closed https://github.com/getumbrel/umbrel-apps/issues/4116


nmfretz:
Wonderful, thanks for the quick turnaround @duckaxe! Reviewed and tested, including access via tunnel.

Going live.

duckaxe:
Thanks 👍

PR #4194, opened 2025-12-06
Update Bassin to 2.1.0

duckaxe:
Key highlights in this release:

### UI
* Improved output of mining numbers
* Added theme switch (light/dark)

### System
* Set `btcsig` to `"/mined by Bassin on Umbrel/"`
* Do not overwrite the configuration file

@nmfretz Please test on your AMD64 and ARM64 machines. I only tested on my Raspi.

### DIFF
--- a/bassin/hooks/pre-start
+++ b/bassin/hooks/pre-start
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+# This script updates the btcsig in ckpool.conf for existing users
+# who installed when the default was "/mined by Bassin/"
+# Only runs once to avoid overriding user's intentional changes
+
+APP_DATA_DIR="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/..")"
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, great work @duckaxe! I have reviewed and tested. It is working well, including on amd64 devices without the avx2 instruction set.

I added a pre-start hook here https://github.com/getumbrel/umbrel-apps/pull/4194/commits/4af2d52668419054c1a2d6b9fcaf0ff512cc9ee1

It's a one-time migration that updates the coinbase signature from `/mined by Bassin/` to `/mined by Bassin on Umbrel/` for existing installs. I've used a flag file to ensure it only runs once for existing users who have exactly `/mined by Bassin/`, so if they later customize their btcsig it won't get overwritten.

I saw you've also made sure that the entire config isn't overwritten each start 👌

duckaxe:
Thanks for app testing and adding the pre-start hook 👍

nmfretz:
Whoops, I completely forgot to merge 😅

PR #4164, opened 2025-12-02
Update stirling-pdf to version v2.0.2

al-lac:
🤖 This is an automated pull request to update stirling-pdf to version v2.0.2. This PR must be reviewed and tested before merging.

**Auto-generated update considerations**:
- Analytics configuration in `settings.yml` may have changed.
- The Docker image base has changed from Alpine to Debian, which is a breaking change for the Docker runtime and may affect custom configurations or paths.
- This is a major release (v2.0.0); it is strongly recommended to back up the config directory before upgrading.
- System settings can now be customized via the UI, potentially deprecating direct interaction with server config files.
- The application now supports split deployments with a separated frontend and backend, which may require changes to Docker Compose configurations.
- Licensing changes are introduced for organizations with more than 5 users, though existing users are grandfathered.
- The application now includes frontend code within the main JAR for easier single-jar hosting, and a separate `-Server.jar` is available for backend-only deployments, which may require adjusting Docker commands.
- A new option to save signatures to the server (if login is enabled) has been introduced, which could impact data persistence requirements for server-side saving.

[Original GitHub release notes](https://github.com/Stirling-Tools/Stirling-PDF/releases/tag/v2.0.2)

PR #4236, opened 2025-12-11
Update Stirling PDF to v2.1.3

sharknoon:
This PR updates Stirling PDF to v2.1.3.

The upgrade is luckily very smooth. Some settings were without a replacement removed (such as Google Drive support) and others are now exclusively configurable in the UI instead of a settings file. These can't be migrated automatically. All other settings are being migrated by Stirling fully automatically, which is really nice :)

I have tested the app upgrade and it worked without issues on my Raspberry Pi 5.

See previous PR from @al-lac https://github.com/getumbrel/umbrel-apps/pull/4164

Edit: More info on the Migration can be found here: https://docs.stirlingpdf.com/Migration/Overview

### DIFF
--- a/stirling-pdf/docker-compose.yml
+++ b/stirling-pdf/docker-compose.yml
@@ -6,9 +6,8 @@ services:
       APP_HOST: stirling-pdf_app_1
       APP_PORT: 8080
       PROXY_AUTH_WHITELIST: "/api/*"
-  
   app:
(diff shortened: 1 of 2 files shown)


al-lac:
Hey @sharknoon, thanks a lot for looking into this big stirling-pdf update! 💯 

I tested it on my Umbrel dev instance and everything went super smooth and is working as it should.

Going live! 🚀

PR #4269, opened 2025-12-17
canary: Update to v1.1.0

schjonhaug:
## Summary
- Added support for custom ntfy server configuration, allowing users to use their own self-hosted ntfy instance for notifications
- Removed redundant bitcoin dependency (Canary communicates exclusively through Electrum via BDK; since electrs already declares bitcoin as its dependency, listing both is redundant)

## Test plan
- [x] Tested upgrade from v1.0.1 to v1.1.0 on local Umbrel
- [x] Verified ntfy custom server configuration works

### DIFF
--- a/canary/docker-compose.yml
+++ b/canary/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api/*"
 
   backend:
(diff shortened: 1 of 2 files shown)


al-lac:
Hey @schjonhaug, thanks a lot for updating the canary app with some really nice new features. Tested on my Umbrel dev instance and everything is just working nicely. Going live!

PR #4313, opened 2025-12-20
Fix: worker version in Twenty App

cergmin:
Hey, @al-lac, I noticed that we merged Twenty with wrong worker version (we merged `1.13.0@{hash from version 1.13.7}`), so I think we need to fix that

### DIFF
--- a/twenty/umbrel-app.yml
+++ b/twenty/umbrel-app.yml
@@ -39,12 +39,17 @@ gallery:
   - 3.jpg
   - 4.jpg
 releaseNotes: >-
-  This major release includes extensive new features, improvements and bug fixes:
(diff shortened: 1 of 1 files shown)


twenty/umbrel-app.yml
```diff
@@ -3,7 +3,7 @@ id: twenty
 name: Twenty
 tagline: An adaptable CRM solution for seamless team collaboration and growth
 category: files
-version: "1.13.7"
+version: "1.13.8"
```
Mr-Sunglasses:
```suggestion
version: "1.13.7-sha-fix"
```

It’s generally a good practice to keep the version aligned with the Docker image version. So instead of bumping the version, I suggest retaining the existing version and simply appending a `sha-fix` tag to indicate the update.

PR #4339, opened 2025-12-25
Update btcpay-server to version 2.3.2

colder-is-better:


al-lac:
Hey @colder-is-better, thanks for providing this update to btcpay-server, really appreciate it! 💯 

I updated it to the latest version and tested it on my dev instance, going live now.

PR #4341, opened 2025-12-27
Fix: electrs docker-compose string interpolation for ELECTRS_DAEMON_P2P_ADDR environment variable

ipasechnikov:
Found this issue with the environment variable while I was trying to connect electrs to a bitcoin node. It kept failing to establish the connection. Turns out that the connection string was built incorrectly.


kaichappel:
Sorry for the delay in responding, and thanks for the fix @ipasechnikov.

This makes sense to me as a cleanup, but I don’t think this should currently affect the Bitcoin node apps in the app store.

Bitcoin Node and Bitcoin Knots both export `APP_BITCOIN_P2P_WHITEBIND_PORT` with a non-empty value, so Electrs should receive a valid whitebind P2P address with those. Libre Relay doesn’t expose a separate whitebind port, so it doesn’t export `APP_BITCOIN_P2P_WHITEBIND_PORT` at all. In that case, the existing `${APP_BITCOIN_P2P_WHITEBIND_PORT-${APP_BITCOIN_P2P_PORT}}` syntax already falls back to `APP_BITCOIN_P2P_PORT`.

This change would additionally handle the case where a Bitcoin-compatible app explicitly exports `APP_BITCOIN_P2P_WHITEBIND_PORT=""`. That seems like a reasonable cleanup, but I’d like to understand where you ran into the connection issue before we merge it.

Were you using Bitcoin Node, Bitcoin Knots, Libre Relay, or another app implementing the Bitcoin dependency?

ipasechnikov:
@kaichappel If I remember correctly, I was trying to connect to Bitcoin Node app. Electrs app was failing constantly to establish connection with the node. I looked into the issue and found that there was something wrong with environment variables of Electrs.

kaichappel:
Thanks @ipasechnikov, and sorry for the back-and-forth here. We talked this through a bit more and I’m going to walk back my earlier thinking.

For the supported Bitcoin apps, `APP_BITCOIN_P2P_WHITEBIND_PORT` should either be set to a valid port or not exported at all. If it’s exported as an empty value, that should fail rather than silently fall back, since it means the Bitcoin provider export is in a bad state.

Your original issue may have come from a misconfigured or corrupted `exports.sh` on that device. If you hit this again, please open an issue or reach out to us through support and we’ll prioritize looking into it.

Thanks again for taking the time to dig into this and open the PR.

PR #4346, opened 2025-12-30
Kokoro: Correct wrong link

dennysubke:
The linked website isn’t associated with the developer. I’ve updated the “official website” to the source referenced by the dev: https://huggingface.co/hexgrad/Kokoro-82M


al-lac:
Great catch, thanks for the fix @dennysubke!

PR #4352, opened 2026-01-04
Update InfluxDB to 1.12.2

phelipebf:
Update InfluxDB to 1.12.2 and Chronograf to 1.10.8


al-lac:
Hey @phelipebf, thanks for providing this update to influxdb, we appreciate it! 🤝 

Due to a change chronograf version 10.10.5, it is [not running as root anymore](https://hub.docker.com/_/chronograf#running-as-root) which can cause issues when updating from an older version. I made some adaptions to fix this scenario.

Tested on my Umbrel dev instance.

PR #4357, opened 2026-01-07
update to datum gateway v0.4.1

Retropex:


al-lac:
Hey @Retropex, thanks for providing this update to datum, we appreciate it! 🤝 

Tested on my dev instance and going live now.

PR #4422, opened 2026-01-09
enable authenticated UI instead of Internal unauthenticated UI

SaschaNes:
From the Frigate documentation:
Port 5000: Internal unauthenticated UI and API access. Access to this port should be limited. Intended to be used within the docker network for services that integrate with Frigate.

Port 8971: Authenticated UI and API access without TLS. Reverse proxies should use this port. ______________________________________________________________ 
Frigate’s port 5000 is intended for internal, unauthenticated access only. Umbrel currently exposes this port to users, which unintentionally publishes an unauthenticated UI. This PR switches the exposed port to 8971, the officially documented authenticated interface. The change aligns Umbrel with Frigate’s security model and provides safer defaults. No functionality is removed; only the user-facing port is corrected.

### DIFF
--- a/frigate/docker-compose.yml
+++ b/frigate/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   app_proxy:
     environment:
       APP_HOST: frigate_web_1
-      APP_PORT: 5000
+      APP_PORT: 8971
       PROXY_AUTH_ADD: "false"
   
   web:


al-lac:
Hey @SaschaNes, thanks for your contribution, we appreciate it! 🤝 

I just tried this out and it is sadly not as simple as just changing the port. We would also need to disable TLS in the config and handle the admin user credentials somehow. Right now the credentials are only shown in the logs after the first startup, if they should be reset, users would have to manually change the config.

I will discuss with the team on how we could handle this improvement and get back to you.

SaschaNes:
Hello @al-lac, disabling TLS shouldnt be an issue as in the documentation of Frigate its meant to be set-up with a SSL-Proxy infront of it. I personally use Zoraxy, before that I used Nginx Proxy Manager.
Currently, without any authentication, its pretty risky to use Frigate in such state, as some people may use umbrelos on a vRoot-Server that is public avaible. With that everybody who has the IP+Port, can access cameras that are connected to Frigate.
With that in mind, running this app without TLS would be more secure than now, because the unauthenticaed port is automaticaly loged in with admin rights

al-lac:
Hey @SaschaNes, got it. TLS is not the big issue in my opinion as well, it is more the admin user migration / creation that makes it harder to update this. Hope I have some time soon to look into it.

kaichappel:
Thanks @SaschaNes, and sorry for the delay on our end. We've finally gotten around to this.

I opened a replacement PR that builds on this fix and keeps you as co-author: https://github.com/getumbrel/umbrel-apps/pull/5700

The package now points Umbrel’s app proxy at Frigate’s documented reverse-proxy port `8971` instead of the internal unauthenticated `5000` port. It also adds a small pre-start hook/default config update so existing installs get the needed Frigate proxy/auth/TLS settings for Umbrel’s auth-proxied setup.

Thanks again for catching this.

PR #4465, opened 2026-01-13
Fix lightning detection of albyhub

al-lac:
albyhub failed to start if other apps with the name `lightning` are installed as well. For example if `core-lighting` was installed, alby failed to start as it was searching for lightning node data which is not there.

This updates the grep command to only filter for the `lighting` app itself.

### DIFF
--- a/albyhub/exports.sh
+++ b/albyhub/exports.sh
@@ -3,7 +3,7 @@
 # Check if Lightning Node app is installed and export required variables if so
 installed_apps=$("${UMBREL_ROOT}/scripts/app" ls-installed)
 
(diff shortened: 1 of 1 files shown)


al-lac:
Would be great if you could give this a quick test @NodeDiver or @nmfretz.

sharknoon:
I have tested it and installed the Bitcoin and the Core Lightning App. After that I have installed this new Alby Hub App and it started just fine. Good work @al-lac 🚀

al-lac:
Awesome, thanks for testing @sharknoon! 🔝 

Going live now! 🚀

PR #4529, opened 2026-01-21
Update PeerSwap to v5.0.2

Impa10r:
Avoid failed swaps when some Liquid UTXOs are locked


al-lac:
Hey @Impa10r, thanks for keeping PeerSwap up-to-date on the Umbrel App Store, we really appreciate it! 🏅 

Tested successfully on my Umbrel dev instance, going live!

PR #4555, opened 2026-01-24
Update papra to version 26.0.0

cergmin:
I have tested new version on Raspberry Pi 5

Added `PROXY_AUTH_ADD: "false"` because Papra has its own auth and Umbrel's auth cause problems, when I'm trying to open the app via Cloudflare Tunnel

### DIFF
--- a/papra/docker-compose.yml
+++ b/papra/docker-compose.yml
@@ -3,7 +3,7 @@ services:
     environment:
       APP_HOST: papra_server_1
       APP_PORT: 1221
-      PROXY_AUTH_ADD: "false"
+      PROXY_AUTH_WHITELIST: "/api/*"
 
   server:
     image: ghcr.io/papra-hq/papra:26.0.0-rootless@sha256:e41d2237a59b072063e4059a85492cb11fe4aefb706851becbbce88c7c2ae072
--- a/papra/umbrel-app.yml
+++ b/papra/umbrel-app.yml
@@ -24,20 +24,16 @@ gallery:
   - 2.jpg
   - 3.jpg
 releaseNotes: >-
-  - Added support for two factor authentication
-
-  - The documents page can now be used with advanced search queries
-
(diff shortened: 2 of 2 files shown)


al-lac:
Hey @cergmin,

thanks a lot for providing this update for the papra app. We appreciate it a lot! 🥇 

How are you adding the app to Cloudflare?

You can just bypass the proxy by configuring the following address: `papra_server_1:1221`

cergmin:
> How are you adding the app to Cloudflare?
> 
> You can just bypass the proxy by configuring the following address: `papra_server_1:1221`

@al-lac , wow, `papra_server_1:1221` really works! Before your suggest I added the app to Cloudflare as `umbrel.local:8112`

I still think it's better to disable Umbrel's auth for this app, since it already has its own auth with 2FA. What do you think?

al-lac:
Great to hear that it works this way! When you use the addresses like this, you skip the umbrel app proxy.

We usually prefer to keep the proxy on, for papra it would make sense as well as the sign-up is open for the public.
So I would leave it on for now at least.

cergmin:
@al-lac, Umbrel's auth also breaks other local scenarios, like API usage

```
curl -H "Authorization: Bearer {api_token}" \
>   http://umbrel.local:8112/api/organizations
Found. Redirecting to http://umbrel.local:2000/?origin=host&app=papra&path=%2Fapi%2Forganizations
```

al-lac:
@cergmin, we could handle the /api this way:
`PROXY_AUTH_WHITELIST: "/api/*"`

Would this solve it for you?

al-lac:
Hey @cergmin, going live with this one now. If we should change anything else about it, let me know! Thanks again for the update! 🔝

cergmin:
@al-lac, yes, `PROXY_AUTH_WHITELIST` works great, thanks!

PR #4580, opened 2026-01-28
App Submission: HandBrake

Knufle:
# App Submission
[jlesage/docker-handbrake](https://github.com/jlesage/docker-handbrake)

### 256x256 SVG icon
handbrake

I only have this 512x512 PNG image. Would you guys have a proper way to convert it to SVG?

### Gallery images

chrome_78lvuuNS9s
chrome_tv3vu33QCF
chrome_ULQ1RLssOA
chrome_kteLxPxewp

### Community Store

I created this community store just to test the app, and it worked as expected. I'm gonna leave the [link here](https://github.com/Knufle/knufle-umbrel) in case someone wants an easy way to try it.

I have tested my app on:
* [x] umbrelOS on my server Laptop

### DIFF
--- a/handbrake/docker-compose.yml
+++ b/handbrake/docker-compose.yml
@@ -1,0 +1,21 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: handbrake_server_1
+      APP_PORT: 5850
+
+  server:
+    image: jlesage/handbrake:v26.01.1@sha256:b9da4f2be8c7c289ff0fd3f9e63110cc651c803fca2727c35c6f17047ad8765e
+    restart: on-failure
+    volumes:
+      - ${APP_DATA_DIR}/config:/config:rw
+      - ${UMBREL_ROOT}/data/storage/downloads/handbrake/input:/storage:ro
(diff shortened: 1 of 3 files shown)


Knufle:
I wasn't aware of the arm64+amd64 requirement. If that's a hard requirement, then I think this won't be merged for now.

This makes me wonder, though, are there plans to be able to have non-multi-arch apps in the store and just indicate compatibility through the UI?

al-lac:
Hey @Knufle, thanks for submitting handbrake to the Umbrel App Store! 🔝 🍍 

Right now arm64+amd64 images are a hard requirement. So far there are no plans to support amd64 only apps in the store. Will let you know once that changes here.

nmfretz:
Hey @Knufle, sorry this sat open for so long. We’re doing a bit of cleanup on older PRs, so I checked back on the upstream HandBrake images and they still look amd64-only, including the current latest/stable tags.

So the blocker here is still the same as @al-lac mentioned earlier: apps need to support both amd64 and arm64 before we can ship them in the official App Store, and there isn’t anything on the short-term umbrelOS roadmap for architecture-specific App Store listings.

I’m going to close this for now, but if upstream ships multi-arch images in the future, feel free to open a fresh PR and we’d be happy to take another look.

PR #4618, opened 2026-01-30
Add pre-start hook to remove stale docker.pid file in portainer

al-lac:
Tested on Umbrel dev instance, pending additional tests.


al-lac:
Tested update on Umbrel Home.

PR #4620, opened 2026-01-31
Update btcpay-server to version 2.3.4

colder-is-better:


al-lac:
Hey @colder-is-better, thanks a lot for keeping btcpay-server updated on the Umbrel App Store, really appreciated! 🔝 

Tested on my dev instance and going live!

PR #4622, opened 2026-01-31
App  Submission: pogolo

0xf0xx0:
pogolo is a go-based solo bitcoin mining pool.

# App Submission

### App name
pogolo

### 256x256 SVG icon

i threw together a gopher png, but i dont have a way to make it into a svg
gopher

### Gallery images
pogolo is a cli app and doesnt have a webui, so im usin ~~dozzle~~ gotty

image
image

### I have tested my app on:
- [x] umbrelOS on Linux VM


pogolo/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+      APP_HOST: pogolo_dozzle_1
+      APP_PORT: 5663
+
+  dozzle:
+    image: amir20/dozzle:v9@sha256:2d59d09503f2c947133a4f231fdeb1f4c99e23aa0d4137f81bb1a6318e0e06cf
+    restart: on-failure
+    stop_grace_period: 30s
+    # needs to be root
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
```
al-lac: Using the docker socket here is not allowed due to security risks.

If the logs of the pogolo container are important to show, another way to show them has to be looked into. See the `snowflake` app for an example.

pogolo/hooks/pre-start
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+APP_DATA_DIR="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/..")"
+chown -R 1000:1000 "${APP_DATA_DIR}"
```
al-lac: Not really needed in this case.

al-lac (changes requested):
Hey @0xf0xx0, thanks for submitting your app to the Umbrel App Store, we really appreciate it! 🥇 

Sadly, I could not quite get the app to run for me, the container fails with this error on an arm64 instance:
`exec /pogolo: no such file or directory`

Can you look into this?

Also added some comments on what should be adjusted to get the app ready for the store.

pogolo/docker-compose.yml
0xf0xx0: hmm, i'll have to slap together a webui then :thinking: is there a log viewer in umbrel i can depend on that supports ansi coloring before i do so?

0xf0xx0:
> Sadly, I could not quite get the app to run for me, the container fails with this error on an arm64 instance: `exec /pogolo: no such file or directory`
> 
> Can you look into this?

does the latest image work? i dont have an arm device to test, unless i can run the arm docker image on x64 somehow x3

pogolo/docker-compose.yml
al-lac: No built in log viewer with ansi coloring as of yet in the OS.

al-lac:
> > Sadly, I could not quite get the app to run for me, the container fails with this error on an arm64 instance: `exec /pogolo: no such file or directory`
> > Can you look into this?
> 
> does the latest image work? i dont have an arm device to test, unless i can run the arm docker image on x64 somehow x3

Still same issue in the latest version sadly: `pogolo_pogolo_1  | exec /pogolo: no such file or directory`

0xf0xx0:
hm, im not able to reproduce it myself as qemu-user-static refuses to work. do i also need to build the image for armv7? can you `ls /` in the image?

al-lac:
@0xf0xx0, I think the arm64 image is just broken / not built correctly:
```
docker run -it --entrypoint /bin/sh 0xf0xx0/pogolo:v1.1.1
exec /bin/sh: exec format error
```

0xf0xx0:
hmm, ill take a second look at buildx.

0xf0xx0:
ok, im confused. if i build for just arm64 with `docker buildx build --platform linux/arm64 -f contrib/pkgs/docker/Dockerfile --tag 0xf0xx0/pogolo:testing --output "type=registry" .`, it works on an oracle cloud ampere instance. once i use `--platform linux/arm64,linux/amd64` it breaks. am i doing something wrong?

E: nevermind, it no longer works at all! i dont understand docker buildx

0xf0xx0:
ok this *should* work

al-lac:
Hey @0xf0xx0 and thanks for your patience, can confirm that it works on my arm64 based umbrel now. Great job! 🔝 

Sadly I don't have a Bitaxe or Miner to test further, maybe @nmfretz or @sharknoon can help out here.

pogolo/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
...
+  pogolo:
+    image: 0xf0xx0/pogolo:v1.x@sha256:298c6e7fc40109a649a451e5825a9157f9b8ae4098fb2251170c7009063fcf06
+    restart: on-failure
+    user: "1000:1000"
+    stop_grace_period: 30s
+    labels:
+      - logopo
+    ports:
+      - 5661:5661 # stratum
+      - 5662:5662 # api
```
kaichappel, line 27: This API port does not look like it needs to be host-published. The widget sidecar is already using Docker DNS to reach `pogolo_pogolo_1:5662`, and external miners only need the stratum port, so keeping the API internal avoids exposing it on the user's LAN/Tailscale.

```suggestion
    ports:
      - 5661:5661 # stratum
```

kaichappel (changes requested):
Nice work @0xf0xx0, this is close.

I gave the Umbrel package a test with Bitcoin Node. It installs and opens through Umbrel, and I was able to connect to the stratum port with a regtest mining address; pogolo accepted subscribe/authorize and showed the gopher in the logs. It also came back up cleanly after restart.

The only packaging change I think we should make before this goes live is to keep the API port internal. The widget sidecar can reach it over Docker DNS, and miners only need the stratum port. Publishing `5662` exposes the unauthenticated API on the user's LAN/Tailscale unnecessarily. I left a small inline suggestion for that, but please push back if you intended the API to be reachable directly from other devices.

Also, thanks for including the icon and screenshots in the PR body. We'll get the final app-store assets sorted on our side before launch.

Once that is adjusted, I can re-test.

pogolo/docker-compose.yml
0xf0xx0: yeah, id like to expose it to the user for whatever use they may have for it. is there a better way to do so?

kaichappel: Ah got it, thanks. For direct access from other devices, not really. Publishing the port is the straightforward way to do that.

The tradeoff is that `5662` will be reachable directly on the user's local network, outside the Umbrel app route/auth layer. So anything exposed by that API is available there too.

If you're comfortable with that risk profile then just leave the port exposed

0xf0xx0: it should be fine, it's read-only uwu

kaichappel: Perfect, thanks for confirming. This looks good to me now.

nmfretz:
Nice work @0xf0xx0. Tested locally with a bitaxe and working great:

image

Gallery assets are now ready so we can go live. Welcome to the app store 🎉

image

0xf0xx0:
thanks for havin me!

PR #4647, opened 2026-02-02
datum: switch to new env variables

Retropex:
Just to avoid keeping [these variables](https://github.com/getumbrel/umbrel-apps/blob/a3a66cec8eff947041699dcf6558d3683a37c277/bitcoin-knots/exports.sh#L17C1-L19C50) around for nothing.
Once merged I will remove them in the next Bitcoin Knots update.

### DIFF
--- a/datum/hooks/pre-start
+++ b/datum/hooks/pre-start
@@ -10,7 +10,7 @@ if [ -f "${DATUM_CONFIG_FILE}" ]; then
   echo "Checking DATUM configuration file."
   
   # Check if Knots environment variables are set
-  if [ -z "${APP_BITCOIN_KNOTS_RPC_USER}" ] || [ -z "${APP_BITCOIN_KNOTS_RPC_PASS}" ] || [ -z "${APP_BITCOIN_KNOTS_NODE_IP}" ] || [ -z "${APP_BITCOIN_KNOTS_INTERNAL_RPC_PORT}" ]; then
+  if [ -z "${APP_BITCOIN_KNOTS_RPC_USER}" ] || [ -z "${APP_BITCOIN_KNOTS_RPC_PASS}" ] || [ -z "${APP_BITCOIN_KNOTS_NODE_IP}" ] || [ -z "${APP_BITCOIN_KNOTS_RPC_PORT}" ]; then
     echo "Missing Bitcoin Knots environment variables. Exiting."
     exit 1
   fi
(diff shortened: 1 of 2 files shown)


Retropex:
Hey @al-lac, I plan to update knots to v29.3 tomorrow. It would be very nice if this could be merged to allow me to remove these variables.

nmfretz:
Hey @Retropex, nothing I love better than code cleanup 🤌

But one thing to consider is that by doing this we lock users in to needing to upgrade the Knots app with this DATUM update.

If they don't then their DATUM install will be broken, because it will fail when checking for `APP_BITCOIN_KNOTS_INTERNAL_RPC_PORT`.

We'd need to warn users in the release notes that they need to upgrade to the latest Bitcoin Knots update from the app store.

What do you think?

Retropex:
I'm not sure to follow, on the contrary there is no need to update Knots at the same time because datum will now use the same env variables as all app that depend on bitcoind

nmfretz:
ah sorry, was a bit tired and explained it really poorly. What I mean is that if we remove these variables for the apps, then we lock a subset of users into needing to update both. It seems reasonable to do I just want you to be aware. If you decide to do it then I think it's worth telling users they may need to update both apps to maintain compatibility in some cases.

Consider these scenarios:

### scenario 1
A user is running an older version of the Knots app when the internal and external ports were different. At that time we had this if you recall:
```
export APP_BITCOIN_KNOTS_INTERNAL_RPC_PORT="8332"
export APP_BITCOIN_KNOTS_RPC_PORT="9332"
```

Now for whatever reason let's say that person does not want to update their Knots app (maybe they like the old UI before the refresh), then if they update DATUM with cleaned env vars it will reference `APP_BITCOIN_KNOTS_RPC_PORT` which will be the wrong port for their app and datum will be broken.

### scenario 2
A user updates the Knots app (with env vars cleaned), but for some reason doesn't want to update DATUM or forgets or something. Then their DATUM pre-start hook exits early when trying to access `APP_BITCOIN_KNOTS_INTERNAL_RPC_PORT`. Not the end of the world because they would have already written the config out, but still kinda non-ideal.

Retropex:
Thanks for clarifying, to avoid burden I think I will cancel this change for now.

nmfretz:
okay sounds good @Retropex. We can keep this in mind down the line.

PR #4648, opened 2026-02-02
App Submission: Holesail Switchboard

oren-z0:
# App Submission

### App name
Holesail Switchboard

### 256x256 SVG icon

Created with ChatGPT and some GIMP, I don't know how to turn this to an SVG:

holesail-switchboard-logo

### I have tested my app on:
- [ ] umbrelOS on Linux VM

This app  uses ports 3161-3170 for its clients. For example when the user connects client 3161 to a remote Holesail Server "hs://0000...", he could access that server via `http://umbrel.local:3161`.

### DIFF
--- a/holesail-switchboard/docker-compose.yml
+++ b/holesail-switchboard/docker-compose.yml
@@ -1,0 +1,31 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: holesail-switchboard_web_1
+      APP_PORT: 3160
+  web:
+    image: orenz0/holesail-switchboard:v2.0.8@sha256:87bee9070543a0ec929289fc1d6ebe37b899350b4844d35ff1f2b38e689868ad
+    restart: on-failure
+    stop_grace_period: 1m
+    ports:
+      - 3161:3161
+      - 3162:3162
+      - 3163:3163
+      - 3164:3164
+      - 3165:3165
+      - 3166:3166
+      - 3167:3167
+      - 3168:3168
+      - 3169:3169
+      - 3170:3170
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+    environment:
+      HSSB_DATA_FILE: /data/data.json
+      HSSB_HOST: 0.0.0.0
+      HSSB_PORT: 3160
+      HSSB_CLIENT_HOST: 0.0.0.0
+      HSSB_FIXED_CLIENT_PORTS: 3161-3170
+      HSSB_SUBTITLE: "For more information, visit: https://github.com/oren-z0/holesail-switchboard/blob/main/docs/Umbrel.md"
--- a/holesail-switchboard/docker-compose.yml
+++ b/holesail-switchboard/docker-compose.yml
@@ -5,6 +5,7 @@ services:
     environment:
       APP_HOST: holesail-switchboard_web_1
       APP_PORT: 3160
+
   web:
     image: orenz0/holesail-switchboard:v2.0.8@sha256:87bee9070543a0ec929289fc1d6ebe37b899350b4844d35ff1f2b38e689868ad
     restart: on-failure
--- a/holesail-switchboard/umbrel-app.yml
+++ b/holesail-switchboard/umbrel-app.yml
@@ -13,23 +13,16 @@ description: >-
 
 
   Features:
-
-  - Create and manage multiple P2P servers and clients from a single dashboard
-
-  - QR code generation for easy sharing of server connection URLs
-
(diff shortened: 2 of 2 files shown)


holesail-switchboard/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: holesail-switchboard_web_1
+      APP_PORT: 3160
+  web:
```
al-lac, line 9: Empty line in between for better formatting.
```suggestion
      APP_PORT: 3160

  web:
```

holesail-switchboard/umbrel-app.yml
```diff
@@ -0,0 +1,45 @@
...
+  - Persistent configuration that survives restarts
+
+
+  Use cases:
+
+  - Expose your Umbrel apps to the internet without port forwarding
+
+  - Share access to local services with friends and family
+
+  - Connect to remote services securely
```
al-lac: Better and shorter formatting for lists.
```suggestion
  Features:
    - Create and manage multiple P2P servers and clients from a single dashboard
    - QR code generation for easy sharing of server connection URLs
    - Secure mode support for private connections
    - Persistent configuration that survives restarts


  Use cases:
    - Expose your Umbrel apps to the internet without port forwarding
    - Share access to local services with friends and family
    - Connect to remote services securely
```

al-lac (changes requested):
Hey @oren-z0,

thanks for this really great and useful submission! 🔝 🔌

Tested it on my dev instance and everything is working nicely. Was able to connect to the UI itself via the Holesail app on my Phone. ⚡ 

Could you explain to me what the clients actually do or maybe add this to the Guide?

Other than that I only have some minor suggestions about formatting, really great job on the submission! 👏

oren-z0:
Thanks for the review @al-lac , I've added a link in the dashboard to: https://github.com/oren-z0/holesail-switchboard/blob/main/docs/Umbrel.md , where I explain how to use the tool.
I've also added the link to the description.

The clients allow you to connect to other Holesail-servers, that could be running on other Umbrels.
So for example you could connect two Bitcoin Core apps P2P even if they are both behind NAT, and maybe even lightning channels.

Disclaimer: I'm not part of Holesail's core team. I just really loved the project so I (vibe) coded a UI around it.
I also published [LNbits Holesail Proxy](https://apps.umbrel.com/app/lnbits-holesail-proxy) in the past, back when I did not fully understood how Umbrel works - I thought apps can only communicate with other app containers if they set them as dependencies. Then I realized the dependency configuration is just for the boot order, and the communication between apps is much more permissive - so a global Holesail Switchboard app could work.

al-lac:
Hey @oren-z0,

Thanks a lot for making the requested changes! ⚡ 

And got it, now I also get what the clients are for.

I still remember the [LNbits Holesail Proxy](https://apps.umbrel.com/app/lnbits-holesail-proxy) app, would you say the new app is kind of a replacement for it as it support all apps?

oren-z0:
@al-lac Yes, it makes LNbits Holesail Proxy replaceable by setting a server to point to lnbits_web_1 on port 3007.
Do you consider ever limiting which containers can talk to which? Maybe by creating other sub-networks other than umbrel_main_network? I've noticed that the node-red-standalone app is not on that sub network.

Also, it would be nice to publish a table mapping app names to their web-container name + port, and possibly other internal containers that the user might want to connect remotely (probably not database, but a connection to Bitcoin Core RPC/P2P could be relevant).

al-lac:
@oren-z0, great, maybe we can phase this app out then.

Usually everything is in one sub network, but apps that use host networking are of course a different thing. (like node-red-standalone).

Good idea, we can keep this in mind for the future @nmfretz.

oren-z0:
@al-lac Do you need anything else from my side?

al-lac:
Hey @oren-z0, thanks for your patience. All good from your side, we will reach back out once the gallery assets are ready.

al-lac (approved):

al-lac:
Hey @oren-z0, thanks again for providing your app to the Umbrel App Store! 🔝 

We appreciate the work you put into it 🤝 

I added the gallery assets and we are ready to go live now 🚀 

Feel free to use any of these assets:
[badge-light](https://apps.umbrel.com/app/holesail-switchboard)
[badge-dark](https://apps.umbrel.com/app/holesail-switchboard)
[badge-light-logo](https://apps.umbrel.com/app/holesail-switchboard)
[badge-dark-logo](https://apps.umbrel.com/app/holesail-switchboard)

PR #4650, opened 2026-02-03
Update Urbit to v4.2

yapishu:
- Updated Urbit to official v4.2 binary: https://github.com/urbit/vere/releases


al-lac:
Hey @yapishu, thanks for the urbit update! 🔝 

I just tested it and I cannot seem to boot ships after updating from 4.0 to 4.2:
```
~
urbit 4.2
boot: home is /data/piers/comet-48168
loom: mapped 2048MB
lite: arvo formula 4ce68411
lite: core 641296f
lite: final state 641296f
disk: loaded epoch 0i1515
loom: mapped 2048MB
boot: protected loom
live: mapped: MB/762.511.360
boot: installed 1770 jets
stale snapshot, downgrade runtime to replay
pier: serf unexpectedly shut down
```

When booting fresh ships on Urbit 4.2, the access key does not seem to work, I am stuck on the login screen.

Can you take a look at this?

arguser:
Hey @al-lac that it's probably a local case, could be an old pier trying to boot without going migration or exited abruptly with a previous runtime. The version is good to go.

al-lac:
Got it @arguser! Then lets go live.

PR #4652, opened 2026-02-04
canary: Update to v1.3.1

schjonhaug:
## Summary
Update Canary to v1.3.1

## What's New
Fix: Fixed translation keys showing instead of localized content in notifications, improved BTC amount formatting with locale-aware number display, improved contact editing to preserve unchanged notification methods and added Electrum connectivity checks to prevent wallet creation when the server is unavailable.

## Test plan
- [x] Tested on local Umbrel

### DIFF
--- a/canary/umbrel-app.yml
+++ b/canary/umbrel-app.yml
@@ -27,7 +27,11 @@ gallery:
   - 2.jpg
   - 3.jpg
 releaseNotes: >-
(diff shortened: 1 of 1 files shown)


al-lac:
Hey @schjonhaug, thanks for keeping canary up-to-date on the Umbrel App Store, we really appreciate it! 💯 

Updated the release notes and tested it on my Umbrel dev instance.

PR #4751, opened 2026-02-13
lightning-terminal: update to `v0.16.1-alpha`

ViktorT-11:
In this PR we bump Litd to `v0.16.1-alpha`.

See the release notes here: https://github.com/lightninglabs/lightning-terminal/blob/master/docs/release-notes/release-notes-0.16.1.md

### DIFF
--- a/lightning-terminal/docker-compose.yml
+++ b/lightning-terminal/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3004
 
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @ViktorT-11. I've tested with the current version of LND in the app store (v0.20.0-beta) and it is working well.

Also, checked the release notes to confirm this is the case:
> NOTE that the minimum version of lnd that can be used in --lnd-mode=remote is v0.19.0-beta.

Going live ⚡️

PR #4821, opened 2026-02-19
Update btcpay-server to version 2.3.5

colder-is-better:


al-lac:
Hey @colder-is-better, appreciate you taking care of this btcpay-server update! Thanks! 🔝 

Tested and going live.

PR #4919, opened 2026-03-01
tunnelsats: add TunnelSats v3.4.2

TrezorHannes:
### Tunnel⚡Sats App Submission

Official submission for **TunnelSats** featuring full Umbrel App Store security hardening, 2-service `app_proxy` architecture, and in-app WireGuard configuration backups.

---

### 🚀 Major Architectural Milestone: 2-Service `app_proxy` Split (v3.4.0)

In response to maintainer review, **v3.4.0** refactors TunnelSats from a single container into a 2-service model:

1. **`tunnelsats-web`**:
   - Unprivileged frontend web container running as non-root (`user: "1000:1000"`).
   - Drops all Linux capabilities (`cap_drop: [ALL]`) and enforces `security_opt: [no-new-privileges:true]`.
   - Attached to standard Docker bridge network (`tunnelsats_default`) and target port `9740` served behind Umbrel's `app_proxy`.
   - Protected with CSRF token authentication on management endpoints.
2. **`tunnelsats-daemon`**:
   - Dedicated host-networked container (`network_mode: "host"`).
   - Holds `cap_add: [NET_ADMIN]` exclusively to configure kernel WireGuard (`wg0`) interfaces and `nftables`/`iptables` policy routing.
3. **IPC Transport**:
   - `tunnelsats-web` communicates with `tunnelsats-daemon` over a permissioned Unix domain socket (`/data/daemon.sock`) using structured JSON IPC payloads.

---

### 💾 WireGuard Configuration Backup & Recovery Flow (v3.3.8 / v3.4.0)

- **In-App Config Download**: Direct **WireGuard Configuration Download & Copy Modal** (`tunnelsats.conf`) built into the UI dashboard (under Backup / Configuration tab), allowing users to back up their active WireGuard config at any time.
- **Pre-Uninstall Guidance**: Updated `umbrel-app.yml` and user instructions to explicitly explain that uninstalling removes local data, advising users to download `tunnelsats.conf` prior to uninstallation.

---

### 🔒 Store Hardening & Container Security (`SECURE_MODE=true`)

- **WireGuard Configuration Hardening**: Sanitizes downloaded WireGuard configuration parameters directly at the API/entrypoint layer.
- **Privilege Reduction**:
  - Removed `container_name`, `NET_RAW`, `security_opt: [apparmor:unconfined]`, and host Docker socket mounts (`/var/run/docker.sock`).
  - Stripped `migration_source` mount (relevant only for community store upgrades).
  - Explicitly marked Lightning node mounts (`lnd` / `cln`) as `:ro` (read-only).
  - Preserved `NET_ADMIN` capability strictly for applying WireGuard tunnel routing and `nftables` rules in `tunnelsats-daemon`.
- **Cleanup**: Stripped internal host diagnostic scripts (`verify.sh`) and removed empty `icon` field from `umbrel-app.yml`.

---

### 🔍 Monorepo Linter Justifications (`npm run lint:apps -- tunnelsats`)

The official linter passes with **0 error(s)** and 3 expected warnings:
1. **`access.security_opt`** (`tunnelsats-web`): `no-new-privileges:true` hardens the unprivileged web dashboard against privilege escalation.
2. **`access.host_network`** (`tunnelsats-daemon`): Required for kernel WireGuard (`wg0`) interface creation and host `iptables`/`nftables` policy routing for Lightning Network P2P daemons (LND/CLN).
3. **`access.cap_add`** (`tunnelsats-daemon`): `cap_add: [NET_ADMIN]` is required exclusively for WireGuard interface configuration and policy routing rules.


al-lac:
Hey @TrezorHannes and thanks for submitting your app to the Umbrel App Store! We appreciate the effort you put into this! 🤝 

Before taking a closer look at this, is the docker socket required for the app to function? Because apps don't get access to it for security reasons.

The other two security exceptions (host networking and root user) would be fine, but if possible otherwise, even better.

Also, the standard pattern for all LND-dependent apps in the store is to mount LND's data directory read-only. For your app, that would look like this:
```
volumes:
```

Also you can point to your data directory like this:
```
${APP_DATA_DIR}/data:/data
```

Nothing besides downloads should be stored outside, as uninstalling the app should remove all app data.

Please take a look at these points and let me know if you have more questions.

TrezorHannes:
Hey @al-lac, thanks for reviewing our submission and for the feedback! 🤝

We're happy to provide more context regarding these architectural choices. Our goal with TunnelSats is to provide a "one-click" privacy-preserving VPN experience for Umbrel users without requiring them to ever use SSH or manually configure their routing tables.

### 1. Docker Socket (`/var/run/docker.sock`)
The app strictly requires the Docker socket to:
1. **Dynamically resolve IPs:** It queries the internal Docker bridge IP addresses of the running LND/CLN containers. This is required to establish exact `nftables` DNAT and routing rules without relying on static IPs, which can be fragile.
2. **Automate configuration application:** When a user toggles the VPN in our UI, the app automatically applies the necessary changes to `lnd.conf` (or CLN `config`) and then sequentially restarts the daemon (and middleware for LND) via the Docker socket so the changes take effect immediately. Without socket access, users would have to manually edit configs and run `docker restart` commands via SSH, breaking the seamless UI experience.

### 2. Host Networking and Root User
Since umbrelOS is immutable and doesn't persist host-level Wireguard services across reboots natively, TunnelSats brings its own full routing dataplane *inside* the container. It creates policy routing tables (table 51820) and applies `nftables` rules directly onto the host network namespace to prevent Clearnet IP leaks. `network_mode: "host"`, `NET_ADMIN`, and `root` privileges are hard requirements for `wg-quick` and `nftables` to manage the host's routing tables. 

### 3. Mounting LND's Data Directory Read-Only
Mounting the LND directory read-only breaks the automated UI flow. As mentioned above, when the VPN is enabled, our app automatically injects required VPN properties (like `externalhosts=`) directly into `lnd.conf`. If it's read-only, we can't do this, and users would be forced to SSH into their Umbrel node to manually paste configuration snippets before the VPN functions.

### 4. Pointing to `:data` instead of `../tunnelsats-data`
We originally used `../tunnelsats-data` specifically to protect our users' purchased subscriptions. The configuration files stored here (`tunnelsats.conf`, `tunnelsats-meta.json`) contain paid WireGuard keys. 
We've observed that Umbrel users frequently uninstall and reinstall apps as a first-line troubleshooting step. If we use `${APP_DATA_DIR}/data`, the uninstall action wipes the directory, resulting in the user losing their paid VPN configuration. Because the uninstall action happens in the native Umbrel UI outside of our app, we cannot intercept it to warn the user to export their keys first. Mounting to `../tunnelsats-data` ensures their paid configuration survives a basic reinstall cycle.
Does umbrelOS have a supported alternative mechanism for preserving specific license/config data during a reinstall? If not, we're open to suggestions on how to best handle this while protecting the user's purchase.

We understand these are exceptions to the standard app store patterns, but they are strictly necessary to deliver a secure, automated, and user-friendly VPN experience. We would be happy to add this context directly into the PR description if it helps future reviewers!

Let us know your thoughts or if you need any more details.

TrezorHannes:
Bumping this up @al-lac for your review please

al-lac:
Hey @TrezorHannes, thanks for your patience on this.

Access to the docker socket is usually a non-negotiable. But @kaichappel or @nmfretz should reach out soon with some more details!

Filouman:
bump. It would be great to see Tunnel Sats included in the Umbrel App Store when someone has a chance to review this PR. cc: @kaichappel @nmfretz

nmfretz:
Hey @TrezorHannes. Huge apologies for the wait here, I left you hanging way too long.

TunnelSats looks awesome.

@al-lac is right that the Docker socket is the main blocker here that we can't move on. We can’t allow App Store apps to mount `/var/run/docker.sock`, even read-only because it is effectively root-level access to the host. A container with access to it can ask Docker to start new containers, mount arbitrary host paths, join or modify networks, inspect other apps, restart services, and bypass most of the isolation Docker is supposed to provide. In other words, if the app or any dependency inside it were compromised, the blast radius would be the entire Umbrel, not just the TunnelSats container. This isn’t about distrust of TunnelSats specifically; it’s just not a level of access we can give to app containers on a user’s Umbrel. We don't even allow it for things as well known as the Portainer app... instead it uses docker-in-docker to remain isolated https://github.com/getumbrel/umbrel-apps/blob/master/portainer/docker-compose.yml

It's also pretty risky to allow writable access to LND/Core Lightning data dirs. Existing Lightning-dependent apps in the store mount these directories read-only. One idea here would be to use any supported RPC/API surface where it exists, instead of writing directly into LND/CLN data directories. I know that may not cover every setting TunnelSats needs, especially for listener/proxy/startup config. For anything that can’t be changed through a supported API, the safer pattern would be to show the user the config snippet/instructions and have them apply/restart through Umbrel. Editing config files is much easier now that umbrelOS supports file editing through Files, so this may still be a reasonable UX without giving TunnelSats writable access to another app’s data directory.

What about something like this:

- Avoid Docker discovery by targeting Umbrel’s known Lightning network addresses directly: LND at `10.21.21.9:9735`, and Core Lightning’s daemon at `10.21.21.96:9735`. TunnelSats could detect availability by probing those endpoints instead of inspecting Docker containers or attaching them to a separate Docker network.
- Mount Lightning/Core Lightning data read-only only if needed for certs/macaroons/config inspection.
- Don’t mutate LND/CLN config or restart those apps from inside TunnelSats. For settings that can’t be changed through a supported RPC/API, the app could generate the config snippet/instructions and have the user apply/restart through Umbrel.
- Keep TunnelSats-owned state under `${APP_DATA_DIR}/data`. I get the goal of protecting paid WireGuard config across reinstall, but writing to a sibling app-data directory bypasses Umbrel’s app lifecycle and cleanup model. For now, this would need to be handled at the app level with something like export/import or account recovery. We’re also working on better app-data management in umbrelOS, which should make it easier for users to preserve or move app data.

Happy to help think through architecture here because the app itself looks really valuable.

TrezorHannes:
Thanks @nmfretz coming back to us. No worries for the delay and appreciate the counter proposals, too.

We've checked what it would take, and while the risks you outlined are totally valid concerns, and we wouldn't handle those differently, the trade-off for our customers are real, mainly

1. no automated lnd.conf / cln.conf updates on install and uninstall
2. manual CLN / LND restart on install and uninstall

Which are not deal-breakers, but we have kicked off a mini poll in our social media and [GH here](https://github.com/orgs/Tunnelsats/discussions/9) to see what they would vote for: Trust vs Convenience.

Since the changes in our approach and code architecture would be quite substantial, I wouldn't want to introduce a forked branch to have one for the community vs official app store each. Ideal for the customer, headache to maintain[^1].

Let's see what we get, I'll come back here again. 
Thanks again

[^1]: Checking whether a toggle'd approach is feasible and not as much headache as I would have assumed

TrezorHannes:
Hey @nmfretz,

Following up on this, we've successfully implemented a toggleable architecture that supports both models seamlessly without maintaining separate forks!

We have just updated this PR to version `3.3.0` which runs in **Secure Mode** by default (`SECURE_MODE=true` set in the environment). 

Here is what we have changed to meet the App Store sandbox constraints:
1. **Docker Socket Removed:** We have completely removed the `/var/run/docker.sock` volume mount from `docker-compose.yml`. The app no longer has read or write access to the host Docker daemon.
2. **Dynamic Lightning Node Probing:** To replace Docker container discovery, the app now dynamically checks LND or Core Lightning availability by probing their default addresses (`10.21.21.9:9735` for LND and `10.21.21.96:9736` for CLN) and querying configuration paths.
3. **Manual Node Configuration & Reconcile Fallbacks:** Since we cannot modify configurations or restart the node containers from inside the app, the frontend UI has been updated to automatically detect this sandboxed mode. If in Secure Mode, the UI provides step-by-step instructions and one-click copy-to-clipboard blocks for the required `lnd.conf` / `config` properties, letting the user apply them and perform restarts manually.
4. **Subscription Preservation:** We aligned our persistent data storage mount to `${APP_DATA_DIR}/../tunnelsats-data` as requested to conform to Umbrel's lifecycle while keeping paid WireGuard configuration keys safe across standard reinstall cycles.

For users installing via our Community Store, the app defaults to `SECURE_MODE=false` (with socket mounted) to preserve the fully automated experience, but the submission here is locked down in Secure Mode.

Please perform a re-review when you have a moment. Thanks again for the architectural feedback!

TrezorHannes:
@al-lac We should be good to merge, can you do a re-review please. :raising_hand: 
Thanks

al-lac:
Hey @TrezorHannes, this will be handled by @nmfretz and @kaichappel.

Looking forward to seeing the app on the app store soon! 🚀

tunnelsats/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+version: "3.7"
+
+services:
+  tunnelsats:
```
nmfretz: If I’m understanding the current architecture correctly, this container serves the web UI/API and performs the privileged networking work.

Would it be possible to split this into an unprivileged web service behind `app_proxy` and a separate host-networked daemon? The web service could communicate with the daemon through a narrow, authenticated interface, with the daemon accepting only the structured settings it needs rather than arbitrary configuration files.

Once we’ve settled the structure, could you also update the PR description? It still describes the original Docker socket architecture.

tunnelsats/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+    # Use the specific version tag for production
+    image: tunnelsats/ts-umbrel-app:3.3.0@sha256:1c71c42724d0ac67ab08608f442bfe94d9abe24f215b905d6f5a0f86bd00d6c6
+    container_name: tunnelsats
```
nmfretz: Can we remove the explicit `container_name`? Umbrel injects this during installation, and I don’t think Secure Mode depends on the fixed name.

tunnelsats/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+    restart: on-failure
+    network_mode: "host"
+    cap_add:
```
nmfretz, line 36: Can you confirm what still requires `NET_RAW` and `apparmor:unconfined`? From what I can tell, the networking work may only require `NET_ADMIN`, but I could be missing something. After the service split, these should also apply only to the networking daemon.

tunnelsats/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      - NET_ADMIN
+      - NET_RAW
+    security_opt:
+      - apparmor:unconfined
+    environment:
+      - SECURE_MODE=${SECURE_MODE:-true}
+    volumes:
+      # Persistent peer directory (Uninstall-safe)
```
nmfretz, line 46: Just checking that I understand the recovery flow: since this is under `${APP_DATA_DIR}`, Umbrel will remove it when the app is uninstalled. Can the user recover their subscription and server configuration afterward? If not, I think the UI should make that clear before uninstall.

tunnelsats/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      - ${APP_DATA_DIR}/data:/data
+      # Migration source support for legacy users
+      - ${APP_DATA_DIR}/.:/migration_source:ro
```
nmfretz: Can we remove this migration mount? This is a new official app, so there’s no previous official installation to migrate. The Community Store app is a separate installation, so its data won’t be available here either.

tunnelsats/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      # Lightning Node config mounts (Graceful discovery via relative paths)
+      # These use relative paths to avoid the strict Umbrel "AND" dependency requirement
+      - ${APP_DATA_DIR}/../lightning/data/lnd:/lightning-data/lnd:ro
```
nmfretz, line 49: I wanted to ask about these mounts, because supporting either LND or Core Lightning makes this a little tricky.

Since both mounts are unconditional, one of the source paths will normally not exist. Docker’s short bind-mount syntax creates missing source directories automatically, so installing TunnelSats with only LND can create an empty `app-data/core-lightning/...` tree, and vice versa. That happens outside the other app’s lifecycle and may cause ownership or installation issues if the user installs it later.

Looking through Secure Mode, am I right that the actual routing already detects running nodes through TCP probes, while these mounts are mainly used for config-based UI status, stopped-node detection, and restore instructions?

If so, could Secure Mode use the existing probes, remember the user’s LND/CLN choice in TunnelSats’ own data, and generate the configure/restore instructions from that choice? The UI could show the node configuration as unverified rather than reading it directly. This would also avoid giving TunnelSats access to the complete Lightning data directories when it appears to only need the config files.

Does that seem workable from your side, or is there another requirement for these mounts that I’ve missed? Happy to consider another approach if there’s a cleaner fit.

tunnelsats/scripts/verify.sh
```diff
@@ -0,0 +1,171 @@
+#!/bin/bash
```
nmfretz: Can we remove this script? It isn’t used by the Umbrel lifecycle and still expects Docker access and paths from the previous architecture.

tunnelsats/umbrel-app.yml
```diff
@@ -0,0 +1,32 @@
...
+releaseNotes: ""
+developer: TunnelSats Team
+website: https://tunnelsats.com
+submitter: Tunnelsats
+submission: https://github.com/getumbrel/umbrel-apps/pull/4919
+dependencies: []
+repo: https://github.com/Tunnelsats/ts-umbrel-app
+support: https://tunnelsats.com/join-telegram
+port: 9739
+icon: ""
```
nmfretz: Can we remove the empty `icon` field? We host the official App Store icon separately and add it before release.

nmfretz (changes requested):
Hey @TrezorHannes, sorry for the delay! Thanks a ton for adapting based on the feedback.

While looking through the v3.3.0 source, I found an upstream security issue that I’d like to discuss privately. Since this release is already available through the TunnelSats Community App Store, I don’t want to share the details publicly until an updated image is available.

I realize this has turned into a much more involved review than either of us expected, especially after the restructuring you’ve already done. TunnelSats is a tricky App Store package because privileged networking is central to how it works, so we need to be extra careful about how the user-facing part is separated from those privileges. I really appreciate you working through this with us.

For the official app, I think the cleanest path may be to put a small unprivileged web service behind `app_proxy`—Umbrel’s authenticated reverse proxy—and keep the WireGuard/firewall work in a separate host-networked daemon. This would keep browser requests behind the user’s Umbrel login and away from the root networking process.

Am I right that port `9739` is currently served by the same root, host-networked process that applies the networking changes? If so, would that split work with how you’ve structured the app? It can still be the same codebase and image with separate entrypoints, so this shouldn’t require maintaining another fork.

I’ve left some details and a few smaller questions inline. There may be context I’ve missed, so please correct me where needed.

I know this is another fairly substantial ask, so I’m happy to work through the proposed structure with you before you dive into another round of changes. Thanks again for sticking with this!

nmfretz:
@TrezorHannes Just confirming that I’ve now sent the private report to the security address listed in the TunnelSats repository. Please let me know if it hasn’t come through.

tunnelsats/docker-compose.yml
TrezorHannes: Thanks Nathan. We agree the service split (`app_proxy` + privileged daemon) is the right long-term target architecture. For this initial store submission, `SECURE_MODE=true` (the default in this official compose) already achieves the core security isolation: no Docker socket mount, no `NET_RAW`, no `apparmor:unconfined`, no config writes to Lightning node mounts, and all node discovery via TCP probes.

We'd like to land this submission with the current single-container `SECURE_MODE` architecture (now at v3.3.1 with the WireGuard configuration sanitization fix), and follow up with the `app_proxy` + daemon split as a v4.0 milestone. Would that be acceptable?

tunnelsats/docker-compose.yml
TrezorHannes: Addressed — our promote pipeline now strips `container_name` from the official submission automatically.

tunnelsats/docker-compose.yml
TrezorHannes: Both stripped from the official submission. In Secure Mode, only `NET_ADMIN` is required for nftables/iptables routing rules. `NET_RAW` was only used by legacy killswitch ping probes (disabled in Secure Mode), and `apparmor:unconfined` was needed for Docker socket access (removed in Secure Mode).

tunnelsats/docker-compose.yml
TrezorHannes: Our data volume uses a peer directory (`../tunnelsats-data`) in the community compose, which survives Umbrel's app uninstall. For the official store submission, our promote pipeline maps this to `${APP_DATA_DIR}/data`.

Could you confirm how Umbrel handles `${APP_DATA_DIR}/data` on uninstall — is this path preserved or cleaned up? If it's cleaned up on uninstall, we can revert to the peer directory pattern to protect the user's subscription key and WireGuard configuration across reinstalls. Happy to adjust based on store expectations!

tunnelsats/docker-compose.yml
TrezorHannes: Good catch — removed from the official submission. The migration source path is only relevant for Community Store upgrades.

tunnelsats/docker-compose.yml
TrezorHannes: You're right that node discovery in Secure Mode primarily relies on TCP probes. The read-only mounts (`:ro`) serve two secondary fallback functions:
1. **Stopped-node detection** — when a node is installed but not currently running, the TCP probe fails, but the config file check allows us to show it in the setup UI for manual configuration instructions.
2. **CLN network detection** — determining bitcoin vs testnet vs signet directly from the CLN data directory.

Neither mount is ever written to in Secure Mode. We've updated both mounts to explicitly specify `:ro` in the submission. If phantom directory creation is a concern for Umbrel, we can track a follow-up to store user node selection in our own metadata file.

tunnelsats/scripts/verify.sh
TrezorHannes: Removed from the official submission. Users needing host-level connectivity verification can run the diagnostic script directly via our community documentation.

tunnelsats/umbrel-app.yml
TrezorHannes: Done — empty `icon` field deleted from `umbrel-app.yml`.

nmfretz:
Hey @TrezorHannes I see you pushed a commit. Ping me when you're ready for me to take a look!

TrezorHannes:
> Hey @TrezorHannes I see you pushed a commit. Ping me when you're ready for me to take a look!

Hey @nmfretz , thanks so much, yes, see my inline comments, the commit is ready for review, please

TrezorHannes:
@nmfretz Updated PR to version `3.3.4` (commit `02afd2c6`) with pinned multi-arch Docker digest `sha256:c16d75ae8c2bd86eede339b516b39cb22b980e3c2fb14029fa0b3aec1a1a8bd4` and hardened policy routing fallback rules. Ready for re-review whenever you have a moment!

tunnelsats/docker-compose.yml
nmfretz: Thanks for walking me through this! I’m happy with the single-container approach here. The config allowlist addresses the issue I was concerned about, so I don’t think we need to block this on splitting the web UI and service.

tunnelsats/docker-compose.yml
nmfretz: Perfect, thanks for removing `NET_RAW` and the AppArmor override!

tunnelsats/docker-compose.yml
nmfretz: Thanks for checking! `${APP_DATA_DIR}/data` is removed along with the rest of the app’s data when a user uninstalls it.

The peer directory pattern isn’t supported for official apps. Apps on Umbrel are expected to clean up after themselves when deleted, which is also the normal user expectation when uninstalling an app.

If TunnelSats can recover the user’s subscription and WireGuard configuration after reinstalling, one option would be to explain that recovery flow. Otherwise, the app could make it clear that this data is removed on uninstall and suggest setting up Backups in umbrelOS so it can be restored.

We’re also adding user-configurable mount paths in an upcoming umbrelOS release, so users who intentionally want to keep this data somewhere else will have a supported way to do that.

What would you prefer to do here?

tunnelsats/docker-compose.yml
nmfretz: Ah yep, that makes sense. I’m good with keeping these mounts read-only. I’ve added a small comment in the compose file to explain the empty directory case, since umbrelOS will populate that path normally if the Lightning app is installed later.

nmfretz:
Thanks @TrezorHannes! I’ve gone through the latest changes and replied inline. This is looking much closer now. I think the only thing left to settle is the uninstall and recovery flow. Thanks again for all the work you’ve put into this!

TrezorHannes:
> Thanks @TrezorHannes! I’ve gone through the latest changes and replied inline. This is looking much closer now. I think the only thing left to settle is the uninstall and recovery flow. Thanks again for all the work you’ve put into this!

Hi @nmfretz -

thanks so much. We will have both remaining items ready tomorrow:
- the app_proxy auth (pending PR https://github.com/Tunnelsats/ts-umbrel-app/pull/138)
- the tunnelsats.conf download (already merged into [3.3.8 ](https://github.com/Tunnelsats/ts-umbrel-app/releases/tag/v3.3.8) )

We'll push to the mono-repo and re-request once done. Much appreciate your partnership on this. :clap:

TrezorHannes:
Hi @nmfretz 

Thank you so much for your guidance and partnership throughout this submission! Your feedback directly shaped what we believe is a significantly cleaner, more secure architecture.

We have updated this PR to **v3.4.0** (commit `caf048ce`, multi-arch digest `sha256:fc34f625282a0a149d44ea6483879fd7770417de20889d65feed46ce555b909b`) addressing both remaining items:

---

### 1. 🛡️ 2-Service Architecture & `app_proxy` Authentication
We implemented the full 2-service split as discussed:
- **`tunnelsats-web`**: An unprivileged frontend container running as non-root (`user: "1000:1000"`), with all capabilities dropped (`cap_drop: [ALL]`) and `no-new-privileges: true`. It attaches to the standard bridge network and serves port `9740` directly to Umbrel's `app_proxy`. All management endpoints are protected with session CSRF token validation.
- **`tunnelsats-daemon`**: A dedicated host-networked daemon container that performs WireGuard interface (`wg0`) configuration and `nftables`/`iptables` policy routing.
- **IPC Transport**: `tunnelsats-web` communicates with `tunnelsats-daemon` over a permissioned Unix domain socket (`/data/daemon.sock`) using structured JSON request payloads.

---

### 2. 💾 Uninstall & WireGuard Configuration Recovery Flow
- **In-App Config Download**: Added a direct **WireGuard Configuration Download & Copy Modal** (`tunnelsats.conf`) in the UI dashboard (under the Backup / Configuration tab) so users can download and backup their WireGuard config file at any time.
- **Pre-Uninstall Guidance**: Updated `umbrel-app.yml` and user instructions to explicitly explain that uninstalling the app removes local data, and advising users to download their `tunnelsats.conf` file prior to uninstallation so their active subscription can be recovered or used with external WireGuard clients.

---

### 3. 🔍 Monorepo Linter Justifications (`npm run lint:apps -- tunnelsats`)
The official linter passes with **0 error(s)** and 3 expected warnings:
1. **`access.security_opt`** (`tunnelsats-web`): `no-new-privileges:true` hardens the unprivileged web dashboard against privilege escalation.
2. **`access.host_network`** (`tunnelsats-daemon`): Required for kernel WireGuard (`wg0`) interface creation and host `iptables`/`nftables` policy routing for Lightning Network P2P daemons (LND/CLN).
3. **`access.cap_add`** (`tunnelsats-daemon`): `cap_add: [NET_ADMIN]` is required exclusively for WireGuard interface configuration and policy routing rules.

---

This submission is ready for your re-review whenever you have a moment. Thanks again for all your help getting this to where it is today! 🙌

tunnelsats/docker-compose.yml
TrezorHannes: Hi @nmfretz —

Thanks! For now, we've implemented the UI integration:
1. An in-app **WireGuard Configuration Download & Copy Modal** (`tunnelsats.conf`) directly in the UI dashboard (under the Backup / Configuration tab) so users can easily back up their WireGuard config file at any time.
2. Clear pre-uninstall guidance in `umbrel-app.yml` and `instructions.md` explaining that local app data is removed upon uninstallation and encouraging users to download their `tunnelsats.conf` backup prior to uninstalling.

We'd also be very happy to implement user-configurable mount paths once that feature lands in umbrelOS!

TrezorHannes:
@nmfretz Kindly bumping this for your re-review

TrezorHannes:
Updated to **v3.4.2** (`@sha256:cd6a635faa06d51b9495acad73d0072eb46c63947c5f1acac5495fd1b8b43690`).

### What's new in v3.4.2: umbrelOS 2.0 Beta Compatibility
- **umbrelOS 2.0 Host Gateway Support**: In umbrelOS 2.0 beta, `umbreld` introduces a host-level `AppGateway` reverse proxy. `v3.4.2` updates the management security boundary to recognize and authenticate incoming proxy traffic originating from the host gateway interface, alongside legacy containerized `app_proxy` sidecars.
- **Full Dual-Version Backward Compatibility**: Seamlessly supports users running both umbrelOS 1.x (with containerized `app_proxy`) and users upgrading to umbrelOS 2.0 beta.
- **Tested on Live Node**: Verified directly on a live umbrelOS 2.0 beta test node across both standard and hardened secure modes (`SECURE_MODE=true|false`).
- **Store Linter**: Passes official app linter cleanly with 0 errors.

PR #5006, opened 2026-03-06
Update boltz-client to v2.11.1

jackstar12:


nmfretz:
Wonderful, thanks @jackstar12. I've bumped the ride-the-lightning containers to 0.15.8-beta and updated the release notes and version so existing installs get the udpate. Tested and going live.

PR #5070, opened 2026-03-12
Update Monero daemon to v0.18.4.6

colder-is-better:
This is a bug-fix release.

Reference: https://www.getmonero.org/2026/03/04/monero-0.18.4.6-released.html


al-lac:
Hey @colder-is-better, thanks a lot for another Monero update, really appreciated! 💯 

Tested it on my dev instance, going live!

PR #5084, opened 2026-03-17
Update Thunderhub image version to 0.15.4

apotdevin:


al-lac:
Hey @apotdevin, thanks a lot for keeping Thunderhub updated on the Umbrel App Store, we value this a lot! 🥇 

Tested on my dev instance and going live.

PR #5111, opened 2026-03-18
Update Core Lightning - fix CLNRest host binding and add provider contract exports

satwise:
CLNRest was bound to the container-internal daemon IP, making it unreachable from other containers (RTL, LNbits). This change:

- Adds `CLNREST_HOST=0.0.0.0` to exports.sh so consumers can reference the bind address without hardcoding it
- Passes `--clnrest-host=${CLNREST_HOST}` to lightningd (was daemon IP, now 0.0.0.0 so CLNRest listens on all interfaces)
- Exports `CLNREST_URL`, `CLNREST_RUNE_PATH`, `CLNREST_PORT`, and TLS cert paths as the provider contract for CLNRest consumers (RTL, cln-application)
- Replaces `--clnrest-port=${CORE_LIGHTNING_REST_PORT}` with the canonical `CLNREST_PORT`; keeps `CORE_LIGHTNING_REST_PORT` as a backward-compat alias
- Adds `depends_on` for startup ordering (lightningd → tor, app → lightningd)

## Testing
Tested on Pi5 running umbrelOS 1.5.0 with Core Lightning v25.09.3. CLNRest is now reachable from RTL and LNbits containers.

## Breaking Changes
None. Backward-compatible aliases (`CORE_LIGHTNING_REST_PORT`, `COMMANDO_CONFIG`) are preserved.

### DIFF
--- a/core-lightning/docker-compose.yml
+++ b/core-lightning/docker-compose.yml
@@ -9,6 +9,8 @@ services:
     image: ghcr.io/elementsproject/cln-application:25.07.3@sha256:af22cebd21e1175651049d09cf2dd547da569fddd55e4c7766e5956bd47e91fa
     command: npm run start
     restart: on-failure
+    depends_on:
+      - lightningd
     volumes:
       - ${APP_DATA_DIR}/data/app:${APP_CONFIG_DIR}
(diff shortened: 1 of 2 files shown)


satwise:
Folded into #5112. The CLNRest binding fix is now included in the LNbits PR as a combined submission.

PR #5115, opened 2026-03-20
Add ZMQ publishing to monerod command

brainchainz:
ZMQ publishing is required for Monero's p2pool decentralized mining network

### DIFF
--- a/monero/docker-compose.yml
+++ b/monero/docker-compose.yml
@@ -43,7 +43,7 @@ services:
     # "unless-stopped" ensures monerod restarts via UI-triggered daemon restarts, not just on failure
     restart: unless-stopped
     stop_grace_period: 15m
-    command: "${APP_MONERO_COMMAND}"
+    command: "${APP_MONERO_COMMAND}" --zmq-pub=tcp://0.0.0.0:18083
     image: ghcr.io/sethforprivacy/simple-monerod:v0.18.4.6@sha256:ed9299d73634a13f10de85e63d104276bf5e4ce5175ee4cc16f64f4bbeacbbd4
     ports:
       - "${APP_MONERO_P2P_PORT}:${APP_MONERO_P2P_PORT}"


aries-xyz:
We could also add the --prune-blockchain option while we're at it.
I enabled it and it worked :)

al-lac:
Hey @brainchainz and @retrocausalidad, these changes should probably be done on the Monero application itself. 

Best to open an issue [here](https://github.com/sethforprivacy/simple-monerod-docker) or maybe @sethforprivacy can chime in here as well.

brainchainz:
I run my zmq publishing right after the "${APP_MONERO_COMMAND}" in the docker compose and it works great.

nmfretz:
Hey @brainchainz, sorry for the delay here. This is a nice addition.

The YAML parsing error is because the flag is appended outside the quoted string. But we actually don't need to touch the compose file at all. The monerod command is built in `exports.sh` via `BIN_ARGS`, so the ZMQ flag can be added there:

```bash
BIN_ARGS+=( "--zmq-pub=tcp://0.0.0.0:18083" )
```

And we should export the port so future apps can discover it:

```bash
export APP_MONERO_ZMQ_PORT="18083"
```

Note that this wouldn't expose the port publicly, only on the internal Docker network. That should be fine since a future p2pool Umbrel app would connect to monerod's ZMQ over the Docker network.

nmfretz:
@retrocausalidad pruning is a good option for saving disk space, but it's also a one-way operation and comes with some tradeoffs, so it should be a user choice rather than something we enable by default. It might be worth opening a feature request on the [Monero Umbrel app repo](https://github.com/deverickapollo/umbrel-monero) to add a pruning toggle in the UI.

brainchainz:
> Hey @brainchainz, sorry for the delay here. This is a nice addition.
> 
> The YAML parsing error is because the flag is appended outside the quoted string. But we actually don't need to touch the compose file at all. The monerod command is built in `exports.sh` via `BIN_ARGS`, so the ZMQ flag can be added there:
> 
> ```shell
> BIN_ARGS+=( "--zmq-pub=tcp://0.0.0.0:18083" )
> ```
> 
> And we should export the port so future apps can discover it:
> 
> ```shell
> export APP_MONERO_ZMQ_PORT="18083"
> ```
> 
> Note that this wouldn't expose the port publicly, only on the internal Docker network. That should be fine since a future p2pool Umbrel app would connect to monerod's ZMQ over the Docker network.

sounds good, there is a umbrel p2pool app called Superbrain and the users would love to not have to add this flag manually, and after every update. If that bin_args achieves the same effect, lets do it! Thanks again

kaichappel:
Thanks @brainchainz, and sorry for the delay on this. We’ve merged the update and included you as a co-author on the commit: https://github.com/getumbrel/umbrel-apps/commit/f42a5d8e30ff5fb82b1d0de8e265d764f32ce55a

Really appreciate the contribution here. This is a great improvement for Monero and p2pool users. We added the ZMQ publisher via `BIN_ARGS` in `exports.sh` and exported `APP_MONERO_ZMQ_PORT="18083"`, instead of appending the flag in `docker-compose.yml`.

brainchainz:
Hi @kaichappel and @nmfretz - wonderful! great approach, and thank you for swinging back around on this.

PR #5142, opened 2026-03-23
Update rusty-kaspad to version v1.1.0-1

elldeeone:
## Summary

This updates `rusty-kaspad` to app version `v1.1.0-1`.

The bundled Rusty Kaspa node remains on `v1.1.0`. The `-1` suffix is an Umbrel app/package revision for the new frontend/runtime packaging.

## Changes

- update the frontend image to `elldee/kaspa-frontend:v0.4.0@sha256:a419270a224df638d10cef7bc60edd3a930a4a08dda4c14aee1eaaf892a59ef4`
- remove the old Python API/log-buffer architecture
- keep the app on Rusty Kaspa `v1.1.0`
- fix the kaspad data mount to use `/app/data`, which is where the official image stores `.rusty-kaspa`
- add a `pre-start` migration hook to move legacy anonymous-volume node data into Umbrel-managed app data for existing installs
- bump `manifestVersion` to `1.1` for hooks support
- update packaged-by URL to `https://dunshea.au`

## Why this is needed

The previous package mounted `${APP_DATA_DIR}/data/app` to `/data`, but the official `kaspanet/rusty-kaspad` image stores the actual database under `/app/data/.rusty-kaspa`.

That meant existing installs could keep their real node database in an anonymous Docker volume instead of Umbrel-managed app data.

This PR corrects the mount and adds a one-time move-based migration for existing installs.

## Validation

Tested on a local x86_64 Umbrel VM (`umbrel.local`):

- verified the live app runs correctly with the new frontend package
- verified kaspad runs from `${APP_DATA_DIR}/data/app -> /app/data`
- verified the old anonymous data volume is no longer used after migration
- verified the node continued syncing from migrated data instead of starting fresh
- verified the dashboard reached healthy connected/synced state
- ran a synthetic end-to-end legacy migration test for the new `pre-start` hook:
  - legacy anonymous `/app/data` volume detected
  - `.rusty-kaspa` moved into Umbrel app data
  - second run correctly no-ops

## Images

- `elldee/kaspa-frontend:v0.4.0@sha256:a419270a224df638d10cef7bc60edd3a930a4a08dda4c14aee1eaaf892a59ef4`
- `kaspanet/rusty-kaspad:v1.1.0@sha256:150ffa0c0dcc6497951f9199d1ba3e48c7d91ef958846b28c9fb5045459a2dca`

### DIFF
--- a/rusty-kaspad/docker-compose.yml
+++ b/rusty-kaspad/docker-compose.yml
@@ -3,7 +3,7 @@ version: "3.7"
 services:
   app_proxy:
     environment:
-      APP_HOST: rusty-kaspad-frontend-1
+      APP_HOST: rusty-kaspad_frontend_1
       APP_PORT: 4321
 
   frontend:


rusty-kaspad/data/app/.gitkeep
al-lac, line 1: This should be kept for new installs as the folder is still in use later on.

rusty-kaspad/hooks/pre-start
al-lac, line 1: Is this really needed?

The `/data/app` directory was also used before the update.

If the directory in the container changes, this does not have any effect here.

elldeeone: I believe this is needed, and it’s essentially correcting an oversight in my original packaging (sorry!).

In my original submission, I mounted `${APP_DATA_DIR}/data/app` to `/data` without realising that `kaspanet/rusty-kaspad` image actually uses `/app/data` as `RUSTY_HOME`.

I re-checked this again against the current upstream package and the exact image in use:
- the current package mounts `${APP_DATA_DIR}/data/app:/data`
- the official image uses `RUSTY_HOME=/app/data`

On an existing install, that means the active database can end up under `/app/data/.rusty-kaspa` in an anonymous Docker volume, while the Umbrel-managed app-data path stays effectively unused.

That was the case on my test Umbrel instance, which is why I added the hook. Without it, existing users could update into an empty bind mount and start a fresh database instead of continuing from their existing one.

The hook is intended to be very conservative and only act in that legacy case:
- if `${APP_DATA_DIR}/data/app/.rusty-kaspa` already exists, it exits
- if there is no legacy anonymous `/app/data` volume, it also exits
- otherwise it moves the legacy database into Umbrel-managed app data

If you see a cleaner or more Umbrel-native way to handle that migration path, I’m very happy to adjust it.

rusty-kaspad/data/app/.gitkeep
elldeeone: Agree, will fix.

rusty-kaspad/hooks/pre-start
al-lac: Just looked into this some more, when the app is started, it always mounts a new volume. So this will always be empty. Thus the script will never migrate anything.

However, it is important we get this out soon as the app is currently in a broken state because of the mounting issue.

elldeeone: Thanks @al-lac 

If Umbrel’s startup path always ends up attaching a fresh volume, then I agree the migration hook isn’t the right approach and I’m happy to drop it so we can get the mount fix out quickly.

Would there be an Umbrel-supported way to clean up the legacy anonymous `/app/data` volume from previous installs, or would you prefer that I keep this PR focused on the mount fix only?

My main concern is avoiding the old Kaspa db being left behind, but I agree that’s secondary to getting the broken mount corrected.

al-lac: I think we can have this PR just focus on the mount fix. The cleanup of the volumes should anyways happen on the OS level.

elldeeone: @al-lac thanks - i've now made changes as discussed.

al-lac:
@elldeeone, looking better now! Thanks for making the adaptions.
CleanShot 2026-03-27 at 12 56 08@2x

Data is also being stored in the correct place now:
```
root@umbrel-dev:/home/umbrel/umbrel/app-data/rusty-kaspad/data# du -sch ./*
1.1G	./app
36K	./logs
1.1G	total
```

But restarts still seem to be an issue (starts fresh):
```
root@umbrel-dev:/home/umbrel/umbrel/app-data/rusty-kaspad/data# du -sch ./*
388M	./app
36K	./logs
388M	total
```

We should probably add something like this to exclude the blockchain from backups:
```
backupIgnore:
  - data/app
```

Can you test this some more and let me know when it is ready?

al-lac:
Hey @elldeeone, I managed to fully sync a node now. Tried a restart and it is taking forever, we probably should add something like this so it get force stopped / restarted after some minutes. But maybe you can look into it so it shuts down faster and clean per default.

Stop workaround:
```
stop_grace_period: 3m
```

elldeeone:
Thanks @al-lac, I did another round of testing and pushed a follow up update.

  To clarify the restart concern:
  - During early on in the Kaspa nodes Initial Block Download, restarting can make the node appear to start fresh again (which is why you will see db size change). That is normal behaviour during IBD, not a remaining issue with the mount fix.
  - After a full sync, restart resumes correctly.

  On the shutdown side:
  - I also looked into the long stop/restart time on a fully synced node. Part of that was the frontend not handling `SIGTERM` cleanly, so I fixed that in the frontend and updated the app to `elldee/kaspa-frontend:v0.4.1`.
  - After that change, `kaspad` was still hitting `SIGKILL` after 1 minute on stop, so I added `stop_grace_period: 3m` for `kaspad`.
  - With the current PR state, a stop on my synced Umbrel test node completed in about 17s and `kaspad` exited cleanly with exit code 0 instead of being force-killed.

I also introduced explicit container names in the compose, as I found the app proxy behaviour was a little inconsistent without them and this makes rusty-kaspad_frontend_1 resolve reliably.

  This should now be ready for another look when you have a moment.

al-lac:
Hey @elldeeone, thanks a lot for making the adaptions. Looking good from my point of view!

I removed the container names again, never had any issues with them until now. Please specify what issues you saw exactly so we can investigate this.

Other than that, we would be ready to go now.

elldeeone:
@al-lac I rechecked this from scratch and it looks like the issue i saw earlier was a byproduct of how i was debugging it, so it's a non-issue.

I had manually replaced the generated `app-data` compose during testing, which is not the same path as a normal Umbrel install. When installed normally, I could not reproduce the hostname issue again.

From my side this should be good to go now. Thanks!

al-lac:
Awesome job @elldeeone. Great to work with you on this.

Tested it one more time and also good from my side, going live! 🚀

PR #5246, opened 2026-03-30
Update qbittorrent to version 5.2.0 and change download path to "/downloads"

### LINKED ISSUE #3841: Radarr and Qbittorrent folders conflict
I just installed Radarr and would like to use Qbitorrent as the default download client instead of Transmission. I'm getting the following error message. I believe this will cause issues when importing with hard links.

Help please

Image

### LINKED ISSUE #4315: Disable Umbrel's Proxy Auth and AuthSubnetWhitelist for qBitTorrent
### Part 1: Enabling `AuthSubnetWhitelist` by default to be able to autoconfigure Radarr / Sonarr / etc (cc @nmfretz)

Code, that was added 1.5 years ago, enabled `AuthSubnetWhitelist` after installation.

This caused qBitTorrent to bypass authentication entirely, as it sees requests coming from the Umbrel Proxy's internal IP (which is whitelisted).
 
https://github.com/getumbrel/umbrel-apps/pull/1194

### Part 2: Enabling Umbrel's Proxy Auth (cc @Mr-Sunglasses)

@Mr-Sunglasses noticed that there is no password required to open qBitTorrent, and enabled Proxy Auth. Now Umbrel's proxy asks for password.

It fixed the security issue, but caused another problem, that you can't connect to qBitTorrent on your Umbrel from 3rd-party apps (like qRemote on iOS), because they are unable to pass Umbrel's Auth.

https://github.com/getumbrel/umbrel-apps/pull/3947

## Proposed Solution

I propose to remove code that enables AuthSubnetWhitelist by default and disable Umbrel's Proxy Auth.

This will break zero-config setup for *Arr apps, but you will still be able to configure it manually, if I understand it correctly.
And you will be able to connect 3rd-party apps.

I believe supporting external clients is more important than preserving autoconfiguration.

@nmfretz, could you, please, clarify how disabling AuthSubnetWhitelist would affect the *Arr apps?

al-lac:
This should fix #3841.

Tested update and fresh install.

### DIFF
--- a/qbittorrent/docker-compose.yml
+++ b/qbittorrent/docker-compose.yml
@@ -15,6 +15,5 @@ services:
       # - UMASK=002
     volumes:
       - ${APP_DATA_DIR}/data/config:/config
-      # /app/qBittorrent/downloads is the default path set in qBittorrent
-      - ${UMBREL_ROOT}/data/storage/downloads:/app/qBittorrent/downloads
+      - ${UMBREL_ROOT}/data/storage/downloads:/downloads
     restart: on-failure
--- a/qbittorrent/docker-compose.yml
+++ b/qbittorrent/docker-compose.yml
@@ -6,9 +6,10 @@ services:
       APP_HOST: qbittorrent_server_1
       APP_PORT: 8080
       PROXY_AUTH_ADD: "true"
+      PROXY_AUTH_WHITELIST: "/api/*"
 
   server:
-    image: ghcr.io/hotio/qbittorrent:release-5.1.4@sha256:b547db036748d449c5cbd31635dc3b695ccc70424775c44ea91df18bd1da1ea4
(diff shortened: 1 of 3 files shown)


nmfretz:
Hey @al-lac, nice work on this. The path alignment with the *arr apps is the right fix.

One thing to flag: the migration updates `DefaultSavePath` in the config, but that only affects new torrents. Existing torrents store their save path individually in qBittorrent's resume data, so after the update they'll still reference `/app/qBittorrent/downloads`, which no longer exists in the container. Those torrents would show "Missing files" and stop seeding.

I think the simplest approach is to just call this out in the release notes so users know to select all existing torrents and set their location to `/downloads` after updating.

If we did want to handle it automatically, a couple options:
- Keep both volume mounts temporarily (mounting the same host dir at both `/downloads` and `/app/qBittorrent/downloads`) so existing torrents still resolve.
- Potentially migrate the resume data directly via qBittorrent's SQLite database in the post-start hook, though that would need careful testing.

PR #5258, opened 2026-04-01
Add PyBLOCK: Terminal-based Bitcoin & Lightning node dashboard

### LINKED ISSUE #639: If you want to build it:
# App Submission

### App name
PyBLOCK
...

### 256x256 SVG icon
https://pbs.twimg.com/media/E3dw96iXMAIRpx5.png

...

### I have tested my app on:
- [ ] [Umbrel dev environment](https://github.com/getumbrel/umbrel-dev) Yes
- [ ] [Umbrel OS on a Raspberry Pi 4](https://github.com/getumbrel/umbrel-os) Yes
- [ ] [Custom Umbrel install on Linux](https://github.com/getumbrel/umbrel#-installation) Yes

...

If you want make your own app:
https://github.com/curly60e/pyblock/blob/master/dockerfile

...

We are running PyBLOCK in other way installation but you can make your own PyBLOCK App for Umbrel if you want using our dockerfile or maybe the Umbrel Team can make the app by default in the Umbrel App store.

[PyBLOCK](https://pbs.twimg.com/media/E3dw96iXMAIRpx5.png)

GaltRanch:
# App Submission

### App name
PyBLOCK

### 256x256 SVG icon
Included in `pyblock/icon.svg` — cyberpunk Bitcoin theme, no rounded corners.

### Gallery images
4 screenshots included (PNG):
- Main menu with Rich categorized panels
- Bitcoin submenu with 4-column layout
- Block visualizer with fee-rate treemap
- TUI dashboard with live data

### About
PyBLOCK is a terminal-based Bitcoin dashboard with a cyberpunk aesthetic that runs in the browser via a ttyd web terminal. It connects directly to Umbrel's Bitcoin Core and LND nodes with zero configuration needed.

**Features:**
- Real-time block height and BTC price display
- Lightning Network: channel management, invoicing, keysend, rebalancing
- Interactive block visualizer with fee-rate colored treemap
- Mempool monitor with fee estimates
- OP_RETURN message viewer
- Mining pool stats (Ocean, Braiins, CKPool, PyBLOCK Pool)
- Moscow Time converter
- CoinGecko price feeds
- Nostr console integration
- ColdCore hardware wallet support
- Weather, rate.sx charts
- And 40+ more features

**Technical:**
- Auto-configures from Umbrel's `APP_BITCOIN_*` and `APP_LIGHTNING_*` env vars
- Runs via ttyd on port 6969
- Docker image: `curly60e/pyblock:v4.0.0`
- Dependencies: Bitcoin Core + LND

**Website:** https://pyblock.xyz:8443
**Repository:** https://github.com/curly60e/pyblock

### I have tested my app on:
- [x] Docker container with simulated Umbrel environment variables
- [ ] Umbrel OS on a Raspberry Pi 4
- [ ] Custom Umbrel install on Linux


SatoshiNakamotoBitcoin:
Great.

pyblock/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+  app_proxy:
+    environment:
+      APP_HOST: pyblock_web_1
+      APP_PORT: 6969
+
+  web:
+    image: galtranch/pyblock:v4.0.0@sha256:3b9ed2849732755277d626a39da1df01eb76d7449f5e4a5dd05c846e6f57f6c0
+    restart: on-failure
+    stop_grace_period: 1m
+    user: "1000:1000"
```
al-lac, line 15: The app did not start when running rootless, failed with this error:
```
/app/entrypoint.sh: line 10: /app/pyblock/pybitblock/config/bclock.conf: Permission denied
```

Please adapt the app or let it run with root privileges.

pyblock/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
```
al-lac: Please create the `/data` directory and add a `.gitkeep` file in there.

pyblock/icon.svg
al-lac, line 1: Please keep logos and screenshots in the PR details.

al-lac (changes requested):
Hey @GaltRanch, thanks a lot for submitting PyBLOCK to the Umbrel App Store! 💯 

Looks like a really nice dashboard and fits nicely into the Umbrel Ecosystem.

I did not get the app to run properly on my test device sadly. When running with `user: "1000:1000"` it would give me a permission error, without it I would get the following error shown by the app: `No such file or directory: 'config/selection.conf'`

Please also make sure to add the logo and screenshots to the PR details.

al-lac:
Hey @GaltRanch, any update on this or can we help out in any way?

PyBLOCK-Bitcoin:
Hello @al-lac, yes, we are fixing it.

Thanks in advance.

GaltRanch:
Hi @al-lac — thanks again for testing. Both failure modes you hit should be fixed in the latest push to this branch.

**What was wrong**

1. **UID mismatch** — the previous image was built from `ubuntu:24.04`, which ships a pre-existing `ubuntu` user at UID 1000. The Dockerfile then created `pyblock` without pinning a UID, so it got 1001. When Umbrel forced `user: "1000:1000"` the process ran as UID 1000 with no write access to anything chowned to 1001 → permission denied.

2. **Wrong bind-mount target** — the compose file mounted `${APP_DATA_DIR}/data:/data`, but the app reads/writes its config from `/app/pyblock/pybitblock/config`. Configs were silently going into the container overlay (lost on restart) and the no-user fallback path tripped on a relative-path lookup → `No such file or directory: 'config/selection.conf'`.

**What changed in this push**

- New image: `galtranch/pyblock:v4.0.1` (pinned by sha256). Built from [curly60e/pyblock@933af3b](https://github.com/curly60e/pyblock/commit/933af3b) which pins the container user to UID/GID 1000 and adds an early fail-fast writability check on the config dir.
- Bind-mount target corrected: `${APP_DATA_DIR}/data/config:/app/pyblock/pybitblock/config`, plus `${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro` for the LND tls/macaroon.
- Bumped `version: "4.0.1"` with release notes, and added the 4 gallery screenshots.

**Local verification**

Ran `docker run --user 1000:1000 -v /tmp/empty:/app/pyblock/pybitblock/config ...` simulating the Umbrel env:
- All 5 config files generated by the entrypoint into the bind-mount as UID 1000 ✅
- `ttyd` serves on `:6969`, HTTP healthcheck passes ✅
- No permission errors

Ready for another round of testing whenever you have a moment 🙏

al-lac:
Hey @GaltRanch and @PyBLOCK-Bitcoin, thanks a lot for making the adaptions 💯 

I changed the config file path and now it seems to come up, the thing that I still see when it starts up is this error before it switches to light mode, is this expected?

`│ Bitcoin CLI path not configured. Redirecting to Lite Mode. │`

GaltRanch:
Hi @al-lac, thanks for testing!
 
Yes, that fallback is currently expected with this submission. PyBLOCK's modes A (Bitcoin + Lightning) and B (Bitcoin only) shell out to a local bitcoin-cli binary for most node queries — they're not pure-RPC modes. Our Umbrel image doesn't bundle bitcoin-cli, so when the app starts in mode A, the binary check fails and it redirects to Lite Mode (which still works against public APIs).

For full mode A to work in Umbrel, the app container needs either:

1. A bitcoin-cli binary available inside the container (e.g. mounted from the Bitcoin Core app, or shipped in our image), and 
2. BITCOIN_CLI_PATH set in the environment (our entrypoint already wires it through).

Is there a recommended pattern on the Umbrel side for exposing bitcoin-cli to dependent apps — e.g. mounting the Bitcoin Core app's binary, or a standard APP_BITCOIN_CLI_PATH env var we should consume? 
Happy to adapt the compose file once we know which pattern you prefer.

al-lac:
Hey @GaltRanch, got it, thanks for the detailed feedback!

Will discuss with the team and get back to you.

PyBLOCK-Bitcoin:
We believe that to use option A (Bitcoin + Lightning) or option B (Bitcoin only), some kind of connection to the Bitcoin and Lightning containers would be required. We've also been told that Knots uses RPC port 9332 instead of the standard 8332.

When you launch PyBLOCK Dashboard on the same machine where you have Bitcoin and Lightning installed, you only need to type *bitcoin-cli and lncli* in the configuration steps for option A and *bitcoin-cli* in the configuration steps for option B.

SatoshiNakamotoBitcoin:
#639 Closed.

PyBLOCK-Bitcoin:
> #639 Closed.

Thanks.

nmfretz:
@GaltRanch super sorry for leaving you hanging! I completely dropped the ball on this.

> We believe that to use option A (Bitcoin + Lightning) or option B (Bitcoin only), some kind of connection to the Bitcoin and Lightning containers would be required. We've also been told that Knots uses RPC port 9332 instead of the standard 8332.
>
> When you launch PyBLOCK Dashboard on the same machine where you have Bitcoin and Lightning installed, you only need to type bitcoin-cli and lncli in the configuration steps for option A and bitcoin-cli in the configuration steps for option B.

Ya for your options A/B, you need to connect from the PyBLOCK container to the selected Bitcoin/Lightning dependency containers. The compose file is already using the right Umbrel mechanism for that: umbrelOS injects the selected provider details via `APP_BITCOIN_*` / `APP_LIGHTNING_*`, and the PyBLOCK container connects over the Docker network.

> We've also been told that Knots uses RPC port 9332 instead of the standard 8332.

So for this, PyBLOCK does not need separate logic for Core vs Knots. `APP_BITCOIN_RPC_PORT` handles both; Umbrel sets it to the RPC port for whichever Bitcoin provider the user is running.

If I understand correctly, the remaining issue here is that PyBLOCK’s mode A/B still treats `bitcoincli` / `lncli` as required CLI paths right? Like in your current image, `bitcoin-cli` and `lncli` are not included, so those paths stay empty and PyBLOCK falls back to Lite Mode even though the RPC/LND connection details are present.

umbrelOS doesn't expose a standard `APP_BITCOIN_CLI_PATH` or `APP_LNCLI_PATH`, and I wouldn't recommend mounting CLI binaries out of the Bitcoin/Lightning app containers.

I think the fix could one of these:

1. (I prefer this one personally): in the Umbrel/Docker auto-config path, use the existing injected RPC/LND connection details directly, without requiring `bitcoin-cli` / `lncli`.

2. include `bitcoin-cli` / `lncli` inside the PyBLOCK image and set `BITCOIN_CLI_PATH` / `LND_CLI_PATH` to binaries or wrapper scripts inside the PyBLOCK container.

Let me know if that's confusing. And sorry again for the delay, I'll be much more on top of it now.

pyblock/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+      APP_HOST: pyblock_web_1
+      APP_PORT: 6969
+
+  web:
+    image: galtranch/pyblock:v4.0.1@sha256:09aba662460b42ecef5fdbbb158971fbde01da32b78cf613c3e9db6b912f5387
+    restart: on-failure
+    stop_grace_period: 1m
+    user: "1000:1000"
+    volumes:
+      - ${APP_DATA_DIR}/data/config:/app/config
```
nmfretz: just noticed as I was looking at this... the image appears to write config to `/app/pyblock/pybitblock/config`, but this mounts app data at `/app/config`.

Can you doublecheck that?

GaltRanch:
Hi @nmfretz, thanks for the detailed reply and for confirming the architecture — no worries about the delay 🙏

I went with **option 2** (bundle the binaries inside the PyBLOCK image with RPC-injecting wrappers) because it's a self-contained ~70MB change today, while option 1 would mean refactoring all 341 `path["bitcoincli"]` and 47 `lncli` call sites in PyBLOCK. I'd like to keep option 1 on the roadmap as a follow-up improvement.

**What's in the new image (`galtranch/pyblock:v4.0.2`, multi-arch amd64+arm64):**

- `bitcoin-cli` from **Bitcoin Knots 28.1.knots20250305** (downloaded from bitcoinknots.org, SHA256SUMS-verified at build time) — RPC-compatible with whichever Bitcoin variant the user picks
- `lncli` from **LND v0.20.1-beta** (matching what `getumbrel/umbrel-lightning` ships)
- Real binaries at `/usr/local/bin/{bitcoin-cli,lncli}.bin`
- Thin shell wrappers at `/usr/local/bin/{bitcoin-cli,lncli}` that `exec` the real binary with `-rpcconnect`/`-rpcuser`/`-rpcpassword` (and `--rpcserver`/`--tlscertpath`/`--macaroonpath` for lncli) pulled from the `BITCOIN_RPC_*` / `LND_*` env vars our entrypoint already re-exports from Umbrel's `APP_BITCOIN_*` / `APP_LIGHTNING_*`
- Wrappers fail loud if those env vars are missing, so misconfigurations surface immediately instead of silently bouncing to Lite Mode

**Also in this push:**

- Reverted the config bind-mount target from `/app/config` back to `/app/pyblock/pybitblock/config`. The app reads/writes its configs at the latter path; the previous mount left the host directory unused and configs were going to the ephemeral container overlay (lost on every restart).
- Bumped version to v4.0.2 with release notes.

The Knots binary specifically (since you mentioned Knots uses RPC port 9332) means the bundled `bitcoin-cli` itself is Knots-branded; the RPC port comes from `APP_BITCOIN_RPC_PORT` so we don't hardcode 8332 vs 9332.

Ready for another round of testing whenever you have time 🙏

nmfretz:
> I went with option 2 (bundle the binaries inside the PyBLOCK image with RPC-injecting wrappers) because it's a self-contained ~70MB change today, while option 1 would mean refactoring all 341 path["bitcoincli"] and 47 lncli call sites in PyBLOCK. I'd like to keep option 1 on the roadmap as a follow-up improvement.

Nice, ya option 1 sounds like it'd be a massive refactor for you guys.

Okay, so for option 2 then don't you need to be using `admin.macaroon` instead of `readonly.macaroon`? The app description talks about `Lightning Network channel management and invoicing...`. You may need to do:

 ```yaml
  LND_MACAROON_PATH: /lnd/data/chain/bitcoin/${APP_BITCOIN_NETWORK}/admin.macaroon
```

FYI @GaltRanch I'm away starting tomorrow for a week.

**Comment by nmfretz** [comment]
Sorry for the delay here, @GaltRanch.

I picked this back up and pushed two packaging fixes to the PR branch:

- Set the working directory so PyBLOCK can find its config files.
- Switched LND from the read-only macaroon to the admin macaroon, since the app includes Lightning write actions.

While testing on my Umbrel, opening Lightning with `L` and selecting `N` for Node Info failed with:

```text
[Errno 13] Permission denied: ''
```

Codex suggested checking how the `lncli` path is loaded and passed into `_run_ln()` in `pybitblock/nodeconnection.py`, as it looks like PyBLOCK is trying to run an empty command.

It also flagged that `Image.ANTIALIAS` needs to be replaced with `Image.Resampling.LANCZOS` for compatibility with the current Pillow version.

With those changes applied, Node Info works for me. Could you update the code and publish a new versioned multi-arch image? Once the PR has the new version and digest, I’ll give it another test.

PyBLOCK-Bitcoin:
> With those changes applied, Node Info works for me. Could you update the code and publish a new versioned multi-arch image? Once the PR has the new version and digest, I’ll give it another test.


How is this going, @GaltRanch? Is there any news?

GaltRanch:
Thanks for testing and for the packaging fixes @mfretz!
Both issues are now fixed upstream in curly60e/pyblock#742:

Permission denied: '' – exactly what Codex suspected: every Lightning function loaded the lncli path into a local variable, while _run_ln() read a module-level default that never got updated, so it executed an empty command. _run_ln() now resolves the configured lncli path at call time and falls back to lncli on PATH (the bundled wrapper in the image).
Pillow – replaced all Image.ANTIALIAS uses with Image.Resampling.LANCZOS.

I've published the new multi-arch image (linux/amd64 + linux/arm64) and updated this PR branch with the new version and digest:
galtranch/pyblock:v4.0.3@sha256:3503deced29cdeb1b5699f3b1ef928ef394f17a10baa55eaea2cccf25d3e9a13
Ready for another test round whenever you have a chance – thanks again!

PyBLOCK-Bitcoin:
@GaltRanch 

Maybe wrong @ to Nathan @nmfretz

PyBLOCK-Bitcoin:
Any news? @GaltRanch @nmfretz

PR #5263, opened 2026-04-02
fix(datum): raise Stratum fd ulimit to 65536/524288

satwise:
## Summary
Raises Datum Stratum service file-descriptor limits to improve stability under higher connection/concurrency load.

## Change
- Update `datum/docker-compose.yml` to set `nofile` ulimit to:
  - soft: `65536`
  - hard: `524288`

## Why
Prevents avoidable fd exhaustion behavior in Stratum-heavy scenarios and keeps the fix isolated from unrelated app changes.

## Scope
- `datum/docker-compose.yml` only

Draft only for now; not ready for review yet.

### DIFF
--- a/datum/docker-compose.yml
+++ b/datum/docker-compose.yml
@@ -12,6 +12,10 @@ services:
     command: ["--config=/app/conf/datum_gateway_config.json"]
     user: 1000:1000
     restart: on-failure
+    ulimits:
+      nofile:
+        soft: 65536
+        hard: 524288
     volumes:
       - ${APP_DATA_DIR}/data/settings/:/app/conf/
     ports:
--- a/datum/docker-compose.yml
+++ b/datum/docker-compose.yml
@@ -12,10 +12,11 @@ services:
     command: ["--config=/app/conf/datum_gateway_config.json"]
     user: 1000:1000
     restart: on-failure
+    # Match Debian's DATUM Gateway systemd service (LimitNOFILE=65535).
     ulimits:
       nofile:
-        soft: 65536
-        hard: 524288
+        soft: 65535
+        hard: 65535
     volumes:
       - ${APP_DATA_DIR}/data/settings/:/app/conf/
     ports:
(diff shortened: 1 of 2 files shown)


nmfretz:
Hey @satwise I see this is still a draft PR since the beginning of April. Is this something you're still working on.

Pinging @Retropex for a second set of eyes in case this is something we should/shouldn't be considering.

Retropex (approved):
I'm fine with it, that's something I do on the [official debian package](https://sources.debian.org/src/datum-gateway/0.4.1~beta-5/debian/datum-gateway.service#L11).

nmfretz:
Thanks @Retropex, we can wait for the next DATUM update and include it.

satwise:
Thank you all for looking at this and sorry I was unavailable for a period of time to answer in a timely manner

nmfretz:
No worries at all @satwise. The plan is to wait until the next DATUM update, unless @Retropex thinks it is worth a hotfix release now. Have there been user reports of issues?

Retropex:
I have not seen anyone complaining on umbrel because we don't expect Eh/s on umbrel device but lets not wait that someone hit the issue and merge this! 🚀

nmfretz:
> I have not seen anyone complaining on umbrel because we don't expect Eh/s on umbrel device but lets not wait that someone hit the issue and merge this! 🚀

Ya same, haven't received any user reports of DATUM hitting fd exhaustion on umbrelOS.

@satwise I've tweaked this slightly before merging so it matches the Debian DATUM Gateway service more closely. Debian sets `LimitNOFILE=65535` for DATUM, so in Docker Compose the closest equivalent is setting both the soft and hard `nofile` limits to `65535`. That gives DATUM plenty of room above the default fd limit without giving the container a much wider hard ceiling than it needs.

Since we haven’t received user reports of DATUM hitting fd exhaustion on umbrelOS, I think the safer path is to keep this change conservative and aligned with the Debian package. This does not reserve memory or file descriptors up front; it just raises the maximum DATUM can use if it actually has that many open sockets/files.

I also bumped version and added release notes.

PR #5264, opened 2026-04-02
fix(blockstream-blind-oracle): avoid AdGuard host-port 8095 collision

satwise:
## Summary
Moves Blockstream Blind Oracle web port off `8095` to avoid collision with AdGuard Home.

## Root cause
- `adguard-home` runs in `network_mode: host` with web UI bound to `0.0.0.0:8095`
- `blockstream-blind-oracle` exported `APP_PINSERVER_WEB_PORT=8095`
- This causes host port collision and downstream startup/connectivity instability (including Tor-facing app path issues)

## Change
- `blockstream-blind-oracle/exports.sh`
  - `APP_PINSERVER_WEB_PORT`: `8095` -> `8102`

## Scope
Single file, single setting change to keep review noise minimal.

Draft only for now; not ready for review yet.

### DIFF
--- a/blockstream-blind-oracle/exports.sh
+++ b/blockstream-blind-oracle/exports.sh
@@ -1,5 +1,5 @@
 export APP_PINSERVER_PORT="8097"
-export APP_PINSERVER_WEB_PORT="8095"
+export APP_PINSERVER_WEB_PORT="8102"
 
 local app_pinserver_hidden_service_file="${EXPORTS_TOR_DATA_DIR}/app-${EXPORTS_APP_ID}-node/hostname"
 


kaichappel:
Thanks @satwise, and sorry for the delay here. We've merged the AdGuard Home conflict fix and included you as a co-author on the commit: https://github.com/getumbrel/umbrel-apps/commit/a4fb239db46b2759a53794c9453988ed27427318

Really appreciate the report. The reason we didn't merge this PR as-is is that changing `APP_PINSERVER_WEB_PORT` would break Blind Oracle installs/updates: that variable is also the internal/app_proxy target, while the `pinserver_web` image still listens inside the container on `8095`.

We adjusted the implementation to keep the internal/app_proxy web port at `8095` and only move the host-published Web UI bind to `8102`.

That frees `8095` for AdGuard Home, while keeping the normal Umbrel app route and the Jade PIN server endpoint on `8097` unchanged.

This PR is covered by the merged fix.

PR #5295, opened 2026-04-05
Update btcpay-server to version 2.3.7

colder-is-better:
This release is the first one using .NET 10, and comes with new features and bug fixes.


al-lac:
Hey @colder-is-better, thanks for providing another btcpay-server update, we value the effort! 🔝 

Tested and ready to go live!

PR #5299, opened 2026-04-07
Enable third-party app access to qBittorrent API

### LINKED ISSUE #4315: Disable Umbrel's Proxy Auth and AuthSubnetWhitelist for qBitTorrent

nmfretz:
## Summary
- Enables qBittorrent's reverse proxy support so it reads the real client IP from `X-Forwarded-For` and checks it against `AuthSubnetWhitelist` instead of the direct connection IP
- Adds `PROXY_AUTH_WHITELIST: "/api/*"` so third-party apps can reach the API without Umbrel's proxy auth
- Includes migration for existing installs

This allows third-party apps (e.g. qRemote on iOS) to authenticate with qBittorrent natively while preserving the subnet whitelist for internal auto-configuration of Radarr, Sonarr, etc.

Fixes #4315

## How it works

- **Third-party app from LAN** (192.168.1.x) → proxy lets `/api/*` through, sets `X-Forwarded-For: 192.168.1.x` → doesn't match `10.21.0.0/16` → qBittorrent requires native login
- **Radarr/Sonarr/MAC** (10.21.x.x) → connects directly, no proxy involved → matches subnet whitelist → no auth needed
- **Browser via Umbrel dashboard** → `PROXY_AUTH_ADD` handles it as usual

@al-lac I've tested this on my Umbrel and confirmed all three scenarios work. No need to merge this PR directly, feel free to bring these changes into your existing qBittorrent PR if that's easier.

### DIFF
--- a/qbittorrent/docker-compose.yml
+++ b/qbittorrent/docker-compose.yml
@@ -6,6 +6,7 @@ services:
       APP_HOST: qbittorrent_server_1
       APP_PORT: 8080
       PROXY_AUTH_ADD: "true"
+      PROXY_AUTH_WHITELIST: "/api/*"
 
   server:
     image: ghcr.io/hotio/qbittorrent:release-5.1.4@sha256:b547db036748d449c5cbd31635dc3b695ccc70424775c44ea91df18bd1da1ea4
--- a/qbittorrent/hooks/post-start
+++ b/qbittorrent/hooks/post-start
@@ -12,10 +12,15 @@ echo "inside qbittorrent/hooks/post-start"
 APP_DATA_DIR="$(readlink -f "$(dirname "${BASH_SOURCE[0]}")/..")"
 QBITTORRENT_CONF_FILE="${APP_DATA_DIR}/data/config/config/qBittorrent.conf"
 UMBREL_QBITTORRENT_CONFIG_FLAG="${APP_DATA_DIR}/HAS_BEEN_CONFIGURED_BY_UMBREL"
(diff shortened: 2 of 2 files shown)


al-lac:
Hey @nmfretz, I integrated the changes into #5246

PR #5301, opened 2026-04-08
Fix SearXNG host trust error and update to 2026.4.7

nmfretz:
## Summary
- Fixed "host not trusted" 400 error when other containers (e.g. Open WebUI) call SearXNG — caused by underscores in the Docker-generated hostname `searxng_web_1` which Werkzeug's host validation rejects
- Set `container_name: searxng-web` (hyphens instead of underscores) and updated `APP_HOST` to match
- Added `SEARXNG_LIMITER: "false"` to disable bot detection, since container-to-container calls lack proxy headers
- Bundled a minimal `settings.yml` using `use_default_settings: true` to enable JSON search format for API consumers like Open WebUI
- Added a pre-start hook to one-time migrate existing installs from the auto-generated 3000-line `settings.yml` to our minimal overlay
- Updated SearXNG from 2026.3.26 to 2026.4.7

### DIFF
--- a/searxng/data/config/settings.yml
+++ b/searxng/data/config/settings.yml
@@ -1,0 +1,8 @@
+# Overlay on top of SearXNG's built-in defaults (avoids shipping the full 3000-line config).
+# JSON format is required for API consumers like Open WebUI.
+use_default_settings: true
+
+search:
+  formats:
+    - html
+    - json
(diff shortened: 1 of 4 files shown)


nmfretz:
Tested fresh install and app update.

PR #5304, opened 2026-04-09
Disable Samba

aidencole98:
## App name
Samba

umbrelOS now includes built-in Samba-based file sharing, so keeping the Samba app in the store creates two ways to do the same thing. The app also has less functionality than the native implementation, which makes it more confusing than useful for users.

This PR disables Samba in two ways: it keeps `disabled: true` for umbrelOS v1.1+ and bumps `manifestVersion` to `1.2` with a removal notice in the description so legacy versions of umbrelOS also can't install it and can see why.

### DIFF
--- a/samba/umbrel-app.yml
+++ b/samba/umbrel-app.yml
@@ -1,4 +1,4 @@
-manifestVersion: 1
+manifestVersion: 1.2
 id: samba
 name: Samba
 tagline: Make your storage accessible using Samba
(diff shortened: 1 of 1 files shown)


nmfretz:
Looks great @aidencole98! Existing installs of Samba will be unaffected. If anyone is making their way to this PR and wondering how to file share on umbrelOS please check out our article:
https://umbrel.com/support/files/accessing-your-files-from-other-devices

PR #5307, opened 2026-04-12
Upgrade to portainer-ce:2.40.0-alpine

### LINKED ISSUE #5306: [ Replace Container ] Portainer with Alpine-based version
Hello,

I can see that there is a new version of Portainer, so I guess it's just a matter of days to get an update:

https://hub.docker.com/r/portainer/portainer-ce/tags

I would like to request that, as part of this new update, we get Alpine-based images instead of the current ones.

This is because Portainer is intended to manage apps that aren't yet supported, or that won't be supported for a variety of reasons. However, the current image does not include either `bash` or `sh`, which means I cannot even log into Portainer to perform manual actions if one of the stacks I am installing requires me to run a CLI command to generate admin credentials.

Please advise,

DavidGarciaCat:
Close: https://github.com/getumbrel/umbrel-apps/issues/5306

Upgraded to the newer Portainer version

I couldn't test this update, as I cannot replicate right now an upgrade from the previous version to the new one while replacing the base image for the new tag using an Alpine-based image

Some help testing this new update would be highly appreciated

### DIFF
--- a/portainer/docker-compose.yml
+++ b/portainer/docker-compose.yml
@@ -27,7 +27,7 @@ services:
       - ${APP_DATA_DIR}/data/docker:/data
 
   portainer:
-    image: portainer/portainer-ce:2.39.1@sha256:1ae8e65d50ca5498cb2c33e617495a1e3ef245b0d2392b4a44c70ae09b822891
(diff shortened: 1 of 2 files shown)


DavidGarciaCat:
Closing this PR, aiming to preserve the current LTS version
Please review the new proposal instead https://github.com/getumbrel/umbrel-apps/pull/5341

PR #5311, opened 2026-04-13
Update tailscale to v1.96.5

colder-is-better:


al-lac:
Thanks a lot for providing this tailscale update @colder-is-better, we appreciate it! 🤝 

Really great job, going live! 🚀

PR #5341, opened 2026-04-14
Portainer CE 2.39.1 LTS through Alpine-based image and without the built-in Auth Proxy

### LINKED ISSUE #5306: [ Replace Container ] Portainer with Alpine-based version

DavidGarciaCat:
Closes https://github.com/getumbrel/umbrel-apps/issues/5306

This PR sets two main changes:
- Disables the Umbrel OS Auth Proxy because:
  - It prevents machine-to-machine comms
  - Portainer already has an internal database of users + it can connect to other OAuth/OpenID services
- Replaces the current docker image with a new one, keeping the same LTS version but based on Alpine Linux

Related links:
- Issue https://github.com/getumbrel/umbrel-apps/issues/5306
- PR https://github.com/getumbrel/umbrel-apps/issues/5307, which will be closed to preserve the LTS version

### DIFF
--- a/portainer/docker-compose.yml
+++ b/portainer/docker-compose.yml
@@ -27,7 +27,7 @@ services:
       - ${APP_DATA_DIR}/data/docker:/data
 
   portainer:
-    image: portainer/portainer-ce:2.39.1@sha256:1ae8e65d50ca5498cb2c33e617495a1e3ef245b0d2392b4a44c70ae09b822891
+    image: portainer/portainer-ce:2.39.1-alpine@sha256:2d423f0e4598833b6cdccca4baf59b552522b68d90a8906f11044bbe91e427f6
     command: --host unix:///var/run/docker.sock --admin-password-file=/default-password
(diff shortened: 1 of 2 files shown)


al-lac:
Hey @DavidGarciaCat, thanks for this! Makes sense to disable the app proxy in that case. I updated from a non-alpine based image to this one, and everything worked fine afterwards. Going live!

PR #5378, opened 2026-04-18
App Submission: HermitStash

dotCooCoo:
## HermitStash

Post-quantum encrypted self-hosted file sharing server.

- **Image:** `ghcr.io/dotcoocoo/hermitstash:1.6.1` (pinned with `@sha256:` digest)
- **Port:** 3080
- **Category:** files
- **Architectures:** amd64, arm64
- **Volumes:** `data/db` → `/app/data`, `data/uploads` → `/app/uploads`
- **Shared memory:** 256MB (in-memory scratch space for encryption operations)

### What it does

HermitStash encrypts uploaded files with ML-KEM-1024, XChaCha20-Poly1305, and Argon2id before they touch disk. It provides shareable download links with expiry and download limits, WebAuthn passkey authentication, S3-compatible storage backends, and an admin panel.

Website: https://hermitstash.com
Source: https://github.com/dotCooCoo/hermitstash
License: AGPL-3.0-or-later

### DIFF
--- a/hermitstash/docker-compose.yml
+++ b/hermitstash/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
-    image: ghcr.io/dotcoocoo/hermitstash:1.7.15@sha256:44238b69447ec2b25492b12a8da03eb7dff224a5e1e909b14dab22088d044355
+    image: ghcr.io/dotcoocoo/hermitstash:1.8.5@sha256:689645120be24f016b4e2741b9d37640f26a7dcd1c149a7eec502e618d3bcff9
     init: true
     restart: on-failure
     stop_grace_period: 1m
--- a/hermitstash/docker-compose.yml
+++ b/hermitstash/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
-    image: ghcr.io/dotcoocoo/hermitstash:1.8.24@sha256:b9f7a0fcf3d23e69c9085c289bd1d3be6dd5734483fd34257942fc2090be352b
+    image: ghcr.io/dotcoocoo/hermitstash:1.9.2@sha256:a50bbbd5fc1f973668a7956afc71453633b9502ac09a00608904a322cb4e4200
     init: true
     restart: on-failure
     stop_grace_period: 1m
--- a/hermitstash/docker-compose.yml
+++ b/hermitstash/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
-    image: ghcr.io/dotcoocoo/hermitstash:1.9.4@sha256:a4400df45f95f0f913fe4dcc4e15cbeb262d9c470c692526418a6884410f7026
+    image: ghcr.io/dotcoocoo/hermitstash:1.9.5@sha256:8cbb72a688bb4afd816db096fa429d63ac46aff37160504d24732ef136b63b40
     init: true
     restart: on-failure
     stop_grace_period: 1m
--- a/hermitstash/docker-compose.yml
+++ b/hermitstash/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
-    image: ghcr.io/dotcoocoo/hermitstash:1.9.5@sha256:8cbb72a688bb4afd816db096fa429d63ac46aff37160504d24732ef136b63b40
+    image: ghcr.io/dotcoocoo/hermitstash:1.9.6@sha256:7055504d30d16cc9b66640188295d6fb2a14879ce812497ac695ed53ce4a88d5
     init: true
     restart: on-failure
     stop_grace_period: 1m
--- a/hermitstash/docker-compose.yml
+++ b/hermitstash/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
-    image: ghcr.io/dotcoocoo/hermitstash:1.8.5@sha256:689645120be24f016b4e2741b9d37640f26a7dcd1c149a7eec502e618d3bcff9
+    image: ghcr.io/dotcoocoo/hermitstash:1.8.15@sha256:7090cc8d150d592c9911b636e87bc0c62a160e1af35c37e1e5c4b66a4db083a9
     init: true
     restart: on-failure
     stop_grace_period: 1m
--- a/hermitstash/docker-compose.yml
+++ b/hermitstash/docker-compose.yml
@@ -1,0 +1,19 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: hermitstash_web_1
+      APP_PORT: 3000
+
+  web:
+    image: ghcr.io/dotcoocoo/hermitstash:1.6.1@sha256:6e9a0f874bebc9e45dac75dcf18e6dc3f5fc4e6f15d86baa3af7a0fb3bb05de7
+    restart: on-failure
+    stop_grace_period: 1m
+    shm_size: "256m"
+    volumes:
+      - ${APP_DATA_DIR}/data/db:/app/data
+      - ${APP_DATA_DIR}/data/uploads:/app/uploads
+    environment:
+      NODE_ENV: production
+      TRUST_PROXY: "true"
--- a/hermitstash/docker-compose.yml
+++ b/hermitstash/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
-    image: ghcr.io/dotcoocoo/hermitstash:1.9.9@sha256:22802449e071e8a572a7340c9ea34c494c9e05797112e3d577e0e09cabafdc1a
+    image: ghcr.io/dotcoocoo/hermitstash:1.9.11@sha256:e989f1edd59bf3616584662d833175727f15a2c22c9ff0f450e6e3f946542e86
     init: true
     restart: on-failure
     stop_grace_period: 1m
--- a/hermitstash/docker-compose.yml
+++ b/hermitstash/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
-    image: ghcr.io/dotcoocoo/hermitstash:1.9.2@sha256:a50bbbd5fc1f973668a7956afc71453633b9502ac09a00608904a322cb4e4200
+    image: ghcr.io/dotcoocoo/hermitstash:1.9.4@sha256:a4400df45f95f0f913fe4dcc4e15cbeb262d9c470c692526418a6884410f7026
     init: true
     restart: on-failure
     stop_grace_period: 1m
--- a/hermitstash/docker-compose.yml
+++ b/hermitstash/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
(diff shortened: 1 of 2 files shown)


dotCooCoo:
Added gallery entries for 5 screenshots. Source images are committed to the HermitStash public repo and directly viewable here if helpful for review:

- https://raw.githubusercontent.com/dotCooCoo/hermitstash/main/public/img/screenshots/1.png
- https://raw.githubusercontent.com/dotCooCoo/hermitstash/main/public/img/screenshots/2.png
- https://raw.githubusercontent.com/dotCooCoo/hermitstash/main/public/img/screenshots/3.png
- https://raw.githubusercontent.com/dotCooCoo/hermitstash/main/public/img/screenshots/4.png
- https://raw.githubusercontent.com/dotCooCoo/hermitstash/main/public/img/screenshots/5.png

All are 1638×832 PNG. Happy to re-export to any target dimension/format if needed.

hermitstash/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
...
+version: "1.9.11"
+tagline: Post-quantum encrypted self-hosted file sharing
+description: >-
+  HermitStash is a self-hosted file upload server with post-quantum encryption.
+  Files are sealed with ML-KEM-1024, XChaCha20-Poly1305, and Argon2id before
+  touching disk. Supports shareable download links with expiry and download
+  limits, WebAuthn passkey authentication, S3-compatible storage backends, and
+  an admin panel for managing users, uploads, and server settings. The TLS
+  layer negotiates X25519MLKEM768 for quantum-resistant key exchange on
+  supported browsers.
```
al-lac: Fix description formatting. Feel free to add more paragraphs (needs 2 empty new lines in between).
```suggestion
  HermitStash is a self-hosted file upload server with post-quantum encryption.
  
  
  Files are sealed with ML-KEM-1024, XChaCha20-Poly1305, and Argon2id before touching disk. Supports shareable download links with expiry and download limits, WebAuthn passkey authentication, S3-compatible storage backends, and an admin panel for managing users, uploads, and server settings. The TLS layer negotiates X25519MLKEM768 for quantum-resistant key exchange on supported browsers.
```

hermitstash/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: hermitstash_web_1
+      APP_PORT: 3000
+
+  web:
+    image: ghcr.io/dotcoocoo/hermitstash:1.9.11@sha256:e989f1edd59bf3616584662d833175727f15a2c22c9ff0f450e6e3f946542e86
```
al-lac: Run rootless. Maybe the PUID and PGID below can be removed then.
```suggestion
    image: ghcr.io/dotcoocoo/hermitstash:1.9.11@sha256:e989f1edd59bf3616584662d833175727f15a2c22c9ff0f450e6e3f946542e86
    user: "1000:1000"
```

al-lac (changes requested):
Hey @dotCooCoo and thanks for submitting your App to the Umbrel App Store 🔝 

Really cool use case and a nice user interface to go with it! 💯 

I noted down some things that could be adapted.

During testing I noticed two things:
* The signup apparently sends an email for confirmation, can this be adapted so users can easily register without a confirmation?
* After uploading a file, the link points to "http://localhost:3000", ideally this should point to whatever umbrel has set as its hostname using "${DEVICE_DOMAIN_NAME}"

Thanks again, feel free to reach out if there are any questions.

dotCooCoo:
Thanks for the review, and sorry for missing these in the initial submission — both turned out to be existing HermitStash config knobs that just needed to be wired through the compose file:

- `EMAIL_VERIFICATION: "false"` — turns off the verification-email round-trip on signup. Operators who want it can flip it back from the admin UI.
- `RP_ORIGIN: "http://${DEVICE_DOMAIN_NAME}:3080"` — anchors share URLs (and the WebAuthn relying-party origin) to the device's domain instead of localhost.

Also bumped the image from `1.9.11` to `1.10.2` to pick up the latest release.

dotCooCoo:
Second round pushed. All four review items addressed in `b7071c43`:

- **Description formatting** — split into four paragraphs with blank-line separators using the YAML folded-scalar form.
- **Rootless** — switched to `user: "1000:1000"` and dropped the `cap_add` block (CHOWN/SETUID/SETGID/DAC_OVERRIDE were only there to let the root-mode entrypoint chown volumes + `setpriv` to a non-root user). PUID/PGID env vars removed for the same reason.
- **Custom-branding writable surface** — the in-image `/app/public/img/custom` directory is owned by HermitStash's `hermit` system user (low UID), which isn't writable as 1000. Added a `${APP_DATA_DIR}/data/custom` bind mount on top so the rootless container can land admin-uploaded logos/favicons. Empty `data/custom/.gitkeep` ships the template directory.
- **`releaseNotes`** — reset to `""` per the linter rule for new app submissions.

Lint is green; let me know if anything else needs tweaking.

dotCooCoo:
Bumped to HermitStash v1.11.5 (sha256:0a2c638b…). v1.11.x patch line covers: cert isolation in test infra, dependabot test-dep CVE bumps (brace-expansion / fast-uri — devDeps only, not in runtime image), Docker SIGTERM graceful-shutdown path, and silenced a framework boot-warning that was emitting at error level. No operator-facing config changes since v1.11.1.

al-lac:
Hey @dotCooCoo, thanks for the updates! Sadly I only get this now when opening up the app: `Forbidden`

Not sure if it is because of any of these warnings:
```
hermitstash_web_1        |   Startup warnings:
hermitstash_web_1        |   ⚠ Session secret is the default value. Change it in admin settings.
hermitstash_web_1        |   ⚠ rpOrigin is not HTTPS. Session cookies may not be secure. Use HTTPS in production.
hermitstash_web_1        |   ⚠ Email is not active (no email backend configured (SMTP_HOST/SMTP_USER or RESEND_API_KEY)) — self-serve password reset will silently fail. Admins can generate reset links manually via Admin → Users → Reset link.
hermitstash_web_1        |   ⚠ RP_ORIGIN='http://umbrel-dev.local:3080' is invalid: invalid URL. Using default.
hermitstash_web_1        |   ⚠ data/ directory has loose permissions (755). Consider chmod 700.
```

Could you take a look at this?

dotCooCoo:
Hey @al-lac — sorry about that, and thanks for the detailed paste. The one that actually matters is the `RP_ORIGIN` line:

```
⚠ RP_ORIGIN='http://umbrel-dev.local:3080' is invalid: invalid URL. Using default.
```

The validator was being too strict and rejecting any `http://` URL, so the server quietly ignored the origin Umbrel was passing in and fell back to `http://localhost:3000`. Once that happened, every request you sent from `umbrel-dev.local:3080` looked cross-origin to it and got bounced as `Forbidden`. Same page-load, every time.

Fix shipped in HermitStash v1.11.16 and I just pushed the bump on this PR — pull it down and re-run the install and you should land on the login page instead.

The other warnings aren't what's blocking you, but for what it's worth:

- **Session secret default** — worth swapping in admin settings once you're in (it's a hardening item, not the blocker).
- **`data/` permissions 755** — just a local-fs nudge.
- **Email not active** — only affects self-serve password reset. Admin → Users → Reset link still works.
- **rpOrigin is not HTTPS** — a soft nudge toward TLS, not the refusal.

Let me know if it's still misbehaving after the bump.

al-lac:
Hey @dotCooCoo, thanks for the fix! Sadly I still get the same error. Even though the warning from before is now gone:

```
hermitstash_web_1  |   Startup warnings:
hermitstash_web_1  |   ⚠ Session secret is the default value. Change it in admin settings.
hermitstash_web_1  |   ⚠ rpOrigin is not HTTPS. Session cookies may not be secure. Use HTTPS in production.
hermitstash_web_1  |   ⚠ Email is not active (no email backend configured (SMTP_HOST/SMTP_USER or RESEND_API_KEY)) — self-serve password reset will silently fail. Admins can generate reset links manually via Admin → Users → Reset link.
hermitstash_web_1  |   ⚠ data/ directory has loose permissions (755). Consider chmod 700.
```

dotCooCoo:
Thanks for re-testing, and for confirming the earlier `RP_ORIGIN` warning is gone. The four warnings you're still seeing are all non-blocking (session secret, the HTTPS nudge, email backend, data-dir perms); none of them produce the `Forbidden`.

I tracked down what did. HermitStash runs a lightweight bot filter on page loads that keys off the browser's `Sec-Fetch-*` request headers. Per the Fetch Metadata spec, browsers only send those headers to a "secure context" — an HTTPS origin, or `localhost`. An Umbrel app is reached over plain HTTP at a `.local` hostname (`http://umbrel-dev.local:3080`), which isn't a secure context, so your browser correctly omits `Sec-Fetch-Mode` — and the filter was misreading that absence as a bot and returning `Forbidden`. It never surfaced in my own testing because that runs over `localhost`, where the header is present. (The same trap would catch older Safari, which omits those headers even over HTTPS, so it was worth fixing regardless of Umbrel.)

The fix makes a missing `Sec-Fetch-Mode` advisory-only — it can no longer refuse a request on its own, while genuine bots are still caught by the other checks. It's shipped, and I've bumped the image on this PR. Pull it down and re-run the install and you should land on the login page — let me know if anything's still off.

al-lac:
Hey @dotCooCoo, awesome, thanks for taking the time and effort to look into and fix this! 💯 

I tested it again and it looks a lot better now! Was able to create an account and upload and download files like one would expect to use the app. Great job! 🔝 

The only thing that does not seem to work right now are the passkeys, I assume this is also because of the HTTPS / localhost issue: `Failed: Cannot read properties of undefined (reading 'create')`

Maybe this feature could be disabled with a flag until Umbrel supports HTTPS.

We will get started on creating the gallery assets!

dotCooCoo:
You're exactly right — passkeys (WebAuthn) are only exposed to the browser in a secure context (HTTPS, or `localhost`), so there's no server-side switch that can turn them on over plain HTTP. That's a browser platform rule, not something the app controls.

What I've fixed is the bad behaviour around it: instead of throwing `Cannot read properties of undefined`, the passkey option now detects the missing API and hides itself with a short "Passkeys require a secure (HTTPS) connection" note. It comes back automatically if/when the app is served over HTTPS — nothing to toggle. (There's also a `PASSKEY_ENABLED` setting in the admin panel if you ever want it off explicitly.)

While I was in there I checked for the same trap elsewhere and fixed two more: the copy-link buttons and multi-file uploads were leaning on secure-context-only browser APIs too, so they now fall back to paths that work over plain HTTP.

All shipped in v1.11.18 and I've bumped the image on this PR — pull it and the passkey crash should be gone, copy-link should work, and password sign-in works as before. Thanks for the thorough testing on this one.

al-lac:
Hey @dotCooCoo, thanks a lot for your efforts in getting HermitStash ready. We have now added the gallery assets and are ready to go live.

Happy to have you as part of the community, welcome to the Umbrel App Store! 🚀 

CleanShot 2026-05-29 at 14 37 19@2x

dotCooCoo:
Thank you for all of your efforts as well! @al-lac

al-lac:
Thanks @dotCooCoo! 

If you want to promote the fact that your app is on the Umbrel App Store, feel free to use any of these assets:

[badge-light](https://apps.umbrel.com/app/hermitstash)
[badge-dark](https://apps.umbrel.com/app/hermitstash)
[badge-light-logo](https://apps.umbrel.com/app/hermitstash)
[badge-dark-logo](https://apps.umbrel.com/app/hermitstash)

PR #5382, opened 2026-04-20
Add ZMQ publish option to Monero command, required for p2pool mining

brainchainz:
This publishing is required for Monero nodes to connect with the P2pool decentralized mining protocol. Users currently have to add this in manually. Adding it here would make the mining experience seamless for non-technical users.


kaichappel:
Thanks again @brainchainz. This one is covered by the same Monero ZMQ update we just merged, and I left the details here: https://github.com/getumbrel/umbrel-apps/pull/5115#issuecomment-4588859363

Really appreciate you following up on the p2pool use case. The merged version keeps the flag in `exports.sh` via `BIN_ARGS`, and you’re included as a co-author on the commit.

PR #5429, opened 2026-04-23
App Submission: BOLT12 Pay

Alex71btc:
# App Submission

### App name

BOLT12 Pay

### 256x256 SVG icon

grafik
grafik

### App description

BOLT12 Pay is a self-hosted Lightning payments and identity app for LND nodes.

It supports:

* BOLT12 Offers
* Lightning Address (BIP353)
* LNURL
* BOLT11 fallback
* optional Nostr identity features such as NIP-05 and Zaps

This app is intended for sovereign Bitcoin users who want full control over their payment and identity infrastructure.

Real-world testing confirms successful native BOLT12 payments, including:

* BIP353 (Lightning Address) → BOLT12 Offer resolution
* direct BOLT12 Offer payments
* successful payments to Phoenix wallet offers (ACINQ)
* successful payments from small test amounts up to 21,000 sats
* fully self-hosted routing via Umbrel LND

After establishing direct liquidity to ACINQ, Phoenix offer payments became consistently reliable, demonstrating that the app is production-capable and not merely experimental.

### Notes for reviewers

This app currently depends on a custom LND build with BOLT12-related protocol support enabled. 
(user-friendly instructions are included in the app ui)

Required LND config:

```ini
[protocol]
protocol.custom-message=513
protocol.custom-nodeann=39
protocol.custom-init=39
```

However, LND v0.21.0 is expected to support onion messages natively, which should significantly simplify deployment and remove the need for custom protocol patches.


Alex71btc:
Thanks — linting is now clean and the app has been tested successfully on Umbrel Home and umbrelOS VM.

Happy to make any additional adjustments if needed.

Alex71btc:
I found that the current pinned `bolt12-pay` image was built from the wrong Dockerfile/context and needs to be corrected before review/testing.

I’m pushing a fix shortly.

Alex71btc:
Quick update:

During testing on a real Umbrel Home deployment I found that the previously pinned `bolt12-pay` image had been built from the wrong Docker build context.

This caused the container to start `lndk` directly instead of the FastAPI backend (`/app/entrypoint.sh`), which would have broken real deployments.

This is now fixed:

* correct Docker build context used
* proper multi-arch image rebuilt
* pinned sha256 digests updated
* linter passes cleanly
* successfully tested on a live Umbrel Home node

The current PR state reflects the corrected production-ready setup.

Alex71btc:
Pushed a critical documentation fix for the Umbrel LND configuration snippet.

The required keys now use the correct protocol.* format:

[protocol]
protocol.custom-message=513
protocol.custom-nodeann=39
protocol.custom-init=39

bolt12-pay/screenshots/1.png
al-lac, line 1: Please keep screenshots and icons only in the PR details.

bolt12-pay/docker-compose.yml
```diff
@@ -0,0 +1,70 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: bolt12-pay_web_1
+      APP_PORT: 8081
+      PROXY_AUTH_WHITELIST: "*"
+      PROXY_AUTH_BLACKLIST: "/pay,/pay-login"
+  web:
```
al-lac: Add new line for better structure.
```suggestion
      PROXY_AUTH_WHITELIST: "*"
      PROXY_AUTH_BLACKLIST: "/pay,/pay-login"

  web:
```

bolt12-pay/docker-compose.yml
```diff
@@ -0,0 +1,70 @@
...
+      LNURL_MIN_SENDABLE_MSAT: "1000"
+      LNURL_MAX_SENDABLE_MSAT: "1000000000"
+      LNURL_COMMENT_ALLOWED: "120"
+      LNURL_ALIAS_MODE: "shared"
+      LNURL_SHARED_DESCRIPTION: "LNURL payment"
+      LNURL_DEFAULT_DESCRIPTION: "Lightning payment"
+      LNURL_ALIAS_MAP: ""
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${UMBREL_ROOT}/app-data/lightning/data/lnd:/lnd:ro
```
al-lac: Use of exported path from the lightning app. Best double check if correct path.
```suggestion
      - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
```

bolt12-pay/docker-compose.yml
```diff
@@ -0,0 +1,70 @@
...
+      LNDK_GRPC_HOST: https://lndk
+      LNDK_GRPC_PORT: "7000"
+      LNDK_CERT_PATH: /data/lndk/tls-cert.pem
+      LNDK_MACAROON_PATH: /lnd/data/chain/bitcoin/mainnet/admin.macaroon
+      LNDK_TIMEOUT_SECONDS: "30"
+      ALLOW_PAY_OFFER: "true"
+
+      LND_REST_URL: https://lightning_lnd_1:8080
+      LND_TLS_CERT_PATH: /lnd/tls.cert
+      LND_MACAROON_PATH: /lnd/data/chain/bitcoin/mainnet/admin.macaroon
```
al-lac: The active network could be used here.
```suggestion
      LND_MACAROON_PATH: /lnd/data/chain/bitcoin/${APP_BITCOIN_NETWORK}/admin.macaroon
```

bolt12-pay/docker-compose.yml
```diff
@@ -0,0 +1,70 @@
...
+      SECRETS_JSON_PATH: /data/config/secrets.json
+      PAY_UI_SESSION_TTL: "1800"
+      CORS_ORIGINS: "*"
+
+      LNDK_CLI: /usr/local/bin/lndk-cli
+      LNDK_NETWORK: bitcoin
+      LNDK_GRPC_HOST: https://lndk
+      LNDK_GRPC_PORT: "7000"
+      LNDK_CERT_PATH: /data/lndk/tls-cert.pem
+      LNDK_MACAROON_PATH: /lnd/data/chain/bitcoin/mainnet/admin.macaroon
```
al-lac: The active network could be used here.
```suggestion
      LNDK_MACAROON_PATH: /lnd/data/chain/bitcoin/${APP_BITCOIN_NETWORK}/admin.macaroon
```

bolt12-pay/docker-compose.yml
```diff
@@ -0,0 +1,70 @@
...
+  lndk:
+    image: alex71btc/lndk:stable-good-20260401@sha256:ff16de59cacb5e5a9d53b76e879dcaaaa9b204537dc26697275b737c489f2939
+    restart: on-failure
+    entrypoint:
+      - /bin/sh
+      - -c
+    command:
+      - |
+        echo "Waiting for LND TLS + macaroon..."
+        while [ ! -f /lnd/tls.cert ] || [ ! -f /lnd/data/chain/bitcoin/mainnet/admin.macaroon ]; do
```
al-lac: The active network could be used here.
```suggestion
        while [ ! -f /lnd/tls.cert ] || [ ! -f /lnd/data/chain/bitcoin/${APP_BITCOIN_NETWORK}/admin.macaroon ]; do
```

bolt12-pay/docker-compose.yml
```diff
@@ -0,0 +1,70 @@
...
+          echo "Waiting for cert/macaroon..."
+          sleep 5
+        done
+
+        echo "Starting LNDK with retry loop..."
+        while true; do
+          lndk --address=https://lightning_lnd_1:10009 --cert-path=/lnd/tls.cert --macaroon-path=/lnd/data/chain/bitcoin/mainnet/admin.macaroon --data-dir=/data/lndk --grpc-host=0.0.0.0 --grpc-port=7000
```
al-lac: The active network could be used here.
```suggestion
          lndk --address=https://lightning_lnd_1:10009 --cert-path=/lnd/tls.cert --macaroon-path=/lnd/data/chain/bitcoin/${APP_BITCOIN_NETWORK}/admin.macaroon --data-dir=/data/lndk --grpc-host=0.0.0.0 --grpc-port=7000
```

bolt12-pay/docker-compose.yml
```diff
@@ -0,0 +1,70 @@
...
+          echo "LNDK exited or failed. Retrying in 10s..."
+          sleep 10
+        done
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${UMBREL_ROOT}/app-data/lightning/data/lnd:/lnd:ro
```
al-lac: Use of exported path from the lightning app. Best double check if correct path.
```suggestion
      - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
```

bolt12-pay/icon.svg
al-lac, line 1: Keep icon only in PR description.

bolt12-pay/LICENSE
al-lac, line 1: Not needed here.

bolt12-pay/README.md
al-lac, line 1: Not needed here.

bolt12-pay/umbrel-app.yml
```diff
@@ -0,0 +1,43 @@
+manifestVersion: 1
+id: bolt12-pay
+name: BOLT12 Pay
+icon: ""
```
al-lac: Remove icon as it will be filled automatically.
```suggestion
```

bolt12-pay/umbrel-app.yml
```diff
@@ -0,0 +1,43 @@
...
+    protocol.custom-message=513
+    protocol.custom-nodeann=39
+    protocol.custom-init=39
+
+  Restart Lightning afterwards. Without this, BOLT12 functionality will not work.
+
+  BOLT12 support on LND is still evolving, wallet compatibility may vary,
+  and this app uses cutting-edge Lightning features.
+
+  Made for sovereign Bitcoin users.
```
al-lac, line 33: Improve description formatting.
```suggestion
  BOLT12 Pay is a self-hosted Lightning payments and identity app for LND nodes.
  It supports BOLT12 Offers, Lightning Address (BIP353), LNURL, and BOLT11 fallback,
  with optional Nostr identity features such as NIP-05 and Zaps.


  This app currently depends on a custom LND build with BOLT12-related custom
  message support enabled.


  In your LND configuration, add:

    [protocol]
    protocol.custom-message=513
    protocol.custom-nodeann=39
    protocol.custom-init=39


  Restart Lightning afterwards. Without this, BOLT12 functionality will not work.


  BOLT12 support on LND is still evolving, wallet compatibility may vary,
  and this app uses cutting-edge Lightning features.


  Made for sovereign Bitcoin users.
```

al-lac (changes requested):
Hey @Alex71btc and thanks a lot for submitting your app to the Umbrel App Store! 🔝 

This looks like a great fit for Umbrel and we appreciate the work you put into this PR. 🤝 

I took a look at it and some things need some fine tuning, but for the most part it looks really good already!

What I noticed during the setup is that it switched to german after clicking the setup button, is this maybe a bug?

Please take a look at the suggestions and feel free to reach out in case of open questions.

Alex71btc:
Thanks a lot for the detailed review!
 
Please let me know if anything else should be adjusted.
I applied all requested changes and resolved the conversations and re-requested review.
German was the default language but after switsching to english after setup it should stay english.

Alex71btc:
Thanks again for the detailed review!

I applied the requested changes, resolved the conversations, and moved the icon/screenshots to the PR details as suggested.

Please let me know if anything else should be adjusted.

al-lac:
Hey @Alex71btc,

thanks a lot for making the suggested adaptions 🔝 

Any way you could get the app to stick to english or the browser language per default, that would improve the UX a lot. I will also try to get @nmfretz to review this app and get back to you soon on this.

Alex71btc:
Thanks a lot for the detailed review and the positive feedback — really appreciated! 🙏

Regarding the language behavior: German is currently the historical default language because the app originally evolved around German text/content and English support was added later on.

I suspect what you saw may be related to browser language detection and the setup flow opening a fresh session/window. I’ll definitely take another look at improving the default UX here and see if I can make it consistently prefer English or the browser language.

Also as a quick update:

* BOLT12 Pay has already been published in the official Start9 Community Registry
* Umbrel team members including @mayank have also been very positive about the project so far
* I will get to work making English the default.
* you can test the complete app already here: https://github.com/Alex71btc/umbrel-community-store/tree/master/alex-bolt12-pay

Really excited about the possibility of bringing native BOLT12 support to Umbrel users 🚀

Alex71btc:
Implemented the requested UX improvements now 🙂

This update includes:

* browser language detection, English as default.
* persistent language selection
* fixed translation runtime behaviour
* removed inconsistent German fallback behaviour after setup

Fresh sessions should now correctly respect browser locale while still allowing manual language switching.

al-lac:
Hey @Alex71btc,

thanks a lot for making the adaptions, sadly it still uses the german language as a default even though my browser local is `en_US`. Any clue what the issue here could still be?

For now I will contact the team about creating the required gallery assets.

Alex71btc:
Well I am sure the setup page starts in English now with the latest image this behaviour showed in my tests. Also the selected language should stay English once selected. But I will investigate this for further updates as soon as possible!
Great to hear the progress.

Alex71btc:
Just try a browser hard refresh then the app should behave with English as default!

Alex71btc:
Looked into this again. 

I think this may now be caused by persisted localStorage state from previous sessions.

The app currently prioritizes an explicitly selected language stored in localStorage over browser locale detection, which is intentional so users keep their chosen language across sessions.

Could you maybe test once in:

- a private/incognito window
- or after clearing localStorage/site data for the app

Fresh sessions should now default to the browser language correctly.

al-lac:
Hey @Alex71btc, thanks for looking into this again!

It works fine for the initial screen, but once I press `Open App`, or any other button, the next screen is in german. Private windows and clearing the cache / cookies did not help here sadly.

URL: http://umbrel.local:8367/app

Alex71btc:
Ok will try to improve this behaviour in the next update!

Alex71btc:
Thanks again for the detailed testing 🙂

I found the root cause now:
some public runtime translation fallbacks still defaulted to German internally even in English mode.

This is now fixed in 0.2.93:

I re-tested successfully with:

fresh Firefox EN session
fresh Chromium profile
clean localStorage
setup → app → pay navigation flow

The app now correctly stays in English across the full flow.

Updated Docker image + Umbrel submission are pushed now.

I hope I made all necessary changes so the app now uses the correct docker image 0.2.93!

al-lac:
Hey @Alex71btc,

Great! Thanks for taking a closer look at this and finding a solution.

Also tested successfully on my side. ⚡ 

Will let you know once the gallery assets have been created.

Alex71btc:
Awesome!

Alex71btc:
Updated the submission to v0.2.94.

This includes fixes for:

* Landing page copy button
* DE/EN language switching
* Amount label translation
* Improved toast styling

The updated multi-arch Docker image has been published and the app manifest has been bumped accordingly.

nmfretz:
Gallery assets are ready @Alex71btc!

I just gave the current package a look and test. I noticed one auth-boundary issue we may need to resolve before this can go live.

You've set `PROXY_AUTH_WHITELIST: "*"` which makes almost the whole app reachable without Umbrel authentication, with only `/pay` and `/pay-login` added back behind auth via `PROXY_AUTH_BLACKLIST`. So a ton of the backend API surface is still reachable directly.

The app has access to the user's Lightning node macaroon and payment/configuration state, so setup, config, admin, alias, DNS, and payment-control routes probably need to be treated as sensitive. If those routes are callable without a real authenticated session, someone who can reach the app URL could read or change app/payment configuration, interfere with aliases or DNS-backed payment addresses, and potentially trigger Lightning payment actions through the app.

If I'm misunderstanding the intended upstream auth flow, please push back. But I think we need to narrow the app proxy whitelist to only the routes that genuinely need to be public, and keep setup/admin/payment-control APIs behind Umbrel auth and/or the app's own server-side auth layer.

What do you think?

Alex71btc:
@nmfretz Thanks for taking a close look at this.

A few clarifications about the current auth model:

- Initial setup is only available until a Pay password has been configured.
- Admin/Pay functionality is protected by the app's own server-side password authentication.
- Sensitive actions require a valid authenticated session cookie.
- Login attempts are rate limited.
- CSRF protection is enabled for state-changing requests.
- Alias/public payment pages are intentionally reachable without authentication because they are intended to be public payment endpoints.

That said, I agree the current Umbrel proxy configuration ("PROXY_AUTH_WHITELIST="*"") deserves another review.

I'll audit the exposed routes and verify that setup, admin, config, DNS, alias-management and payment-control endpoints cannot be used without a valid authenticated session. If necessary I'll narrow the proxy whitelist to only the routes that genuinely need to be public.

Thanks for catching this.

Alex71btc:
I just reproduced the issue locally and confirmed that /api/setup/config is currently reachable without authentication and exposes configuration data.

You're right that this needs to be tightened before release.


Although the  app already has its own password-protected session layer for setup, admin and payment functionality, but I agree that sensitive configuration endpoints should not be exposed through the Umbrel proxy either.

I'm currently auditing all setup/admin/config routes and will restrict access so that only the intentionally public endpoints (landing page, aliases, LNURL and payment receiver endpoints) remain accessible without authentication.

I'll push an update shortly.The app already has its own password-protected session layer for setup, admin and payment functionality, but I agree that sensitive configuration endpoints should not be exposed through the Umbrel proxy either.

I'm currently auditing all setup/admin/config routes and will restrict access so that only the intentionally public endpoints (landing page, aliases, LNURL and payment receiver endpoints) remain accessible without authentication.

I'll push an update shortly.

Alex71btc:
Implemented the requested security changes and completed end-to-end testing.

Changes:

* Moved public alias pages to `/alias/*`
* Restricted public access to only:

  * `/`
  * `/links`
  * `/alias/*`
  * `/.well-known/*`
  * LNURL endpoints
  * QR endpoints
  * required public assets
* Admin, setup and management routes are no longer publicly exposed
* Added dedicated alias routing to avoid future route collisions

Verified:

* Public alias pages
* BOLT12 payments
* Lightning Address / BIP353
* LNURL payments
* NIP-05
* NWC connections
* Nostr zap notifications
* Umbrel deployment
* StartOS deployment

One observation: when accessing protected routes such as `/app` or `/pay` through an external Cloudflare Tunnel hostname, Umbrel auth redirects to:

`http://.:2000/?origin=host&app=bolt12-pay...`

which results in NXDOMAIN externally. Public payment functionality works correctly, but external Umbrel-authenticated routes appear to be limited by the current app-proxy behavior rather than the application itself.

Overall, the public attack surface is now significantly reduced while preserving all intended public payment functionality.

Alex71btc:
@nmfretz , @al-lac  I investigated the auth model further and moved protection of the sensitive functionality into the backend itself.

The following routes now require an authenticated pay session:

* setup configuration
* alias management
* Cloudflare/BIP353 management
* admin and NWC management endpoints

I verified this from outside the Umbrel UI against the public Cloudflare URL and http://umbrel.local:8367:

* GET /api/setup/config → 401 Authentication required
* GET /api/alias → 401 Authentication required
* GET /api/admin/nostr-status → 401 Authentication required
* GET /api/admin/nwc/connections → 401 Authentication required

I also tested the complete user flow (pay UI, app UI, offer creation/payment, alias pages, NWC and PWA installation).

Keeping the proxy whitelist heavily restricted caused a number of issues with the public pay UI and authentication flow. With the current build, the public payment experience works correctly while sensitive configuration and management functionality is protected server-side.

Alex71btc:
@nmfretz @al-lac Thanks again for the review and the auth-boundary feedback.

I've implemented additional protections and updated the package to v0.2.98.

### Changes

* Added server-side authentication checks (`require_pay_auth`) for sensitive endpoints:

  * `/api/setup/config`
  * `/api/alias`
  * `/api/alias/*`
  * `/api/cloudflare/create-bip353`
  * Nostr admin endpoints
  * NWC management endpoints

* Verified unauthenticated access now returns `401 Authentication required` for alias, admin and management APIs.

* Added Umbrel auth protection for `/api/setup/config`.

### First-run setup scenario

One concern I had after your review was the initial setup flow.

Previously, on a fresh installation, an unauthenticated user who could reach the app URL could potentially access setup/configuration endpoints before the owner completed setup.

With the current package:

* `/api/setup/config` is no longer directly accessible without Umbrel authentication.
* The setup wizard was updated to gracefully handle blocked config access.
* Public payment functionality continues to work as expected.

This means a user can no longer directly access the setup configuration endpoint and claim/configure a fresh installation before the owner completes setup.

### Verification

Tested locally:

* `/api/alias` → `401`
* `/api/admin/nostr-status` → `401`
* `/api/admin/nwc/*` → `401`
* `/api/setup/config` → Umbrel auth redirect (`302`)

Public payment flows, alias pages, LNURL, BOLT12 payments, Nostr functionality and the PWA install flow remain operational.

Updated image:

`alex71btc/bolt12-pay:0.2.98`

The app now follows a layered model:

* Public payment and discovery endpoints remain accessible.
* Management and configuration endpoints require a valid authenticated pay session. (rate limiting)
* Initial setup configuration access is additionally protected by Umbrel authentication.

Alex71btc:
Small update for the submission:

This bumps BOLT12 Pay to v0.2.99 and updates the image digest.

Changes include:

* Improved setup reliability during onboarding when `/api/setup/config` is not yet available.
* Simplified StartOS onboarding following the new LND Auto-Configure workflow for Onion Messages.
* Verified upgrade and fresh-install testing on Umbrel.

The setup endpoint remains protected and returns `401 Unauthorized` when accessed without authentication.

Alex71btc:
Hey, @al-lac , did you find some time to review the security changes I implemented? Gallery assets seem to be ready...

Alex71btc:
Additional update pushed while testing the security fixes.

I discovered that legacy default relay settings could still cause newly created Nostr identities to inherit `relay.alex71btc.com` on upgraded installations, even though the relay had already been removed from the current defaults.

This update:

* removes `relay.alex71btc.com` from legacy default relay sets during migration
* preserves user-customized relay configurations
* keeps manual relay entry fully supported
* ensures new identities use the current neutral default relay set

The goal is to avoid new installations and upgraded deployments unintentionally depending on a third-party relay while preserving user choice.

al-lac:
Hey @Alex71btc, thanks a lot for all the updates and improvements! 💯 

@kaichappel and @nmfretz will take take over reviewing this PR, they should reach out soon.

nmfretz:
@Alex71btc awesome, glad to see the changes!

Looks like there's still a packaging issue left over that circumvents your backend changes though. The compose still has:

```yaml
PROXY_AUTH_WHITELIST: "*"
PROXY_AUTH_BLACKLIST: "/api/setup/config"
```

That means everything except `/api/setup/config` bypasses Umbrel auth and relies on the app backend to protect each sensitive route correctly.

Could you remove that setup and replace it with a narrow `PROXY_AUTH_WHITELIST` for only the routes that intentionally need to be public? That way public payment/discovery pages can still work, while setup/admin/config/management/payment-control routes require Umbrel auth by default.

Once that’s tightened up, I can re-test.

nmfretz:
@Alex71btc some pretty sick gallery assets are ready to go ⚡️⚡️⚡️!

https://github.com/getumbrel/umbrel-apps-gallery/pull/79/changes

Alex71btc:
@nmfretz   Thanks for the review.

One concern I have is that I previously tested a much stricter Umbrel auth configuration and it broke the public payment flow completely when accessing the app through a public HTTPS endpoint (for example Cloudflare Tunnel). That happened because umbrel authis not showing when accessing over a public https domain e.g. pay.alex71btc.com...

BOLT12 Pay is intentionally both an admin application and a public payment service, so endpoints used for BOLT12, LNURL, BIP353 and public payment pages must partley remain reachable without an Umbrel user session.


Before I make further changes, could you clarify which specific routes you expect to remain behind Umbrel auth?

I fully agree that setup, admin and management functionality should be protected, but I want to avoid breaking public payment functionality that external wallets depend on.

One thing that may not be obvious from the compose configuration alone is that the current backend no longer exposes sensitive functionality through public unauthenticated routes.

The app originally relied more heavily on Umbrel proxy auth, but that broke public payment flows ( LNURL, BIP353 and Cloudflare Tunnel based access only Bolt12 worked), so the security model was moved into the backend itself.

Since then I added additional hardening:

- server-side authenticated sessions
- password hashing
- CSRF protection
- login rate limiting
- setup-state validation
- route-level authentication checks
- separation between public payment pages and management/admin functionality

As a result, setup, configuration, admin and payment-management actions are not accessible without a valid application session, even though the proxy configuration itself is broader.

I'm happy to further reduce the proxy-exposed surface where possible, but I wanted to clarify that the current implementation already blocks unauthenticated access to sensitive functionality at the application layer.

Another solution using umbrel auth would be that the app is only reachable over the umbrel.local address, the intended public user interface over domain (e.g. pay.alex71btc.com) would be not usable, still have to test if lightning lnurl addresses would still work using that domain despite of this. 

That being said I will consider seperating public payment pages and wallet pages and test narrowed whitelist again and if lnurl receiving still works this can be mentioned in the app description so users still can use an Umbrel authenticated session to initiate payments. But this will give a bad user experience when invoking the app over public domain since payments would fail then, due to no possibility of umbrel auth session. Or could this be solved by a redirect?

nmfretz:
ya sorry for all the confusion @Alex71btc! I really don't want to be an annoyance. I just want to protect you from the footguns of using `PROXY_AUTH_WHITELIST: "*"`,  because that makes every currently-missed and future sensitive API route public unless backend auth catches it. It's a recipe for a really nasty slip-through for an app with access to the LND admin macaroon. 

Like for example, I just asked codex to check for currently unprotected endpoints and one of them is `/api/pay-bolt11`. 

So imagine someone tricks a user into going to `http://evil.com/free-sats` while on the same LAN as their umbrel. The malicious page runs something like:
```
<script>
fetch("http://<victim-umbrel-app-url>/api/pay-bolt11", {
  method: "POST",
  headers: { "Content-Type": "application/json" },
  body: JSON.stringify({
    invoice: "lnbc100000n1p...attacker_invoice..."
  })
});
</script>
```

and they've just paid the attacker since:
- Umbrel proxy auth is bypassed by PROXY_AUTH_WHITELIST: "*"
- App backend does not check app session/auth on /api/pay-bolt11
- CORS_ORIGINS: "*" is set in compose
- Endpoint does not require CSRF token either
- Backend then pays via LND using the mounted admin macaroon

Potentially the best fix is to group the app's intentionally public routes under a small public surface, then whitelist only that.

For example: public receive/discovery pages and APIs could live under `/.well-known/*`, `/public/*`, `/receive/*`, or `/api/public/*`. Then payment/admin/config/alias routes like `/api/pay-*`, `/api/setup/*`, `/api/alias*`, `/api/auth/*`, `/app`, `/pay`, etc. stay private by default behind Umbrel auth.

That's kinda outside my domain here though since I'm really only reviewing these submissions from a packaging standpoint, not their internals. You will know much much better than me what needs to be public/private.

Alex71btc:
@nmfretz Really thanks again for the review and for pointing out the payment endpoint issue.

I went through the payment and payment-preview surface and added additional hardening:

### Backend hardening

Added authenticated pay-session checks to:

* `/api/pay-offer`
* `/api/pay-address`
* `/api/pay-bolt11`
* `/api/pay-lnurl`
* `/api/preview-pay-target`

These routes now require a valid authenticated application session and are no longer callable anonymously.

### CSRF protection

Added CSRF validation to payment execution endpoints:

* `/api/pay-offer`
* `/api/pay-address`
* `/api/pay-bolt11`
* `/api/pay-lnurl`

I verified that unauthenticated requests now return:

`401 Authentication required`

and payment requests without a valid CSRF token are rejected.

### Proxy surface reduction

I also reduced the Umbrel proxy whitelist from a broad configuration to only the routes required for public receive/discovery functionality:

* landing page (`/`)
* public alias pages (`/alias/*`)
* LNURL/BIP353 discovery (`/.well-known/*`, `/api/lnurl/*`)
* QR generation (`/api/qr/*`)
* static assets required by public pages

Everything else, including:

* `/app`
* `/pay`
* setup routes
* admin routes
* payment-control APIs

now requires Umbrel authentication by default.

### Validation

I re-tested:

* public alias pages
* QR generation
* LNURL/BIP353 receive flow
* authenticated payment flow
* unauthenticated access to payment APIs

The previously identified unauthenticated payment execution path is no longer possible.

Could you please give the updated package another look?

Alex71btc:
I just had to remove additional experimantal database additions for further tx history updates, now the docker image is clean with just the security hardening

nmfretz:
Woop! Great work @Alex71btc, thanks for putting up with me and digging in to that. Really appreciate it. Let's get this live!

You can always make updates to simplify the whitelist in the future if you get the itch.

Welcome to the app store 🎉 

image

Alex71btc:
Hey @nmfretz @al-lac this is totally awesome! Thank you so much, feeling totally honored! 
I came a long way getting in to bitcoin during covid with zero coding skills, then building my first node with umbrel on raspberry pi and now running multiple Servers with self hosted self developed apps, BOLT12 Pay app is now live on umbrel app store and Startos community  marketplace. 🙏❤️

Alex71btc:
@nmfretz Thanks again for working through this with me.
After testing the narrowed whitelist setup, I noticed one tradeoff worth documenting:
Public payment functionality (LNURL, BIP353, public payment pages) continues to work, but browser features that depend on a secure public context (such as PWA installation flows and QR camera scanning) are now limited when accessing the app through Umbrel-local routes.
The backend now enforces authenticated sessions, CSRF protection, rate limiting and route-level authorization for all management and payment-control functionality, so the original security concerns are addressed at the application layer as well.
In the future it might be interesting to explore whether Umbrel could support a more explicit separation between public payment routes and authenticated management routes for apps that intentionally expose payment services.

nmfretz:
> Public payment functionality (LNURL, BIP353, public payment pages) continues to work, but browser features that depend on a secure public context (such as PWA installation flows and QR camera scanning) are now limited when accessing the app through Umbrel-local routes.

Good news on this front @Alex71btc, we're shipping out of the box https in just a matter of weeks, so apis that require a secure context will end up working v soon!

Alex71btc:
That's more than totally awesome!!🙏👍😁

PR #5449, opened 2026-04-23
Add Go Brrr Pool

gobrrrme:
### App ID
  `gobrrr-pool`

  ### 256x256 SVG icon
  Already committed to the fork at `gobrrr-pool/icon.svg`.

  ### Description
  Go Brrr Pool is a self-hosted solo Bitcoin mining pool powered by ckpool-solo. Connect your miners directly to
  your own node and keep 100% of block rewards. No third-party pool, no KYC, no fees, no middleman.

  Features a real-time web dashboard with hashrate monitoring, worker statistics, pool-wide leaderboard, efficiency
   dashboard, and local mempool API integration. ZMQ enables instant block notifications and automatic difficulty
  adjustment accepts diff proposals from miners.

  Two stratum ports: port 21420 for standard mining, port 21422 for high-difficulty mode (min diff 1,000,000)
  suited for powerful machines and hashrate rentals. Workers connect using their Bitcoin address as the username —
  any block found pays out directly and entirely to the miner's own wallet.

  Built on ckpool, the same battle-tested open-source pool software powering solo.ckpool.org.

  ### Gallery images
  Already committed to the fork at `gobrrr-pool/1.jpg`, `2.jpg`, `3.jpg`, `4.png`.

  ### I have tested my app on:
  - [x] umbrelOS on Linux VM

  ### Additional notes
  - Dependencies: `bitcoin`, `mempool`
  - Stratum ports: 21420 (standard), 21422 (high-difficulty)
  - Includes a homescreen widget for pool stats via a lightweight ETL sidecar container (`gobrrr-stats`)
  - App repo: https://github.com/gobrrrme/GoBrrr-Pool


gobrrr-pool/umbrel-app.yml
```diff
@@ -0,0 +1,42 @@
...
+dependencies:
+  - bitcoin
+  - mempool
+repo: https://github.com/gobrrrme/GoBrrr-Pool
+support: https://github.com/gobrrrme/GoBrrr-Pool/issues
+port: 21421
+gallery: []
+path: ""
+defaultUsername: ""
+defaultPassword: ""
```
kaichappel, line 48: Is this sidecar intended to power an Umbrel homescreen widget? If yes, this is the manifest wiring to add. The endpoint points directly at the sidecar service, which is how umbreld fetches app widget data.

```suggestion
defaultPassword: ""
widgets:
  - id: "stats"
    type: "four-stats"
    refresh: "30s"
    endpoint: "gobrrr-stats:23002/widget-api/widget/stats"
    link: ""
    example:
      type: "four-stats"
      link: ""
      items:
        - title: "Hash Rate"
          text: "0"
          subtext: "H/s"
        - title: "Miners"
          text: "0"
          subtext: "Active"
        - title: "Best Share"
          text: "0"
          subtext: "Diff"
        - title: "Share Rate"
          text: "0.0"
          subtext: "/min"
```

gobrrr-pool/docker-compose.yml
```diff
@@ -0,0 +1,75 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: gobrrr-pool_webui_1
+      APP_PORT: 21421
+      PROXY_APP_EXTRAS: gobrrr-stats:23002
+      PROXY_AUTH_WHITELIST: /api/*,/widget-api/*
```
kaichappel: If the stats sidecar is for an Umbrel homescreen widget, this can be reduced to the main web UI auth whitelist. `PROXY_APP_EXTRAS` is not used by app_proxy for widget routing, and `/widget-api/*` should be fetched from the manifest endpoint instead.

```suggestion
      PROXY_AUTH_WHITELIST: /api/*
```

gobrrr-pool/docker-compose.yml
```diff
@@ -0,0 +1,75 @@
...
+      container_name: gobrrr-pool_webui_1
+
+  gobrrr-stats:
+    image: rodb2008com/gobrrr-pool-api-etl:v1@sha256:bc86f35c5cd526f0a2464574e764540fbe78173dcaaaa6bafdbdc9a818501948
+    restart: on-failure
+    depends_on:
+      ckpool:
+        condition: service_healthy
+    ports:
+      - '23002:23002'
```
kaichappel: With the widget endpoint declared in `umbrel-app.yml`, the stats sidecar does not need a host-published port. Umbrel can fetch it over the internal container network from `gobrrr-stats:23002/widget-api/widget/stats`, and leaving `23002` published exposes an internal widget API directly on the Umbrel host.

```suggestion
```

kaichappel (changes requested):
Thanks for the submission @gobrrrme.

I gave the Umbrel package a test. It installs and opens cleanly, and both stratum ports accepted miner subscribe/authorize requests.

One thing I wanted to check: it looks like the `gobrrr-stats` sidecar is meant to power an Umbrel homescreen widget. Is that the goal?

If yes, could you wire it like this?

- Add a `widgets:` entry in `umbrel-app.yml` with `endpoint: "gobrrr-stats:23002/widget-api/widget/stats"`, using the `four-stats` shape the sidecar already returns. I left an inline suggestion with the full block.
- Remove `PROXY_APP_EXTRAS`; app_proxy will not expose the sidecar through that env var.
- Change `PROXY_AUTH_WHITELIST` back to `/api/*`, since `/widget-api/*` should be fetched through the widget manifest endpoint rather than the main web UI.
- Remove the `23002:23002` host port from `gobrrr-stats`; Umbrel can fetch the widget over the internal container network from the manifest endpoint.

The other thing: could you delete the submitted icon and screenshots from this PR (`icon.svg`, `1.jpg`, `2.jpg`, `3.jpg`, and `4.png`)? We'll add the final app-store icon and gallery screenshots on our side before this goes live.

Once those changes are made, I can re-test.

gobrrr-pool/docker-compose.yml
gobrrrme: Thanks for the review and for the suggestions! Yes, the sidecar was meant as homescreen widget. I'll get those changes done and update once finished.

gobrrrme:
@kaichappel thanks for your feedback, changes were made.

  - Added widgets: manifest block in umbrel-app.yml (four-stats, endpoint gobrrr-stats:23002/widget-api/widget/stats)
  - Removed PROXY_APP_EXTRAS and reduced PROXY_AUTH_WHITELIST to /api/*
  - Removed 23002 host port from gobrrr-stats
  - Deleted icon.svg, 1.jpg, 2.jpg, 3.jpg, 4.png
  - Fixed formatting error in widget
  
  (I have not tested again after these changes, please let me know if something is off).
  
  Thanks for your help with this and have a nice weekend.

kaichappel (approved):
Great work @gobrrrme, this looks good from our side now. We will finalize the gallery assets on our side. Thanks again!

nmfretz:
Thanks again for the submission @gobrrrme! I made a few small packaging tweaks to get this finalized:

- Removed unnecessary `container_name` entries... these are managed by umbrelOS
- Switched the Mempool API URL to use Umbrel’s Mempool exports which is more robust.
- Added the missing `data/` directories required by the bind mounts.
- Changed the stats sidecar to mount the ckpool logs directory instead of binding a specific log file which can have some unintended consequences in Docker.
- Cleaned up a couple of YAML/manifest formatting details.

I connected a miner successfully and confirmed that it's working (as well as the widget), so we're good to go from an umbrel packaging persepective!

Gallery assets are done, so we're going live. Welcome to the app store 🎉
Screenshot 2026-06-09 at 12 06 43 PM

nmfretz:
Also @gobrrrme one consideration for you is whether you truly want the Mempool app to be a required dependency. Because it's quite a heavy app that not everyone will want to run in order to mine... so it may prevent some users from installing and using your app.

In a future app update you could always remove the mempool dependency. Looks like it's only used for dashboard/network stats right? so we can extremely easily just check to see if a user has mempool installed via an exports.sh file. If so you use their local mempool api. If not you default to a public mempool instance.

What do you think?

nmfretz:
Really love having all the stats 🔥.

So you're saying there's a chance:

image

gobrrrme:
> Also @gobrrrme one consideration for you is whether you truly want the Mempool app to be a required dependency. Because it's quite a heavy app that not everyone will want to run in order to mine... so it may prevent some users from installing and using your app.
> 
> In a future app update you could always remove the mempool dependency. Looks like it's only used for dashboard/network stats right? so we can extremely easily just check to see if a user has mempool installed via an exports.sh file. If so you use their local mempool api. If not you default to a public mempool instance.
> 
> What do you think?

Hi @nmfretz, thank you so much for merging my first app to the official app store!

On the mempool dependency, I would love to make that "optional", exactly as you described, but I wasn't sure how to achieve that. As in, defaulting to mempool dot space api. I know mempool can put a lot of load on the machines, so that would be a worthwile update.

While I've got you here. Another thing I would have loved to implement would be changing the pool's headline on the main page from "Go Brrr Pool" to "Umbreluser's Pool", using the username like the Umbrel homescreen does. I couldn't figure out how to do that. Maybe you can give me a pointer.

Thanks for your help!

nmfretz:
Ya no worries we can make mempool optional. If you open a PR for an app update I can push the required changes and you can then test.

> While I've got you here. Another thing I would have loved to implement would be changing the pool's headline on the main page from "Go Brrr Pool" to "Umbreluser's Pool", using the username like the Umbrel homescreen does. I couldn't figure out how to do that. Maybe you can give me a pointer.

Ah ya, for this one we don't have a way in umbrelOS right now to pass in the user's name. That would be pretty cool.

PR #5451, opened 2026-04-23
App Submission: Satsbook

brewgator:
# App Submission

### App name
Satsbook

### 256x256 SVG icon
icon

#### Home Page Dashboard
Screenshot 2026-04-28 at 8 35 07 AM

#### Lightning Node Stats
Screenshot 2026-04-28 at 8 35 22 AM
Screenshot 2026-04-28 at 8 36 03 AM

#### P&L Stats
Screenshot 2026-04-28 at 8 35 52 AM

#### Cold Storage Wallets
Screenshot 2026-04-28 at 8 35 41 AM

### I have tested my app on:
- [x] umbrelOS on Linux VM

### DIFF
--- a/satsbook/docker-compose.yml
+++ b/satsbook/docker-compose.yml
@@ -1,0 +1,24 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: satsbook_server_1
+      APP_PORT: 3015
+
+  server:
+    image: ghcr.io/satsbook/satsbook:1.1.3@sha256:751e33b956a645c45e6faebf71fd54e24c979e9984a3f625362e18347d7c92b3
+    restart: on-failure
+    stop_grace_period: 1m
+    init: true
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${APP_LIGHTNING_NODE_DATA_DIR:-/dev/null}:/lnd:ro
+    environment:
+      SATSBOOK_LND_HOST: ${APP_LIGHTNING_NODE_IP:-localhost}
+      SATSBOOK_LND_PORT: ${APP_LIGHTNING_NODE_GRPC_PORT:-10009}
+      SATSBOOK_LND_MACAROON_PATH: /lnd/data/chain/bitcoin/mainnet/readonly.macaroon
+      SATSBOOK_LND_TLS_CERT_PATH: /lnd/tls.cert
+      SATSBOOK_DATABASE_PATH: /data/satsbook.db
+      SATSBOOK_APP_PORT: 3015
+      SATSBOOK_LOG_LEVEL: info
--- a/satsbook/docker-compose.yml
+++ b/satsbook/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   app_proxy:
     environment:
       APP_HOST: satsbook_server_1
-      APP_PORT: 3015
+      APP_PORT: 3017
 
   server:
     image: ghcr.io/satsbook/satsbook:1.1.3@sha256:751e33b956a645c45e6faebf71fd54e24c979e9984a3f625362e18347d7c92b3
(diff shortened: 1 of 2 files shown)


satsbook/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+      APP_PORT: 3015
+
+  server:
+    image: ghcr.io/satsbook/satsbook:1.1.3@sha256:751e33b956a645c45e6faebf71fd54e24c979e9984a3f625362e18347d7c92b3
+    restart: on-failure
+    stop_grace_period: 1m
+    init: true
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${APP_LIGHTNING_NODE_DATA_DIR:-/dev/null}:/lnd:ro
```
kaichappel: Since the manifest already declares the `lightning` dependency, this should use the required Lightning data-dir env var directly. The read-only directory mount is fine; the risky part is the `/dev/null` fallback, because it can hide a missing dependency/configuration problem and let the app start without a real LND mount.

```suggestion
      - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
```

satsbook/docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+    environment:
+      SATSBOOK_LND_HOST: ${APP_LIGHTNING_NODE_IP:-localhost}
+      SATSBOOK_LND_PORT: ${APP_LIGHTNING_NODE_GRPC_PORT:-10009}
+      SATSBOOK_LND_MACAROON_PATH: /lnd/data/chain/bitcoin/mainnet/readonly.macaroon
```
kaichappel: Umbrel exposes `APP_BITCOIN_NETWORK` so Bitcoin/Lightning apps can match the user's node network instead of assuming mainnet.

Satsbook looks mainnet-only, so the hardcoded `mainnet` path may be right here. Can you confirm that's intended? If Satsbook should instead follow the user's Umbrel network, this path should use `${APP_BITCOIN_NETWORK}`.

The `localhost` / `10009` fallbacks above should be removed either way, since `lightning` is declared as a dependency.

satsbook/umbrel-app.yml
```diff
@@ -0,0 +1,26 @@
...
+  to your LND node and provides a privacy-first dashboard with fee
+  analytics, portfolio tracking, and CSV import from Strike, River,
+  Coinbase, and Swan. No cloud services required.
+developer: brewgator
+website: https://github.com/satsbook/satsbook
+repo: https://github.com/satsbook/satsbook
+support: https://github.com/satsbook/satsbook/issues
+dependencies:
+  - lightning
+port: 3015
```
kaichappel: `3015` is already used by another app, so Satsbook needs a unique app_proxy port before this can merge. `3017` is available.

```suggestion
port: 3017
```

kaichappel (changes requested):
Thanks for the submission @brewgator. Satsbook is a useful fit for Umbrel, and the package installs and opens cleanly in my testing, but there are a few things we need to address before this can go live.

The main Lightning packaging issue is the fallback values in `docker-compose.yml`. Since the manifest already declares the `lightning` dependency, the compose file should rely on Umbrel's Lightning-provided env vars rather than falling back to `/dev/null`, `localhost`, or `10009`. Those fallbacks can mask a missing dependency/configuration problem and let the app start with a broken LND connection. The read-only LND directory mount itself is fine.

One question on the `mainnet` macaroon path: Umbrel exposes `APP_BITCOIN_NETWORK` so Bitcoin/Lightning apps can match the user's node network. Satsbook looks mainnet-only, so the hardcoded `mainnet` path may be right here. Can you confirm that's intended? If Satsbook is meant to support the user's Umbrel network, this should use `${APP_BITCOIN_NETWORK}` instead.

I also left comments for two package cleanup items:

- `port: 3015` conflicts with an existing app-store port, so this needs a unique app_proxy port.
- `${APP_DATA_DIR}/data` is mounted, but the PR doesn't include the corresponding `satsbook/data` scaffold. Please add `satsbook/data/.gitkeep` so the mount exists cleanly.

satsbook/docker-compose.yml
brewgator: Fixed — removed the `/dev/null` fallback. Using `${APP_LIGHTNING_NODE_DATA_DIR}` directly since `lightning` is a declared dependency.

satsbook/docker-compose.yml
brewgator: Good call — updated the macaroon path to use `${APP_BITCOIN_NETWORK}` so it follows the user's Umbrel network. Also removed the `localhost`/`10009` fallbacks on host and port.

satsbook/umbrel-app.yml
brewgator: Done — changed to `3017` across both docker-compose.yml and umbrel-app.yml.

brewgator:
@kaichappel all concerns have been addressed, please review at your earliest convenience.

kaichappel:
Thanks @brewgator, appreciate you working through the feedback here.

I gave this another test and it's working nicely for me. The only remaining thing we need from your side is the empty data directory scaffold for the bind mount in the compose file:

```bash
mkdir -p satsbook/data
touch satsbook/data/.gitkeep
git add satsbook/data/.gitkeep
git commit -m "Add Satsbook data directory scaffold"
git push
```

The `.gitkeep` file is just there so Git includes the otherwise-empty `satsbook/data` directory. That way `${APP_DATA_DIR}/data:/data` has the expected host directory when the app is installed.

Once that's pushed, we can take it from there and finalize the gallery assets on our side before launch.

brewgator:
@kaichappel Done - added the `satsbook/data/.gitkeep` scaffold. Ready for gallery finalization on your side 👍

nmfretz:
Excellent work @brewgator! Gallery assets are now done. Welcome to the app store 🎉

image

PR #5458, opened 2026-04-26
Update miner-sentinel to version 1.0.1

dcbert:
## What

Updated the `miner-sentinel` Docker image to the latest version.

## Changes
- Updated image reference in `docker-compose.yml` (new tag / sha256)
- Bumped app version in `umbrel-app.yml`
- Added release notes

## Why
This includes the fix for the issue where the app would only work correctly with IP-based URLs (instead of domain-based ones). Users can now use either without problems.

Related issue: https://github.com/dcbert/miner-sentinel/issues/10

## Testing
- Tested on Umbrel OS (latest)
- App starts correctly with both IP and domain URLs
- No breaking changes

---

**Ready for review**

### DIFF
--- a/miner-sentinel/docker-compose.yml
+++ b/miner-sentinel/docker-compose.yml
@@ -24,7 +24,7 @@ services:
       retries: 5
 
   backend:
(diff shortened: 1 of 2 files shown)


al-lac:
Hey @dcbert, thanks a lot for keeping miner-sentinel up-to-date on the Umbrel App Store 🔝 

I changed the postgres env vars to the new values and tested it. Everything looks good! Going live 🥳

PR #5462, opened 2026-04-27
feat(nostr-relay): upgrade to nostr-rs-relay 0.9.0 with NIP-42, NIP-44, NIP-50, NIP-59, NIP-77 and ARM64 support

satwise:
## Upgrades nostr-relay from 0.8.1 to 0.9.0

Bumps `nostr-rs-relay` to `0.9.0-no-jemalloc` (ARM64-compatible build) and resolves two relay-proxy crash bugs discovered when upgrading a live install.

---

### Changes

| File | Change |
|------|--------|
| `umbrel-app.yml` | Bump version to `1.4.0`, add `manifestVersion: 1.1` |
| `docker-compose.yml` | New relay image digest; add `RelayManager.js` volume mount |
| `data/relay/config.toml` | Replace literal `0.0.0.0` with `{{RELAY_HOST}}` placeholder |
| `hooks/pre-start` | Guard `store.json`; inject CLNRest URL; document RelayManager.js fix |
| `data/relay-proxy/RelayManager.js` | Patched copy — see Bug #2 below |

---

### Bug fixes (upgrading from 0.8.1)

These issues will affect **any install upgrading from 0.8.1 → 0.9.0 with an npub or NIP-05 identifier set**.

#### Bug 1 — `store.json` crash loop on startup

**Symptom:** `relay-proxy` container enters a crash loop immediately on start, with `SyntaxError: Unexpected end of JSON input` in logs. `docker ps` shows 400+ restarts.

**Root cause:** An existing `store.json` from 0.8.1 may lack the `"relays"` key that v1.1.0 of the relay-proxy image requires.

**Fix:** `hooks/pre-start` now validates `store.json` before containers start. If the file is missing or lacks the `relays` key, it is initialised to `{"relays":[]}` (preserving `identifier` if present).

---

#### Bug 2 — `relay-proxy` crashes 60 minutes after startup (npub identifier path)

**Symptom:** `relay-proxy` runs cleanly for ~60 minutes, then crashes with:
```
SyntaxError: Unexpected token w in JSON at position 0
    at JSON.parse (<anonymous>)
    at RelayManager.removeOutdatedRelays
```
Container restarts, runs for another 60 minutes, crashes again — indefinitely.

**Root cause (in `getumbrel/umbrel-nostr-relay-proxy:v1.1.0`):**

`getPublicKeyAndRelaysFromIdentifier()` returns an **array** of relay URLs when the identifier is an npub (NIP-19). The result is stored as:
```js
this.latestRelays = { relays: ["wss://relay.damus.io", ...], timestamp: 0 };
```

60 minutes later the cleanup timer fires and calls `removeOutdatedRelays()`, which has:
```js
// Bug: relays is an array here, not a JSON string
const latestRelays = this.latestRelays
  ? Object.keys(JSON.parse(this.latestRelays.relays))
  : [];
```
`JSON.parse(["wss://relay.damus.io"])` → `SyntaxError`.

`relays` only becomes a JSON string _after_ a Kind 3 Contact List event is received (from `event.content`). Users on npub who haven't broadcast a Kind 3 event will crash on every interval tick.

**Fix:** `data/relay-proxy/RelayManager.js` ships a patched copy that guards `removeOutdatedRelays()` with type and Array checks:
```js
let latestRelays = [];
try {
  if (this.latestRelays && typeof this.latestRelays.relays === 'string' && this.latestRelays.relays) {
    latestRelays = Object.keys(JSON.parse(this.latestRelays.relays));
  } else if (Array.isArray(this.latestRelays?.relays)) {
    latestRelays = this.latestRelays.relays;
  }
} catch (e) { /* ignore parse errors */ }
```

This file is volume-mounted over `/app/RelayManager.js` in `docker-compose.yml`:
```yaml
volumes:
  - ${APP_DATA_DIR}/data/relay-proxy:/app/data
  - ${APP_DATA_DIR}/data/relay-proxy/RelayManager.js:/app/RelayManager.js:ro
```

The fix applies automatically on every app start (including after umbrelOS updates) since the volume mount always uses the file shipped in the app directory.

> **Note for upstream:** Both bugs exist in `getumbrel/umbrel-nostr-relay-proxy:v1.1.0`. Bug 1 is a data-migration issue; Bug 2 is a logic error in `RelayManager.removeOutdatedRelays()` that affects all npub users regardless of migration state.

---

### Testing

Validated on Raspberry Pi 5 (umbrelOS, arm64) upgrading from 0.8.1:
- relay-proxy starts cleanly (0 restarts) ✅
- Connects to `relay.damus.io`, `nos.lol` within seconds ✅
- Runs past the 60-minute mark without crash ✅
- `nostr-rs-relay` WAL checkpoints clean, events persisting ✅


kaichappel:
Thanks for the PR @satwise.

For the Umbrel Nostr Relay app, development happens in the dedicated app repo here: https://github.com/getumbrel/umbrel-nostr-relay. Changes to the app itself should be opened there first, and then we can pull the relevant app-store packaging changes into `getumbrel/umbrel-apps` as needed.

This PR also includes quite a few files outside the usual Umbrel app packaging surface, including app implementation code, docs, ops notes, and other repo-level files. That makes it difficult for us to review and merge here.

I’m going to close this PR for now, but if you’d like to continue with the 0.9.0 update, the best next step is to open a focused PR against `getumbrel/umbrel-nostr-relay`.

Thanks again.

PR #5261, opened 2026-04-02
gap(core-lightning): provide *native* V3 API 'CLNREST_HOST' data contract for modern CLN umbrel-app consumers *and* for full DR recovery into Umbrel harness

satwise:
## Summary

Two additive gaps in the `core-lightning` packaging since #3931 retired `core-lightning-rest` (V2) and moved CLNRest into `lightningd` itself (V3, built-in since v23.08):

1. **DR-safe bind gap** — CLNRest binds to the static container IP (`${APP_CORE_LIGHTNING_DAEMON_IP}`) instead of `0.0.0.0`. In steady state these are equivalent — port 2107 is published via `ports:` and reachable from other containers and the LAN. The difference surfaces on failure paths: a DR restore or fresh install can start `lightningd` before the static IP is attached to its interface, causing a hard `EADDRNOTAVAIL` bind failure. `0.0.0.0` removes that race at zero exposure cost (container netns remains the security boundary).

2. **Provider contract gap** — `exports.sh` exports zero CLNRest-native keys (`CLNREST_HOST`, `CLNREST_URL`, `CLNREST_RUNE_PATH`). CLNRest-native consumers (LNbits, Boltz, Zap Bridge) expect these keys natively. Without them, every consumer must carry an Umbrel-specific V2→V3 translation internally, and umbrelOS cannot reconstruct the V3 environment during DR recovery — consumers fall back to empty strings or hardcoded IPs that drift on reinstall.

The same reason `bitcoin` exports `BITCOIN_RPC_*` and `electrs` exports `ELECTRUM_*`: the Provider publishes the native contract its Consumers expect. CLN is the outlier still shaped around the deprecated V2 naming, six months after #3931 retired the binary.

---

### `core-lightning/docker-compose.yml`
- `--clnrest-host=${CLNREST_HOST}` — resolves to `0.0.0.0`; DR-safe, no exposure delta
- `container_name` pins on `lightningd`, `app`, and `tor` only:
  - `lightningd` — V3 consumer-facing API surface; downstream consumers race Docker DNS on parallel recreation
  - `app` — Commando rune lifecycle owner; already hardcoded as `APP_HOST` in this same compose file
  - `tor` — hard startup dependency of `lightningd`; race documented in #5529
  - `app_proxy` left to umbrelOS auto-naming

### `core-lightning/exports.sh`
Purely additive — V2 Umbrel-native names preserved verbatim. New exports:
```bash
# DNS-stable container hostnames (survive IP drift across restarts / DR recovery)
[...]

Issue #4785, opened 2026-02-17
gap(core-lightning): missing native V3 CLNRest provider contract blocks DR recovery and modern consumer apps


satwise:
## Summary

Two related gaps in the `core-lightning` migration from `core-lightning-rest` (V2 API) to the native CLNRest (V3) API make modern CLN consumer apps like LNbits fragile under DR recovery and break Umbrel's provider contract for CLNRest-native consumers:

1. **Bind gap** — CLNRest binds to a single static container IP instead of `0.0.0.0`.
   In steady-state this is functionally equivalent (port 2107 is published via `ports:`
   and reachable from other containers and the LAN). The difference surfaces on failure
   paths: a DR restore or fresh install can start `lightningd` before the static IP is
   attached to its interface, causing a hard `EADDRNOTAVAIL` bind failure. Binding to
   `0.0.0.0` removes that race entirely.

2. **Provider contract gap** — `exports.sh` exports zero CLNRest-native keys
   (`CLNREST_HOST`, `CLNREST_URL`, `CLNREST_RUNE_PATH`). CLNRest-native apps (LNbits,
   Boltz) cannot autodiscover the endpoint, and umbrelOS cannot reconstruct the V3
   environment during DR recovery — consumers fall back to empty strings or hardcoded
   V2 patterns that are wrong by construction on a fresh reinstall.

CLNRest (V3) has been built into `lightningd` since v23.08, but the Umbrel packaging
still reflects V2 (`core-lightning-rest`) naming conventions throughout.

---

## Gap 1 — `docker-compose.yml`: CLNRest bind host

**Before (master):**
```yaml
- --clnrest-host=${APP_CORE_LIGHTNING_DAEMON_IP}   # binds to 10.21.21.96 only
```

**After (PR #5261):**
```yaml
- --clnrest-host=${CLNREST_HOST}   # exported as 0.0.0.0 → all interfaces
```

With `container_name` pins on `lightningd`, `app`, and `tor`:
```yaml
lightningd:
  container_name: core-lightning_lightningd_1
app:
  container_name: core-lightning_app_1
tor:
  container_name: core-lightning_tor_1
```

### Impact — DR-safe bind address

`lightningd` honors `--clnrest-host` literally. Binding to the daemon's static
container IP is safe in steady state but fails on two recovery paths:

- **DR / fresh-install ordering** — `lightningd` can start before the static IP is
[...]

PR #3780, opened 2025-10-13
update to bitcoin knots v29.2

Retropex:

PR #5465, opened 2026-04-27
feat(umbrel-lnbits-cln): add LNbits for Core Lightning

### LINKED ISSUE #4753: Umbrel Store LNbits does not offer CLN as 'required app'  preventing install
Please see the attached screenshots demonstrating Umbrel Core Lightning Node (CLN) missing as additional Required App for LNbits in the Umbrel store: https://apps.umbrel.com/app/lnbits
 
Request for https://github.com/getumbrel/umbrel-apps/tree/master/lnbits Umbrel Store App:
As CLN is fully supported by LNbits as confirmed in open request https://github.com/lnbits/lnbits/issues/3780, please include https://apps.umbrel.com/app/core-lightning Umbrel Store App as a Required App required to install LNDhub /app-store/lnbits for the Umbrel CLN community.

Screenshot and related LNbits issue confirming LNbits is fully compatibility with CLN: https://github.com/lnbits/lnbits/issues/3780

Image

satwise:
## Summary
Adds a dedicated LNbits app for Core Lightning (`umbrel-lnbits-cln`) using CLNRest and rune-based permissions.

## What changed
- New app manifest, compose stack, exports, and pre-start rune provisioning hook
- Runtime wallet backend selection persistence for CLN-focused deployment
- Gallery assets + app data scaffolding
- `container_name: lnbits-cln_web_1` explicit — prevents Compose v2 dash-naming from breaking reverse proxy upstreams

## Dependency
Depends on Draft PR #5261 for provider-contract and baseline transition of new Core Lightning API, CLNRest (v3) from core-lightning-rest (v2).

## Related
- Closes #4753
- Relates to #4785
- Relates to [#3780](https://github.com/lnbits/lnbits/issues/3780)


nmfretz:
Thanks for working on this @satwise! I see you have this marked as draft still, but I've taken a look now to help with some direction.

I totally agree that having LNbits available for Core Lightning users makes sense. Ideally, the current LNbits app in the app store should not be conceptually tied to LND, But the practical constraint is Umbrel’s current app/dependency model: the app compose file is static, and the dependency is what gives us concrete auto-configuration for a local node, including exported IPs/ports, network names, and read-only mounts for the relevant credentials/socket.

For LND, the app needs the LND cert/macaroon paths. For Core Lightning, it needs CLN connection details and either the `lightning-rpc` socket or CLNRest credentials. Umbrel currently does not have a clean way for one app package to say “if the user selected LND, mount/configure this; if they selected Core Lightning, mount/configure that; if neither, skip the mount and let the user configure an external funding source.”

A big part of the Umbrel app-store experience is one-click install with sensible auto-configuration. So for now, I think two separate apps like you are doing here is the cleanest practical solution:

- `lnbits`: current LND-backed LNbits app
- `lnbits-cln`: LNbits configured for Core Lightning (your PR here)

For this PR specifically, I don’t think the current implementation is the right shape to merge as-is. `CLNRestWallet` is a valid LNbits backend, especially if the goal is least-privilege access via restricted runes. For Umbrel’s LNbits CLN app, though, I think we could start from the simpler `CoreLightningWallet` path unless there is a specific reason that will not work here or something else I'm missing.

My main concerns are:
- The CLNRest/provider-contract direction broadens the scope beyond simply adding LNbits for Core Lightning.
- The pre-start hook and generated rune env file add extra lifecycle complexity.
- Sourcing the generated rune file from app data is risky, since that directory is also mounted into the LNbits container and could turn container-writable state into host-side shell execution.
- The long startup command you have in the compose file directly reads and mutates LNbits settings in SQLite, which is brittle and should not be needed for an app package.
- This PR appears tied to broader Core Lightning provider-contract changes, but LNbits can work with Core Lightning without those changes.

A much simpler approach is to follow the existing LNbits app and swap the funding source from LND REST to LNbits’ documented `CoreLightningWallet` backend. The tradeoff is that this gives LNbits broad local RPC access to the CLN node, so it is not a least-privilege rune-based design. But that is similar in spirit to the current LND-backed LNbits app using the LND admin macaroon, and it keeps the Umbrel app simple and auto-configured.

For example, the compose can be roughly this simple and won't need any lifecycle hooks or exports:

```yaml
version: "3.7"

services:
  app_proxy:
    environment:
      APP_HOST: lnbits-cln_web_1
      APP_PORT: 3007
      PROXY_AUTH_ADD: "false"

  web:
    image: lnbits/lnbits:v1.5.4@sha256:ce5ca587ada6f4d41d0da8f4fe554c421296dbcc649edda634bb57559046f370
    init: true
    restart: on-failure
    stop_grace_period: 1m
    volumes:
      - ${APP_DATA_DIR}/data:/data
      - ${APP_CORE_LIGHTNING_DATA_DIR}:/cln:ro
    environment:
      # Global
      LNBITS_HOST: "0.0.0.0"
      LNBITS_PORT: 3007
      LNBITS_DATA_FOLDER: "/data"
      AUTH_HTTPS_ONLY: "false"

      # Core Lightning
      LNBITS_BACKEND_WALLET_CLASS: CoreLightningWallet
      CORELIGHTNING_RPC: "/cln/${APP_CORE_LIGHTNING_BITCOIN_NETWORK}/lightning-rpc"

      # App
      LNBITS_SITE_TITLE: "LNbits - Core Lightning"
      LNBITS_DEFAULT_WALLET_NAME: "Core Lightning Wallet"
      LNBITS_DISABLED_EXTENSIONS: "amilk"
      LNBITS_ADMIN_EXTENSIONS: "ngrok"
      LNBITS_ADMIN_UI: "true"
```

This keeps the app close to the existing LNbits package, gives users the expected one-click Umbrel setup, avoids custom hooks and database mutation, and follows LNbits’ documented direct Core Lightning configuration.

My recommendation would be to rework this PR around the simpler `CoreLightningWallet` approach unless there is a specific reason LNbits on Umbrel should use CLNRest/runes here. We could bring in the lnbits and core lightning maintainers for some opinions/direction if you think that would be a good idea.

satwise:
> My recommendation would be to rework this PR around the simpler `CoreLightningWallet` approach unless there is a specific reason LNbits on Umbrel should use CLNRest/runes here. We could bring in the lnbits and core lightning maintainers for some opinions/direction if you think that would be a good idea.

Thank you very much for your detailed and thoughtful review @nmfretz and I welcome any and all maintainers you feel would add value.  For your kind consideration, I did originally stand-up a solid CoreLightningWallet interface on LNbits 1.3.1, however after updating to 1.4.x, the payment path shifted from mostly synchronous-like checks to explicitly asynchronous event processing so settlement observation and notification are now async by design. If a deployment expects immediate synchronous state transitions, JSON-RPC and stream timing can appear as “choking” under load even when eventual settlement is correct, so I had to upgrade to CLNRest due to:

** Concrete operational limitations when running LNbits on CoreLightning JSON-RPC in a Nostr-relay stack supporting MSat Micropayments **

1. Request success is not settlement finality.
A successful JSON-RPC response confirms command handling, not irreversible payment completion on the network.

2. Settlement detection is pull-driven, not push-guaranteed.
LNbits typically depends on short timeout polling loops and repeated status checks, which creates observation lag under load.

3. Race window between settle and observe.
If a worker is delayed, restarted, or disconnected, an invoice can settle before LNbits records it, causing temporary pending drift.

4. At-least-once processing risk.
Reconnect and retry behavior can replay the same settlement signal; without strict idempotency, duplicates propagate to webhook and relay layers.

5. No atomic boundary across CLN, LNbits, and Nostr publish.
JSON-RPC state update, LNbits ledger mutation, and Nostr event emission are separate operations, so partial completion is possible.

6. High-frequency MSat traffic amplifies RPC backpressure.
Large volumes of tiny payments increase queue depth and status-check contention, widening pending windows and timeout frequency.

7. Outgoing payment status can be briefly inconsistent.
Submission, route attempt, and final status can appear in different cycles, so immediate status reads can misclassify state.

8. Restart recovery depends on durable cursors and replay logic.
If last-seen indexes and pending sets are not persisted and replayed, restarts can miss or double-handle settlements.

9. MSat accounting precision does not remove routing constraints.
Internal MSat ledgers are precise, but external routing minima, fee floors, and rounding still affect final settlement behavior for micropayments.

10. Nostr zap correctness depends on authoritative post-settlement reconciliation.
Publishing receipts before confirmed settlement creates false positives; publishing too late without dedupe can create duplicates.

May I propose:
1. Keep the minimal JSON-RPC CoreLightningWallet path as the default profile.
2. Keep CLNRest as an explicit opt-in profile for operators who need separation between services and rune-managed auth boundaries.
    a. current LNbits for LND is already REST-wallet shaped in docker-compose.yml.
    b. I have proven my Core Lightning can export CLN_REST app key contracts in exports.sh and docker-compose.yml reliably for LNbits, Boltz, or emerging apps that expect native keys, as we discovered implementing LNBits 'CLNRest Plug In'
3. Perhaps split delivery into two steps: minimal path first, optional CLNRest profile plus docs second.
4. DR behavior and recovery are documented from real tests in DR-RESTART-TEST.md and CLN-V26.04.1-AB-LNBITS-DR.md.
5. This keeps defaults simple while preserving a realistic operational path for deployments that need stronger DR and boundary controls

I also have confirmed that both interfaces switch seamlessly, as only CoreLightning interface exposes Node Information Tab in LNbits, so I have used swicthed the source often in my instance

satwise:
Recent updates addressed this issue

satwise:
Accidently closed. I have made the requested changes @nmfretz . Is there anything else you need from me to add LNbits for Core Lightning to Umbrel App Store?

satwise:
@nmfretz, may I propose:

Keep the minimal JSON-RPC CoreLightningWallet path as the default profile.
Keep CLNRest as an explicit opt-in profile for operators who need separation between services and rune-managed auth boundaries.

a. current LNbits for LND is already REST-wallet shaped in docker-compose.yml.
b. I have proven my Core Lightning can export CLN_REST app key contracts in exports.sh and docker-compose.yml reliably for LNbits, Boltz, or emerging apps that expect native keys, as we discovered implementing LNBits 'CLNRest Plug In'
Perhaps split delivery into two steps: minimal path first, optional CLNRest profile plus docs second.

DR behavior and recovery are documented from real tests in DR-RESTART-TEST.md and CLN-V26.04.1-AB-LNBITS-DR.md.
This keeps defaults simple while preserving a realistic operational path for deployments that need stronger DR and boundary controls

I also have confirmed that both interfaces switch seamlessly, as only CoreLightning interface exposes Node Information Tab in LNbits, so I have switched sources often in my instance

PR #5467, opened 2026-04-28
Update OpenClaw to 2026.4.27

nmfretz:
waiting for next openclaw release to see if the performance (speed) regressions have been addressed.

### DIFF
--- a/openclaw/docker-compose.yml
+++ b/openclaw/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       APP_PORT: 18789
 
   gateway:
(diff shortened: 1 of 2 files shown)


al-lac (approved):

nmfretz:
Bumped to 2026.4.27. Still has quite a large startup delay of 2-3 minutes. 

Did the following to mitigate and give user's feedback:
- We now skip shell completion setup during onboarding. Upstream runs it as a blocking step with up to a 30s timeout, but it often times out ... and isn’t useful for this one-time quickstart onboarding
- Added an “Almost there...” message after the security notice because OpenClaw still sits silently for ~2-3 minutes while finishing upstream first-run setup work.
- We now don't run `doctor` immediately after first onboarding. Normal restarts still run doctor.

PR #5520, opened 2026-04-29
App Submission: bitBoard

egzola:
# App Submission

### App name
bitBoard

### 256x256 SVG icon
logo256

### Gallery images
screen1

screen2

screen3

### I have tested my app on:
- [x] umbrelOS on Linux VM


egzola:
Hi @al-lac ,

As mentioned earlier, after finishing bitBalance I’m now submitting a new app: bitBoard.

Let me know if anything needs to be adjusted. Thanks!

bitboard/umbrel-app.yml
```diff
@@ -0,0 +1,26 @@
+manifestVersion: 1.1
```
kaichappel: Since this package doesn't use hooks, we should keep this on `manifestVersion: 1` so it doesn't unnecessarily require a newer umbrelOS manifest version.

```suggestion
manifestVersion: 1
```

bitboard/data/.gitkeep
```diff
@@ -0,0 +1 @@
+
```
kaichappel, line 1: Could you rename this placeholder to `data/.gitkeep`? Umbrel strips `.gitkeep` during install, but `.gitkeep.txt` is copied into `/data` as a real user-data file. The app still worked in testing, but the package should avoid leaving placeholder files behind.

kaichappel (changes requested):
Thanks for the submission @egzola. I gave the Umbrel package a test and the app itself is working nicely. It installs and opens through Umbrel, and I was able to add SOL to the dashboard with the change persisting as expected.

There are a couple of package cleanups before this can go live. I left inline comments for the manifest version and the data placeholder.

Thanks for including the icon/gallery assets too. We'll help get the final gallery assets ready on our side before launch.

egzola:
Thanks @kaichappel  — I’ve updated the package cleanup items:

- changed `manifestVersion` to `1`
- renamed the data placeholder to `data/.gitkeep`

Thanks again for the review!

nmfretz:
Alright @egzola the gallery assets are ready. Welcome to the app store 🎉

image

egzola:
Thanks @nmfretz !

I appreciate the review and all the help throughout the submission process.

Happy to be part of the Umbrel App Store.

PR #5521, opened 2026-04-29
App Submission: AuraLog

nillawafa:
## App submission: AuraLog

AuraLog is a privacy-first journaling app with AI-powered insights, mood tracking, and analytics. Entries are stored locally on the Umbrel device. AI features are optional and require an Anthropic API key.

### Checklist
- [x] App runs on umbrelOS on a Raspberry Pi 4
- [x] No data leaves the device unless AI features are explicitly triggered
- [x] Images published to GHCR with sha256 digests pinned in docker-compose.yml
- [x] App tested: install, create entries, verify persistence across restarts

### Links
- Repo: https://github.com/nillawafa/Auralog
- GHCR (web): ghcr.io/nillawafa/auralog:v1.0.0
- GHCR (proxy): ghcr.io/nillawafa/auralog-proxy:v1.0.0


Oni-giri:
Hi, the repo doesn't seem to be public?

nillawafa:
Public now, thanks
    On Sunday, May 24, 2026 at 06:24:56 AM EDT, Yakitori ***@***.***> wrote:  
 
 Oni-giri left a comment (getumbrel/umbrel-apps#5521)
Hi, the repo doesn't seem to be public?

—
Reply to this email directly, view it on GitHub, or unsubscribe.
You are receiving this because you authored the thread.Message ID: ***@***.***>
   

|  | Virus-free.www.avast.com |

auralog/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+
+services:
+
+  app_proxy:
+    environment:
+      APP_HOST: auralog_web_1
+      APP_PORT: 80
+
+  web:
+    image: ghcr.io/nillawafa/auralog:v1.0.0@sha256:acfb4da8af0237a77e91c525829152a2942b6fa4c90bfbb6304c7590525fc055
```
kaichappel: Blocker: this pinned web image is crash-looping under nginx. In testing, the `web` container repeatedly exited with:

```text
unknown directive "_header" in /etc/nginx/conf.d/default.conf:1
```

That leaves app_proxy resetting the connection before the UI can load. Could you publish a fixed `auralog` image and update this digest? The upstream `docker/nginx.conf` appears to start with `_header ...`, so the image needs a valid nginx config before we can re-test the app.

auralog/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+      - proxy
+    networks:
+      - default
+
+  proxy:
+    image: ghcr.io/nillawafa/auralog-proxy:v1.0.0@sha256:240a38c7250cfdafeb6d008e457389ccd90a620ad4672cbb8a34fe47dfdf945a
+    restart: on-failure
+    stop_grace_period: 1m
+    environment:
+      - ANTHROPIC_API_KEY=${APP_PASSWORD}
```
kaichappel: It looks like this may be trying to enable AuraLog's AI features by passing Umbrel's generated `${APP_PASSWORD}` into `ANTHROPIC_API_KEY`. Is that the intended setup, or should users enter their Anthropic key somewhere inside AuraLog after install?

If the intended flow is in-app API-key setup, I would suggest leaving `ANTHROPIC_API_KEY` unset here and persisting the user-provided key under `${APP_DATA_DIR}`. If you were aiming for install-time setup instead, we will need a different configuration path, since Umbrel does not currently prompt users for arbitrary provider API keys during install.

auralog/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+      - PORT=3851
+      - NODE_ENV=production
+      - DATA_DIR=/data
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
```
kaichappel: Could you add a checked-in scaffold for this bind mount, for example `auralog/data/.gitkeep`? The linter is flagging the missing host path, and letting Docker create it implicitly can lead to permissions surprises later.

auralog/umbrel-app.yml
```diff
@@ -0,0 +1,61 @@
...
+
+  🔒 **Complete Privacy** — Journal entries are stored on your Umbrel device
+  at a persistent data path. Optional encryption. Zero telemetry.
+
+
+  **Journey categories:** Personal, Work, Reading, Creative, Learning, Health
+
+
+  **AI features require an Anthropic API key.** All journaling and analytics
+  work without one. Enter your key in the app password field at install time.
```
kaichappel: Can you clarify the intended API-key setup flow here? This says users should enter their key in an app password field at install time, but I could not verify the in-app flow while the web image is not launching.

If users are meant to add their Anthropic key inside AuraLog after install, I would suggest updating this copy to point them to that in-app setting. If the key is meant to be collected by Umbrel during install, I do not think the current package supports that yet.

auralog/umbrel-app.yml
```diff
@@ -0,0 +1,61 @@
...
+
+developer: Robert Kobatzki
+website: https://github.com/nillawafa/Auralog
+repo: https://github.com/nillawafa/Auralog
+support: https://github.com/nillawafa/Auralog/issues
+gallery: []
```
kaichappel, line 38: Small cleanup before launch: please remove the committed `icon.png` and screenshot PNGs from the app package. We will add/finalize the app-store icon and gallery assets on our side; `gallery: []` can stay empty here for now.

kaichappel (changes requested):
Thanks for the submission @nillawafa. I gave the Umbrel package a test, but it does not launch cleanly for me yet.

The web container is crash-looping on nginx startup, so the app_proxy route resets the connection before the UI loads. I left an inline comment with the exact error and the image line that needs to be updated.

There are a few other things we need to work through before this can go live:

- I left a question about the Anthropic API-key flow. It looks like the package may be trying to use Umbrel's generated `${APP_PASSWORD}` as `ANTHROPIC_API_KEY`, but I could not verify the intended in-app setup while the UI is not launching. If the goal is for users to enter their Anthropic key inside AuraLog after install, I suggested leaving the env var unset in compose and persisting that user-provided key under `${APP_DATA_DIR}` instead;
- the app data bind mount needs a checked-in scaffold directory;
- the app package should not include the root-level icon/screenshot PNGs;
- the upstream repo currently has no license detected by GitHub, so please add a license there before we publish it.

kaichappel:
Thanks @nillawafa, this is much closer now. The app installs and opens cleanly for me, and most of the earlier packaging issues look addressed.

I found one remaining issue: journal entries are not reliably being saved to AuraLog's server-side storage when the app is opened through common Umbrel URLs.

The short version is that AuraLog appears to only accept API requests when the browser Origin is exactly `http://umbrel.local:3850`. But Umbrel users may open apps from other valid addresses, for example:

- `http://umbrel.local:3850`
- `http://my-umbrel.local:3850`
- `http://192.168.1.50:3850`
- a Tailscale/MagicDNS hostname
- a custom local domain

In my test, an Origin like `http://my-umbrel.local:3850` caused the save request to `/api/entries` to return `403`. The UI still showed the new journal entry because it fell back to browser `localStorage`, but the entry was not saved to AuraLog's app data on the Umbrel. So it can look like the journal entry was saved, while a fresh browser/device or server-side API check shows no entry.

Is this something you can look into fixing before we go live?

nillawafa:
Thanks for the feedback. I'll test asap but my umbrel is down currently because I'm still running on a Pi 4. Might just cave and get the Home

nmfretz:
Hey @nillawafa just checking in to see how it's going. Anything we can help with on your Pi4 front? You can also use the testing agent skill in this repo to get up and running with a dev environment:
https://github.com/getumbrel/umbrel-apps/blob/master/.claude/skills/umbrel-test-app/SKILL.md

nillawafa:
Pushed. Thanks for the tip!

nmfretz:
no worries @nillawafa. Looks like it didn't actually get pushed to this branch though. Can you try again?

nmfretz:
Hey @nillawafa just pinging you again since it looks like the changes weren't pushed

nmfretz:
Sorry for the delay here, @nillawafa—we only just noticed that the follow-up commits landed after our last ping. We’ve had our heads down on a few other things, so thanks for your patience and for sticking with this.

I re-tested the latest changes, and there’s just one last issue to address before we can get AuraLog live. Entries still aren’t saved when the app is opened through MagicDNS or an HTTPS/custom domain because `isLocalOrigin()` in `server/index.js` rejects those valid origins.

Can you update the origin handling so the Umbrel App Store deployment accepts requests when the Origin hostname matches the hostname AuraLog was opened on, while still returning `403` for malformed or different origins? If the existing stricter behavior is needed for standalone installs, this can be configurable and enabled in `auralog/docker-compose.yml`.

Once that’s fixed, please publish a new multi-arch `auralog-proxy` image and update its tag and digest here. I’ll re-test it and get this live. Thanks again!

nmfretz:
Hey @nillawafa just checking in to see if you've seen my previous message and are still interested in getting AuraLog on the app store!

PR #5532, opened 2026-05-03
Update downtify to version 2.6.0

henriquesebastiao:
This PR updates Downtify to version 2.6.0 and also adds to the `docker-compose.yml` file the configuration that maps the container's `/data` directory to the Umbrel host machine, since in the latest versions of Downtify the functionality of permanence of configurations was implemented, which is lost with each container update if the `/data` directory is not mapped to the machine.


al-lac:
Hey @henriquesebastiao, thanks for another update to downtify! Awesome job 🔝 

Added the data folder, and now we can go live.

Issue #4452, opened 2026-01-13
[App Request] - Readeck


sashazykov:
https://readeck.org/en/download

nmfretz:
Readeck is now live in the app store! https://github.com/getumbrel/umbrel-apps/commit/cfe4382653408a13035752f2874713e60c1d7109

PR #5580, opened 2026-05-05
App Submission: Readeck

vitto4:
# App Submission

### App name
Readeck

### 256x256 SVG icon
https://codeberg.org/readeck/readeck/src/branch/main/web/media/logo.svg

### Gallery images
https://codeberg.org/readeck/readeck/src/branch/main/screenshots/bookmark-list.webp
https://codeberg.org/readeck/readeck.org/src/branch/main/src/en/blog/202602-readeck-22/bookmark-edit.webp
https://codeberg.org/readeck/readeck.org/src/branch/main/src/en/blog/202402-readeck-12/fonts.webp
https://codeberg.org/readeck/readeck.org/src/branch/main/src/media/img/highlights.webp
https://codeberg.org/readeck/readeck.org/src/branch/main/src/media/img/label-list.webp

### Notes
- I believe port `8000` clashes with _ZeroNote_, so it might be better to change it ?
- For the category I wasn't too sure, is `file` ok ? (same as _Karakeep_)
- The tagline from the official repo (`Readeck is a simple web application that lets you save the precious readable content of web pages you like and want to keep forever.`) felt a bit long by Umbrel app store standards, so I tried to shorten it.
- Readeck's [configuration page](https://readeck.org/en/docs/configuration) recommends :
  > If you're running Readeck and a reverse proxy on the same host, you can set `trusted_proxies` to ["127.0.0.1"].

  But if I'm not mistaken, the Umbrel app proxy operates on `10.0.0.0/8`, which is included in readeck's default, so I left it unchanged.
- First time packaging for umbrelOS, hopefully I got it right hehe

### I have tested my app on:
- [ ] umbrelOS on Linux VM

### DIFF
--- a/readeck/umbrel-app.yml
+++ b/readeck/umbrel-app.yml
@@ -1,0 +1,48 @@
+manifestVersion: 1
+id: readeck
+category: files
+name: Readeck
+version: "0.22.3"
+tagline: Save web pages to read and keep forever
+description: >-
+  Readeck is a simple web application that lets you save the precious readable content of web pages you like and want to keep forever. See it as a bookmark manager and a read later tool.
+
+
+  **Features**
+
+  - **🔖 Bookmarks** - Like a page you're reading? Paste the link in Readeck and you're done!
+  
+  - **📸 Articles, pictures and videos** - Readeck saves the readable content of web pages for you to read later. It also detects when a page is an image or a video and adapts its process accordingly.
+  
+  - **⭐ Labels, favorites, archives** - Move bookmarks to archives or favorites and add as many labels as you want.
+  
+  - **🖍️ Highlights** - Highlight the important content of your bookmarks to easily find it later.
+  
+  - **🗃️ Collections** - If you need a dedicated section with all your bookmarks from the past 2 weeks labeled with "cat", Readeck lets you save this search query into a collection so you can access it later.
+  
+  - **🧩 Browser Extension** - Want to keep something for later while browsing? No need to copy and paste a link. Install the browser extension and save bookmarks in one click!
+  
+  - **📖 E-Book export** - What's better than reading your collected articles on your e-reader? You can export any article to an e-book file (EPUB). You can even export a collection to a single book!
+
+    On top of that, you can directly access Readeck's catalog and collections from your e-reader if it supports OPDS.
+  
+  - **🔎 Full text search** - Whether you need to find a vague piece of text from an article, or all the articles with a specific label or from a specific website, we've got you covered!
+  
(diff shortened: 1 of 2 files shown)


vitto4:
Ah also I forgot, this PR should close https://github.com/getumbrel/umbrel-apps/issues/4452.

And about the root user lint warning, I'm unsure what's best here, but possibly related issue [Readeck #428](https://codeberg.org/readeck/readeck/issues/428) :
> Hi, do you specify a user (or user id) when you start the container? If so, you don't need to, you're already unprivileged.

vitto4:
Update : after some testing, it seems Readeck runs just fine as a regular user within the container 👍

readeck/umbrel-app.yml
```diff
@@ -0,0 +1,48 @@
...
+  - **🔒 Built for your privacy and long term archival** - Will this article you like be online next year? In 10 year? Maybe not; maybe it's all gone, text and images. For this reason, and for your privacy, text and images are all stored in your Readeck instance the moment you save a link.
+
+    With the exception of videos, not a single request is made from your browser to an external website.
+releaseNotes: ""
+developer: Readeck
+website: https://readeck.org
+dependencies: []
+repo: https://codeberg.org/readeck/readeck
+support: https://readeck.org/en/docs/faq
+port: 8000
```
kaichappel: Could we move only the Umbrel/app_proxy port off `8000`? `8000` is already used by ZeroNote, while Readeck can still listen internally on `APP_PORT: 8000` in `docker-compose.yml`.

```suggestion
port: 8026
```

kaichappel (changes requested):
Nice work @vitto4, this is really close.

I gave the Umbrel package a test. It installs and opens cleanly, and I was able to complete the Readeck onboarding and save a bookmark without any issues.

The only blocker I found is the app-store port. The manifest currently uses `8000`, which conflicts with ZeroNote in the app store. I left a small inline suggestion to move the Umbrel/app_proxy port to `8026`.

The internal `APP_PORT` in `docker-compose.yml` can stay at `8000`, since that is just Readeck's internal web-service port.

Thanks also for including the gallery image and icon references. We'll get the final gallery images and icon together on our side before this goes live.

Once the port is changed, I can re-test and we should be in good shape.

vitto4:
Hey @kaichappel, thanks for the review !
Agreed, I updated the port to `8026`

kaichappel:
Thanks for making the change @vitto4, we're good to go live!

nmfretz:
Great work @vitto4! We've added the gallery assets and are going live to the app store 🎉

image

PR #5583, opened 2026-05-06
App Submission: am-i.exposed

Copexit:
# App Submission
Am I Exposed? It's a beautiful and feature rich privacy tool. The feature set compares to everything the defunct KYCP.org and OXT by the Samourai team. It can be tested at  https://am-i.exposed . Running directly on your UmbrelOS it will only query your Mempool API.

We created the [copexit-umbrel-app-store](https://github.com/Copexit/copexit-umbrel-app-store/) a couple of months ago and it seems to have quite some installs, but also quite some people confused about how to add an external store on Umbrel. Therefore, it seems logical to submit it for inclusion on the main store.

 ### App name                                                                                                                                                                          
 am-i.exposed
                                                                                                                                                                                        
 ### 256x256 SVG icon                                                                                  
icon

 ### Gallery images                                 

                                                                           

### I have tested my app on:                 
      
  - [x] umbrelOS 1.7.2 on an NUC i5
  - [x] umbrelOS on Linux VM


Salinas25lz:
Strongly in favor of including this in the official store. 

I've been using am-i.exposed for a while and it has become my go-to tool for auditing my own on-chain footprint and spotting privacy mistakes. The educational layer is one of its strongest points — you don't just get a verdict, you actually learn *why* something is a leak and what to do differently next time. That's rare in this space.

The decisive argument for the official store, in my view, is the local Mempool integration. Auditing your own UTXOs against a third-party hosted service is self-defeating — you leak the very metadata you're trying to evaluate. Running this against your own Umbrel node is the only way the analysis is privacy-coherent end-to-end.

KYCP and OXT mattered to a lot of us, and there hasn't been a credible self-hostable replacement until now. The community-store route adds real friction: first-time Umbrel users tend to read "not in the official store" as "not vetted", which is the opposite of what's true here.

Thank you,

am-i-exposed/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: am-i-exposed_web_1
+      APP_PORT: 8080
+      PROXY_AUTH_ADD: "false"
```
kaichappel: Could you clarify whether this is intentional? `PROXY_AUTH_ADD: "false"` means this route bypasses Umbrel auth. Does am-i.exposed provide its own auth layer, or is the goal for the app route to be accessible without Umbrel auth? If there isn't a separate auth flow, we'd prefer to remove this override so Umbrel protects the UI and the proxied local Mempool/Tor routes.

am-i-exposed/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+
+  web:
+    image: ghcr.io/copexit/am-i-exposed-umbrel:v0.35.7@sha256:809a579e475d36377b41c162346973ccf37b05cac540098660981f29dd273fd9
+    user: "101:101"
```
kaichappel, line 11: I checked our usual preference for `user: "1000:1000"`, but the current image doesn't appear to support it: nginx can't write its templated config under that UID, so no server block is rendered and `/health` is refused. So I'm not asking for that change unless you want to adjust image ownership upstream. Using UID 101, or the image's default nginx user, seems intentional for this image.

am-i-exposed/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+    init: true
+    restart: on-failure
+    stop_grace_period: 30s
+    environment:
+      APP_MEMPOOL_IP: ${APP_MEMPOOL_IP}
+      APP_MEMPOOL_PORT: ${APP_MEMPOOL_PORT}
+      APP_MEMPOOL_HIDDEN_SERVICE: ${APP_MEMPOOL_HIDDEN_SERVICE}
+      APP_TOR_PROXY_IP: ${APP_AM_I_EXPOSED_TOR_PROXY_IP}
```
kaichappel: Could you clarify whether the fixed container IPs are intentional here? From my testing, the web container can reach the sidecar by Docker service name (`APP_TOR_PROXY_IP=tor-proxy`) and `/tor-proxy/health` still returns `ok`. If there isn't a specific reason for static IPs, we'd prefer to use the service name and remove `exports.sh` plus the `ipv4_address` blocks below.

```suggestion
      APP_TOR_PROXY_IP: tor-proxy
```

am-i-exposed/docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+      APP_HOST: am-i-exposed_web_1
+      APP_PORT: 8080
+      PROXY_AUTH_ADD: "false"
+
+  web:
+    image: ghcr.io/copexit/am-i-exposed-umbrel:v0.35.7@sha256:809a579e475d36377b41c162346973ccf37b05cac540098660981f29dd273fd9
+    user: "101:101"
+    init: true
+    restart: on-failure
+    stop_grace_period: 30s
```
kaichappel, line 14: The image's built-in healthcheck uses `http://localhost:8080/health`, which resolves to IPv6 first in this container while nginx is only listening on IPv4. The app still opens, but Docker leaves `am-i-exposed_web_1` marked `unhealthy`. Could you override the healthcheck here to probe IPv4 explicitly?

```suggestion
    stop_grace_period: 30s
    healthcheck:
      test: ["CMD-SHELL", "wget -qO- http://127.0.0.1:8080/health || exit 1"]
      interval: 30s
      timeout: 3s
```

kaichappel (changes requested):
Thanks for the submission @Copexit. I gave the Umbrel package a test: it installs, opens through Umbrel, and I was able to complete a scan against the local Mempool API.

Also, thanks for including the icon and screenshots in the PR body. We'll get the gallery assets finalized on our side before this goes live.

There are a few packaging details I'd like to clarify before this can go live. Some of these may be intentional, so please push back if I'm missing context:

- It looks like Umbrel auth is disabled with `PROXY_AUTH_ADD: "false"`. Does am-i.exposed have its own auth layer, or is the intent for the app route to be accessible without Umbrel auth? If there isn't a separate auth flow, we'd prefer to remove that override so Umbrel protects the UI and the proxied local Mempool/Tor routes.
- I noticed the app exports fixed container IPs and uses them for the web/tor-proxy containers. Is there a reason this needs static addressing? From my testing, the web container can reach the sidecar by Docker service name, so if there isn't a specific reason for fixed IPs, we'd prefer to use the service name and drop the `exports.sh`/`ipv4_address` assignments.
- The web container healthcheck is the one concrete runtime issue I hit. The app itself opens, but Docker marks the container unhealthy because the image healthcheck probes `localhost`, which resolves to IPv6 first in this container while nginx is listening on IPv4. I left a suggestion to override that probe to `127.0.0.1`.

One note on the web container user: I also checked our usual preference for `user: "1000:1000"`, but the current image doesn't appear to support it because nginx can't write its templated config under that UID. So I'm not asking for that change unless you want to adjust image ownership upstream. Using UID 101, or the image's default nginx user, seems intentional for this image.

Once those bits are clarified/updated, I can re-test.

am-i-exposed/docker-compose.yml
Copexit: Yes, sounds like indeed this should be removed and be protected by Umbrel auth.

Copexit: Done - dropped `PROXY_AUTH_ADD: "false"` in 5e61816. Umbrel auth will now protect the UI and the proxied Mempool/Tor routes.

am-i-exposed/docker-compose.yml
Copexit: Confirmed - the image's `USER nginx` (UID 101) is intentional. nginx-unprivileged's `/etc/nginx/conf.d` is owned by UID 101, and forcing UID 1000 blocks the entrypoint's envsubst step. We'll consider chowning the writable paths in the Dockerfile to enable `user: "1000:1000"` in a future release, but keeping `101:101` for now since it matches what the image supports out of the box.

am-i-exposed/docker-compose.yml
Copexit: Good catch - applied in 5e61816. Switched `APP_TOR_PROXY_IP` to the literal `tor-proxy` service name and dropped both `networks:` blocks and the `exports.sh` file entirely. There was no specific reason for static IPs - earlier in our testing we hit a container-startup bug (the UID 1000 / nginx-unprivileged conf.d issue, now fixed in v0.35.7) that masked whether service-name DNS worked, and we defaulted to IP-pinning. Re-testing confirms service-name DNS resolves fine. Cleaner result, thanks.

am-i-exposed/docker-compose.yml
Copexit: Applied your suggestion in 5e61816 - the web service now overrides the healthcheck to probe `http://127.0.0.1:8080/health` explicitly. We'll also patch the `HEALTHCHECK` directive in `Dockerfile.umbrel` in the next image release so this isn't needed at the compose layer, but the override fixes the unhealthy state today without requiring an image rebuild.

nmfretz:
Great work @Copexit! Gallery assets are now complete. Welcome to the official app store 🎉

image

besoeasy:
thats a good ui ! nice

PR #5585, opened 2026-05-06
App Submission: OpenThread Border Router for Home Assistant

vnc0:
# App Submission

### OpenThread Border Router for Home Assistant
OpenThread Border Router (OTBR) lets you bridge a Thread mesh network to your
local Wi-Fi/Ethernet network using a supported Thread radio dongle. 

Important: It is necessary to enable IPv6 on both your home server and your network.

**Why `hass-otbr-docker` instead of the official `openthread/border-router`?**

This Umbrel app targets users running OTBR alongside Home Assistant or Matter. `hass-otbr-docker` is a thin wrapper around the official OpenThread Border Router project that adds the integrations Home Assistant expects out of the box:

- The `otbr-agent-rest-discovery` service so HA's OTBR integration auto-discovers the router via mDNS.
- A `socat` TCP bridge that HA's OTBR integration needs to talk to `otbr-agent`.
- `universal-silabs-flasher` for one-shot firmware flashing of Silicon Labs dongles, which is the most common hardware Umbrel users will plug in.
- Sensible env-var defaults (`OTBR_REST_PORT`, `BACKBONE_IF`, NAT64, IPv6 firewall).

The official `openthread/border-router` image only recently gained a built-in web UI ([ot-br-posix#3200](https://github.com/openthread/ot-br-posix/pull/3200), merged Jan 2026) and would still require us to script the HA-side glue ourselves. Picking `hass-otbr-docker` keeps the app focused on its intended use case — Thread devices reaching Home Assistant — without reimplementing the integration layer.

### 256x256 SVG icon

icon

### Gallery images
SCR-20260506-udsj
SCR-20260506-udni
SCR-20260506-udkz
SCR-20260506-udiw
SCR-20260506-uesi

### I have tested my app on:
- [x] umbrelOS on an Umbrel Home with an SONOFF Zigbee 3.0 USB Dongle Plus MG24 and several Ikea Thread devices
- [ ] umbrelOS on Linux VM

### DIFF
--- a/openthread-border-router/docker-compose.yml
+++ b/openthread-border-router/docker-compose.yml
@@ -1,21 +1,67 @@
 version: "3.7"
 
 services:
+  setup:
+    image: python:3.12-alpine@sha256:236173eb74001afe2f60862de935b74fcbd00adfca247b2c27051a70a6a39a2d
+    # Host networking so the wizard sees the real host serial devices and the
+    # host's interface/IPv6 state (via /proc and /sys), not a container namespace.
+    network_mode: host
+    restart: on-failure
+    stop_grace_period: 10s
+    environment:
+      SETUP_PORT: "7586"
+      OTBR_WEB_PORT: "7587"
+      OTBR_REST_PORT: "8083"
+    volumes:
+      - ${APP_DATA_DIR}/hooks:/app:ro
+      - ${APP_DATA_DIR}/data:/data
+      - /dev:/dev:ro
+    command: ["python3", "/app/server.py"]
+
+  # OpenThread Border Router. Waits for the setup wizard to write a configured
+  # settings file before launching, so it never crash-loops on the default
+  # device/interface values on a fresh install.
   server:
     image: ghcr.io/ownbee/hass-otbr-docker:v0.3.0@sha256:5ca7008462e94d4c937aca8c1c9a60e709d6d300625363c161b946683032d5b1
     network_mode: host
     privileged: true
     restart: on-failure
     stop_grace_period: 1m
-    env_file:
-      - ${APP_DATA_DIR}/data/settings.env
+    depends_on:
+      - setup
+    environment:
+      OTBR_WEB_PORT: "7587"
+      OTBR_REST_PORT: "8083"
     volumes:
-      - ${APP_DATA_DIR}/data/thread:/data/thread
+      - ${APP_DATA_DIR}/data:/data
       - /dev:/dev
     entrypoint:
       - /bin/bash
       - -c
       - |
+        set -uo pipefail
+        SETTINGS=/data/settings.env
+        load_settings() {
+          CONFIGURED=0; DEVICE=""; BAUDRATE=""; FLOW_CONTROL=""; BACKBONE_IF=""
+          FIREWALL=""; NAT64=""; AUTOFLASH_FIRMWARE=""
+          [ -f "$$SETTINGS" ] || return
+          while IFS='=' read -r key val; do
+            case "$$key" in
+              DEVICE|BAUDRATE|FLOW_CONTROL|BACKBONE_IF|FIREWALL|NAT64|AUTOFLASH_FIRMWARE|CONFIGURED)
+                export "$$key=$$val" ;;
+            esac
+          done < "$$SETTINGS"
+        }
+        echo "[otbr] waiting for first-run setup to complete (open the app to configure)..."
+        while true; do
+          load_settings
+          if [ "$${CONFIGURED:-0}" = "1" ] && [ -n "$${DEVICE:-}" ] && [ -e "$${DEVICE:-/nonexistent}" ]; then
+            break
+          fi
+          echo "[otbr] not configured yet — retrying in 5s"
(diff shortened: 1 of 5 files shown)


openthread-border-router/umbrel-app.yml
```diff
@@ -0,0 +1,60 @@
...
+
+
+  1. Plug your Thread radio dongle into your Umbrel device before installing.
+
+  2. On first start the app creates a settings file at
+     `app-data/openthread-border-router/data/settings.env`. Defaults assume
+     a dongle at `/dev/ttyUSB0` and a backbone interface of `eth0`. Edit
+     this file (e.g. set `DEVICE=/dev/ttyACM0` or `BACKBONE_IF=enp1s0` for
+     Umbrel Home) and restart the app to apply changes. The file is
+     preserved across app updates.
```
kaichappel: This setup flow is the main blocker. It asks users to plug in a dongle and then edit `settings.env`, but if the default radio or backbone interface is wrong, OTBR exits before the app UI is reachable.

Could you add a minimal safe setup/placeholder path before launching OTBR? This doesn't need to be a polished upstream settings UI, but the Umbrel app should install, start, and open cleanly enough to tell users what to configure, then launch OTBR once those values are valid.

openthread-border-router/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+  server:
+    image: ghcr.io/ownbee/hass-otbr-docker:v0.3.0@sha256:5ca7008462e94d4c937aca8c1c9a60e709d6d300625363c161b946683032d5b1
+    network_mode: host
+    privileged: true
+    restart: on-failure
+    stop_grace_period: 1m
+    env_file:
+      - ${APP_DATA_DIR}/data/settings.env
+    volumes:
+      - ${APP_DATA_DIR}/data/thread:/data/thread
```
kaichappel: This bind-mount source needs a checked-in scaffold. Could you add `openthread-border-router/data/thread/.gitkeep` so `${APP_DATA_DIR}/data/thread` exists before Docker mounts it?

kaichappel (changes requested):
Thanks for the submission @vnc0.

I think the main blocker is first-run behavior. OTBR expects a configured Thread radio and backbone interface before `otbr-agent` starts, and Home Assistant handles that with an add-on configuration step before start. On Umbrel, we need the package to provide a small safe setup/placeholder layer before launching OTBR. This does not need to be a full upstream settings UI, but a missing or differently named radio/interface should open a clear setup state instead of leaving the app unreachable.

A few things before this can go live:

- Add a minimal safe setup/placeholder path before launching OTBR, instead of seeding active `/dev/ttyUSB0` / `enp1s0` defaults that can fail before users can fix them. If the cleanest fix belongs in `hass-otbr-docker`, an upstream issue or PR is reasonable, but the Umbrel package needs that safe first-run behavior before merge.
- Add `openthread-border-router/data/thread/.gitkeep` so the `/data/thread` bind-mount source exists in app data before Docker mounts it.

Also, thanks for including the icon and screenshots. We'll finalize the gallery assets on our side before this goes live.

openthread-border-router/umbrel-app.yml
vnc0: Setup just got easier. With the new setup wizard, you can configure everything just by choosing your device from the list.
image
image

openthread-border-router/docker-compose.yml
vnc0: Fixed.

kaichappel:
Really great work on this @vnc0. The new setup flow is a big improvement.

One small issue remains. If I enter a manual device path that does not exist, for example `/dev/ttyOTBRmissing`, the wrapper saves it and shows “Configured / The border router is running”, but OTBR is still waiting for the device and the dashboard/API ports never come up.

Could you keep the wrapper in a waiting/setup state until the configured device exists and OTBR has actually started, or validate the device path before marking it configured? The main thing is that the app should not tell users the border router is running while OTBR is still waiting.

kaichappel:
Saw you pushed an update @vnc0. I re-tested this on an Umbrel device without Thread radio hardware, and the behaviour looks good! We'll get gallery assets finalized and then go live!

nmfretz:
Thanks for the submission @vnc0! I made a few changes to get this finalized:

- removed the top-level `data/.gitkeep`... `data/thread/.gitkeep` is enough to keep the required data directory.
- moved the setup wizard from `hooks/server.py` to `server.py.template` and mounted the rendered `server.py` in Docker. This keeps runtime app code out of lifecycle hooks... but we still ensure that updates are applied on existing installs (umbrelOS brings down *template files).
- tweaked the instructions slightly

Also, because the setup UI you added is reachable on the LAN and writes config for the privileged OTBR container, I tightened manual device input slightly to only accept expected serial radio paths. I don’t think the UI meaningfully changes the app’s security model, since OTBR’s own web and REST interfaces are already unauthenticated and host networking means the app can’t sit cleanly behind Umbrel’s built-in auth.

Welcome to the app store 🎉

image

PR #5589, opened 2026-05-07
Update etherpad to v2.7.3

kaichappel:
Updates etherpad to v2.7.3.

### DIFF
--- a/etherpad/docker-compose.yml
+++ b/etherpad/docker-compose.yml
@@ -35,7 +35,7 @@ services:
       SHOW_SETTINGS_IN_ADMIN_PAGE: "true"
       SUPPRESS_ERRORS_IN_PAD_TEXT: "false"   
     restart: on-failure
-    user: "0:0"
+    user: "5001:0"
 
   postgres:
     image: postgres:15-alpine@sha256:93a6166360be8f1d116b7ea97ebb8dea9e2c9c746349e952738bbba72a961d7c
--- a/etherpad/docker-compose.yml
+++ b/etherpad/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 9001
       
   app:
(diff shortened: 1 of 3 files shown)


etherpad/docker-compose.yml
```diff
@@ -34,7 +35,7 @@ services:
       SHOW_SETTINGS_IN_ADMIN_PAGE: "true"
       SUPPRESS_ERRORS_IN_PAD_TEXT: "false"   
     restart: on-failure
-    user: "1000:1000"
+    user: "0:0"
```
al-lac: Great catch @kaichappel! I would go with the id "5001" here and use a pre-start hook to update existing directories.

```suggestion
    user: "5001:0"
```

al-lac (changes requested):

etherpad/docker-compose.yml
kaichappel: Thanks — agreed. I pushed `5001:0` and added a scoped executable `hooks/pre-start` that migrates only Etherpad’s writable app-data mounts (`data/plugins` and `data/etherpad-var`) to `5001:0`, leaving `data/postgres_data` at `1000:1000` for Postgres. Re-tested update path from 2.7.2 plus fresh install/restarts on the amd64 Umbrel device; Etherpad is healthy and the ownership migration behaved as expected.

etherpad/umbrel-app.yml
al-lac: @kaichappel this needs to be bumped to `1.1` when hooks are used.

kaichappel: Good catch, thanks. Bumped `manifestVersion` to `1.1` for the hook, and also polished `hooks/pre-start` to match the repo pattern (`#!/usr/bin/env bash` + `set -euo pipefail`, named paths, and only chown when ownership differs). CI is green again.

al-lac (approved):

PR #5601, opened 2026-05-08
Update fizzy to 5f45e54

kaichappel:
Updates Fizzy from 754aaa5 to 5f45e54.

### DIFF
--- a/fizzy/docker-compose.yml
+++ b/fizzy/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: 'false'
 
   server:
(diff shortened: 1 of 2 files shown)


kaichappel:
Pausing this as draft while we sort out the image source.

The current PR switches to Basecamp's official image, but `fizzy-umbrel` carries an Umbrel-important sign-in patch that shows the magic code in the UI so users don't need to dig through container logs. The official image does not currently expose an equivalent production-safe env flag, so the image switch is not ready as-is.

Awaiting either:

- a bumped `ghcr.io/mauricewipf/fizzy-umbrel` image for the latest Fizzy release, or
- an upstream/official image path that preserves the Umbrel-friendly sign-in flow.

kaichappel:
Closing this since #5665 supersedes it.

PR #5615, opened 2026-05-08
Update Formicaio to v0.8.4

bochaco:
Since v0.6.2, Formicaio has received a significant number of UX and UI improvements across multiple releases (see updated screenshots below).
The interface was redesigned with a fresh visual style, and a range of new features were added to give users more visibility and control over their nodes, earnings, and AI tooling:
  - Redesigned look and feel with an updated GUI visual style
  - Nodes view now supports both tiles and list layouts, togglable by the user
  - Disk usage reported per node and in aggregate, with a configurable polling frequency
  - Rewards addresses panel on the dashboard listing each address and its balance
  - Earnings analytics on the dashboard showing rewards payments and totals per period, with a toggle to enable/disable monitoring
  - New AI section with an MCP server status page, activation instructions, and a dedicated Agent page for interacting with an LLM agent that can manage your nodes
  - Agent chat displays formatted tool output and responses, with its own settings and background task support
  - Nodes can be sorted by binary version
  - Single-node actions applied directly, bypassing the batch queue
  - Batch failures stay in the queue until dismissed, with more detail on in-progress failures
  - Persistent notifications history so users can review past notifications at any time
  - MCP server status and information page under the AI section
  - AI Agent page with its own settings, background task management, and formatted chat output
  - Log level option when adding new nodes
  - Clicking the "total nodes" dashboard card navigates to the nodes list
  - Charts now show an empty-state message when switching nodes instead of stale data
  - Placeholder cards show a title while nodes are being created
  - Various icon and styling fixes across light and dark themes

### Gallery images
- Dashboard screen
Dashboard screen

- Nodes list screen (tiles mode)
Nodes list screen

- Nodes list screen (list mode)
Nodes list screen

- Nodes creation (general tab)
Nodes creation

- Nodes creation (advanced tab)
Nodes creation

- App settings panel
App settings panel

- Multiple nodes actions (tiles mode)
Multiple nodes actions

- Multiple nodes actions (list mode)
Multiple nodes actions

- Terminal screen
Terminal screen

### DIFF
--- a/formicaio/docker-compose.yml
+++ b/formicaio/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       - host.docker.internal:host-gateway
 
   app:
-    image: bochaco/formicaio:0.6.2-native-lcd-disabled@sha256:45d03ee6f65cfc294ba718e5d9e6607009b87331e9886965c4a14714de5a2148
+    image: bochaco/formicaio:0.8.4-native-lcd-disabled@sha256:9f745c0b3dac2bae47de23cb4c444d2c04ce83610a84a26c784534011f4e01f7
     restart: on-failure
     user: 1000:1000
     network_mode: host
--- a/formicaio/umbrel-app.yml
+++ b/formicaio/umbrel-app.yml
@@ -3,7 +3,7 @@ id: formicaio
 name: Formicaio
 tagline: Run nodes on the Autonomi peer-to-peer (P2P) network
 category: crypto
-version: "0.6.2"
+version: "0.8.4"
 port: 52100
 description: >-
   ## Simplify your decentralized experience with Formicaio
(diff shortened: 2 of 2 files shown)


kaichappel:
Thanks @bochaco, and sorry for the delay getting to this.

I tested this on an amd64 Umbrel device. A fresh install of v0.8.2 works and the new dashboard loads nicely, but I don’t think we can merge this as-is because the update path from the current app store version fails on startup.

After updating from v0.6.2 to this PR’s v0.8.2 package, the app container keeps restarting with this migration error:

```text
while executing migration 20260322000000: error returned from database: (code: 1) table settings has no column named node_list_page_size
```

Could you take a look at the migration path from the existing v0.6.2 database? Once that’s fixed, I’m happy to re-test.

bochaco:
> Could you take a look at the migration path from the existing v0.6.2 database? Once that’s fixed, I’m happy to re-test.

Thanks a lot @kaichappel , I'll take a look and update this PR accordingly.

kaichappel:
Thanks @bochaco, looks like this is now updated to v0.8.4. Quick check before I kick off another round of device testing: is this ready for us to test again?

bochaco:
Hi @kaichappel , yes, it's ready, I was just doing some final tests.
I've tested it on both amd64 (Linux) and arm64 (Pi4) and it is upgrading fine now from v0.6.2 to v0.8.4. 

Note that in case you do a fresh v0.6.2 install, it won't start up since it cannot download the node binary from where it used to since it is now unavailable at that URL. The node binary has been moved by the publisher (Autonomi) to a different location (there was a big set of breaking changes made by them in recent releases). But even so, you can still upgrade from such a fresh install to this new v0.8.4 successfully as per my tests.

kaichappel:
Thanks @bochaco, I retested on an amd64 Umbrel device and the update path looks good now.

I did hit the fresh v0.6.2 startup issue you mentioned, but after updating that install to v0.8.4 the migration completed, the containers stayed up, and the dashboard loaded through Umbrel.

I saw a one-off `sqlx-sqlite-worker` panic during migration, but the app continued normally, so I don’t think it blocks the update.

Thanks again for fixing this. Going live!

bochaco:
Awesome, thanks a lot for your help @kaichappel !

PR #5616, opened 2026-05-09
Add Pironman5

cavonlee:
# App Submission

### App name
Pironman 5

### 256x256 SVG icon
_(Upload an icon with no rounded corners as it will be dynamically rounded with CSS.)_

pironman5

### Gallery images

### I have tested my app on:
- [*] umbrelOS on a Raspberry Pi
- [ ] umbrelOS on Linux VM

Note: This is a app for Pironman 5 Case hardware, only for Raspberry Pi. so no need for other system


kaichappel:
Thanks for these submissions @cavonlee, and sorry for the delay in reviewing them.

After looking through the packages more closely, we can’t accept the Pironman submissions as-is. The blocker isn’t the Pironman hardware itself. It’s the amount of host access the apps require.

Both packages currently include a very broad set of host permissions:

```yaml
network_mode: host
pid: host
privileged: true
volumes:
  - /dev:/dev
  - /lib/modules:/lib/modules:ro
  - /:/host:ro
  - /proc:/proc:ro
  - /sys:/sys:ro
```

We understand why a hardware-control app may need deeper system access than a typical web app, and we do make exceptions for some hardware-heavy apps when the access is narrowly scoped. But this combines several host-level permissions at once: host networking, the host PID namespace, full privileged mode, raw device access, read access to the full host filesystem, and host `/proc`/`/sys`.

In practical terms, this means the container is much closer to having control over the Umbrel host itself than being a sandboxed app running on Umbrel. If the app image or any code inside it were ever compromised, the impact could extend well beyond the app’s own data.

On top of that, the image entrypoint crosses into host namespace control so the app can trigger host shutdown/reboot behavior. That gives the container a level of control over the Umbrel host that we can’t ship through the official app store.

So we won’t be able to merge these submissions in their current form.

If there’s a way to redesign the packages with a much narrower permission model, for example only the specific devices/capabilities needed and without full privileged mode, host PID access, or broad host filesystem mounts, we’d be happy to take another look. But as submitted, this is outside what we can accept for the official Umbrel app store.

nmfretz:
Thanks again for taking the time to submit this. I’m going to close this PR for now since I haven’t heard back after the review, and it looks like it may no longer be active.

If you’re able to revisit the package with a narrower host-permission model, please feel free to open a fresh PR anytime. We’d be happy to take another look.

PR #5635, opened 2026-05-11
Add Sensei app

KingParmenides:
## Summary
- Add a Sensei Umbrel app manifest and compose file
- Connect Sensei to the installed Umbrel Bitcoin node through the standard app environment variables
- Persist Sensei data under the app data directory and expose the web admin at /admin/nodes

## Notes
- Uses the existing multi-arch louneskmt/sensei:v0.2.2-beta2 image pinned to manifest digest sha256:7dd160b52baa0c528f58ea8c3be77a73d257004e6c5d25d6c6bbd166d4055688
- This follows up on L2-Technology/sensei#20 and the earlier getumbrel/umbrel#1339 attempt, which was closed after apps moved to this repo

## Validation
- Parsed sensei/docker-compose.yml and sensei/umbrel-app.yml with Ruby YAML
- Confirmed id sensei and port 5401 are unique in the app store checkout
- Confirmed the pinned Docker digest resolves as an application/vnd.docker.distribution.manifest.list.v2+json multi-arch manifest
- Ran git diff --check

Docker runtime testing was not run locally because Docker is not installed in this environment.

## Payout
If L2-Technology/sensei#20 is accepted for the bounty, BTC payout address: 39Q34P8A7g375yqEr8buvNJkUbRgKfbKQZ


KingParmenides:
Pushed a small follow-up to set `NETWORK` from `APP_BITCOIN_NETWORK_ELECTRS` instead of `APP_BITCOIN_NETWORK`. Sensei parses Rust Bitcoin network names, so Umbrel's default `mainnet` value needs the existing Bitcoin export mapping to `bitcoin` to start correctly.\n\nThe external `15401-15420` port mapping is intentional: `app_proxy` handles the admin UI on 5401, while Sensei allocates node ports from that range for Lightning node listeners.

tnull:
Last commit on sensei's `main` branch is ~4 years ago. It's probably fair to say that it's not being actively maintained at this point. So I doubt adding Umbrel support *now* makes sense.

kaichappel:
Thanks for taking the time to put this together @KingParmenides, and thanks @tnull for flagging the maintenance concern.

After looking into Sensei a bit more, I don’t think this is a good fit for the official Umbrel app store right now. The upstream project does not appear to be actively maintained, the Docker image is from 2022, and the project site/docs no longer seem to be available.

For a Lightning app especially, we need a maintained upstream with a clear support and security posture, since users may end up putting real funds on it.

That said, Umbrel supports community app stores, so anyone who wants to run Sensei on their own Umbrel can still package and install it that way.

I’m going to close this for now, but we’d be happy to revisit if Sensei becomes actively maintained again and has a current release/image we can test properly.

PR #5660, opened 2026-05-12
Update Bitcoin Knots to v1.2.12

Retropex:


kaichappel:
Thanks @Retropex, this looks good from my side.

I tested the update and a fresh install on an Umbrel device. Both worked as expected, and the BIP 110/RDTS warning popup shows correctly before the upgrade is applied.

Going live!

Issue #560, opened 2021-01-18
Bounty: Mastodon app for Umbrel


WhiteRabbitBTC:
Would be great to see a Mastodon app available for Umbrel. I'll stake $500 USD (paid in bitcoin) to anyone who can help make this a reality.

LuckyCoders:
5 years have passed, and yet nothing has changed.

We have **[Akkoma](https://akkoma.social)** and **[GoToSocial](https://gotosocial.org/)**, which run easily on low-power hardware. However, we need to keep experimenting and adapt their Docker containers to work with Cloudflare Tunnels "out of the box."

I would like to do this someday.

P.s
It would also be nice if Pixelfed adapted or integrated it with Immich (already supported in Umbrel)

Asobu01:
I can take getumbrel/umbrel-apps#560.
First step: reproducing the reported issue first, then applying a focused fix with clear validation.
Draft PR target: ~8h.
Will include repro + root-cause summary, focused patch (plus tests/docs if needed), and review-ready PR with validation notes.
I can post a reproducible checklist/log snippet in ~30 minutes so you can verify execution speed before the full PR.
@Asobu01

Asobu01:
Quick follow-up on `getumbrel/umbrel-apps#560`. I am still ready to start immediately and can post the first repro/checklist note before the draft PR if helpful. If you want me on this one, I can begin now. (@Asobu01)

poofeth:
Submitted a Mastodon app package in #5657 for this bounty request. It includes the official Mastodon image, Postgres, Redis, web/streaming/Sidekiq services, and a first-run bootstrap for secrets, database setup, and the Umbrel admin account.

caydyan:
Submitted a clean Mastodon app package for this bounty in #5734.

It addresses the review feedback from the earlier submission path: pre-install env-file creation, executable hooks, preserved domain/SMTP settings across restarts, first Owner bootstrap via Umbrel deterministic app password, and app-package cleanup. `Lint apps` is passing.

BTC payout address if accepted for the bounty: `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

Pkaulsjs:
Hi! This bounty is now implemented: **PR #6006** packages Mastodon v4.7.0 as an Umbrel app.

What it includes:
[...]

PR #5662, opened 2026-05-12
Add Mastodon app

weishen612:
## Summary

This adds a Mastodon app package for Umbrel.

Refs #560.

The app runs the Mastodon web process, streaming service, Sidekiq, PostgreSQL, Redis, and a small Nginx proxy that keeps the normal web routes and streaming API on the same Umbrel app port. The pre-start hook prepares the persistent data directories and creates the Mastodon secrets on first install, while preserving them across restarts.

## Local Validation

I tested this in two passes.

First, I ran the app directly with Docker Compose using temporary app data and a small local overlay for the Umbrel app proxy placeholder. The migration completed, the web UI returned 200 with the Mastodon title, the streaming health endpoint returned `OK`, and restarting the stack kept the generated `mastodon.env` unchanged.

Then I installed it through a local `umbrel-dev` app-store flow. The install mutation returned `true`, the app state was `ready`, Postgres and Redis were healthy, the migration container exited 0, and the web, streaming, Sidekiq, proxy, and app proxy containers stayed up. After an app restart, `mastodon.env` kept the same hash and the web UI plus streaming health endpoint still responded.

Checks run:

```sh
TMPDIR=/tmp TEMP=/tmp TMP=/tmp python3 -m pytest tests/test_mastodon_app.py -q
TMPDIR=/tmp TEMP=/tmp TMP=/tmp python3 -m pytest -q
bash -n mastodon/hooks/pre-start
```

Results:

- Mastodon app tests: `9 passed`
- Local project tests: `23 passed`
- Hook syntax check: passed

## Notes

- The manifest leaves `releaseNotes` and `gallery` empty because this is a new app submission.
- Federation and reliable email still need a public domain and SMTP settings outside the default local Umbrel install.


mastodon/docker-compose.yml
```diff
@@ -0,0 +1,113 @@
...
+      streaming:
+        condition: service_started
+
+  web:
+    image: *mastodon_image
+    restart: on-failure
+    stop_grace_period: 1m
+    command: ["bundle", "exec", "puma", "-C", "config/puma.rb"]
+    env_file:
+      - ${APP_DATA_DIR}/mastodon.env
```
kaichappel, line 30: Fresh install fails here because this env file does not exist yet. Compose reads `env_file` before the app reaches `pre-start`, so the install stops with:

```text
env file /home/umbrel/umbrel/app-data/mastodon/mastodon.env not found
```

Could you either add a non-secret placeholder file that the hook can rewrite, or generate it earlier with an executable `pre-install` hook?

mastodon/hooks/pre-start
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env bash
```
kaichappel, line 1: `hooks/pre-start` needs the executable bit. Right now it is checked in as `100644`, so Umbrel will not run it.

Please `chmod +x mastodon/hooks/pre-start`.

mastodon/hooks/pre-start
```diff
@@ -0,0 +1,129 @@
...
+
+if [[ -z "${ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT}" ]]; then
+  ACTIVE_RECORD_ENCRYPTION_KEY_DERIVATION_SALT="$(rails_secret)"
+fi
+
+if [[ -z "${ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY}" ]]; then
+  ACTIVE_RECORD_ENCRYPTION_PRIMARY_KEY="$(rails_secret)"
+fi
+
+cat > "${ENV_FILE}" <<EOF
```
kaichappel, line 95: This currently rewrites `mastodon.env` on every start. That is risky for Mastodon because SMTP/domain settings are user-supplied, and `LOCAL_DOMAIN` / `WEB_DOMAIN` become part of the instance identity. If those values change when the Umbrel hostname changes, the instance can break in ways that are hard to repair.

Could this preserve existing user/domain/SMTP values instead of recreating the whole env file each start? It also needs to support normal Umbrel launch hosts, not only one hardcoded host/port.

mastodon/umbrel-app.yml
```diff
@@ -0,0 +1,28 @@
...
+releaseNotes: ""
+developer: Mastodon gGmbH
+website: https://joinmastodon.org
+dependencies: []
+repo: https://github.com/mastodon/mastodon
+support: https://github.com/mastodon/mastodon/discussions
+port: 8388
+gallery: []
+path: ""
+defaultUsername: ""
```
kaichappel, line 25: A fresh install currently lands on the Mastodon sign-in page, but there is no Owner account yet. Since Mastodon normally creates the first Owner with `tootctl`, Umbrel users need a browser-friendly onboarding/bootstrap path here.

Could you add a tested first-admin flow, such as deterministic admin credentials or another supported setup flow?

mastodon/icon.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="512" height="512" viewBox="0 0 512 512" fill="none" xmlns="http://www.w3.org/2000/svg">
```
kaichappel, line 1: Could you remove this from the app directory? We handle final icon/gallery assets outside `umbrel-apps` before launch.

tests/test_mastodon_app.py
```diff
@@ -0,0 +1,156 @@
+from __future__ import annotations
```
kaichappel, line 1: Could you remove this test file from the app package? This repo does not run per-submission pytest files, so the validation details are better kept in the PR body.

kaichappel (changes requested):
Thanks for the submission @weishen612. Mastodon is a bit different from most apps we review, so I think we should first agree on what this package is meant to offer.

If this is meant to be a full Mastodon server, users need a public domain with HTTPS, working SMTP, and federation endpoints that are reachable without normal Umbrel app auth. Those choices matter because Mastodon's domain becomes part of the account identity and is hard to change later. That is difficult to turn into a normal one-click Umbrel install.

If this is meant to be a local/private sandbox, that may be a different path, but the app should be very clear about that scope so users do not expect a normal federated instance.

I left the package-specific issues inline. The current package still needs a clean fresh install path, safer domain handling for Umbrel access, and a first-admin setup flow before we can retest.

nmfretz:
Thanks again for the Mastodon submission @weishen612!

Just a heads up: #5734 was opened as another Mastodon submission and it calls out the review feedback from this PR.

Since this PR has already been reviewed and is still open, I wanted to give you first chance to follow up here. Are you planning to make the requested changes?

I'll wait until Monday, June 22, 2026 before deciding whether to continue review on the other PR instead.

There's also the bounty on #560, so I want to keep the process clear for everyone.

nmfretz:
Closing due to inactivity, and in favour of https://github.com/getumbrel/umbrel-apps/pull/5734

PR #5664, opened 2026-05-13
Update Monero daemon to v0.18.5.0

colder-is-better:
This is the v0.18.5.0 release of the Monero software.
This recommended release adds SOCKS v5 support, removes UPnP support, and includes a large number of bug fixes.

Reference: https://github.com/monero-project/monero/releases/tag/v0.18.5.0

### DIFF
--- a/monero/docker-compose.yml
+++ b/monero/docker-compose.yml
@@ -44,7 +44,7 @@ services:
     restart: unless-stopped
     stop_grace_period: 15m
     command: "${APP_MONERO_COMMAND}"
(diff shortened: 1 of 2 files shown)


kaichappel:
Thanks for the update @colder-is-better.

I tested this on a device and everything is working as expected. Fresh install, update, restart, and uninstall all looked good, and Monero came up on `0.18.5.0`.

One small thing came up while testing the testnet setting, but it looks like an existing upstream app issue rather than something from this update. I opened that here: https://github.com/deverickapollo/umbrel-monero/issues/160

Not holding this PR up.

colder-is-better:
@kaichappel @nmfretz Thank you for testing and merging -- and also for opening that upstream issue!

PR #5665, opened 2026-05-13
Update Fizzy

mauricewipf:
Release notes: https://github.com/basecamp/fizzy/releases


kaichappel:
Thanks @mauricewipf, I tested both the update path and a fresh install. Both worked as expected, including after restart.

Thanks for maintaining the image!

PR #5666, opened 2026-05-13
Update PeerSwap to v5.0.3

Impa10r:
Enable swaps with Strike node


kaichappel:
Thanks @Impa10r, I reviewed and tested the update and it looks good. Going live ⚡️

PR #5457, opened 2026-04-26
refactor(core-lightning-rtl): adopt CLNRest V3 provider contract

satwise:
## Summary

Refactor `core-lightning-rtl` to consume the CLNRest V3 provider contract (introduced in PR #5261) instead of hardcoding the CLN daemon's internal IP, port, and rune path.

## What changed

`core-lightning-rtl/docker-compose.yml`:

```diff
- LN_SERVER_URL: "https://$APP_CORE_LIGHTNING_DAEMON_IP:$CORE_LIGHTNING_REST_PORT"
- RUNE_PATH: "/root/.lightning/.commando-env"
+ LN_SERVER_URL: "${CLNREST_URL}"
+ RUNE_PATH: "${CLNREST_RUNE_PATH}"
```

## Why

Decouples RTL from CLN packaging internals. The provider contract becomes the single source of truth for CLNRest endpoint and credentials, so future changes to bind host, port, or rune location flow through automatically without RTL needing edits.

## Scope

- `core-lightning-rtl/docker-compose.yml` — 2 line change
- `core-lightning-rtl/umbrel-app.yml` — release notes bullet

## Dependency

Depends on **PR #5261** (provider contract exports). Without it, `${CLNREST_URL}` and `${CLNREST_RUNE_PATH}` are undefined.

## Related

- #4785 — Underlying packaging defect
- #4823 — RTL ECONNRESET (root-caused by same bind issue)
- PR #5261 — Provider contract fix (prerequisite)
- PR #5262 — LNbits for CLN (parallel consumer)

Issue #4823, opened 2026-02-20
bug(core-lightning): RTL ECONNRESET — CLNRest bound to unreachable daemon IP


satwise:
### Bug

RTL crashes with `ECONNRESET` when attempting to connect to CLNRest.

### Root Cause (Revised)

> **Note (May 2026):** The original framing that CLNRest was unreachable from other containers was retracted after @nmfretz confirmed in [#4785](https://github.com/getumbrel/umbrel-apps/issues/4785#issuecomment-4366406526) that `ports: 2107:2107` does publish the port on LAN. The ECONNRESET symptom is real; the underlying cause is more nuanced.

The actual root causes are:

1. **Daemon-IP bind creates a DR/HA race** - CLNRest binds to `\` (a dynamically-assigned container IP). On a fresh install or IP reassignment, lightningd may not have obtained that IP yet when CLNRest attempts to bind, surfacing as `EADDRNOTAVAIL` and causing the process to crash silently. RTL's ECONNRESET is a downstream symptom.

2. **Missing native V3 provider contract** - `core-lightning/exports.sh` does not export `CLNREST_HOST`, `CLNREST_URL`, or `CLNREST_RUNE_PATH`. Consumer apps (RTL, LNbits, Boltz) must carry Umbrel-specific V2->V3 translation internally rather than consuming canonical keys. Any mismatch or stale IP reference causes connection failures.

### Impact

- RTL cannot reliably connect to CLN after fresh install or container restart ordering races
- LNbits CLN backend cannot connect without app-internal translation shims
- Any consumer app requiring CLNRest lacks stable, canonical endpoint references

### Fix

**PR #5261** delivers the fix in `core-lightning/`:

1. **docker-compose.yml** - bind `--clnrest-host` to `0.0.0.0` (DR-safe, avoids EADDRNOTAVAIL race)
2. **exports.sh** - add native V3 exports: `CLNREST_HOST`, `CLNREST_URL`, `CLNREST_RUNE_PATH`\n3. **exports.sh** - add DNS-stable hostname exports: `APP_CORE_LIGHTNING_DAEMON_HOST`, `APP_CORE_LIGHTNING_APP_HOST`\n4. **docker-compose.yml** - pin `container_name` on lightningd, app, tor for startup dependency stability

### Related

- **PR #5261** - DR-safe CLNRest bind + native V3 provider contract (closes this issue on merge)
[...]

PR #5667, opened 2026-05-15
Update Core Lightning to 26.04.1

kaichappel:
Updates Core Lightning to `26.04.1` and the CLN Application to `26.04`.

Tested on an amd64 Umbrel device with Bitcoin Node. Core Lightning started cleanly, opened through Umbrel, and basic on-chain and Lightning wallet functionality worked as expected.

Startup logs looked normal, and the app did not hit the websocket localhost connection issue that can happen without the explicit websocket host env vars.

### DIFF
--- a/core-lightning/docker-compose.yml
+++ b/core-lightning/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       APP_HOST: $APP_CORE_LIGHTNING_IP
       APP_PORT: $APP_CORE_LIGHTNING_PORT
   app:
-    image: ghcr.io/elementsproject/cln-application:25.07.3@sha256:af22cebd21e1175651049d09cf2dd547da569fddd55e4c7766e5956bd47e91fa
+    image: ghcr.io/elementsproject/cln-application:26.04@sha256:78f92a973be7e756ad6b3325262ff689512bd2711c80fd5aac5e7f56eb9fb6d7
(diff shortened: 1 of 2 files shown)


nmfretz:
Hey @ShahanaFarooqui could you please give this a once-over when you have a chance?

@satwise thanks for digging into CLN and opening the related PRs/issues. I went through the Core Lightning-related pieces and here is where I think things stand:

- https://github.com/getumbrel/umbrel-apps/pull/5261 / https://github.com/getumbrel/umbrel-apps/issues/4785  
I don't think we actually need to add a new Core Lightning REST/provider contract as part of this update. From what I checked, the broader provider-contract change does not look necessary for normal Umbrel installs or recovery. What do you think @ShahanaFarooqui?

- https://github.com/getumbrel/umbrel-apps/pull/5457 / https://github.com/getumbrel/umbrel-apps/issues/4823  
I don't think the RTL change is needed based on what we have tested so far. If there is still a reproducible RTL failure after this Core Lightning update, we can look at that separately with logs.

If Shahana spots anything else in the Core Lightning setup, we can address it before merging. Otherwise I think this PR is the right focused update.

ShahanaFarooqui:
LGTM.

nmfretz:
Perfect, thanks @ShahanaFarooqui. Going live with this now .

PR #5685, opened 2026-05-22
App Submission: Hashrate Autopilot

rdouma:
# App Submission

### App name
Hashrate Autopilot

### 256x256 SVG icon
icon-256

### I have tested my app on:
- [ ] umbrelOS on Linux VM

---

Hashrate Autopilot rents hashrate on the Braiins Hashpower marketplace and routes it to a self-hosted Ocean non-custodial mining pool. Declares Bitcoin Node, Electrs, and Datum Gateway as dependencies; auto-discovers all three over Umbrel's shared Docker network, so first-boot setup only requires the operator's public stratum address (typically a DDNS hostname). Everything else is one click.

Previously distributed via a community app store at https://github.com/rdouma/hashrate-autopilot; the same daemon image (multi-arch amd64 + arm64, currently v1.8.5) has been running across the v1.5–v1.8 release series.


rdouma:
Regarding the "specify user: 1000:1000" notice: running as root here is deliberate, not an oversight. Non-root broke the SQLite bind-mount on Umbrel in past testing; documented in the Dockerfile.

I provided the icon in PNG format; I understood this was ok too? I don't have an SVG version.

hashrate-autopilot/umbrel-app.yml
```diff
@@ -0,0 +1,34 @@
...
+
+developer: Remco Douma
+website: https://github.com/rdouma/hashrate-autopilot
+dependencies:
+  - bitcoin
+  - electrs
+  - datum
+repo: https://github.com/rdouma/hashrate-autopilot
+support: https://github.com/rdouma/hashrate-autopilot/issues
+port: 3016
```
kaichappel: The app-store port `3016` conflicts with ChainForensics, so this needs a different `umbrel-app.yml` `port:` before merge. I re-checked the current app-store port inventory, and `3012` is free.

The compose `APP_PORT` can stay `3010`; this is only the Umbrel/app_proxy port.

```suggestion
port: 3012
```

hashrate-autopilot/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: hashrate-autopilot_web_1
+      APP_PORT: 3010
+      PROXY_AUTH_ADD: "false"
```
kaichappel: I noticed `PROXY_AUTH_ADD: "false"` disables Umbrel's outer auth for the app. In my test, the first-run setup flow was reachable before a dashboard password had been created, so anyone who can reach the app URL could get to that setup page and complete it.

Is that intentional, and does that behavior make sense for this app? If I'm missing part of the intended setup/auth flow, please push back.

kaichappel (changes requested):
Nice work @rdouma, and thanks for the submission. Sorry for the delay in reviewing.

I gave the Umbrel package a test and it's already looking good. It installs, the first-run setup flow completes, and the dashboard opens cleanly. I haven't tested live bidding yet.

There are a couple of things to look at before this can go live. I left comments inline.

rdouma:
Thanks for the review. Both addressed:

- Port — 3018. Your suggested 3012 turned out to also be taken (back-that-mac-up); 3018 is the first free port I found.
- `PROXY_AUTH_ADD`: You're right, dropped it. Wanted to avoid a potential double login but it's really not important and less secure.

Ready when you are for a re-test.

rdouma:
I just dutifully report every new version here, although at this point I'm starting to doubt it somehow someone is waiting for me and I misunderstood?

nmfretz:
Sorry for the delay @rdouma we're just busy at the moment! Gallery assets are now ready so I'll do a final pass before going live today!

rdouma:
Ah okay @nmfretz, thanks for the answer.

nmfretz:
Excellent work @rdouma! 

> Thanks for the review. Both addressed:
> 
> Port — 3018. Your suggested 3012 turned out to also be taken (back-that-mac-up); 3018 is the first free port I found.
> PROXY_AUTH_ADD: You're right, dropped it. Wanted to avoid a potential double login but it's really not important and less secure.

Good call adding back in the auth. Another bonus of doing that is users get 2FA for free for your app if they have it enabled on their Umbrel.

Gallery assets are done and we're going live. Welcome to the app store 🎉
image

rdouma:
Thanks! 🥳

PR #5695, opened 2026-05-26
Add ArchiveBox

Oni-giri:
# App Submission

### App name
ArchiveBox

### What is it?
ArchiveBox is a powerful, self-hosted internet archiving tool. You give it URLs and it captures permanent, offline snapshots of those pages, so you keep your copy even after the original goes offline, changes, or disappears behind a paywall.

Every snapshot is saved in multiple redundant formats at once: raw HTML, a SingleFile capture, a full-page screenshot, a printable PDF, a readability “reader view”, a WARC web archive, media downloaded via yt-dlp, plus extracted favicons, titles, and links. Everything lives as plain files on your disk, no vendor lock-in. URLs can be added one at a time, pasted as a list, or bulk-imported from bookmarks, Pocket, Pinboard, RSS, and more, and a built-in web UI lets you browse, full-text search, and manage the collection.

- Website: https://archivebox.io
- Source: https://github.com/ArchiveBox/ArchiveBox
- Docker image: `archivebox/archivebox:0.7.4` (pinned by digest)

### 256x256 SVG icon
icon

### Gallery images
archivebox-01-archive-the-web-1440x900
archivebox-02-save-every-format-1440x900
archivebox-03-import-urls-in-bulk-1440x900

### Notes for reviewers
- New app, single container, persists to `${APP_DATA_DIR}/data:/data`.
- Port `8006` chosen to avoid the existing conflict on `8000` (zeronote), happy to have it reassigned.
- An admin account (`admin` / `archivebox`) is auto-created on first launch; the manifest surfaces this as the default username/password and prompts the user to change it.
- `CSRF_TRUSTED_ORIGINS` is set to `http://${DEVICE_DOMAIN_NAME}:8006` to match the app's origin; users reaching it via LAN IP may need to add that origin.
- First boot runs DB init/migrations, so the UI takes a minute or two to come up, reflected in the health check `start_period`.

### I have tested my app on:
- [ ] umbrelOS on Linux VM

### DIFF
--- a/archivebox/docker-compose.yml
+++ b/archivebox/docker-compose.yml
@@ -1,0 +1,36 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: archivebox_archivebox_1
+      APP_PORT: 8000
+
+  archivebox:
+    image: archivebox/archivebox:0.7.4@sha256:1a5a37331091d9df865ead2b9c231aa5a892fc26fe0422ce6140d9e2d9532327
+    restart: on-failure
+    stop_grace_period: 1m
+    # The image's default command is `archivebox server --quick-init 0.0.0.0:8000`,
+    # which runs DB init/migrations on every boot and binds to all interfaces so the
+    # app_proxy can reach it. No manual `archivebox init` step is needed.
+    shm_size: "1gb" # headless Chrome needs a decent shared-memory pool while archiving
+    environment:
+      ALLOWED_HOSTS: "*"
+      # CSRF check for admin login / the add-URL form: must match the origin you open
+      # the app from. This covers the default umbrel.local hostname on the app's port;
+      # if you reach the app by LAN IP instead, add that origin here too (space-separated).
+      CSRF_TRUSTED_ORIGINS: "http://${DEVICE_DOMAIN_NAME}:8006"
+      # Admin user auto-created on first run only. Change the password after logging in.
+      ADMIN_USERNAME: admin
+      ADMIN_PASSWORD: archivebox
+      PUBLIC_INDEX: "True" # anyone on your network can browse the snapshot list
+      PUBLIC_SNAPSHOTS: "True" # ...and view archived snapshot content
+      PUBLIC_ADD_VIEW: "False" # ...but only logged-in admins can add new URLs
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+    healthcheck:
+      test: ["CMD-SHELL", "wget -q --spider http://127.0.0.1:8000/ || exit 1"]
+      interval: 30s
+      timeout: 10s
+      retries: 5
+      start_period: 180s
--- a/archivebox/docker-compose.yml
+++ b/archivebox/docker-compose.yml
@@ -22,7 +22,7 @@ services:
       CSRF_TRUSTED_ORIGINS: "http://${DEVICE_DOMAIN_NAME}:8006"
       # Admin user auto-created on first run only. Change the password after logging in.
       ADMIN_USERNAME: admin
-      ADMIN_PASSWORD: archivebox
+      ADMIN_PASSWORD: "${APP_PASSWORD}"
       PUBLIC_INDEX: "True" # anyone on your network can browse the snapshot list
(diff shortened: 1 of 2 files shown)


besoeasy:
wow this looks good will test today

archivebox/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
...
+dependencies: []
+repo: https://github.com/ArchiveBox/ArchiveBox
+support: https://github.com/ArchiveBox/ArchiveBox/issues
+port: 8006
+gallery: []
+path: ""
+defaultUsername: admin
+defaultPassword: archivebox
+submitter: Oni-giri
+submission: ""
```
kaichappel: ArchiveBox creates this admin user on first run, and users can change the password after logging in. Since this is the initial credential shown in Umbrel, would you be open to using Umbrel's generated per-install password here instead? That way each install gets its own initial admin password rather than sharing the same default.

The `submission` value should also point to this public PR.

If you want to go that route, this should do it:

```suggestion
defaultUsername: admin
deterministicPassword: true
submitter: Oni-giri
submission: https://github.com/getumbrel/umbrel-apps/pull/5695
```

archivebox/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+    shm_size: "1gb" # headless Chrome needs a decent shared-memory pool while archiving
+    environment:
+      ALLOWED_HOSTS: "*"
+      # CSRF check for admin login / the add-URL form: must match the origin you open
+      # the app from. This covers the default umbrel.local hostname on the app's port;
+      # if you reach the app by LAN IP instead, add that origin here too (space-separated).
+      CSRF_TRUSTED_ORIGINS: "http://${DEVICE_DOMAIN_NAME}:8006"
+      # Admin user auto-created on first run only. Change the password after logging in.
+      ADMIN_USERNAME: admin
+      ADMIN_PASSWORD: archivebox
```
kaichappel: If you do want to use the per-install password pattern above, `${APP_PASSWORD}` is the value Umbrel generates and shows to the user. Passing it here lets ArchiveBox create the first admin account with that same password.

```suggestion
      ADMIN_PASSWORD: "${APP_PASSWORD}"
```

archivebox/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+      PUBLIC_INDEX: "True" # anyone on your network can browse the snapshot list
+      PUBLIC_SNAPSHOTS: "True" # ...and view archived snapshot content
+      PUBLIC_ADD_VIEW: "False" # ...but only logged-in admins can add new URLs
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+    healthcheck:
+      test: ["CMD-SHELL", "wget -q --spider http://127.0.0.1:8000/ || exit 1"]
```
kaichappel: Small log cleanup: `wget --spider` sends a `HEAD` request, which is leaving repeated Django tracebacks in the app logs. This GET probe still checks the web server without that noise.

```suggestion
      test: ["CMD-SHELL", "wget -q -O /dev/null http://127.0.0.1:8000/public/ || exit 1"]
```

kaichappel (changes requested):
Nice work @Oni-giri, this is close.

I gave the Umbrel package a test and it's working nicely for me. I was able to log in and archive a URL without any issues.

Also, thanks for including the icon and screenshots in the PR body. We'll finalize those app-store assets on our side before this goes live.

I left a few small package comments inline: `submission` should point to this PR, there's an optional note about using Umbrel's generated password for the initial ArchiveBox admin user, and there's a small healthcheck cleanup to avoid noisy Django tracebacks in the logs.

Once those are sorted, I can re-test.

Oni-giri:
@kaichappel Resolved, thank you so much for the detailed walkthrough!

nmfretz:
Awesome, thanks @Oni-giri! I've tweaked a few things to get this finalized. 

- Removed unused CSRF_TRUSTED_ORIGINS config, since ArchiveBox doesn’t actaully read it.
- Clarified the description to avoid implying scheduled captures are available from the web UI.
- Updated the tagline to ArchiveBox’s official wording.
- Added gallery assets.

Going live! Welcome to the app store 🎉
image

PR #5698, opened 2026-05-27
Add LoneStrike — zero-fee solo Bitcoin mining pool

danhaus93-ops:
## What is this?

SoloStrike is a self-hosted solo Bitcoin mining pool for Umbrel, powered by the battle-tested ckpool-solo engine. It lets users connect their ASICs directly to their own Bitcoin node and keep 100% of every block found — no pool fees, no custody risk, no middleman.

The app has been running in a community store (<https://github.com/danhaus93-ops/solostrike-umbrel>) since v1.4 with daily users on multiple Umbrel deployments, currently at **v2.0.4** release for the official store.

## Submission checklist

- App ID is `solostrike`
- `port: 32221`
- `gallery: []` and `icon: ""` empty for first submission (per docs)
- `releaseNotes: ""` empty for first submission (per docs)
- All Docker images pinned by SHA256 digest, multi-arch (amd64 + arm64)
- Healthchecks configured for ckpool and api
- AGPL-3.0 licensed with separate trademark policy (in source repo)
- Full container hardening — every runtime container runs non-root (ckpool + api UID 1000, stunnel UID 65534/nobody, UI nginx-unprivileged UID 101), with `no-new-privileges: true` on all services and explicit `user:` directives in compose. The only root container is a one-shot init that fixes data-dir permissions and exits.

## What the user gets

- 🔒 **Private Mode** — disables third-party API calls when enabled
- 🔐 **Hardened by default** — fully non-root, no-new-privileges, SHA-pinned multi-arch images
- 🌐 **SoloStrike Pulse** — opt-in anonymous community census over nostr (globe / mesh / nonce-field views)
- 🏆 **Top Strikers** — crowdsourced tuning benchmarks inside Pulse: the most efficient real-world settings (frequency, core voltage, fan) for your exact miner model, ranked by sustained J/TH. Opt-in and anonymized — shares hardware model + performance numbers only, never wallet address, IP, hostname, or worker names. Gated behind a risk disclaimer because overclock settings can damage hardware.
- 🛡️ **Tor routing for Pulse** — broadcasts route through Umbrel’s tor_proxy
- 🎯 **The Reckoning** — strike forecast simulator with hashrate slider and global rank
- 🔥 **The Burn** — power-cost integration: net profit per strike, break-even rate
- ⚙ **Multi-port stratum** — 3333 ASICs, 3334 hobby miners, 4333 TLS via stunnel
- 🪨 **The Hunt** — block-potential card: subsidy, fee tiers, statistical earnings
- ⚡ **Firepower + Strike Velocity** — real-time hashrate and shares/sec
- 🖥️ **New desktop layout** — a dedicated multi-page desktop experience (The Field, Intel, Pool Internals, Luck & Analytics) with full-width charts, fleet table, and analytics cards, distinct from the mobile carousel
- 🌍 **19-language internationalization** — near-complete UI translation across 19 languages (desktop fully translated; mobile card mode still in progress)
- 🎨 **8 themes** including Paper Light + Performance Mode for low-power devices
- **Auto miner-type detection** (Bitaxe, NerdQaxe++, Antminer S19/S21, Avalon, Whatsminer, Braiins)

## Support

- Issues: <https://github.com/danhaus93-ops/solostrike-umbrel/issues>
- Community store (active): <https://github.com/danhaus93-ops/solostrike-umbrel>
- Tested on umbrelOS 1.x (x86 + arm64)
- License: AGPL-3.0 with separate trademark policy

## Notes for reviewers

Hardening is complete: all four runtime containers (ckpool, api, stunnel, ui) run non-root with explicit `user:` directives + no-new-privileges; images are SHA-pinned and multi-arch. The stratum ports (3333 / 3334 / 4333) are mapped to the host because external ASICs connect to them directly — that’s the one intentional port exposure.

**Top Strikers** (new in the 2.0.x line) is opt-in benchmark sharing that rides the **existing** nostr Pulse channel — it adds no new outbound endpoints, transmits only hardware model + tuning/performance numbers under a random handle, stays disabled unless the user turns on Pulse, and respects Private Mode and Tor routing exactly like the rest of Pulse.

Happy to adjust copy, gallery, port, or any field.

Thanks!

### DIFF
--- a/lonestrike/docker-compose.yml
+++ b/lonestrike/docker-compose.yml
@@ -226,7 +226,7 @@ services:
     # new release: rebuild, push, capture the new digest with
     # `docker inspect <image>:latest --format='{{index .RepoDigests 0}}'`,
     # and replace below.
-    image: ghcr.io/danhaus93-ops/solostrike-stunnel:v3.6.3@sha256:e5aad2b36614f2e4899aad5bb498007aaae90f4eb734ba8d2186448f12fd6bfa
+    image: ghcr.io/danhaus93-ops/solostrike-stunnel:v3.6.4@sha256:f057f7593e530f3c3eef8fd9004cd0d8bb2210d4cf80d5ecaf4a9d7e2e571c7d
     restart: on-failure
     security_opt:
       - no-new-privileges:true
@@ -246,7 +246,7 @@ services:
       - ${APP_DATA_DIR}/data/stunnel:/certs
 
   api:
-    image: ghcr.io/danhaus93-ops/solostrike-api:v3.6.3@sha256:6177fdd027a6c75078a68399e78404e2fa73668e006ca62164c3f284685d038a
+    image: ghcr.io/danhaus93-ops/solostrike-api:v3.6.4@sha256:3f380da8a518e9ad5b6af454a4d30cf5cfa503ca13807ee4b0320681a8396f83
     restart: on-failure
     security_opt:
       - no-new-privileges:true
@@ -293,7 +293,7 @@ services:
       - STRATUM_PORT_TLS=4333
 
   ui:
-    image: ghcr.io/danhaus93-ops/solostrike-ui:v3.6.3@sha256:22e697531598075405fa792346f61112c31f6aa306f2bff28fb5d8c23ef4b44f
+    image: ghcr.io/danhaus93-ops/solostrike-ui:v3.6.4@sha256:c2d377b44c95ef83d13e0259d8f249216cee213c1c509b430cb6cd49242eb046
     restart: on-failure
     security_opt:
       - no-new-privileges:true
--- a/lonestrike/docker-compose.yml
+++ b/lonestrike/docker-compose.yml
@@ -226,7 +226,7 @@ services:
     # new release: rebuild, push, capture the new digest with
     # `docker inspect <image>:latest --format='{{index .RepoDigests 0}}'`,
     # and replace below.
-    image: ghcr.io/danhaus93-ops/solostrike-stunnel:v3.6.5@sha256:fb0c3b5f202ba383c475fe638d8e53551268296a3ab4df0af37c8bd68e8a9835
+    image: ghcr.io/danhaus93-ops/solostrike-stunnel:v3.6.6@sha256:6c02cc3e2cc4f9982f96893ff9de14486c34c253642d9eaa4933ef4e2bc9b8dd
     restart: on-failure
     security_opt:
       - no-new-privileges:true
@@ -246,7 +246,7 @@ services:
       - ${APP_DATA_DIR}/data/stunnel:/certs
 
   api:
-    image: ghcr.io/danhaus93-ops/solostrike-api:v3.6.5@sha256:96d124842e1fd30f2b31d79435ac2eb5514263ec6c68fb9c4ad5957e10338ffe
+    image: ghcr.io/danhaus93-ops/solostrike-api:v3.6.6@sha256:1a833f3adcc5cc9187de0d603f9c8b41fc30da6aec6ed44cfb3bdc47b76dfd39
     restart: on-failure
     security_opt:
       - no-new-privileges:true
@@ -293,7 +293,7 @@ services:
       - STRATUM_PORT_TLS=32224
 
   ui:
-    image: ghcr.io/danhaus93-ops/solostrike-ui:v3.6.5@sha256:5c037099b7cecb09bb8188cc8b66d8910768d55d3010928137f1b1911c588356
+    image: ghcr.io/danhaus93-ops/solostrike-ui:v3.6.6@sha256:38bcd349c117c8372a32252a607055a7825252dca0e5f124d5b044163bfcf21b
     restart: on-failure
     security_opt:
       - no-new-privileges:true
--- a/lonestrike/docker-compose.yml
+++ b/lonestrike/docker-compose.yml
@@ -1,0 +1,306 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: danhaus93-solostrike_ui_1
+      # v1.11.57: UI runs nginx-unprivileged (non-root, UID 101) on 8080.
+      APP_PORT: 8080
+      # v1.8.4: narrow auth whitelist — exposes ONLY read-only stats and the
+      # home-screen widget endpoint without an Umbrel session token. The
+      # widget loader has no session, so without whitelisting /api/widget/*
+      # the home-screen widget shows dashes (this fixes the v1.7.7 widget
+      # bug for real). /metrics is whitelisted for Prometheus/Grafana.
+      # /api/state, /api/health*, and /api/stratum-health are read-only.
+      #
+      # NOT whitelisted (require Umbrel session):
+      #   /api/setup, /api/config, /api/reset-share-stats, /api/webhooks,
+      #   /api/network-stats/* (the POST endpoints under that prefix can
+      #   toggle the Pulse network broadcasting and need protection — and
+      #   the same data is exposed read-only inside /api/state anyway, so
+      #   nothing useful is lost by keeping the network-stats prefix gated).
+      #
+      # Important: Umbrel's app_proxy whitelist does PREFIX matching, so
+      # listing "/api/network-stats" would also expose "/api/network-stats/
+      # enable" etc. Each entry below is either an exact-match path or
+      # ends in /* to mean "this prefix is fine to expose entirely".
+      # v3.2.0: ONLY /metrics bypasses Umbrel auth. Previously /api/state,
+      # /api/health* and the widget endpoint were readable unauthenticated.
+      # The home-screen widget is fetched by umbreld directly at ui:8080
+      # (see `widgets.endpoint` in umbrel-app.yml), not via app_proxy, so it
+      # needs no whitelist entry.
+      PROXY_AUTH_WHITELIST: /metrics
+
+  # ═════════════════════════════════════════════════════════════════════
+  # init-permissions — one-shot install bootstrap (v1.11.55 chown, v1.11.56 conf-gen)
+  # ═════════════════════════════════════════════════════════════════════
+  # Runs as root, does TWO things needed for fresh-install correctness,
+  # then exits. ckpool/api/stunnel wait for this to complete successfully
+  # via depends_on: condition: service_completed_successfully.
+  #
+  # 1. CHOWNS /data to UID 1000 — runtime containers run as 1000 (security
+  #    hardening) but Umbrel creates host dirs as root. Without chown,
+  #    ckpool/api hit permission denied on first boot.
+  #
+  # 2. GENERATES /data/ckpool/config/ckpool.conf if missing — ckpool exits
+  #    immediately if its config doesn't exist. The upstream ckpool-solo
+  #    image has no entrypoint script that creates this file. Without
+  #    this step, fresh installs would crash-loop ckpool forever.
+  #
+  # The btcaddress field is a placeholder; in --btcsolo mode (see ckpool
+  # `command:` below) the actual payout address is taken from each miner's
+  # stratum username (e.g., "bc1q...workername"). The wizard collects the
+  # display address and stores it in /data/config/persist.json; if that
+  # file exists from a previous install, we use its address here for
+  # cosmetic consistency, but mining still works regardless.
+  #
+  # Container exits in ~1 second and shows as "Exited (0)" in docker ps -a.
+  # Runs on every app boot — idempotent, safe to re-execute.
+  init-permissions:
+    image: alpine:3.20@sha256:d9e853e87e55526f6b2917df91a2115c36dd7c696a35be12163d44e6e2a4b6bc
+    user: "0:0"
+    restart: on-failure
+    environment:
+      - BITCOIN_RPC_HOST=${APP_BITCOIN_NODE_IP}
+      - BITCOIN_RPC_PORT=${APP_BITCOIN_RPC_PORT}
+      - BITCOIN_RPC_USER=${APP_BITCOIN_RPC_USER}
+      - BITCOIN_RPC_PASS=${APP_BITCOIN_RPC_PASS}
+      - BITCOIN_ZMQ_HASHBLOCK=tcp://${APP_BITCOIN_NODE_IP}:${APP_BITCOIN_ZMQ_HASHBLOCK_PORT}
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+    # IMPORTANT: every shell variable below uses $$ (double-dollar), not $.
+    # Docker Compose interpolates $VAR / ${VAR} in the compose file — including
+    # inside this command script — at parse time. A single $ would be eaten by
+    # Compose (resolved to empty) before the container runs. $$ escapes to a
+    # literal $ so the in-container shell does the expansion. The container env
+    # vars (BITCOIN_RPC_*) are defined in the `environment:` block above and are
+    # expanded by the shell at runtime via $${...}.
+    command:
+      - sh
+      - -c
+      - |
+        set -e
+        echo "[init-permissions] starting"
+
+        # Step 1: ensure data dirs exist and are owned by the correct UIDs.
+        # ckpool + api run as UID 1000; stunnel runs as UID 65534 (nobody) and
+        # must own /data/stunnel to write its TLS cert. A blanket
+        # chown -R 1000:1000 /data (v1.11.55-57) broke stunnel's cert
+        # generation → TLS port 4333 down. Chown per-service instead.
+        mkdir -p /data/ckpool/config /data/ckpool/logs /data/config /data/stunnel
+        chown 1000:1000 /data
+        chown -R 1000:1000 /data/ckpool /data/config
+        chown -R 65534:65534 /data/stunnel
+        echo "[init-permissions] chowned: ckpool+config=1000, stunnel=65534"
+
+        # Step 2: render ckpool.conf FRESH from the committed template every run.
+        # Rationale: the Bitcoin RPC host/port/credentials and ZMQ endpoint are
+        # Umbrel-provided exports that can change across updates or node
+        # reconfiguration. The previous "generate once, preserve forever"
+        # approach left ckpool pinned to stale connection details if those
+        # exports ever changed. We now regenerate the connection fields each
+        # init from the live env, and preserve ONLY btcaddress (install state).
+        #
+        # btcaddress precedence: (1) api's saved config.json payoutAddress, else
+        # (2) the btcaddress already inside an existing ckpool.conf, else
+        # (3) the unspendable genesis placeholder. In --btcsolo mode the real
+        # payout comes from each miner's stratum username, so the placeholder is
+        # never used for real payouts.
+        CONF=/data/ckpool/config/ckpool.conf
+        TPL=/data/ckpool/config/ckpool.conf.template
+
+        BTC_ADDR=""
+        if [ -f /data/config/config.json ]; then
+          BTC_ADDR=$$(grep -oE '"payoutAddress"[[:space:]]*:[[:space:]]*"[^"]+"' /data/config/config.json 2>/dev/null | sed -E 's/.*"payoutAddress"[[:space:]]*:[[:space:]]*"([^"]+)".*/\1/' || true)
+        fi
+        if [ -z "$$BTC_ADDR" ] && [ -f "$$CONF" ]; then
+          BTC_ADDR=$$(grep -oE '"btcaddress"[[:space:]]*:[[:space:]]*"[^"]+"' "$$CONF" 2>/dev/null | sed -E 's/.*"btcaddress"[[:space:]]*:[[:space:]]*"([^"]+)".*/\1/' || true)
+        fi
+        if [ -z "$$BTC_ADDR" ]; then
+          BTC_ADDR="1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa"
+        fi
+
+        if [ ! -f "$$TPL" ]; then
+          echo "[init-permissions] template missing in /data — writing bundled default" >&2
+          printf '{\n  "btcd" : [\n    {\n      "url" : "__BITCOIN_RPC_HOST__:__BITCOIN_RPC_PORT__",\n      "auth" : "__BITCOIN_RPC_USER__",\n      "pass" : "__BITCOIN_RPC_PASS__"\n    }\n  ],\n  "btcaddress" : "__BTC_ADDR__",\n  "btcsig" : "LoneStrike on Umbrel/",\n  "blockpoll" : 50,\n  "update_interval" : 20,\n  "serverurl" : [\n    "0.0.0.0:3333",\n    "0.0.0.0:3334",\n    "0.0.0.0:4334"\n  ],\n  "mindiff" : 1,\n  "startdiff" : 10000,\n  "maxdiff" : 0,\n  "highdiff" : 500000,\n  "logdir" : "/var/log/ckpool",\n  "zmqblock" : "__BITCOIN_ZMQ_HASHBLOCK__"\n}\n' > "$$TPL"
+        fi
+
+        # Substitute placeholders with live values using awk index()/substr()
+        # LITERAL splicing (NOT sed or gsub): the RPC password is Umbrel
+        # auto-generated and may contain | & / \ which corrupt regex-based
+        # replacers. index()/substr() treats the replacement as an opaque
+        # literal. Tmp file + atomic mv so ckpool never reads a half-written conf.
+        TMP=$$(mktemp)
+        awk \
+          -v rpc_host="$${BITCOIN_RPC_HOST}" \
+          -v rpc_port="$${BITCOIN_RPC_PORT}" \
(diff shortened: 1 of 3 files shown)


danhaus93-ops:
Thanks for the linter feedback! Submission URL, icon field, and restart policies all updated.

Regarding the info-level items:

• **Mount paths:** The `data/ckpool/{config,logs}`, `data/stunnel`, and `data/config` directories are created on first install by a dedicated one-shot `init-permissions` container, which runs before the long-running services start. The linter sees them as missing because no install has happened yet against its checkout — that's expected.

• **External port mappings (3333/3334/4333):** Required for stratum protocol — miners (ASICs, BitAxe, NerdQaxe++) connect directly to these ports from the LAN. Cannot route through `app_proxy` because stratum is a raw TCP protocol, not HTTP. 3333 is the de facto standard mining port across the industry.

• **Unsafe user warnings:** All four runtime containers (ckpool, api, stunnel, ui) now have explicit `user:` directives in `docker-compose.yml` as of v1.11.61 (ckpool + api UID 1000, stunnel UID 65534/nobody, ui UID 101) — the linter notices should clear on the next run.

SoloStrike is designed to coexist alongside the existing Public Pool app — different engine (ckpool-solo vs custom Node.js), different audience (analytics/privacy/TLS vs simplest-possible-setup), so users have a choice depending on what they value. Happy to elaborate on any specific differences if it'd help review.

Looking forward to feedback whenever you have time — thanks for maintaining this store!

nmfretz:
Thanks for the submission @danhaus93-ops! I'm working my way through and will get to this soon.

danhaus93-ops:
> Thanks for the submission @danhaus93-ops! I'm working my way through and will get to this soon.

Thank you for the response, looking forward to the review.

solostrike/docker-compose.yml
```diff
@@ -0,0 +1,272 @@
...
+    # for stratum traffic — same goal as v1.11.12, working implementation.
+    healthcheck:
+      test: ["CMD-SHELL", "test -f /tmp/ckpool/main.pid && test -S /tmp/ckpool/listener"]
+      interval: 5s
+      timeout: 3s
+      retries: 12
+      start_period: 30s
+    ports:
+      - "3333:3333"
+      - "3334:3334"
```
nmfretz: These host ports conflict with existing apps in the repo (`3333` is used by bleskomat-server, `3334` by ghostfolio). The container ports can stay the same; only the LAN-facing published ports need to move.

```suggestion
      - "32222:3333"
      - "32223:3334"
```

solostrike/docker-compose.yml
```diff
@@ -0,0 +1,272 @@
...
+    restart: on-failure
+    security_opt:
+      - no-new-privileges:true
+    # v1.8.4: wait for ckpool to be healthy. stunnel forwards to ckpool:3333
+    # so connecting to a half-booted ckpool would drop early SSL miners.
+    depends_on:
+      ckpool:
+        condition: service_healthy
+    ports:
+      - "4333:4333"
```
nmfretz: Same here: keep the container port as `4333`, but publish it on an available app port.

```suggestion
      - "32224:4333"
```

solostrike/docker-compose.yml
```diff
@@ -0,0 +1,272 @@
...
+      - BITCOIN_ZMQ_HASHBLOCK=tcp://${APP_BITCOIN_NODE_IP}:${APP_BITCOIN_ZMQ_HASHBLOCK_PORT}
+      - POOL_SIGNATURE=SoloStrike on Umbrel/
+      - START_DIFFICULTY=10000
+      - MIN_DIFFICULTY=1
+      - MAX_DIFFICULTY=0
+      - BLOCKPOLL=50
+      - UPDATE_INTERVAL=20
+      - STRATUM_PORT=3333
+      - STRATUM_PORT_HOBBY=3334
+      - STRATUM_PORT_TLS=4333
```
nmfretz: If these env vars are used for user-facing miner connection details, they should match the published host ports. Internal API-to-service checks should still use Docker service ports like `ckpool:3333`, `ckpool:3334`, and `stunnel:4333`.

```suggestion
      - STRATUM_PORT=32222
      - STRATUM_PORT_HOBBY=32223
      - STRATUM_PORT_TLS=32224
```

solostrike/docker-compose.yml
```diff
@@ -0,0 +1,272 @@
...
+
+  ui:
+    image: ghcr.io/danhaus93-ops/solostrike-ui:v2.0.4@sha256:8a4a130d87301e981cdabd885715a8867fb8afbd5ec83ee4dc9b5e6c3463645d
+    restart: on-failure
+    security_opt:
+      - no-new-privileges:true
+    container_name: danhaus93-solostrike_ui_1
```
nmfretz: This hardcoded container name can be removed so the app follows Umbrel’s normal Compose naming.

```suggestion
```

solostrike/umbrel-app.yml
```diff
@@ -0,0 +1,140 @@
...
+
+  • CSV export for workers and found blocks
+
+  • Webhooks — POST events to Discord, ntfy.sh, Home Assistant, Telegram bots, etc.
+
+  • Prometheus /metrics endpoint for Grafana, Home Assistant, and custom monitoring
+
+  • Public read-only API — share your pool stats externally
+
+  • Dual stratum ports — 3333 for ASICs, 3334 for hobby miners (BitAxe, NerdQaxe++)
```
nmfretz: This should match the published miner ports from the compose file.

```suggestion
  • Dual stratum ports — 32222 for ASICs, 32223 for hobby miners (BitAxe, NerdQaxe++), plus TLS on 32224
```

solostrike/umbrel-app.yml
```diff
@@ -0,0 +1,140 @@
...
+        - title: "Pool Hashrate"
+          text: "542.8"
+          subtext: "GH/s"
+        - title: "Workers"
+          text: "3"
+        - title: "Blocks Found"
+          text: "0"
+        - title: "Best Diff"
+          text: "1.2B"
+icon: ""
```
nmfretz: This widget can use the current Umbrel widget shape. The endpoint returns the `four-stats` payload, so the unused `items` / `jsonPath` mapping is not needed in the manifest.

```suggestion
widgets:
  - id: "stats"
    type: "four-stats"
    name: "SoloStrike Stats"
    refresh: "10s"
    endpoint: "ui:8080/api/widget/four-stats"
    link: ""
    example:
      type: "four-stats"
      link: ""
      items:
        - title: "Pool Hashrate"
          text: "542.8"
          subtext: "GH/s"
        - title: "Workers"
          text: "3"
        - title: "Blocks Found"
          text: "0"
        - title: "Best Diff"
          text: "1.2B"
```

nmfretz (changes requested):
Sorry for the wait @danhaus93-ops. Excited to see another solo mining app! I took a look through the packaging and there are a few Umbrel-specific changes needed before I can properly test this.

solostrike/docker-compose.yml
```diff
@@ -0,0 +1,272 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: danhaus93-solostrike_ui_1
```
nmfretz: Umbrel will run this app under the `solostrike` app id, so app_proxy should target the service-derived container name.

```suggestion
      APP_HOST: solostrike_ui_1
```

solostrike/docker-compose.yml
```diff
@@ -0,0 +1,272 @@
...
+      #   /api/network-stats/* (the POST endpoints under that prefix can
+      #   toggle the Pulse network broadcasting and need protection — and
+      #   the same data is exposed read-only inside /api/state anyway, so
+      #   nothing useful is lost by keeping the network-stats prefix gated).
+      #
+      # Important: Umbrel's app_proxy whitelist does PREFIX matching, so
+      # listing "/api/network-stats" would also expose "/api/network-stats/
+      # enable" etc. Each entry below is either an exact-match path or
+      # ends in /* to mean "this prefix is fine to expose entirely".
+      PROXY_AUTH_WHITELIST: /api/widget/four-stats,/api/health,/api/health/detailed,/api/state,/api/stratum-health,/metrics
```
nmfretz: The widget endpoint itself does not need to be whitelisted for Umbrel's home screen widget; umbreld fetches widget data internally from the endpoint in `umbrel-app.yml`. Keeping the other public read-only/metrics routes intact, this can drop just the widget route.

```suggestion
      PROXY_AUTH_WHITELIST: /api/health,/api/health/detailed,/api/state,/api/stratum-health,/metrics
```

danhaus93-ops:
Okay I applied all the suggested changes.

nmfretz:
Awesome, thanks @danhaus93-ops! Is everything working okay for you before I go through again?

danhaus93-ops:
> Awesome, thanks @danhaus93-ops! Is everything working okay for you before I go through again?

I believe so. Should I do a fresh install to see? I click on the apply changes on everything you marked and suggested.  

However it appears to have applied to just the submitted docker-compose and umbrel-app. 

My main repo is still untouched. When merged will the changes apply?

A bit new to this and Github.

nmfretz:
> I believe so. Should I do a fresh install to see? I click on the apply changes on everything you marked and suggested.

Ya, please do! My review here is primarily from an app store packaging perspective. I'll be able to catch obvious issues when checking that the app installs/restarts/persists data etc, but you'll definitely want to confirm for yourself that whatever functionality you've implemented is working as you expect.

> My main repo is still untouched. When merged will the changes apply?
>
> A bit new to this and Github.

No worries. When you clicked “Apply suggestion” here on GitHub, GitHub committed those changes to this PR branch online.

So the PR has the changes now, and if this PR is merged, those changes will be included in the Umbrel app store package.

Your local files on your computer will not update automatically though. To see the same changes locally, you’ll need to pull/sync this PR branch in your local `umbrel-apps` fork. Since Git can be a bit confusing here, you can ask your coding assistant/LLM to help you sync your local `add-solostrike` branch with the latest version from GitHub.

Let me know if anything is confusing!

lonestrike/docker-compose.yml
```diff
@@ -0,0 +1,271 @@
...
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+    # IMPORTANT: every shell variable below uses $$ (double-dollar), not $.
+    # Docker Compose interpolates $VAR / ${VAR} in the compose file — including
+    # inside this command script — at parse time. A single $ would be eaten by
+    # Compose (resolved to empty) before the container runs. $$ escapes to a
+    # literal $ so the in-container shell does the expansion. The container env
+    # vars (BITCOIN_RPC_*) are defined in the `environment:` block above and are
+    # expanded by the shell at runtime via $${...}.
+    command:
```
nmfretz, line 78: Also, just took a look at the init script in the compose, and I think we can make this easier to maintain by moving the generated ckpool config into a template file.

This works as-is, so this isn’t a syntax issue. The concern is that `ckpool.conf` contains Umbrel-provided Bitcoin RPC/ZMQ values, but the script preserves the whole file forever once it exists. If those exports change later, ckpool can keep using stale connection details.

A cleaner shape would be:

- add `solostrike/data/ckpool/config/ckpool.conf.template`
- render `ckpool.conf` from that template during init
- preserve only `btcaddress` from `/data/config/config.json` or the existing `ckpool.conf`
- regenerate the Bitcoin RPC/ZMQ fields each init run
- keep the existing ownership/chmod setup

That also matches the pattern used by `bassin`, which has a committed `ckpool.conf.template` and renders it during init.

danhaus93-ops: Okay, sounds good!

danhaus93-ops: Thanks for the review so far! Two logistics questions for once this merges:

1. Existing community users SoloStrike has been running as a community-store app (`danhaus93-solostrike`) with active users. After the official `solostrike` app goes live, is there a preferred path for moving existing installs over (fresh install vs. any supported data-dir migration), or should I just point users to install the official app fresh?

2. Future updates for shipping new versions of the official app post-merge, I'll open a new PR against "solostrike/" each time with a version bump + re-pinned image digests is that the correct/expected flow, or is there anything else you'd want on update PRs?

One heads-up while it's still in review: the current compose exposes the ASIC/hobby/TLS stratum ports but not the high-difficulty braiins, NiceHash port (container 4334). If that should be exposed too, it's easy to add now happy to push a fix up commit if you'd like it in.

Thanks!

nmfretz:
> Existing community users SoloStrike has been running as a community-store app (danhaus93-solostrike) with active users. After the official solostrike app goes live, is there a preferred path for moving existing installs over (fresh install vs. any supported data-dir migration), or should I just point users to install the official app fresh?

Ya good question. Since the app id changes from `danhaus93-solostrike` to `solostrike`, umbrelOS will treat the official app as a separate install with a separate data directory. There won’t be an automatic migration. You'll want to explicitly test this, but this is a kind of general migration you could tell users to do:

- Back up first (not necessary, but just in case)
- From the umbrelOS dashboard, right-click the community SoloStrike app icon and stop it
- Install the official SoloStrike app from the app store
- Right-click the official SoloStrike app icon and stop it too
- Open Files and drag `Apps/danhaus93-solostrike/<whatever-data-you-or-directory-you-need>` into `Apps/solostrike/data/<or-whatever>`, choosing overwrite if prompted.
- Start the official SoloStrike app and confirm everything works.
- Re-point miners to the official app’s published ports.
- Only delete the community app once the official install is confirmed working.

> Future updates for shipping new versions of the official app post-merge, I'll open a new PR against "solostrike/" each time with a version bump + re-pinned image digests is that the correct/expected flow, or is there anything else you'd want on update PRs?

Yep, updates should be a new PR against `solostrike/`. Just bump the app `version` in `umbrel-app.yml`, update `releaseNotes`, and update/re-pin any image tags and digests.

> One heads-up while it's still in review: the current compose exposes the ASIC/hobby/TLS stratum ports but not the high-difficulty braiins, NiceHash port (container 4334). If that should be exposed too, it's easy to add now happy to push a fix up commit if you'd like it in.

That’s totally up to you!

lonestrike/docker-compose.yml
nmfretz: no worries! answered below https://github.com/getumbrel/umbrel-apps/pull/5698#issuecomment-4724798264

danhaus93-ops:
Awesome, Im going to work on pushing an update and testing the changes.

danhaus93-ops:
Just want to make sure . I am able to push all the recommended changes to my community main repo?

The community app ID can be changed to the recommended without affecting the merge?

danhaus93-ops:
Quick heads-up and a question before this goes further.

After opening this PR I discovered there's another Bitcoin solo-mining project using a very similar name "SoloStrike" with a live site (solostrike.io) and a separate Umbrel community app store. As far as I can tell the naming overlap is coincidental (it's a fairly generic "solo" + "strike" combination), and the two projects are unrelated, but I wanted to flag it proactively rather than have it surface after a merge.

To avoid any confusion in the official store, I'm prepared to rename my app before this is accepted. How would you prefer I handle this?

Rename now and update this PR (new app id + display name), or
Keep this PR as-is and you'd rather advise on a different approach.
I'd rather get the naming right up front than create confusion for users or for the store. Happy to make whatever change you think is cleanest. Thanks for taking the time to review.

nmfretz:
> Just want to make sure . I am able to push all the recommended changes to my community main repo?
>
> The community app ID can be changed to the recommended without affecting the merge?

Ah actually please be very careful with your community app because existing installs are involved. You shouldn't change the existing `id: danhaus93-solostrike` because umbrelOS uses the app id as the app identity, including the app-data directory, compose project name, and generated container names.

Also, for community app stores, the app id must start with the app store id. Your community store has `id: danhaus93` in `umbrel-app-store.yml`, so `danhaus93-solostrike` is the correct id for that store. If you changed it to `solostrike`, existing users would not get an update to the same app. umbrelOS would treat it as a different app with separate data.

Some packaging fixes are safe to copy back to your community app, but only the ones that don’t change the existing app identity or user-facing behavior. In general, safe changes are things like manifest cleanup, widget metadata fixes, comments, or internal packaging cleanup that keeps the same app id, data paths, generated container names, and published ports. Be much more careful with anything that changes the app id, app directory name, `${APP_DATA_DIR}` layout, container names used by `APP_HOST`, or published ports.
  
 > Quick heads-up and a question before this goes further.
> 
> After opening this PR I discovered there's another Bitcoin solo-mining project using a very similar name "SoloStrike" with a live site (solostrike.io) and a separate Umbrel community app store.

Oh good catch on that. Ya pls go ahead and update this PR now with the new name and app ID. We can't change the app ID after submission.

Also, I just added some agent skills to this repo to help make app submissions easier. Feel free to point your coding agent at [AGENT.md](https://github.com/getumbrel/umbrel-apps/blob/master/AGENTS.md)!

danhaus93-ops:
Okay great, I have went ahead and renamed my app "LoneStrike"
Ill update the PR soon!

Thanks for all your help!

nmfretz:
Perfect, ping me when you're ready!

danhaus93-ops (commented):
Okay, I made the new lonestrike folder. Looks like I lost the comment string.

danhaus93-ops:
Is this still being reviewed for merging?

danhaus93-ops:
Not sure whats going on. Its been two weeks without any response. Should we just close the merge request?

nmfretz (changes requested):
Sorry for going quiet here, @danhaus93-ops. I’ve had my head down elsewhere working on umbrelOS after getting back. Thanks for bearing with me.

I’ve now taken another thorough pass over the current submission. On the packaging side, there are a few Umbrel-specific things that still need tightening:

- The UI references the container name used by the community app store package. That name changes when the official app ID is `lonestrike`, so the UI cannot reliably reach the API.
- The TLS proxy uses a generic `ckpool` hostname. Because Umbrel apps share a Docker network, that could resolve to a CKPool container belonging to another installed app. It needs a LoneStrike-specific service name.
- Only the metrics endpoint should bypass Umbrel authentication, and `/metrics` needs to return the actual Prometheus output rather than the UI.
- The generated CKPool configuration needs to work across fresh installs and future updates while preserving user settings.
- The widget, remaining SoloStrike branding, certificate identity, and PR description should all match the current LoneStrike package.

I also had Codex dig through the upstream images as part of this review. It surfaced two issues that cannot be fixed in the package alone:

- The API does not enforce server-side authentication. Umbrel’s app proxy protects browser access, but another app container can reach the API directly and read or change LoneStrike configuration.
- Private Mode still makes external Nostr and registry connections, and enabling it may not stop Pulse if Pulse was previously enabled. That does not match the current description of fully local operation.

Can you please sync this branch with the latest `master` before continuing? I've added `AGENTS.md` along with dedicated `umbrel-package-app` and `umbrel-test-app` skills under `.claude/skills/`. If you’re using a coding agent, I strongly recommend having it use these instructions. They cover the current Umbrel packaging and testing requirements and should be very helpful for working through this.

After the changes are in, please fresh-install the exact `lonestrike` package from this PR and check:

1. The app opens and onboarding completes successfully.
2. Miners can connect through all four ports, including TLS.
3. Settings and the TLS certificate persist after restarting the app.
4. Private Mode stops the external connections described above.
5. Another app container cannot access protected API functionality.
6. The containers settle without recurring errors and `umbrel-cli lint lonestrike` passes.

I know this is a decent-sized round of changes. Once everything is updated and you have tested the exact submission, I’ll be happy to take another look.

danhaus93-ops:
Thanks for the thorough review  all seven items are resolved. The branch is now a single commit on top of current master (per your sync request), and the package is at v3.6.3, which shipped these fixes incrementally to the community store where they’ve been field-tested by real deployments for the past releases.

1. Hardcoded container name (UI→API): nginx.conf is now a template; API_HOST is injected via compose env. No hardcoded names remain in any image.

2. stunnel ckpool alias: same treatment stunnel.conf.template renders @CKPOOL_HOST@ from env at start, fails loudly if unset. Verified end-to-end: openssl s_client stratum subscribe through the TLS port returns mining.notify from the correct pool.

3. /metrics: dedicated location = /metrics route; PROXY_AUTH_WHITELIST tightened to exactly /metrics. Returns real Prometheus text, not the SPA.

4. ckpool conf across installs/updates: the conf is treated as disposable output, regenerated at every container start. Connection details (RPC host/port/credentials, ZMQ) re-resolve from Umbrel’s live exports each time, so node reconfiguration can’t strand ckpool on stale values. Only btcaddress is preserved, with precedence config.json → existing conf → unspendable placeholder (safe in --btcsolo, where payouts derive from miner usernames). The render is atomic (tmp+mv), the template self-heals if missing, and substitution uses literal splicing rather than regex so auto-generated credentials containing metacharacters can’t corrupt the conf.

5. Cert identity: fresh installs generate CN=lonestrike.local / O=LoneStrike; existing installs correctly reuse their persisted cert.

6.  Server-side API auth: every /api/* route and the WebSocket now require a per-install key (generated at first boot, timing-safe comparison; exempt: healthcheck, widget endpoint, one-time TOFU claim). The first browser pairs automatically; additional devices pair via a dedicated UI. Your test: docker exec <any other app container> wget -qO- http://lonestrike_api_1:3001/api/config → 401.

7. Private Mode: now a master kill-switch gates every relay connection, registry fetch, price/fees/blocks poll, and webhook delivery, and tears down live sockets the moment it’s toggled (no restart). Verified: with Private Mode on, netstat inside the api container shows zero external connections only the local node’s RPC/ZMQ and LAN miner polling. v3.6.x additionally adds per-service privacy toggles beneath the master switch; the master always wins and is checked first in every gate.

Happy to address anything further and thanks again, the review made the app materially better.

danhaus93-ops:
@nmfretz Lint is green and the comment above covers all seven review items with verification steps ready for re-review whenever you are. Thanks!

nmfretz (changes requested):
Thanks for working through these updates, @danhaus93-ops. Most of the earlier points are now addressed, but the UI still shows the wrong miner ports.

The official package exposes `32222`, `32223`, `32224`, and `32225`, while the UI shows `3333`, `3334`, `4333`, and `4334`. The current images do not use the new `STRATUM_PORT*` values, so miners following the in-app instructions will not connect.

Can you update the upstream UI, publish new images, and test the exact PR package on umbrelOS by connecting a miner and confirming shares appear? Once that works, I can take another look.

danhaus93-ops:
@nmfretz Okay, ill get it fixed and ping you.

danhaus93-ops:
Thanks @nmfretz the ports are fixed and I've verified it end to end on umbrelOS.

The PR's `docker-compose.yml` was pinning the `ui`/`api`/`stunnel` images at `v3.6.4`, which predate the port fix — so the images genuinely weren't using the `STRATUM_PORT*` values yet. `v3.6.5` adds an `/api/ports` endpoint (the API resolves `STRATUM_PORT*` from the compose env at runtime) and the UI fetches it before first render, so the Stratum card renders whatever host ports the package maps instead of baked-in defaults.

Pushed to `add-solostrike`:
- Bumped `ui`/`api`/`stunnel` to `v3.6.5` (pinned digests).
- Fixed `STRATUM_PORT_TLS`, which was still `4333` — now `32224` to match the published host port.
- Moved the app port `32221` → `32226`, since `32221` collides with the community install; this lets the two run side by side during the migration path in the release notes.

Verified on umbrelOS: installed the exact PR package, pointed a miner at `32222`. The Stratum Connection card now reads `32222 / 32223 / 32224 / 32225`, the worker connects, and shares are landing.

danhaus93-ops:
@nmfretz Follow-up: repinned to v3.6.6 during testing I found the footer port-status indicators were keyed to internal container ports, so they showed ‘unknown’ on the official port scheme. v3.6.6 keys them to the host-facing STRATUM_PORT* values. Re-verified on the PR package: card and app ui footer shows 32222/32223/32224, status green, miner connected, shares landing.

IMG_6918
IMG_6919

danhaus93-ops:
@nmfretz Heads-up: bumped the package to v3.7.2. One feature added since the v3.6.6 port-status fix: a Best Diff Pulse tab that ranks each pool’s highest-difficulty accepted share read from the pool’s own sharelog before broadcast, validated and monotonically merged on receipt. No new endpoints, no new outbound connections; it rides the existing opt-in Pulse event and respects Private Mode like everything else. No packaging changes: same services, ports, mounts, and env as v3.6.6 only the image digests moved. Re-verified the exact PR package on umbrelOS: install, miner on 32222, shares landing, port card and status lights correct.

nmfretz:
Thanks @danhaus93-ops. The main card and mobile footer are fixed now, but I found two more places still using the old ports: the onboarding wizard and the default desktop dashboard.

Can you fix those too, then run through the exact PR package from a clean browser on desktop and connect a miner using the URL the app gives you?

I’m mainly reviewing the Umbrel packaging here. I’ll flag app issues when I spot them, but you’ll need to make sure the complete app experience works for someone installing it fresh before it goes into the official store. If shares show up, we should be ready for finalization.

danhaus93-ops:
@nmfretz Fixed both the onboarding wizard and the desktop dashboard now render the install-aware ports from the same `/api/ports` source as the main card (v3.7.5, digests repinned on the branch).

Re-verified the exact PR package per your checklist: clean browser, fresh install, completed onboarding, connected a miner using the URL the wizard displays shares confirmed. Screenshots attached. Ready for another look!

danhaus93-ops:
@nmfretz We all good to go now?

Issue #4810, opened 2026-02-17
[App Update] Migrate File Drop to Originless - Optimized Codebase with 95% Bandwidth Reduction


besoeasy:
## Summary

The file-drop app should be migrated to the new **Originless** repository. The original file-drop repository now redirects to Originless, which features a completely optimized codebase with significant performance improvements.

**Old Repository:** https://github.com/besoeasy/file-drop (redirects to new repo)  
**New Repository:** https://github.com/besoeasy/Originless

### 1. Light Mode Implementation
- **95% bandwidth reduction** compared to the previous version
- Much lighter resource usage on Umbrel
- Faster sync times

### 2. Optimized Codebase
- Complete rewrite with performance optimizations
- Improved IPFS node efficiency
- Better resource management

### 3. Enhanced Features
- Renamed from "File Drop" to "Originless" 
- More comprehensive documentation
- Better API integration examples
- AI agent integration guide

### Update `umbrel-app.yml`

```yaml
manifestVersion: 1
id: originless  # Updated from file-drop
name: Originless  # Updated from File Drop
version: "0.0.7"  # Latest version
tagline: "Private, decentralized file sharing for Nostr and the web"
category: files
description: "One all-in-one storage backend you can drop into anything: your own apps, screenshot tools, pastebin-style pastes, Nostr clients, Reddit posts, forum embeds—anything that needs durable, anonymous file hosting. Built on IPFS with 95% bandwidth reduction compared to previous versions."
releaseNotes: >-
  Complete rewrite with optimized codebase. 95% bandwidth reduction through light mode implementation. Enhanced performance and faster IPFS syncs.
developer: "besoeasy"
website: "https://github.com/besoeasy/Originless"
repo: "https://github.com/besoeasy/Originless"
support: "https://github.com/besoeasy/Originless/issues"
port: 3232
gallery:
  - 1.jpg
  - 2.jpg
  - 3.jpg
dependencies: []
path: "/"
defaultUsername: ""
defaultPassword: ""
submitter: besoeasy
submission: https://github.com/getumbrel/umbrel-apps/pull/2392
```

### Update `docker-compose.yml`

```yaml
version: "3.7"
services:
  app_proxy:
    environment:
      APP_HOST: originless_app_1
      APP_PORT: 3232

  app:
    image: ghcr.io/besoeasy/originless:0.0.7@sha256:d4eec165c996cebc502b0e31babf0e833c88c78389962698e2c67fb24f0a395e
[...]

PR #5709, opened 2026-05-31
Replace file-drop with originless

besoeasy:
As discussed here https://github.com/getumbrel/umbrel-apps/issues/4810

### DIFF
--- a/file-drop/umbrel-app.yml
+++ b/file-drop/umbrel-app.yml
@@ -4,7 +4,11 @@ name: File Drop
 version: "1.4.3"
 tagline: "Decentralized, Anonymous, and Secure File Sharing"
 category: files
-description: "File Drop is an open-source, decentralized application for anonymously sharing images, videos, and other files. It uses peer-to-peer protocols and the InterPlanetary File System (IPFS) to enable direct, serverless file transfers, enhancing privacy and security"
+description: >-
+  ⚠️ Removal Notice: The Samba app has been removed from the Umbrel App Store as file sharing is now natively integrated into umbrelOS.
+
+
+  File Drop is an open-source, decentralized application for anonymously sharing images, videos, and other files. It uses peer-to-peer protocols and the InterPlanetary File System (IPFS) to enable direct, serverless file transfers, enhancing privacy and security"
 releaseNotes: >-
   This update improves File Drop's IPFS backend with low-power DHT routing,
   persistent IPFS data storage, and a 5 GB default upload limit. It also updates
--- a/file-drop/umbrel-app.yml
+++ b/file-drop/umbrel-app.yml
@@ -1,4 +1,4 @@
-manifestVersion: 1.1
+manifestVersion: 1.2
 id: file-drop
 name: File Drop
 version: "1.4.3"
--- a/file-drop/umbrel-app.yml
+++ b/file-drop/umbrel-app.yml
@@ -24,3 +24,4 @@ defaultUsername: ""
 defaultPassword: ""
 submitter: besoeasy
 submission: https://github.com/getumbrel/umbrel-apps/pull/2392
+disabled: true
--- a/originless/docker-compose.yml
+++ b/originless/docker-compose.yml
@@ -14,7 +14,7 @@ services:
       - "4001:4001/tcp"
       - "4001:4001/udp"
     volumes:
-      - ${APP_DATA_DIR}:/data
+      - ${APP_DATA_DIR}data:/data
     environment:
       - NODE_ENV=production
       - STORAGE_MAX=${STORAGE_MAX:-20GB}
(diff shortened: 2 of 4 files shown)


besoeasy:
@nmfretz i need help with this one, app is same, it avoids a copyright.

renamed from file-drop to originless

kaichappel:
Thanks @besoeasy, that makes sense.

Just want to double-check the tradeoff before we go this route. Adding Originless as a new app and disabling File Drop is clean if the `file-drop` app id/slug itself needs to go away, but existing File Drop users would not be moved to Originless through a normal update. They would keep their existing File Drop app/data, and Originless would be a separate install with separate app data.

The alternative would be to keep the existing `file-drop` app id, change the visible app name to Originless, and update the image/links in place. That would preserve the normal update path for existing users, but Umbrel would still use `file-drop` internally.

Are you sure the internal `file-drop` app id needs to be retired too, not just the public app name?

besoeasy:
yes file-drop can be let go

besoeasy:
@al-lac i need little help with this

file-drop/umbrel-app.yml
al-lac: Bumping to `1.2` also disables the app on older umbrelOS versions.
```suggestion
manifestVersion: 1.2
```

al-lac: Add a message to inform users that the app has been disabled.
```suggestion
description: >-
  ⚠️ Removal Notice: The Samba app has been removed from the Umbrel App Store as file sharing is now natively integrated into umbrelOS.


  File Drop is an open-source, decentralized application for anonymously sharing images, videos, and other files. It uses peer-to-peer protocols and the InterPlanetary File System (IPFS) to enable direct, serverless file transfers, enhancing privacy and security"
```

originless/umbrel-app.yml
```diff
@@ -0,0 +1,23 @@
...
+version: "0.1.0"
+tagline: "Private, decentralized file sharing for Nostr and the web"
+category: files
+description: "One all-in-one storage backend you can drop into anything: your own apps, screenshot tools, pastebin-style pastes, Nostr clients, Reddit posts, forum embeds-anything that needs durable, anonymous file hosting. Built on IPFS with 95% bandwidth reduction compared to previous versions."
+releaseNotes: ""
+developer: "besoeasy"
+website: "https://github.com/besoeasy/Originless"
+repo: "https://github.com/besoeasy/Originless"
+support: "https://github.com/besoeasy/Originless/issues"
+port: 3232
```
al-lac, line 13: We have to use a different port here, otherwise users with an existing install of `file-drop` would not be able to install originless.

originless/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+  app:
+    image: ghcr.io/besoeasy/originless:0.1.0@sha256:f4f232052e2763eb23bd9019da885f5700fae947b8760a695be6774bfc9c38b9
+    user: "1000:1000"
+    restart: on-failure
+    stop_grace_period: 5m
+    ports:
+      - "4001:4001/tcp"
+      - "4001:4001/udp"
+    volumes:
+      - ${APP_DATA_DIR}:/data
```
al-lac: Like the linter suggested, just add a subdirectory to the PR with the name `data` and add an empty `.gitkeep` file in there.
```suggestion
      - ${APP_DATA_DIR}data:/data
```

al-lac (changes requested):
Hey @besoeasy,

Thanks for opening this PR to swap out file-drop for originless, looking forward to seeing the new app live on the store! 🔝 

I added some comments on what should be adjusted.

@kaichappel and @kaichappel will take over the review from here and do some detailed testing.

originless/umbrel-app.yml
besoeasy: aren't we disabling file-drop for new installs ?

besoeasy:
@kaichappel let me know

nmfretz:
Thanks for reviewing @al-lac 💜

@besoeasy sorry for the delay on this. I'll push final changes today to get this to the app store. I'll keep the commits atomic so you can see what's needed.

originless/umbrel-app.yml
nmfretz: We are, but that does not remove file-drop from someone's Umbrel who already has it installed. If they have file-drop installed already, and then install originless without first uninstalling file-drop there will be a host port clash and they will be left with a failed install.

This is completely avoided if we just change file-drop to originless, because then we can keep the port the same and those existing installs just update to the new app.

nmfretz: Let's make a rare exception here and keep the port the same. I'll add a note to the app description. If anyone reaches out to you saying that originless fails to install for them, the reason will likely be that they have file drop still installed.

nmfretz:
@besoeasy looks like I don't have push access to your fork of this repo. I'm closing this and opening a new PR to get this done. Will keep you as author: https://github.com/getumbrel/umbrel-apps/pull/5733

PR #5715, opened 2026-06-02
Add Wazuh

gabrysh:
Adds the Wazuh single-node security platform (indexer, manager, dashboard, v4.14.5) as a one-click app:
- TLS certificates generated on first boot, idempotent across restarts
- Dashboard served over HTTP behind the app_proxy
- node.store.allow_mmap: false on the indexer to avoid raising the host vm.max_map_count (no privileged container needed)
- Manager ports exposed for Wazuh agents (1514/1515/514/55000)

Images pinned by sha256 digest. Tested on Linux (x86_64) and Raspberry Pi.


wazuh/docker-compose.yml
```diff
@@ -0,0 +1,139 @@
...
+        soft: 65536
+        hard: 65536
+    healthcheck:
+      test: ["CMD-SHELL", "curl -fks https://localhost:9200/_cluster/health -u admin:SecretPassword || exit 1"]
+      interval: 20s
+      timeout: 10s
+      retries: 15
+      start_period: 60s
+    volumes:
+      - wazuh-indexer-data:/var/lib/wazuh-indexer
```
nmfretz, line 46: Can you please move these named volumes under `${APP_DATA_DIR}/data/...`?

Umbrel backs up/restores the app data directory, and uninstall removes `${APP_DATA_DIR}`, but Docker named volumes sit outside that flow. So the indexer data, manager state/logs/queue, Filebeat state, and dashboard custom/config data would be missed by backups/restores and left behind on uninstall.

wazuh/docker-compose.yml
```diff
@@ -0,0 +1,139 @@
...
+      - ${APP_DATA_DIR}/config/wazuh_indexer_ssl_certs/root-ca.pem:/usr/share/wazuh-indexer/config/certs/root-ca.pem
+      - ${APP_DATA_DIR}/config/wazuh_indexer_ssl_certs/wazuh.indexer-key.pem:/usr/share/wazuh-indexer/config/certs/wazuh.indexer.key
+      - ${APP_DATA_DIR}/config/wazuh_indexer_ssl_certs/wazuh.indexer.pem:/usr/share/wazuh-indexer/config/certs/wazuh.indexer.pem
+      - ${APP_DATA_DIR}/config/wazuh_indexer_ssl_certs/admin.pem:/usr/share/wazuh-indexer/config/certs/admin.pem
+      - ${APP_DATA_DIR}/config/wazuh_indexer_ssl_certs/admin-key.pem:/usr/share/wazuh-indexer/config/certs/admin-key.pem
+      - ${APP_DATA_DIR}/config/wazuh_indexer/wazuh.indexer.yml:/usr/share/wazuh-indexer/config/opensearch.yml
```
nmfretz, line 52: I think we need an update strategy for these config files before merging.

They are copied into `${APP_DATA_DIR}` on install, but normal app updates only refresh top-level files like `docker-compose.yml`, `exports.sh`, hooks, templates, and `umbrel-app.yml`. The `config/` directory won't be refreshed for existing installs.

Since these look version-coupled to the Wazuh/OpenSearch/dashboard images, future Wazuh updates could leave existing users running new containers with stale config. Can we either keep app-managed config out of persistent app data, or add hook/migration logic that updates app-managed config while preserving any user-managed files?

wazuh/docker-compose.yml
```diff
@@ -0,0 +1,139 @@
...
+      # Agent communication / enrollment / syslog / REST API. Exposed on the
+      # host so Wazuh agents on your network can report in.
+      - "1514:1514"
+      - "1515:1515"
+      - "514:514/udp"
+      - "55000:55000"
+    environment:
+      - INDEXER_URL=https://wazuh.indexer:9200
+      - INDEXER_USERNAME=admin
+      - INDEXER_PASSWORD=SecretPassword
```
nmfretz, line 79: We shouldn't ship this wired up with upstream demo credentials.

The manifest tells users to change `admin / SecretPassword`, but the healthcheck and inter-service env vars still depend on that same hardcoded password. So changing it in Wazuh can break the packaged stack. This is also more sensitive because the Wazuh manager API is reachable from the LAN.

Can we use Umbrel-derived secrets here and generate/render the Wazuh internal users config from those on first start?

nmfretz (changes requested):
Sorry for the wait @gabrysh. Thanks for the submission! I took a close look through the packaging and there are a few Umbrel-specific changes needed before I can properly test this.

The main thing is that Wazuh is a pretty stateful/security-sensitive stack, so we need to be careful about how this maps onto Umbrel's app lifecycle. Right now most of the runtime state is in Docker named volumes, some version-coupled config files live in `config/` where they won't be refreshed by normal app updates, and the stack is wired together with upstream demo credentials.

Can we rework the package so that:

- persistent runtime state lives under `${APP_DATA_DIR}/data/...`
- generated certs/secrets persist across restarts but are not public demo credentials
- app-managed config has a clear update/migration path
- user-managed config is intentionally preserved

Left a few inline comments for the specific spots.

nmfretz:
Hey @gabrysh just pinging you here again in case you missed the review.

gabrysh (commented):
fix conf

nmfretz:
> fix conf

@gabrysh did you mean to push changes?

nmfretz:
Hey @gabrysh just pinging again to see if you are still interested in getting this on the app store.

PR #5718, opened 2026-06-04
Update nostr-vpn to v4.0.57

kaichappel:
Hey @mmalmi we have some automation that lets us know when you've published a new nostr-vpn release. Would you like to continue submitting app updates yourself when you want a new release to be available on the Umbrel app store? Or would you rather we do the updates and testing when we get notified of a new release? Totally up to you.

### DIFF
--- a/nostr-vpn/docker-compose.yml
+++ b/nostr-vpn/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 38080
 
   daemon:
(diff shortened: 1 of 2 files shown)


mmalmi:
If you have automation for updating, it would be very helpful!

I had just submitted PR for 4.0.58 which has the changelog since .47. https://github.com/getumbrel/umbrel-apps/pull/5723

PR #5719, opened 2026-06-04
bffless: update to v0.1.53

toshimoto821:
Updates the BFFless app from CE v0.1.42 to v0.1.53.

### Highlights
- Accept multiple proxy rule sets when creating a deployment
- Choose 301 or 302 status for redirect-type domains
- Custom-domain sign-ups now relay back to the target domain
- Pipeline string response bodies served verbatim instead of JSON-encoded
- Stronger authorization: global Admin treated as project Owner consistently, role lanes enforced when granting project permissions, and global admin honored on the repo feed and project settings
- Create Deployment button and proxy-rule actions hidden/disabled for viewers
- Fixed Azure Blob Storage upload and its setup wizard step
- Surface storage migration errors with a force-switch recovery path
- Include redirectType in the domain response DTO and SSL renewal regeneration

Full release notes: https://github.com/bffless/ce/releases/tag/v0.1.53

### DIFF
--- a/bffless/docker-compose.yml
+++ b/bffless/docker-compose.yml
@@ -9,7 +9,7 @@ services:
   # Nginx reverse proxy - serves frontend, routes API calls
   # Uses custom umbrel image with baked-in config template and setup page
   nginx:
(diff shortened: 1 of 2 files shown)


nmfretz:
Wonderful, thanks for the update @toshimoto821! Tested and works well. Going live

PR #5714, opened 2026-06-02
Add Sage Router app

earlvanze:
# App Submission

### App name
Sage Router

### 256x256 SVG icon
See `sage-router/icon.svg` (1024x1024 SVG source + 512x512 PNG fallback). The icon is a hexagonal routing mesh with a sage-green central hub and cyan/violet/magenta provider nodes around it. No rounded corners in the SVG (CSS rounds it dynamically).

### Gallery images
- `sage-router/1.jpg` (1280x800) — dashboard with provider list, OpenClaw Codex OAuth status, and live routing visualization

Additional gallery images (provider health, intent routing breakdown, billing tiers) can be added during the Umbrel team review per the README.

### Description
Sage Router is a self-hosted, multi-provider AI router. It discovers available LLM providers and routes requests intelligently based on intent, model capabilities, cost, and provider health.

OpenClaw Codex OAuth: connects directly to `chatgpt.com/backend-api/codex/responses` using the same OpenAI OAuth token stored in your OpenClaw auth-profiles. No gateway subprocess hop, no separate token refresh.

Supports:
- OpenAI (OpenAI API + Codex via OpenClaw OAuth)
- Anthropic Claude (via bundled Dario proxy)
- Google Generative AI (Gemini)
- OpenRouter (340+ models, free tier preferred)
- Ollama (local + Ollama Cloud catalog)
- NVIDIA NIM
- Darkbloom
- GitHub Copilot (via OpenClaw gateway bridge)

Features:
- Intent-aware routing (code / frontier / eco / local / agentic)
- OpenAI Chat Completions + Anthropic Messages + Google Generative AI bridges
- Streaming SSE for chat and code completions
- Tool/function calling with Codex Responses API schema translation
- Built-in dashboard with provider health, latency, and live routing visualization
- Per-tenant API keys with subscription tiers (free/lite/pro/max/metered)
- Stripe billing integration

### Source repo
https://github.com/earlvanze/sage-router

### I have tested my app on:
- [ ] umbrelOS on a Raspberry Pi
- [ ] umbrelOS on an Umbrel Home
- [x] umbrelOS on Linux VM (local Docker validation)
- [x] Standalone Docker on Linux x86_64 (docker compose up -d)

### Notes
- The upstream Dockerfile ships a multi-stage Node 22 + Python 3 image that includes bundled Ollama and Dario. For the Umbrel Community App Store image tag, the upstream `ghcr.io/earlvanze/sage-router:latest` is referenced.
[...]

PR #5720, opened 2026-06-04
Add Sage Router

earlvanze:
Adds Sage Router to the Umbrel App Store.

Changes in this update:
- Adds the Sage Router Umbrel app metadata and compose service.
- Makes Sage Router depend on the Ollama Umbrel app.
- Routes Sage Router's Ollama provider through Ollama's exported `APP_OLLAMA_URL` instead of `host.docker.internal:11434`.
- Keeps provider credentials and imported config in Sage Router's own app data under `/config`.

Validation run locally:
- `docker compose config` for `sage-router/docker-compose.yml` with a dummy `app_proxy` service and `APP_OLLAMA_URL=http://ollama_ollama_1:11434`
- `git diff --check --cached` before commit

Related separate PR:
- Ollama dashboard changes are split into #5763.


nmfretz:
Thanks @earlvanze. Is this supposed to superseded https://github.com/getumbrel/umbrel-apps/pull/5714?

nmfretz:
Sorry for the delay @earlvanze. Can you clarify how you're envisioning this package to work? Right now it looks like you're trying to use a host-level OpenClaw/Dario installation.

The current compose mounts:

```yaml
  - ${UMBREL_ROOT}/.openclaw:/root/.openclaw:ro
  - ${UMBREL_ROOT}/.dario:/root/.dario
```

On umbrelOS, apps normally store their own data under:
```
${UMBREL_ROOT}/app-data/<app-id>/data
```

For this app, that would be exposed in compose as soomething like:
```
${APP_DATA_DIR}/data
```

So `${UMBREL_ROOT}/.openclaw` and `${UMBREL_ROOT}/.dario` are not normal app data paths, and they also do not point at the OpenClaw Umbrel app’s data directory.

If Sage Router needs to store its own config/auth state, that should live under${APP_DATA_DIR}/data/.... and if Sage Router is meant to consume state from the OpenClaw app, we need to model that explicitly. Usually that means declaring the dependency and having the dependency expose connection details through exports.sh. Directly bind-mounting another app’s private data directory is not something we generally do, especially for OAuth/JWT material.

Let me know and i can take a closer look!

earlvanze:
Thanks for the review. I reworked this to avoid host-level OpenClaw/Dario mounts.

The updated approach is harness-agnostic and keeps Sage Router state under its own app data:

- The PR compose mounts only `/data` to `/config`; it no longer binds `/.openclaw`, `/.dario`, or another app's private state.
- The router now has `SAGE_ROUTER_CONFIG_SOURCE=auto` plus `harness_discovery.py`, which scans the app-owned config tree for supported harness config shapes: OpenClaw, Hermes, Pi, and generic provider JSON layouts.
- If Sage Router later needs to consume live OpenClaw app state, I agree that should be modeled as an explicit dependency/export contract rather than a private bind mount. This update does not rely on that private mount.
- The image is bumped to `ghcr.io/earlvanze/sage-router-public:v3.28.2@sha256:a4400a4dfa136067c394b6b2f99d8c5aee7c2373086eb66a2e0a871357cb7040`; router and Docker defaults are standardized on port 8790.

Validation:

- `python3 -m py_compile router.py harness_discovery.py`
- `python3 -m unittest discover -s tests` (36 tests)
- Umbrel compose renders with only app-owned `/data` mounted into the app container
- GitHub `Release image` workflow for `v3.28.2` completed successfully

earlvanze:
Follow-up for the lint results:

- Added `sage-router/data/.gitkeep` so the `${APP_DATA_DIR}/data:/config` mount target exists in the app package.
- Made the Sage Router image rootless-compatible by moving runtime HOME/Dario/Ollama state under `/config`, then set the app service to `user: "1000:1000"`.
- Bumped the pinned image to `ghcr.io/earlvanze/sage-router-public:v3.28.4@sha256:a05899c4e30325debed8cf00c8be845b2c84763ce21b5e6e0b4604958a13a754`.
- Also committed the preferred `router-profiles.json` defaults into `earlvanze/sage-router`; the image includes those defaults.

The latest `Lint apps` run is now clean with no errors or warnings: https://github.com/getumbrel/umbrel-apps/actions/runs/27362596553/job/80853647514

nmfretz:
Thanks @earlvanze! Before we continue review, can you clarify the intended first-run/setup flow for Umbrel users and confirm what you have tested end-to-end?

Looks like Sage Router mounts its own app data at `/config` and enables config discovery there. But I don't see an Umbrel package mechanism or in-app setup/import flow that writes provider/OpenClaw/Hermes config into `/config`. If there is a setup I'm missing, can you let me know? Ideally the app would walk users how to set things up. 

I also noticed that the compose currently sets `SAGE_ROUTER_OPENAI_CODEX_AUTH_PROFILE_PATH` to `/config/agents/main/agent/auth-profiles.json`, while the discovery flow appears to support layouts such as `/config/openclaw/...`.

And a  couple of other things:

- Is the model API intended to require a client key? The compose sets `SAGE_ROUTER_API_KEY`, but upstream appears to gate model requests with `SAGE_ROUTER_CLIENT_API_KEYS` / `SAGE_ROUTER_CLIENT_AUTH_REQUIRED`.
- Which providers are expected to work in the Umbrel package as shipped, and which require additional user configuration? The manifest mentions several providers that are disabled in compose, so we should align the package description with the actual default behavior.

earlvanze:
Thanks, this is addressed in the latest push (`99515cbd`).

What changed:

- Added an Umbrel app-data seed at `sage-router/data/openclaw/openclaw.json`, mounted as `/config/openclaw/openclaw.json`, with a default local Ollama provider pointed at `http://host.docker.internal:11434`.
- Added `sage-router/data/SETUP.md` so first install has an app-owned setup/import guide under `/config/SETUP.md`.
- Updated the upstream dashboard in `ghcr.io/earlvanze/sage-router-public:v3.28.5` so when providers/models are not configured it shows a first-run setup panel.
- Stopped mounting or depending on another app's private data. OpenClaw/Hermes-style config is copied/imported into Sage Router's own app data.
- Switched Codex auth discovery to `SAGE_ROUTER_OPENAI_CODEX_AUTH_PROFILE_PATHS` and checks both supported Umbrel layouts: `/config/openclaw/agents/main/agent/auth-profiles.json` and `/config/agents/main/agent/auth-profiles.json`.
- Clarified that Sage Router does not bundle a custom `auth.openai.com/codex/device` OAuth route. Users should use the official Codex/OpenClaw auth flow and import the resulting profile, or provide `APP_SAGE_ROUTER_CODEX_ACCESS_TOKEN` / `APP_SAGE_ROUTER_OPENAI_CODEX_API_KEY`.
- Removed the incorrect `SAGE_ROUTER_API_KEY` setting for model auth. The model API is open inside the app by default with `SAGE_ROUTER_CLIENT_AUTH_REQUIRED=0`; users can opt into client keys with `SAGE_ROUTER_CLIENT_AUTH_REQUIRED=1` and `SAGE_ROUTER_CLIENT_API_KEYS`.
- Aligned the manifest with what ships by default: local host Ollama is the only expected zero-credential provider if the host is running Ollama on `11434`; cloud/quota-bound providers remain disabled until credentials or an explicit export/import contract exists.

Validation run:

- `python3 -m json.tool sage-router/data/openclaw/openclaw.json`
- no-tab and `git diff --check` validation for the app files
- `docker compose config` render with a dummy `app_proxy` image, since Umbrel injects that service at runtime
- upstream router syntax/dashboard checks before building the image
- exact `v3.28.5` image built and pushed as `ghcr.io/earlvanze/sage-router-public:v3.28.5@sha256:28a1ab64cef68c1086a7eda0cf5fa7bd47c73d3949b7cef95d1822a161414e0c`
- PR `Lint apps` check is passing

earlvanze:
Updated this PR to address the app-data/auth concerns from review:

- Removed reliance on mounting host/user or another app's private auth state. Sage Router only mounts its own Umbrel app data at `${APP_DATA_DIR}/data:/config`.
- Added a built-in dashboard setup flow in Sage Router v3.28.7 for provider config and Codex auth JSON import. Imported credentials are written under app-owned `/config` paths.
- Kept Sage Router harness/provider agnostic: OpenClaw is not required, Ollama is optional, and app dependencies remain empty.
- Disabled the OpenClaw/Codex gateway fallback in the Umbrel package until an app-owned credential is configured.
- Kept the app port locked to `8790` and updated the image to `ghcr.io/earlvanze/sage-router-public:v3.28.7@sha256:ed6a98b52bc68af3457690586f7e572047d1c4a4e0065854fda4413b072774f1`.

nmfretz:
Thanks @earlvanze, this is definitely in better shape than the earlier version. A few things still need to be cleaned up before we can continue review:

- Can you please remove the Ollama app/dashboard changes from this PR. It looks like that work has been split into #5763, so this PR should only include the Sage Router app files.
- Please make the Sage Router packaging consistent around Ollama. Right now the app depends on `ollama` and uses `$APP_OLLAMA_URL`, but `sage-router/data/SETUP.md` and the seeded `openclaw.json` still describe optional host Ollama at `http://host.docker.internal:11434`.
- If the dashboard now handles first-run setup/import, please either remove `sage-router/data/SETUP.md` or update it so it exactly matches the current Umbrel package. Users generally won't see files inside app data, so the dashboard and app description should be the source of truth.
- Can you confirm whether `sage-router/data/openclaw/openclaw.json` is intentionally the default app-owned provider config path? If so, please make sure the UI/docs make clear that OpenClaw itself is not required and this is just the internal provider config layout.

Also, I just added some agent skills to this repo to help make app submissions easier. Feel free to point your coding agent at [AGENT.md](https://github.com/getumbrel/umbrel-apps/blob/master/AGENTS.md) which will help you work through things locally before pushing.

earlvanze:
Updated the Sage Router app image pin to `ghcr.io/earlvanze/sage-router-public:v3.28.8@sha256:888a9fdda7c1316d58992ebf5dcbb828318bca781d76e059d3ddbfcb6df65a27`. This release includes the Google Gemini/Vertex tool-call routing fix so Google-compatible providers return structured `tool_calls` and can continue agent turns instead of leaking tool intent as prose.

earlvanze:
Updated Sage Router to `ghcr.io/earlvanze/sage-router-public:v3.28.11@sha256:2b486dddd654a83d82a2c04f9d38482c2069632e8371c5117d6ef3c62f1cbf8d`.

This includes the Google-compatible provider fix for structured tool calls plus the hosted-provider servability guard, so Google/Gemini models are not incorrectly blocked by stale host-local `servable=false` metadata.

Validation:
- `python3 -m py_compile router.py`
- `python3 -m pytest tests/test_tool_call_leak.py -q` -> 28 passed
- Forced `google/gemini-2.5-flash` function-call smoke passes on reachable live routers: Cyber, local Umbrel app, remote Umbrel, and mntd3690.
- Cloud Run deployed to the same digest and `/health` returns 200.
- Tailnet edge reports 7/7 healthy upstreams.
- PR lint is passing.

earlvanze:
Final follow-up: set `path: "/dashboard"` in `sage-router/umbrel-app.yml` so the Umbrel app tile opens the live dashboard instead of the API root. Latest lint run is passing on commit `ab51bfa6`.

earlvanze:
Addressed the remaining packaging/onboarding concerns in `6309a1e6`:

- Rebuilt the branch on current `master`; the PR now contains only `sage-router/**` (three files). All Ollama dashboard/package changes remain isolated to #5763.
- Made Ollama wiring consistent: Sage Router depends on the Umbrel Ollama app and uses only its exported `$APP_OLLAMA_URL`. The host-level `host.docker.internal:11434` seed is gone.
- Removed `data/SETUP.md` and the seeded `data/openclaw/openclaw.json`. The in-app dashboard is now the single first-run setup/import flow and writes only to Sage Router-owned `/config` data.
- Clarified in the manifest that OpenClaw is not required. The compatible config layout is an internal Sage Router detail, not an OpenClaw dependency.
- Removed package-level provider disable defaults. Providers without credentials stay unavailable; credentials added through the dashboard can activate DeepSeek, Anthropic, OpenRouter, NVIDIA NIM, Google, OpenAI-compatible/BYOS, Codex, and other supported routes without editing Compose.
- Pinned the multi-architecture `v3.28.12-slim` image at `sha256:7a6e6c02bc2655322ee7cb5075d3ae6af601f3c84a443311ffc644ae76cd894e` (amd64 + arm64). This release adds provider-wide BYOK key pools while isolating throttling/failures to the affected credential rather than disabling the provider.

Validation:

- Umbrel `Lint apps`: passing
- PR merge state: clean
- Prettier YAML parse/check and `git diff --check`: passing
- Registry manifest verified as an OCI multi-arch index with linux/amd64 and linux/arm64
- Rolled the exact digest to three live Umbrel/Tailnet origins; each is healthy and completed an explicit DeepSeek V4 Flash request with `ROUTE_OK`
- Cloud Run is serving the same digest; authenticated `https://api.sagerouter.dev/health` returned HTTP 200 three times and the same DeepSeek smoke returned `ROUTE_OK`

earlvanze:
Updated Sage Router packaging in `dd20b362` to `v3.28.13-slim` / app version `1.0.6`. This release keeps dashboard-managed BYOK/BYOS credentials in Sage Router-owned `/config`, adds independent credential-pool failover so one 401/402/403/429 does not disable a whole provider, and supports custom OpenAI-, Anthropic-, and Ollama-compatible endpoints alongside the built-in providers. Local validation: `npm run lint:apps -- sage-router --check-images` reports 0 errors (the existing shared-network warning remains intentional) and `git diff --check` passes.

earlvanze:
Withdrawing this submission while Sage Router packaging and release plans are being reworked.

PR #5723, opened 2026-06-04
Update Nostr VPN to v4.0.58

mmalmi:
## Summary

Updates Nostr VPN from `v4.0.47` to `v4.0.58`.

This supersedes the open `v4.0.57` update and includes the `v4.0.58` follow-up direct-path fallback fixes.

## Changes since the previous Umbrel release

- Fixed first-run Umbrel setup so the admin network, local `.nvpn` name, device invites, and self device appear correctly while the daemon is starting.
- Improved direct FIPS connections: mesh/relay is now only fallback transport, while direct UDP keeps retrying in the background.
- Fixed network-change cases where stale private addresses or brief hotspot liveness failures could leave peers stuck on mesh/relay.
- Improved FIPS status output so `nvpn status --json` can show fallback transport and direct probing separately.
- Improved FIPS rekey, liveness, and failover handling under load and during changing network conditions.
- Startup and network-refresh failures now show clearer daemon status instead of stale "Turning VPN on" states.

## Validation

- Verified published image digest: `ghcr.io/mmalmi/nostr-vpn-umbrel:v4.0.58@sha256:cf51622fcc1a794ac80e1962fce0855f21575b1d3f69750cb90de82ee3fb52b5`.
- Verified image index includes `linux/amd64` and `linux/arm64`.
- Parsed `nostr-vpn/umbrel-app.yml` and `nostr-vpn/docker-compose.yml`.
- Ran `git diff --check`.
- Tested the `v4.0.58` image on Umbrel OS on a Raspberry Pi 4: daemon and web containers started, health and web endpoints returned 200, and `nvpn status --json` reported binary version `4.0.58`.

### DIFF
--- a/nostr-vpn/docker-compose.yml
+++ b/nostr-vpn/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 38080
 
   daemon:
-    image: ghcr.io/mmalmi/nostr-vpn-umbrel:v4.0.47@sha256:36948ae2717dd299e534e3762a8b1b73ec33eb643264efe2b7387c060445798a
+    image: ghcr.io/mmalmi/nostr-vpn-umbrel:v4.0.58@sha256:cf51622fcc1a794ac80e1962fce0855f21575b1d3f69750cb90de82ee3fb52b5
     restart: on-failure
     stop_grace_period: 1m
     network_mode: "host"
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks for the update as usual @mmalmi! 

As mentioned in https://github.com/getumbrel/umbrel-apps/pull/5718 by @kaichappel we can take care of updating the umbrel package for you moving forward. We shouldn't be more than half a day or so behind your releases. If we're confused about anything or need your input we'll open a PR here and ping you. But obviously still feel free to open PRs anytime.

Really great work on Nostr VPN.

PR #5724, opened 2026-06-04
add Telegrapho app

egzola:
# App Submission

### App name
Telegrapho

### 256x256 SVG icon
icon

### Gallery images
screen1

screen2

### I have tested my app on:
- [x] umbrelOS on Linux VM


nmfretz:
Thanks for the submission @egzola! I'll should be able to get to this soon. Thanks for your patience.

egzola:
> Thanks for the submission @egzola! I'll should be able to get to this soon. Thanks for your patience.

@nmfretz  Thanks for the update! I appreciate it and I'm happy to answer any questions or make adjustments if needed.

nmfretz (commented):
Sorry for the delay @egzola! Great work on packaging this.

One question: can you clarify whether the `${APP_DATA_DIR}/data:/data` mount is needed? From a quick look at the current image, the server state appears to be in-memory and I don't see anything reading or writing `/data`, so we may be able to drop the volume and `data/.gitkeep` unless persistence is planned.

I also noticed the packaged app currently loads SweetAlert2 from jsDelivr, and the QR code feature sends the current text to `api.qrserver.com`. Since the app copy is centered around sensitive text and avoiding third-party services, can you clarify if that's intentional? It may be better to include those scripts/features directly in the app so text doesn't leave the user's Umbrel.

telegrapho/umbrel-app.yml
```diff
@@ -0,0 +1,25 @@
...
+version: "1.0.0"
+port: 3712
+description: >-
+  Stop pasting sensitive text into third-party apps.
+
+  Telegrapho creates private realtime text channels between your devices using your self-hosted server.
+
+  Copy on mobile and receive instantly on desktop.
+
+  Your text. Your server. Your control.
```
nmfretz, line 18: Small manifest formatting tweak too. With `>-`, we need extra blank lines here so the description keeps paragraph spacing in the app UI:

```suggestion
description: >-
  Stop pasting sensitive text into third-party apps.


  Telegrapho creates private realtime text channels between your devices using your self-hosted server.


  Copy on mobile and receive instantly on desktop.


  Your text. Your server. Your control.
```

egzola:
> Sorry for the delay @egzola! Great work on packaging this.
> 
> One question: can you clarify whether the `${APP_DATA_DIR}/data:/data` mount is needed? From a quick look at the current image, the server state appears to be in-memory and I don't see anything reading or writing `/data`, so we may be able to drop the volume and `data/.gitkeep` unless persistence is planned.
> 
> I also noticed the packaged app currently loads SweetAlert2 from jsDelivr, and the QR code feature sends the current text to `api.qrserver.com`. Since the app copy is centered around sensitive text and avoiding third-party services, can you clarify if that's intentional? It may be better to include those scripts/features directly in the app so text doesn't leave the user's Umbrel.

Thanks for the review and for pointing those out!

You're absolutely right.

The `/data` mount and `data/.gitkeep` were not needed since the current implementation keeps channel state entirely in memory, so I've removed them.

I also removed the external dependencies that were being loaded from jsDelivr and the QR code generation that relied on api.qrserver.com. The required functionality is now bundled locally with the application, so text no longer leaves the user's Umbrel for those features.

I've pushed the changes and updated the image accordingly.

nmfretz:
Excellent, thanks for the changes @egzola. Gallery assets are now done and we're going live. Great work on yet another app!

image

egzola:
Thank you @nmfretz !   I really appreciate the review and feedback.

Your suggestions helped improve the app and make it more aligned with its privacy-focused goals.

Also, the gallery assets look fantastic. They do a great job of communicating the app's purpose at a glance and really elevate the presentation.

Thanks again for your time and for helping bring Telegrapho to the Umbrel App Store.

PR #5725, opened 2026-06-05
Update Nostr VPN to v4.0.62

mmalmi:
## Summary
- Updates Nostr VPN from v4.0.47 to v4.0.62.
- Points both Umbrel services at the pinned v4.0.62 multi-arch image digest.
- Switches the app services to `restart: unless-stopped` and removes the obsolete Compose `version` field.

## Changes since the previous Umbrel release
- Fixed first-run Umbrel setup so the admin network, local `.nvpn` name, device invites, and self device appear reliably while the daemon starts.
- Improved direct peer connectivity: mesh/relay is now fallback transport while direct UDP keeps probing and can recover after hotspot or NAT changes.
- Avoided stale private/hotspot endpoints and brief liveness timeouts leaving peers stuck on mesh.
- Improved daemon liveness, rekey, and failover behavior under load, with clearer startup and network-refresh status.

## Test plan
- Verified the v4.0.62 image manifest includes linux/amd64 and linux/arm64.
- Ran the Umbrel pinned-image Docker web E2E against `ghcr.io/mmalmi/nostr-vpn-umbrel:v4.0.62@sha256:e11c4be353bd6922d1829d9a61073cba3960a985294cafb34869bf2efd5012e1` (4 browser/API tests passed).
- YAML parsed cleanly and `git diff --check` passed.

### DIFF
--- a/nostr-vpn/docker-compose.yml
+++ b/nostr-vpn/docker-compose.yml
@@ -1,14 +1,13 @@
-version: "3.7"
-
 services:
   app_proxy:
+    restart: unless-stopped
     environment:
       APP_HOST: nostr-vpn_web_1
       APP_PORT: 38080
 
   daemon:
-    image: ghcr.io/mmalmi/nostr-vpn-umbrel:v4.0.58@sha256:cf51622fcc1a794ac80e1962fce0855f21575b1d3f69750cb90de82ee3fb52b5
-    restart: on-failure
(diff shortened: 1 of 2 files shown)


nmfretz:
Closing in favour of your latest https://github.com/getumbrel/umbrel-apps/pull/5728

PR #5726, opened 2026-06-05
Update Public Pool server image

benjamin-wilson:
### App
Public Pool

### Summary
- Updates the Public Pool server image from the sethforprivacy fork to the maintained benjamin-wilson/public-pool image.
- Keeps the existing Duckaxe UI image unchanged.
- Bumps the app version to 95565ee and updates release notes.

### Image
`ghcr.io/benjamin-wilson/public-pool:95565ee@sha256:3edffef3bbad23cbd46098904b22ba8a1c14c679ae90cf1e4d97ed98492bbab4`

### Validation
- Upstream image build completed successfully: https://github.com/benjamin-wilson/public-pool/actions/runs/27022260615
- Confirmed image index is multi-arch for linux/amd64 and linux/arm64.
- Ran `git diff --check` for the app-store changes.

### DIFF
--- a/public-pool/docker-compose.yml
+++ b/public-pool/docker-compose.yml
@@ -15,7 +15,7 @@ services:
       - DOMAIN=$DEVICE_DOMAIN_NAME
 
   server:
-    image: ghcr.io/sethforprivacy/public-pool:9938a05@sha256:96347cbbc2ec6f27ec1b282e8353e41197fad6ca56e011873e69e7c3b6555b14
+    image: ghcr.io/benjamin-wilson/public-pool:95565ee@sha256:3edffef3bbad23cbd46098904b22ba8a1c14c679ae90cf1e4d97ed98492bbab4
     restart: on-failure
     stop_grace_period: 30s
     ports:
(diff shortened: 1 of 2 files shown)


nmfretz:
Amazing, thanks for this update @benjamin-wilson! People have been eagerly awaiting these improvements making their way to the self-hosted version. Tested on both amd64 and arm64 and working well.

Is there an official UI image we can use by chance instead of relying on forks? I noticed the UI workflow publishes `ghcr.io/benjamin-wilson/public-pool-ui`, but pulls currently return 403. Is that the intended official UI image, and if so would you be okay making it public?

For Umbrel, the UI just needs to use the local app proxy instead of the baked-in `public-pool.io` URLs. Runtime env var support would be awesome, but if you’d rather not add that, we can handle it entirely on our side with a small nginx `sub_filter` block:

```nginx
location ~* \.js$ {
  proxy_pass http://public-pool_web_1:80;
  proxy_set_header Accept-Encoding "";
  proxy_hide_header ETag;

  sub_filter_types application/javascript text/javascript;
  sub_filter_once off;
  sub_filter '"https://public-pool.io:40557"' '""';
  sub_filter '"public-pool.io:21496"' 'window.location.hostname+":2018"';
  }
  ```

PR #5727, opened 2026-06-05
canary: Update to v1.5.2

schjonhaug:
## Summary
Update Canary Umbrel app to v1.5.2.

## Release summary
Canary v1.5.2 rejects invalid xpub descriptors before they can create wallets that cannot sync correctly, and adds recovery handling so failed or stuck wallet records can be deleted and recreated. It also fixes Genesis wallet and address-watch syncing edge cases.

## Upstream release
https://github.com/schjonhaug/canary/releases/tag/v1.5.2

## Test plan
- [x] Tested on local Umbrel

### DIFF
--- a/canary/docker-compose.yml
+++ b/canary/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api/*"
 
   backend:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for the update @schjonhaug! I've updated the release notes so that existing users see them when going to update the app. Going live.

PR #5728, opened 2026-06-05
Update Nostr VPN to v4.0.67

mmalmi:
## Summary
- Update Nostr VPN from v4.0.58 to v4.0.67.
- Pin the daemon and web containers to the pushed multi-arch v4.0.67 image digest.
- Refresh the Umbrel release notes with the networking fixes since the previous Umbrel release.

## Changes since v4.0.58
- Keeps VPN traffic flowing over mesh fallback when direct UDP is marked link-dead, while direct probing continues and can recover.
- Fixes stale or freshly reconnected UDP paths suppressing better fallback routes after hotspot/NAT changes.
- Prevents fallback transit loops after direct-path loss.
- Warms established sessions after fresh fallback discovery, reducing stalls after liveness timeouts.

## Testing
- Parsed `nostr-vpn/umbrel-app.yml` and `nostr-vpn/docker-compose.yml` as YAML.
- Confirmed the pinned image digest is multi-arch for `linux/amd64` and `linux/arm64`.
- Tested on Umbrel Pi: daemon and web containers ran v4.0.67, the web UI responded through the app proxy, and VPN status reported active.

### DIFF
--- a/nostr-vpn/docker-compose.yml
+++ b/nostr-vpn/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 38080
 
   daemon:
-    image: ghcr.io/mmalmi/nostr-vpn-umbrel:v4.0.58@sha256:cf51622fcc1a794ac80e1962fce0855f21575b1d3f69750cb90de82ee3fb52b5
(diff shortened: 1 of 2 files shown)


nmfretz:
Awesome, thanks for the update @mmalmi 🤝

PR #5730, opened 2026-06-06
Update Public Pool UI image

benjamin-wilson:
## Summary

Updates Public Pool to use the upstream `benjamin-wilson/public-pool-ui` image with runtime configuration support instead of the Duckaxe UI image.

Refs https://github.com/getumbrel/umbrel-apps/pull/5726#issuecomment-4637177262

## Changes

- Switches the Public Pool web image to `ghcr.io/benjamin-wilson/public-pool-ui:5f730f4@sha256:c1eaae042ee179eca87b273c04c54c8b1fba5ceb63a03411a21ec8ac6875be3d`
- Sets `PUBLIC_POOL_API_URL=` so UI API calls use Umbrel's same-origin `/api/...` proxy
- Sets `PUBLIC_POOL_STRATUM_URL=:2018` so the UI displays `<device-hostname>:2018`
- Updates release notes for the UI runtime config change

## Verification

- `ruby -e 'require "yaml"; %w[public-pool/docker-compose.yml public-pool/umbrel-app.yml].each { |f| YAML.load_file(f); puts "OK #{f}" }'`
- `docker buildx imagetools inspect ghcr.io/benjamin-wilson/public-pool-ui:5f730f4`
- Confirmed `public-pool/data/proxy/nginx.conf` routes `/api/...` to the Public Pool server

Note: `docker compose config` cannot fully validate Umbrel apps locally because Umbrel's special `app_proxy` service has no image/build context outside umbrelOS.

### DIFF
--- a/public-pool/docker-compose.yml
+++ b/public-pool/docker-compose.yml
@@ -13,6 +13,7 @@ services:
     stop_grace_period: 30s
     environment:
       - DOMAIN=$DEVICE_DOMAIN_NAME
+      # Intentionally empty so the UI uses same-origin /api routes via Umbrel's proxy.
       - PUBLIC_POOL_API_URL=
       - PUBLIC_POOL_STRATUM_URL=${DEVICE_DOMAIN_NAME}:2018
 
--- a/public-pool/docker-compose.yml
+++ b/public-pool/docker-compose.yml
@@ -14,7 +14,7 @@ services:
     environment:
       - DOMAIN=$DEVICE_DOMAIN_NAME
       - PUBLIC_POOL_API_URL=
-      - PUBLIC_POOL_STRATUM_URL=:2018
+      - PUBLIC_POOL_STRATUM_URL=${DEVICE_DOMAIN_NAME}:2018
 
   server:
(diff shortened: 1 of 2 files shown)


benjamin-wilson:
I would suggest displaying the device IP rather than DEVICE_DOMAIN_NAME. Some networks and or miners have a hard time with mDNS. I've gotten quite a few support requests because of it.

nmfretz:
Ah ya, great call @benjamin-wilson. will make the change and test.

Potentially that will result in more support requests for you due to DHCP-rotation though. The ui will show the new ip on restart, but then users would need to re-configure their miner... and seems a bit risky that someone could have no real indication that they are no longer connected.

nmfretz:
ya, second guessing this a bit. silent failure on dhcp rotation (or silently falling back to a fallback pool if a user has configured it) seems kinda bad.

Any chance you'd be okay with having the UI display multiple stratum endpoints?

nmfretz:
Going live with this as-is for now just to get the update out the door. But let's revisit soon because I don't want you to get pummelled by support questions for this.

Thanks again for sorting out the official image @benjamin-wilson.

PR #5733, opened 2026-06-09
Add Originless (and remove file-drop)

nmfretz:
Orignal PR from @besoeasy: https://github.com/getumbrel/umbrel-apps/pull/5709


nmfretz:
Alrighty @besoeasy this is good to go now. 

Changes for your reference:
- Updated File Drop’s disabled listing to point users to Originless.
- Added a short note for existing File Drop users in the app description. We can delete this in a little bit, but for now hopefully it will prevent you and us from getting support requests for failed installs.
- Removed the pre-start hook and added data/.gitkeep instead.
- Fixed the data mount to ${APP_DATA_DIR}/data:/data.
- Changed the host IPFS swarm port to 4003 to avoid the existing 4001 conflict with another ipfs app.
- Added an advanced IPFS note for router forwarding: external 4001 to Umbrel 4003.

For now I've just kept the gallery assets the same as file drop, but we'll update them shortly with new screenshots from originless

nmfretz:
@besoeasy, new gallery assets!

Screenshot 2026-06-09 at 12 12 34 PM

PR #5734, opened 2026-06-09
Add Mastodon app

caydyan:
## Summary

Adds a Mastodon app package for Umbrel. Refs #560.

This is a clean follow-up to the public Mastodon app attempts and addresses the review feedback visible on #5662:

- adds an executable `pre-install` hook so `mastodon.env` exists before Compose reads `env_file`
- keeps `pre-start` executable and uses it to fill missing secrets/config without overwriting existing `LOCAL_DOMAIN`, `WEB_DOMAIN`, or SMTP settings
- supports the normal Umbrel launch host/port through `DEVICE_DOMAIN_NAME` and `APP_MASTODON_PORT` defaults
- bootstraps the first Owner account (`umbrel@umbrel.local`) with the deterministic Umbrel app password
- removes the app-local icon and custom pytest file that were requested for removal on the earlier submission

## Lint note

`Lint apps` passes with no errors or warnings. It reports one informational note that the `proxy` nginx service uses the image default user.

I left that service on the upstream nginx default intentionally because this container binds port 80 and uses a read-only generated nginx config. The Mastodon application services (`web`, `streaming`, `sidekiq`, `migrate`) already run as the Mastodon user, and the data services run as `1000:1000`. I can switch the proxy to a non-root/high-port layout if maintainers prefer that pattern for this app.

## Validation

```sh
bash -n mastodon/hooks/pre-install
bash -n mastodon/hooks/pre-start
ruby -e 'require "yaml"; ["mastodon/umbrel-app.yml", "mastodon/docker-compose.yml"].each { |p| YAML.load_file(p); puts "#{p}: ok" }'
git diff --check
```

I also ran the lifecycle hooks against a temporary `APP_DATA_DIR` with a stubbed `docker` command to verify that:

- `pre-install` creates `mastodon.env` before app start
- `pre-start` writes required defaults and secrets
- user-edited `LOCAL_DOMAIN`, `WEB_DOMAIN`, and `SMTP_SERVER` values survive later starts

I could not run a full umbrelOS/Docker runtime install from this workstation because the local Docker daemon is not available.

## Payout

Bounty payout address for #560:

`bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`


caydyan:
Quick status check on this Mastodon app submission for #560.

Current contributor-side state:
- PR is open, non-draft, and GitHub reports the branch as clean/mergeable.
- `Lint apps` completed successfully; the only linter note is informational for the nginx proxy container user.
- The package keeps the app self-contained and addresses the earlier review path: pre-install env-file creation, executable hooks, preserved domain/SMTP settings across restarts, first Owner bootstrap via the Umbrel deterministic app password, and app-package cleanup.

If there is any remaining packaging, review, or bounty-routing gate for the $500 BTC offer in #560, please point me at it and I can turn it around.

BTC payout address if accepted for the bounty: `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

caydyan:
Rebased this Mastodon app submission onto current upstream `master` (`cc4cff2a`) and force-with-lease pushed head `604abf66`.

Validation after rebase:
- `bash -n mastodon/hooks/pre-install mastodon/hooks/pre-start`
- Ruby YAML parse for `mastodon/umbrel-app.yml` and `mastodon/docker-compose.yml`
- `git diff --check origin/master...HEAD`
- hosted `Lint apps`: success

The PR is currently open, non-draft, mergeable, and `CLEAN` from GitHub. Ready for maintainer/bounty review.

nmfretz:
Thanks for putting this together @caydyan!

I also appreciate you clearly noting that this builds on the public review feedback from #5662.

Since #5662 is still open and awaiting changes, and #560 has a community bounty attached, I want to handle the overlap carefully so contributor credit and bounty expectations stay clear.

Would you be open to coordinating with @weishen612 on #5662, either by sharing these fixes there or opening a PR against their branch if they're open to it?

I'm going to give the active reviewed PR a reasonable chance to move forward first. If that doesn't happen, I'm happy to continue review here as the current candidate. Either way, there may still be technical review items to work through before Mastodon is ready to merge.

The goal is just to keep the process transparent for everyone involved.

caydyan:
Thanks @nmfretz, that makes sense to me. I do not want this to bypass #5662 or blur contributor/bounty expectations.

@weishen612 I am happy to coordinate in whichever form is easiest for you. The parts from this PR that should be portable back into your branch are:

- pre-install creation for the Mastodon data/env paths so fresh installs avoid missing mount targets;
- executable `pre-install` / `pre-start` hooks;
- preserving `LOCAL_DOMAIN`, `SMTP_*`, and generated secrets across restarts;
- first Owner bootstrap using the Umbrel deterministic app password;
- package cleanup around icons/metadata and current app-store structure.

If you want these folded into #5662, I can either summarize the exact patch here or open a PR against your `mastodon-app` branch so you can review/cherry-pick/modify it under your active submission. Until you say which route you prefer, I will leave #5734 as an alternate reviewed candidate rather than pushing for it to supersede your PR.

For transparency, my branch is currently lint-clean and mergeable, so the coordination work should be mostly about attribution/process and any remaining Mastodon product-scope review items rather than rescuing a broken branch.

caydyan:
Quick follow-up now that #5662 was closed in favor of this PR.

I am still available to work through the remaining Mastodon technical review items here. Current contributor-side state is unchanged: this PR is open, non-draft, mergeable, and the hosted `Lint apps` check is successful. The branch already covers the earlier reviewed install/runtime points from #5662: pre-install env-file/data-dir creation, executable hooks, preserving domain/SMTP/secrets across restarts, first Owner bootstrap, and package cleanup.

Please point me at the next review item you want tightened and I can turn it around on this branch.

BTC payout address for the #560 bounty if accepted: `bc1qev5ant33v5y89qqjvcf4mh9hlax5svqf5xd7gc`

nmfretz:
Thanks @caydyan, will get on this right away!

mastodon/hooks/pre-start
```diff
@@ -0,0 +1,177 @@
...
+    -rwebpush \
+    -e 'v = Webpush.generate_key; puts "VAPID_PRIVATE_KEY=#{v.private_key}"; puts "VAPID_PUBLIC_KEY=#{v.public_key}"'
+}
+
+ensure_present RAILS_ENV production
+ensure_present NODE_ENV production
+ensure_present RAILS_SERVE_STATIC_FILES true
+ensure_present BIND 0.0.0.0
+
+ensure_present LOCAL_DOMAIN "${DEVICE_DOMAIN_NAME}"
```
nmfretz: Can we avoid initializing Mastodon with `umbrel.local` as the default server identity?

For Mastodon, `LOCAL_DOMAIN` / `WEB_DOMAIN` are effectively part of the instance identity. If the user later exposes this through a real domain, changing these after the instance has federated can break remote discovery and federation. I think we should delay the bootstrap until these values are set intentionally, or otherwise make the app wait for a supported setup path.

mastodon/hooks/pre-start
```diff
@@ -0,0 +1,177 @@
...
+FORCE_HTTP_FILE="${APP_DATA_DIR}/force-http.rb"
+MASTODON_IMAGE="ghcr.io/mastodon/mastodon:v4.5.9@sha256:2f4930e3f84a2f6b90b49c4e58329e9abc8bd354344cbd0607484514decff99f"
+
+mkdir -p \
+  "${APP_DATA_DIR}/data/system" \
+  "${APP_DATA_DIR}/data/db" \
+  "${APP_DATA_DIR}/data/redis"
+
+touch "${ENV_FILE}"
+cp "${APP_DIR}/nginx.conf" "${NGINX_CONFIG_FILE}"
```
nmfretz: This copy step fails when the hook runs from the installed app directory. In the Umbrel lifecycle, `APP_DIR` is the installed `${APP_DATA_DIR}`, so this becomes `cp ${APP_DATA_DIR}/nginx.conf ${APP_DATA_DIR}/nginx.conf`; `cp` exits non-zero and `set -e` stops the hook before `mastodon.env` is populated.

Can we remove this copy path and make these files update-safe instead? A clean option would be to use top-level templates or generate the files directly in the hook.

mastodon/docker-compose.yml
```diff
@@ -0,0 +1,117 @@
...
+      redis:
+        condition: service_healthy
+
+  db:
+    image: postgres:17.9-alpine@sha256:c7526c0f6c3f30260a563d7bcf8ad778effac59a44f8ffa86678c35418338609
+    user: "1000:1000"
+    restart: on-failure
+    environment:
+      POSTGRES_USER: mastodon
+      POSTGRES_PASSWORD: mastodon
```
nmfretz: Can we generate a per-install database password instead of shipping the literal `mastodon` password here and in `DB_PASS`?

Umbrel apps share the Docker network, so internal databases should not rely on network isolation alone. Mastodon already supports `DB_PASS`, so this can be generated before Postgres initializes and reused by both the `db` service and the Mastodon services.

mastodon/docker-compose.yml
```diff
@@ -0,0 +1,117 @@
...
+      test: ["CMD-SHELL", "pg_isready -U mastodon -d mastodon_production"]
+      interval: 10s
+      timeout: 5s
+      retries: 10
+
+  redis:
+    image: redis:8.4.2-alpine@sha256:e1b6db24cb4fdd89f4bc9be09f671ea3bec92fbd7042554f76c34aa2be9b59ad
+    user: "1000:1000"
+    restart: on-failure
+    command: ["redis-server", "--save", "60", "1", "--loglevel", "warning"]
```
nmfretz: Can we add Redis authentication too? Mastodon supports `REDIS_PASSWORD` for both Rails and streaming, and Redis can be started with `--requirepass`.

Without that, any other container on the shared Umbrel network can connect to this Redis instance.

nmfretz (commented):
Thanks @caydyan. I really appreciate your patience here, especially while we waited to let the other Mastodon PR play out.

I’ll say up front that I’m not a Mastodon operator myself, so please correct me if I’m misunderstanding part of the setup model. My current understanding is that Mastodon’s `LOCAL_DOMAIN` / `WEB_DOMAIN` are not just normal settings. They become part of the server’s identity, and changing them later can break federation once the instance has been used publicly.

With that in mind, the current packaging flow doesn’t quite line up with the way Mastodon needs to be configured. Right now the package initializes Mastodon automatically with `umbrel.local` defaults. Cloudflare Tunnel or another reverse proxy could make the UI reachable, but that alone would not make the instance correctly federate as a public Mastodon server if Mastodon itself was initialized as `umbrel.local`. Does that understanding sound right?

If so, I think we have a few reasonable paths:

- Delay Mastodon initialization until the required public-instance config exists, then let advanced users set those env vars and restart before the first bootstrap.
- Add a lightweight setup/onboarding flow that collects domain/SMTP config before initializing Mastodon. This is nicer UX, but more package code to maintain.
- Wait for the upcoming umbrelOS work we’re about to launch around out-of-the-box HTTPS and app env-var editing, then target this package at that flow so users can install, set the required env vars, and only then initialize Mastodon.

My preference is one of the delayed-initialization paths. Mastodon is a good fit for advanced Umbrel users, but we should avoid bootstrapping the instance with `umbrel.local` if the user’s goal is a real federated server.

I also found a couple of lower-level packaging issues while testing the package locally. I left those inline.

caydyan:
Thanks @nmfretz. Quick status nudge after the latest hosted run:

- `Lint apps` is still successful as of the 2026-06-30 run.
- The PR remains open and non-draft.
- I am still available to turn around any Mastodon packaging/runtime review items you want addressed here.

No pressure if this is still in review queue; I just wanted to keep the branch visibly ready and responsive.

PR #5736, opened 2026-06-09
hashrate-autopilot: populate releaseNotes with v1.13.0 + community-store migration steps

rdouma:
### What

Populates the empty `releaseNotes:` field on `hashrate-autopilot/umbrel-app.yml` with two things:

1. **A short community-store → official-store migration recipe.** Users who already run Hashrate Autopilot from my community app store (`rdouma-hashrate-autopilot`) need to know that installing this official version creates a second, empty instance side-by-side (different store IDs → different `app-data/` directories → Umbrel treats them as different apps). The notes give them a five-command SSH path to bring their `state.db` forward.

2. **The v1.13.0 user-visible changes** for anyone seeing the app in the official store for the first time. Mirrors the release notes already shipped at https://github.com/rdouma/hashrate-autopilot/releases/tag/v1.13.0.

### Why now

This is the first official-store release and there's a real population of existing users from the community store version (which I've been publishing since v1.7.x). Without the migration notes in the right place, the natural "click Update / Install" path drops them into a fresh database with their entire history sitting one directory over, which can read as "the app lost my data" — a bad first impression of the official store release. The release-notes block is the highest-visibility surface for this guidance because Umbrel surfaces it directly in the App Store UI.

### Scope

`umbrel-app.yml` only. No version bump, no compose change, no manifest schema change. Pure text addition to a field that was empty in the original submission (#5685).

### DIFF
--- a/hashrate-autopilot/umbrel-app.yml
+++ b/hashrate-autopilot/umbrel-app.yml
@@ -6,6 +6,13 @@ version: "1.13.0"
 tagline: Autopilot for renting hashrate to mine on Ocean Mining
 
 description: >-
+  ALREADY RUNNING THE COMMUNITY-STORE VERSION? This official-store app has a
+  different app id than the community-store one (rdouma-hashrate-autopilot),
+  so it installs as a separate app with an empty database. A one-time
+  five-minute migration brings your full history across:
+  https://github.com/rdouma/hashrate-autopilot/blob/main/docs/migrating-from-community-store.md
+
+
   Rent hashrate on the Braiins Hashpower marketplace and route it to your own
   Ocean non-custodial mining pool, automatically. Each tick the autopilot finds
   the cheapest available price for your target hashrate, adds a small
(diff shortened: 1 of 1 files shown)


nmfretz:
Hey @rdouma. Did you want to add a link to instructions in the main app description instead of these long release notes? Because they'll likely not be seen by users when going to install an app.

You could do something like right at the top of the `description` in the umbrel-app.yaml put a front-and-centre note that links to your migration instructions. Then you can easily update those anytime without having to go through us.

Also, i just looked at your instructions real quick, and I think you could just do this all without ssh. umbrelOS gives you access to all app data through Files. so they can just do things like drag and drop. What do you think?

rdouma:
Both done, thanks. The long release notes are gone in favor of a link at the top of the description, pointing at the migration guide in my repo so I can keep it current without bothering you. The guide is now written UI-first.

nmfretz:
Awesome, thanks @rdouma!

PR #5737, opened 2026-06-09
App Submission: MailFlow

Oni-giri:
# App Submission

## App name:

MailFlow

## What is it?

MailFlow is a self-hosted unified webmail client. Instead of juggling multiple browser tabs, you connect all your IMAP/SMTP accounts: Gmail, iCloud, Outlook, or any custom mail server. Read them together in one clean web interface.

It ships with a unified inbox, full-text search, compose/reply/forward, folder navigation, multiple themes, multi-user support with user management, SSO/OIDC, and Microsoft 365 OAuth2.

- Website: https://mailflow.sh
- Source: https://github.com/maathimself/mailflow
- Docker images: ghcr.io/maathimself/mailflow-frontend:1.4.0 and ghcr.io/maathimself/mailflow-backend:1.4.0 (pinned by digest)

## 256x256 SVG icon

icon

## Gallery images

mailflow_01_unified_inboxes_2160x1350
mailflow_02_themes_2160x1350
mailflow_03_accounts_2160x1350

## Notes for reviewers

- New app: nginx frontend + Node.js backend + PostgreSQL + Redis, all self-contained.
- Port 3395 chosen to avoid conflicts.
- No default credentials — the user creates their admin account on first launch.
- SESSION_SECRET and ENCRYPTION_KEY are derived from ${APP_SEED}; internal DB password is a fixed value (not user-facing).
- Backend health check has a start_period of 30s to allow DB init on first boot.

## I have tested my app on:

- [ ] umbrelOS on Linux VM


nmfretz:
Thanks for another submission @Oni-giri! I'll review soon. Thanks for your patience.

mailflow/docker-compose.yml
```diff
@@ -0,0 +1,86 @@
...
+
+  frontend:
+    image: ghcr.io/maathimself/mailflow-frontend:1.4.0@sha256:a143fa899a211ba2fe7fc6664d4e1b19b26b68f238319a773950934c369ac065
+    restart: on-failure
+    stop_grace_period: 1m
+    depends_on:
+      backend:
+        condition: service_healthy
+    volumes:
+      - ${APP_DATA_DIR}/data/certs:/etc/nginx/ssl
```
nmfretz: I think we can drop this mount. Umbrel routes the app proxy to port 80, and the frontend image still starts without persisting `/etc/nginx/ssl` because it generates its self-signed cert inside the container.

```suggestion
```

mailflow/icon.svg
```diff
@@ -0,0 +1,30 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 256 256" width="256" height="256">
```
nmfretz: We can remove this file. App icons are hosted elsewhere for the store, so this doesn't need to be included in the app package.

mailflow/umbrel-app.yml
```diff
@@ -0,0 +1,27 @@
...
+category: networking
+name: MailFlow
+version: "1.4.0"
+tagline: Unified self-hosted webmail client
+description: >-
+  MailFlow is a self-hosted unified webmail client. Connect multiple IMAP/SMTP
+  accounts such as Gmail, iCloud, Outlook, or custom mail servers and read them
+  in one web interface.
+
+  It includes a unified inbox, search, compose/reply/forward, folder navigation,
```
nmfretz, line 13: Since this uses a folded block scalar (`>-`), we need one more blank line here so the rendered description preserves a paragraph break.

```suggestion
  accounts such as Gmail, iCloud, Outlook, or custom mail servers and read them
  in one web interface.


  It includes a unified inbox, search, compose/reply/forward, folder navigation,
```

nmfretz (commented):
Great work @Oni-giri!

Just a few small packaging cleanups before merging:

mailflow/docker-compose.yml
nmfretz: Also, if you end up removing the mount, the mailflow/data/certs dir can be deleted too.

nmfretz:
Oh also @Oni-giri, it looks like `VAPID_PUBLIC_KEY`, `VAPID_PRIVATE_KEY`, and `VAPID_SUBJECT` can only be set via env vars, right? Like there's no way to set this up through the app's UI.

Users could technically add these manually by editing the app’s compose locally, but manual app compose changes on umbrelOS don’t survive app updates. So if someone were to set up push notifications and then update they app they suddenly wouldn't work.

We’re working on env-var configuration support in umbrelOS (coming very soon), so this isn’t a blocker for this submission, but wanted to let you know.

Oni-giri:
> Oh also @Oni-giri, it looks like `VAPID_PUBLIC_KEY`, `VAPID_PRIVATE_KEY`, and `VAPID_SUBJECT` can only be set via env vars, right? Like there's no way to set this up through the app's UI.
> 
> Users could technically add these manually by editing the app’s compose locally, but manual app compose changes on umbrelOS don’t survive app updates. So if someone were to set up push notifications and then update they app they suddenly wouldn't work.
> 
> We’re working on env-var configuration support in umbrelOS (coming very soon), so this isn’t a blocker for this submission, but wanted to let you know.

Those are optional variables used for Web push (useful if you have mailflow as a PWA on your phone for instance). While it's a nice-to-have, it's not a main feature and doesn't prevent deployment I believe.

Anyway, thanks a lot for the review and the explanations!

nmfretz:
> Anyway, thanks a lot for the review and the explanations!

Of course! Gallery assets are ready and we're going live. Great work again @Oni-giri.

image

PR #5739, opened 2026-06-10
chore: update pogolo to v1.1.3

0xf0xx0:
i hope the release notes formatting is ok, i cant see myself cause the umbrel docker dev image isnt building atm


nmfretz:
Thanks for the update @0xf0xx0! Tested and working well.

> i hope the release notes formatting is ok

I added whitespace so they show up like this now:

image

PR #5743, opened 2026-06-11
Update Public Pool: use static IP, and expose it to all umbrel apps

martinbarilik:
Hi, would love to be able to request public pool with my app, i am developing.

Hope it's all good.

### DIFF
--- a/public-pool/docker-compose.yml
+++ b/public-pool/docker-compose.yml
@@ -39,6 +39,9 @@ services:
       - ENABLE_SOLO=true
       - ENABLE_PROXY=false
       - POOL_IDENTIFIER="Public Pool on Umbrel"
+    networks:
+      default:
+        ipv4_address: $APP_PUBLIC_POOL_IP
 
   proxy:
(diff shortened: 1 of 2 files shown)


nmfretz:
Hey @martinbarilik, stoked to see what app you're developing!

Just to clarify the use case here: is the consuming app another Umbrel app that depends on `public-pool`?

If so, do you need a static IP specifically, or could the app connect over Docker’s internal DNS instead? For example:

```yaml
  environment:
    PUBLIC_POOL_API_URL: "http://public-pool_server_1:2019"
```

Just trying to understand whether this needs to become an exported/static endpoint, or whether the existing container DNS name is enough for your app.

martinbarilik:
Hi

> Just to clarify the use case here: is the consuming app another Umbrel app that depends on `public-pool`?

yes, exactly. It's Public Pool's Web.

I am not sure if i need exported/static endpoint, i was just coping a bitcoin implementation.

nmfretz:
Ah ya, you should be fine just using Docker DNS. The Bitcoin Node app has those static IPs set like that for other legacy reasons. 

I'll close this for now. And then we can revisit if you have any issues getting your app to connect. Feel free to open a PR here with your app as soon as you need a bit of direction!

PR #5744, opened 2026-06-11
Update bolt12-pay to v0.2.105.1

Alex71btc:
Fix amountless BOLT12 offer payments.

Previously LNDK used the offer amount when creating invoices. For amountless offers this resulted in a zero-value invoice, causing blinded path creation to fail.

This change uses InvoiceRequest::amount_msats() instead, which correctly contains the payer-selected amount for amountless offers.

Tested successfully with Phoenix Wallet and amountless BOLT12 offers.


Alex71btc:
Included a small follow-up UX fix for amountless offer history display. Functionality unchanged; amountless offers tested successfully on Umbrel.

nmfretz:
Wonderful, thanks for the update @Alex71btc!

PR #5758, opened 2026-06-17
Add Tunnel

miniweed:
## Tunnel

Tunnel exposes a user's Umbrel services to the internet through **their own VPS**
using **WireGuard** (end-to-end encryption) + **Caddy** (automatic HTTPS via
Let's Encrypt). A 100% self-hosted alternative to Cloudflare Tunnel that works
behind CGNAT, with no router port forwarding.

### How it works
Inbound traffic enters through the user's VPS (public IP) and travels an encrypted
WireGuard tunnel to the Umbrel, where Caddy terminates HTTPS and reverse-proxies
to the chosen internal service. The home router never opens a port.

### Notes for reviewers (security)

**Trust boundaries — what runs where.** There are two separate domains, and the
privileged action happens only on the user's own server, never on the Umbrel host:

- **On the Umbrel host (this app):** fully sandboxed. `web` runs with
  `cap_drop: ALL` + `no-new-privileges`; `caddy` adds only `NET_BIND_SERVICE`;
  only `wg` needs `NET_ADMIN` + `SYS_MODULE` for WireGuard — the same model as the
  official **Tailscale** app. No `privileged`, no host network, no host bind
  mounts. The app cannot affect the Umbrel host or other apps.
- **On the user's own VPS:** the generated setup script is run by the user, as root,
  on a server *they* own and rent — the same trust model as any VPS setup guide.
  The app never touches that machine itself.

**The VPS script is fully auditable and safe by design.**
- Built **only from the distro's apt packages** — it never downloads and runs
  remote code (no `curl | sh`).
- **Deterministic** (same config → same script) and prints a **SHA-256** the user
  verifies before running.
- Installs a **restrictive firewall with automatic rollback** (a 120s kill-switch
  reverts the rules if the tunnel doesn't come up, preventing lockout) and
  **SSH hardening** that is skipped safely if no authorized key is present.

**App-side guardrails.**
- Anti-SSRF: reverse-proxy targets are validated; loopback, link-local, cloud
  metadata (169.254.169.254) and control ports (Docker, Caddy admin, the wg API)
  are blocked, with DNS-rebinding pinning. Only RFC1918 internal services can be
  exposed (the app's purpose).
- Secrets (WireGuard private/preshared keys, service targets) are **encrypted at
  rest** (AES-256-GCM, scrypt-derived key); the audit log is a tamper-evident
  SHA-256 hash chain.
- Panel access is protected by Umbrel's authenticated app proxy
  (`PROXY_AUTH_ADD: true`). All images are pinned by multi-arch digest.

### Testing
- [x] Installed and ran on umbrelOS; state persists across app restart.
- [x] End-to-end verified: tunnel up, HTTPS service reachable.
- Platform tested: Linux VM / Umbrel.

### Links
- Source: https://github.com/miniweed/umbrel-app-store
- Icon (256×256 SVG, no rounded corners): https://raw.githubusercontent.com/miniweed/umbrel-app-store/main/miniweed-tunnel/docs/umbrel-submission/icon-256.svg

Tunnel icon

### Gallery (1440×900)
Dashboard
Instructions
Configuration
VPS Setup
Services

### DIFF
--- a/miniweed-tunnel/docker-compose.yml
+++ b/miniweed-tunnel/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       PROXY_AUTH_ADD: "true"
 
   web:
-    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.49@sha256:984ba28eb543e0f8c9d506fc3166c5220e40ebca890536b92bcfb8698b7b7b68
+    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.50@sha256:be804173c119b54181049cb17ae4987795ab425d8908648dcd09de81852674da
     restart: on-failure
     stop_grace_period: 30s
     cap_drop:
@@ -70,7 +70,6 @@ services:
       - ALL
     cap_add:
       - NET_ADMIN
-      - SYS_MODULE
     security_opt:
       - no-new-privileges:true
     sysctls:
--- a/miniweed-tunnel/docker-compose.yml
+++ b/miniweed-tunnel/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       PROXY_AUTH_ADD: "true"
 
   web:
-    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.47@sha256:38b00458b4d2ffba4c9bef18261caa510641cc436fa54b44bfc04599f3c43368
+    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.48@sha256:c4d6af5cf6a2316da0210d44150eeff39e1898fe600f2a296cefd7354a8d01f2
     restart: on-failure
     stop_grace_period: 30s
     cap_drop:
--- a/miniweed-tunnel/docker-compose.yml
+++ b/miniweed-tunnel/docker-compose.yml
@@ -1,7 +1,3 @@
-# docker-compose.yml for the OFFICIAL Umbrel App Store submission.
-# Same as the community-store compose but the web image is pinned by its
-# multi-arch digest (required for the official store). wg and caddy are already
-# pinned. The Umbrel team may further adjust ports/digests during review.
 version: "3.7"
 
 services:
@@ -13,8 +9,7 @@ services:
       PROXY_AUTH_ADD: "true"
 
   web:
-    # 1.6.46 build (multi-arch amd64+arm64). Digest pins the exact tested content.
-    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.46@sha256:1d7aa5c73fc6f204e66c37d392ea39bbf01f49b470a638da481e5268a650eb36
+    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.46@sha256:90207c00ba229b4633d9ef2b65bbd527fc2a51362cc8167bbf582e2521a17cf3
     restart: on-failure
     stop_grace_period: 30s
     cap_drop:
@@ -23,10 +18,12 @@ services:
       - no-new-privileges:true
     volumes:
       - ${APP_DATA_DIR}/data:/data
+      - ${APP_DATA_DIR}/caddy/data:/caddy-data
     environment:
       - DATA_DIR=/data
+      - CADDYFILE_PATH=/caddy-data/Caddyfile
       - WG_API_HOST=wg
-      - WG_API_TOKEN=${APP_SEED}
+      - WG_API_TOKEN=${TUNNEL_WG_TOKEN}
       - PORT=3016
       - APP_SEED=${APP_SEED}
 
@@ -44,24 +41,23 @@ services:
     depends_on:
       - wg
     volumes:
-      - ${APP_DATA_DIR}/data:/data
       - ${APP_DATA_DIR}/caddy/data:/data/caddy
       - ${APP_DATA_DIR}/caddy/config:/config
     environment:
       - XDG_DATA_HOME=/data/caddy
-      - WG_API_TOKEN=${APP_SEED}
+      - WG_API_TOKEN=${TUNNEL_WG_TOKEN}
     command:
       - sh
       - -c
       - |
-        until [ -s /data/Caddyfile ] && wget -q --header="x-wg-api-token: $$WG_API_TOKEN" -O - http://127.0.0.1:8080/status >/dev/null 2>&1; do
+        until [ -s /data/caddy/Caddyfile ] && wget -q --header="x-wg-api-token: $$WG_API_TOKEN" -O - http://127.0.0.1:8080/status >/dev/null 2>&1; do
           echo "[caddy] Waiting for Caddyfile and WireGuard API..."
           sleep 2
         done
-        exec caddy run --config /data/Caddyfile --adapter caddyfile --watch
+        exec caddy run --config /data/caddy/Caddyfile --adapter caddyfile --watch
 
   wg:
-    image: ghcr.io/miniweed/umbrel-tunnel-wg:1.0.5@sha256:2cf7981d175c7e459303816e0bd3553846801b8974494a9b48c36d6638017b36
+    image: ghcr.io/miniweed/umbrel-tunnel-wg:1.0.5@sha256:424562865ec99d64cab96861b3747b3d26bbe0463c672c6aec9cda2e23766a6b
     restart: on-failure
     stop_grace_period: 30s
     cap_drop:
@@ -74,6 +70,6 @@ services:
     sysctls:
       - net.ipv4.conf.all.src_valid_mark=1
     volumes:
-      - ${APP_DATA_DIR}/data:/data
+      - ${APP_DATA_DIR}/data:/data:ro
     environment:
-      - WG_API_TOKEN=${APP_SEED}
+      - WG_API_TOKEN=${TUNNEL_WG_TOKEN}
--- a/miniweed-tunnel/docker-compose.yml
+++ b/miniweed-tunnel/docker-compose.yml
@@ -13,8 +13,8 @@ services:
       PROXY_AUTH_ADD: "true"
 
   web:
-    # 1.6.45 build (multi-arch amd64+arm64). Digest pins the exact tested content.
-    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.45@sha256:15ce2036c6e8a2bb6295c781686025e628cce1db5954c2e614f70340f76c76b1
+    # 1.6.46 build (multi-arch amd64+arm64). Digest pins the exact tested content.
+    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.46@sha256:1d7aa5c73fc6f204e66c37d392ea39bbf01f49b470a638da481e5268a650eb36
     restart: on-failure
     stop_grace_period: 30s
     cap_drop:
--- a/miniweed-tunnel/docker-compose.yml
+++ b/miniweed-tunnel/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       PROXY_AUTH_ADD: "true"
 
   web:
-    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.46@sha256:90207c00ba229b4633d9ef2b65bbd527fc2a51362cc8167bbf582e2521a17cf3
+    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.47@sha256:38b00458b4d2ffba4c9bef18261caa510641cc436fa54b44bfc04599f3c43368
     restart: on-failure
     stop_grace_period: 30s
     cap_drop:
@@ -57,7 +57,7 @@ services:
         exec caddy run --config /data/caddy/Caddyfile --adapter caddyfile --watch
 
   wg:
-    image: ghcr.io/miniweed/umbrel-tunnel-wg:1.0.5@sha256:424562865ec99d64cab96861b3747b3d26bbe0463c672c6aec9cda2e23766a6b
+    image: ghcr.io/miniweed/umbrel-tunnel-wg:1.0.6@sha256:22fbcbc01c31ec70c623ac670f195353c5fa37525ccecb18be86d9df2ed87469
     restart: on-failure
     stop_grace_period: 30s
     cap_drop:
--- a/miniweed-tunnel/docker-compose.yml
+++ b/miniweed-tunnel/docker-compose.yml
@@ -14,7 +14,7 @@ services:
 
   web:
     # 1.6.45 build (multi-arch amd64+arm64). Digest pins the exact tested content.
-    image: ghcr.io/miniweed/umbrel-tunnel-web@sha256:15ce2036c6e8a2bb6295c781686025e628cce1db5954c2e614f70340f76c76b1
+    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.45@sha256:15ce2036c6e8a2bb6295c781686025e628cce1db5954c2e614f70340f76c76b1
     restart: on-failure
     stop_grace_period: 30s
     cap_drop:
--- a/miniweed-tunnel/docker-compose.yml
+++ b/miniweed-tunnel/docker-compose.yml
@@ -9,11 +9,15 @@ services:
       PROXY_AUTH_ADD: "true"
 
   web:
-    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.48@sha256:c4d6af5cf6a2316da0210d44150eeff39e1898fe600f2a296cefd7354a8d01f2
+    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.49@sha256:984ba28eb543e0f8c9d506fc3166c5220e40ebca890536b92bcfb8698b7b7b68
     restart: on-failure
     stop_grace_period: 30s
     cap_drop:
       - ALL
+    cap_add:
+      # Mounted data dirs are owned by 1000:1000 on umbrelOS while the process
+      # runs as root with all caps dropped; without this it cannot write to them.
+      - DAC_OVERRIDE
     security_opt:
       - no-new-privileges:true
     volumes:
@@ -35,6 +39,8 @@ services:
       - ALL
     cap_add:
       - NET_BIND_SERVICE
+      # Same as web: mounted dirs are 1000:1000 and Caddy runs as root.
+      - DAC_OVERRIDE
     security_opt:
       - no-new-privileges:true
     network_mode: "service:wg"
@@ -70,6 +76,8 @@ services:
     sysctls:
       - net.ipv4.conf.all.src_valid_mark=1
     volumes:
-      - ${APP_DATA_DIR}/data:/data:ro
+      # Only the subdir holding the generated wg0.conf: wg does not need to see
+      # config.json/app-seed/etc.
+      - ${APP_DATA_DIR}/data/wg:/data:ro
     environment:
       - WG_API_TOKEN=${TUNNEL_WG_TOKEN}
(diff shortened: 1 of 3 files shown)


miniweed:
Hey team 👋 thanks for taking a look!

Quick context: this came out of something a lot of Umbrel power users have been asking for — a self-hosted alternative to Cloudflare Tunnel that keeps traffic on infrastructure they actually own. Tunnel does that with WireGuard + Caddy through the user's own VPS (works behind CGNAT, no router ports).

I built it prioritizing **security and simplicity**: the app containers are sandboxed (only `wg` needs WireGuard caps, same as Tailscale), the VPS setup script is 100% apt with no remote code — deterministic and SHA-256-verifiable — and the panel sits behind Umbrel's app proxy.

Happy to answer anything or adjust whatever you need 🙏

besoeasy:
hey i am trying to understand the purpose of this ?

miniweed:
Hey @besoeasy it's a self-hosted alternative to Cloudflare Tunnel: it exposes your Umbrel apps on your own domain via an outbound WireGuard tunnel to a VPS you control (Caddy handles reverse-proxy + HTTPS). The point is to get a public URL without opening router ports and without depending on a third party — works even behind CGNAT. Happy to expand on any part!

nmfretz (commented):
Sorry for the delay @miniweed, and thanks for working on this. Super cool idea!

One thing I want to dig into before this can go live is the app-side auth boundary.

I know Umbrel app proxy auth protects the UI route for users, but the backend service is still reachable by other containers on Umbrel's shared Docker network, which is normal for app packages. The concern here is the sensitivity of this particular API surface: it can generate WireGuard keys, update public service routing, and return the VPS setup script.

Can you add a server-side admin token/session for the config/keygen/script/audit endpoints, or explain the intended threat model here? I want to make sure another app container can't silently reconfigure what gets exposed through the tunnel.

I also left a few packaging comments below around splitting the shared data mounts, secret separation, and version/image pinning. Once those are tightened up I can re-test the install path.

miniweed-tunnel/docker-compose.yml
```diff
@@ -0,0 +1,79 @@
...
+      - ALL
+    cap_add:
+      - NET_BIND_SERVICE
+    security_opt:
+      - no-new-privileges:true
+    network_mode: "service:wg"
+    depends_on:
+      - wg
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
```
nmfretz: Can we split the shared data mount so each container only gets the paths it needs?

Right now `web`, `wg`, and the internet-facing `caddy` container all mount the same `${APP_DATA_DIR}/data` directory. Since that directory contains `config.json`, `wg0.conf`, `app-seed`, audit logs, and the generated `Caddyfile`, Caddy can read more sensitive app state than it needs.

A cleaner shape would be to keep all writable state under `${APP_DATA_DIR}/data/...`, but split it into narrower subdirs/files, for example app-owned config under `data/app`, Caddy state under `data/caddy`, and only the generated Caddyfile/WireGuard config shared read-only with the containers that need them. The required bind-mount source dirs/files should also exist before Docker starts, either committed with `.gitkeep` where appropriate or created by an idempotent `pre-start` hook.

miniweed-tunnel/docker-compose.yml
```diff
@@ -0,0 +1,79 @@
...
+      - ${APP_DATA_DIR}/caddy/data:/data/caddy
+      - ${APP_DATA_DIR}/caddy/config:/config
+    environment:
+      - XDG_DATA_HOME=/data/caddy
+      - WG_API_TOKEN=${APP_SEED}
```
nmfretz: Can we avoid passing `${APP_SEED}` directly to Caddy and the WireGuard helper API?

Right now `${APP_SEED}` is used as the config encryption root and also as `WG_API_TOKEN`, which is passed into Caddy. Since Caddy is the internet-facing component, it'd be better to derive a separate helper API token in `exports.sh` and keep the config encryption secret only in the web container.

miniweed-tunnel/umbrel-app.yml
```diff
@@ -0,0 +1,28 @@
+# Manifest for the OFFICIAL Umbrel App Store submission (getumbrel/umbrel-apps).
+# This is NOT the community-store manifest (that one lives in ../../umbrel-app.yml
+# and keeps the 1.6.x version line for update continuity on existing installs).
+# For the official store this app starts fresh at 1.0.0.
+manifestVersion: 1
+id: miniweed-tunnel
+category: networking
+name: Tunnel
+version: "1.0.0"
```
nmfretz: Can we set this to the upstream version being packaged, which looks like `1.6.46`?

Umbrel uses this for the user-visible app version/update history, and both the image tag and upstream community-store manifest point at `1.6.46`.

miniweed-tunnel/docker-compose.yml
```diff
@@ -0,0 +1,79 @@
...
+      - -c
+      - |
+        until [ -s /data/Caddyfile ] && wget -q --header="x-wg-api-token: $$WG_API_TOKEN" -O - http://127.0.0.1:8080/status >/dev/null 2>&1; do
+          echo "[caddy] Waiting for Caddyfile and WireGuard API..."
+          sleep 2
+        done
+        exec caddy run --config /data/Caddyfile --adapter caddyfile --watch
+
+  wg:
+    image: ghcr.io/miniweed/umbrel-tunnel-wg:1.0.5@sha256:2cf7981d175c7e459303816e0bd3553846801b8974494a9b48c36d6638017b36
```
nmfretz: Looks like this tag may have moved since the PR was opened. `ghcr.io/miniweed/umbrel-tunnel-wg:1.0.5` currently resolves to manifest-list digest `sha256:424562865ec99d64cab96861b3747b3d26bbe0463c672c6aec9cda2e23766a6b`, while the package pins `sha256:2cf7981d175c7e459303816e0bd3553846801b8974494a9b48c36d6638017b36`.

Can you re-pin to the current tag digest, or publish/use a tag that resolves to the exact digest you want packaged?

miniweed:
Thanks for the thorough review @nmfretz! All four points addressed in the latest push:

**1. Server-side auth boundary**
The web container now derives an `ADMIN_TOKEN` via `HKDF-SHA256(APP_SEED, "miniweed-tunnel/v1", "tunnel-api-token-v1")` at startup. A middleware guards `GET/POST /api/config`, `/api/keygen`, `/api/vps-setup-script`, `/api/audit`, `/api/audit/verify`, and `/api/health/refresh` — returning 401 to unauthenticated callers. Auth is accepted via an `x-tunnel-api-token` header or an `_t` HttpOnly/SameSite=Strict cookie set on page load. The SPA uses `credentials: same-origin` so the cookie is included automatically in all API calls without JS changes.

**2. Data mount split**
Caddy no longer mounts `${APP_DATA_DIR}/data` — it no longer has access to `config.json`, `app-seed`, or any other app secrets. The Caddyfile moved to `${APP_DATA_DIR}/caddy/data/Caddyfile` (a path Caddy already owns) via a new `caddy/data:/caddy-data` mount on `web` + `CADDYFILE_PATH` env var. The `wg` mount gains `:ro` — it only ever reads `wg0.conf`, never writes.

**3. Secret separation**
`exports.sh` derives `TUNNEL_WG_TOKEN` via `sha256("tunnel-wg-api-v1:${APP_SEED}")`. Caddy and `wg` receive only this derived token. `APP_SEED` is now exclusive to the `web` container.

**4. Digests**
Both images re-pinned:
- `wg` → `sha256:424562...` (tag had moved, now matches current)
- `web` → `sha256:90207c...` (fresh amd64+arm64 build with the above changes)

Let me know if you'd like anything adjusted!

nmfretz:
Thanks for the quick update @miniweed.

I re-checked the auth change and I think this still needs one more pass. The API routes now require the `_t` cookie or `x-tunnel-api-token`, but the app also issues that `_t` cookie on a plain `GET /`.

That means another container on Umbrel's shared Docker network can request `/`, read the `Set-Cookie`, and then call the protected API routes with that cookie. I tested that against the pushed image and was able to get a `200` from `/api/config` after first fetching `/`.

Can you make sure the admin token is not issued from a route that is reachable directly by other containers? The important bit is that another app container shouldn't be able to bootstrap access to the API by itself.

Also, looks like the manifest version is still `1.0.0`; can we set that to the packaged upstream version, `1.6.46`?

miniweed:
Thanks for catching that @nmfretz — you're right, the `_t` cookie approach still allows any container to bootstrap auth with a plain `GET /`.

Before implementing a fix I want to confirm the right approach. Within Umbrel's shared Docker network there's no HTTP-level mechanism that distinguishes a proxied browser request from a direct container request, so the options are:

**A** — Remove the `Set-Cookie` from `GET /`. Embed the admin token in the served HTML as `window.__T__` and have the SPA send it as an `x-tunnel-api-token` header. The container cookie-jar attack (`curl -c/-b`) stops working; exploiting it would require explicit HTML-parsing code, which eliminates the "by itself" bootstrap. Not theoretically perfect, but meaningfully raises the bar.

**B** — Proper login page with a randomly generated per-install admin secret (stored on disk, never served over HTTP). The user authenticates once per session. Containers can't access the secret. More UX friction and more work to implement.

Which direction would you prefer? Happy to implement either. For the version, agreed — will bump to `1.6.46`.

miniweed:
Hi @nmfretz, forget my last message. Rather than picking between A and B, I went with a third option that is stronger than A without B's UX friction, since the underlying requirement was "another app container shouldn't be able to bootstrap access by itself":

**The web server now only accepts connections whose TCP peer is this app's `app_proxy` (or loopback).**

- A global middleware (registered before all routes and static files) resolves `miniweed-tunnel_app_proxy_1` via Docker DNS (30s TTL, forced re-resolution when an unknown peer shows up so an `app_proxy` restart doesn't lock the UI out, fail-closed if resolution fails) and compares it against `req.socket.remoteAddress` — the real TCP peer, which can't be faked via headers, and completing a TCP handshake with a forged source IP on the bridge isn't possible without an active on-path attack.
- Any other container now gets `403` on **every** path: it can't fetch `/`, so the `_t` cookie is never issued to it, and the API routes are unreachable even with a stolen token.
- Browser traffic is unaffected because it enters through `app_proxy`, which already enforces the Umbrel session (`PROXY_AUTH_ADD`), so the gate sees the proxy as the peer. The `SameSite=Strict` cookie stays as CSRF protection.

I reproduced your test against the new image (three containers on one Docker network):

- attacker container → `GET /` → `403`, no `Set-Cookie`; `GET /api/config` → `403`
- container named `miniweed-tunnel_app_proxy_1` → `200` + cookie → API works

Pushed as **1.6.47** (the manifest version now tracks the packaged upstream version, and the stale header comment is gone). Both digests re-pinned and verified against ghcr.io:

- `umbrel-tunnel-web:1.6.47@sha256:38b00458b4d2ffba4c9bef18261caa510641cc436fa54b44bfc04599f3c43368`
- `umbrel-tunnel-wg:1.0.6@sha256:22fbcbc01c31ec70c623ac670f195353c5fa37525ccecb18be86d9df2ed87469` (small robustness fix in the helper API's token comparison)

1.6.47 also includes a few hardening items from an internal re-review: `wg0.conf` is written with `0600` permissions, `/api/status` and `/api/health` now require auth, and key generation is serialized with config saves. Happy to adjust anything!

miniweed:
Small follow-up: pushed `1.6.48` on top — a text-only fix for missing spaces in the Instructions tab, plus refreshed gallery screenshots taken on this version. `wg` is unchanged; the new web digest is `sha256:c4d6af5cf6a2316da0210d44150eeff39e1898fe600f2a296cefd7354a8d01f2`.

miniweed-tunnel/docker-compose.yml
```diff
@@ -0,0 +1,75 @@
...
+    environment:
+      APP_HOST: miniweed-tunnel_web_1
+      APP_PORT: 3016
+      PROXY_AUTH_ADD: "true"
+
+  web:
+    image: ghcr.io/miniweed/umbrel-tunnel-web:1.6.48@sha256:c4d6af5cf6a2316da0210d44150eeff39e1898fe600f2a296cefd7354a8d01f2
+    restart: on-failure
+    stop_grace_period: 30s
+    cap_drop:
```
nmfretz, line 15: Can we add `DAC_OVERRIDE` to both `web` and `caddy`?

On a fresh umbrelOS install, the mounted directories are owned by `1000:1000`, while these containers run as root with all capabilities dropped. That left `web` unable to save its config and Caddy unable to create its files. Adding `DAC_OVERRIDE` fixed both locally.

miniweed-tunnel/docker-compose.yml
```diff
@@ -0,0 +1,75 @@
...
+      - ALL
+    cap_add:
+      - NET_ADMIN
+      - SYS_MODULE
+    security_opt:
+      - no-new-privileges:true
+    sysctls:
+      - net.ipv4.conf.all.src_valid_mark=1
+    volumes:
+      - ${APP_DATA_DIR}/data:/data:ro
```
nmfretz: Can we mount only the generated `wg0.conf` here instead of the entire app data directory?

Since `wg` only reads that file and already has elevated capabilities, it shouldn’t need access to the rest of the app data. A separate output directory or configurable path for `wg0.conf` would keep this much tighter.

nmfretz (commented):
Sorry for the wait here @miniweed, and thanks for the update. I gave `1.6.48` another test, and the app-proxy peer check is working as intended against other containers on the shared network.

I did find a fresh-install write-permissions issue, plus one mount I’d like to tighten up. I left both inline. Once those are updated I can re-test the save and restart flow.

miniweed-tunnel/docker-compose.yml
miniweed: Done in `1.6.49` — added `DAC_OVERRIDE` to both `web` and `caddy` (keeping `cap_drop: ALL` plus caddy's `NET_BIND_SERVICE`).

I reproduced the fresh-install failure locally with `1000:1000`-owned data dirs: without the cap, `web` gets `EACCES` writing `config.json` (500 on save) and Caddy can't create its files; with it, the full save flow works.

miniweed-tunnel/docker-compose.yml
miniweed: Done in `1.6.49` — `wg0.conf` now lives in its own subdir and the `wg` container mounts only `${APP_DATA_DIR}/data/wg` read-only, so it no longer sees `config.json`, `app-seed`, or anything else in the app data dir.

The `wg` image is unchanged (its entrypoint already reads `$DATA_DIR/wg0.conf`, so mounting the subdir at `/data` just works). The web app writes the conf into the new subdir, and on the first boot after the update it migrates an existing `wg0.conf` there — copied rather than renamed, so file ownership normalizes to the web process and the `wg` container (which has no `DAC_OVERRIDE`) can always read the `0600` file. Verified locally: inside the `wg` container, `/data` contains only `wg0.conf` (plus the committed `.gitkeep`) and the tunnel comes up from the migrated config.

miniweed:
Thanks for the re-test @nmfretz! Both points are addressed in **1.6.49**, just pushed to the branch:

- **`DAC_OVERRIDE` on `web` and `caddy`** — reproduced the fresh-install `EACCES` locally (`1000:1000` dirs + root with all caps dropped) and verified saves and Caddy file creation work with the cap added.
- **`wg` now mounts only the generated config** — `wg0.conf` moved to `${APP_DATA_DIR}/data/wg/` and the `wg` service mounts just that subdir read-only. The `wg` image is unchanged; the web app writes to the new location and automatically migrates an existing `wg0.conf` on first boot, so updates from earlier versions keep their tunnel config.

Digests, verified against ghcr.io:

- `umbrel-tunnel-web:1.6.49@sha256:984ba28eb543e0f8c9d506fc3166c5220e40ebca890536b92bcfb8698b7b7b68`
- `umbrel-tunnel-wg:1.0.6@sha256:22fbcbc01c31ec70c623ac670f195353c5fa37525ccecb18be86d9df2ed87469` (unchanged)

Local verification of the save/restart flow you're planning to re-test: fresh-install save (config + Caddyfile + `wg/wg0.conf` written `0600`), update-path migration from the old layout, and the `wg` container bringing the tunnel up while seeing only `wg0.conf` in its mount. Happy to adjust anything else!

nmfretz:
Thanks so much for working through all of this @miniweed — we’re super close now! I really appreciate the quick changes. I re-tested `1.6.49` and it’s looking great.

Sorry to catch one last thing this late: can we drop `SYS_MODULE` from `wg`? I confirmed the full container starts with `NET_ADMIN` alone, and the image doesn’t load kernel modules, so I don’t think it needs host kernel module access.

miniweed:
Thanks @nmfretz! Pushed `1.6.50` on top:

- Dropped `SYS_MODULE` from `wg`'s `cap_add` — confirmed it starts fine with `NET_ADMIN` alone, and the image never loads kernel modules. Only `NET_ADMIN` remains.
- Also included an internal security-hardening pass I did proactively: encrypted migration backups no longer leave a plaintext WireGuard key on disk, the tamper-evident audit log verifies correctly across log rotations, and audit/rate-limit logging uses the real connection IP instead of a spoofable `X-Forwarded-For` header. No changes to the trust boundaries or mounts reviewed in earlier rounds.

New web digest: `sha256:be804173c119b54181049cb17ae4987795ab425d8908648dcd09de81852674da`. `wg` image is unchanged (`1.0.6`, only the compose capability list moved).

Ready whenever you get a chance to re-test — thanks again for all the thorough review rounds!

miniweed:
Hi @nmfretz — hope all is well! Since the last round, umbrelOS 2.0 (beta) landed, and `1.6.50` doesn't work on it: the peer gate you re-tested in rounds 2–4 only admits the `app_proxy` container, which no longer exists on 2.x (the gateway now runs in-process in umbreld). I've pushed **`1.7.0`** so the app works on both generations. Everything you already reviewed is intact; here's what's new.

**Unchanged from the reviewed rounds:** HKDF-derived auth, the mount split (`wg` mounts only `data/wg/` read-only), `DAC_OVERRIDE` on `web`/`caddy`, `wg` with `NET_ADMIN` only, digest pinning, secrets isolation (`APP_SEED` reaches only `web`).

**Peer gate, 1.x vs 2.x.** The gate now also reads the container's default gateway from `/proc/net/route` (the host's IP on the bridge network, where umbreld's in-process gateway and its widget fetches come from). It tells the two OS generations apart with a sticky marker — has `app_proxy` ever resolved since boot:
- On **1.x** the host IP is admitted *only* for `GET /api/widget`; the UI and API still require `app_proxy`, exactly as in `1.6.50`.
- On **2.x** the host IP is the gateway and is trusted for everything (inherent to the platform design; documented as such in the app's SECURITY.md).
- The peer set is rebuilt from scratch on every resolution, so an IP that once belonged to `app_proxy` can't stay authorised after Docker recycles it to another app.

**Widget (the only new surface).** `widgets:` in the manifest with a `three-stats` widget backed by `GET /api/widget` — unauthenticated by design (umbreld fetches it server-side without cookies) and secret-free: tunnel status text, healthy-service count and the public domain. Nothing else is reachable without auth.

**Smaller things:**
- `exports.sh` now uses `derive_entropy` (was a manual `sha256sum` over `APP_SEED`); `PROXY_AUTH_ADD: "true"` dropped since auth is the framework default.
- `backupIgnore` for regenerable data only (Let's Encrypt certs under `caddy/data/`, health snapshots); config and keys are always backed up, and the Caddyfile is regenerated from saved config on startup so a restore never leaves Caddy waiting on a placeholder.
- From an internal audit pass: `GET /api/config` now masks the preshared key as well as the private key (the VPS script, generated server-side, is their only consumer).
- Manifest `port` stays at `3019`; gallery refreshed (5 screenshots).

Tested end-to-end on umbrelOS 2.0 beta (UI via the in-process gateway, widget on the home screen, key regeneration + save, tunnel + HTTPS service reachable) and re-tested on umbrelOS 1.x with no change in behaviour. Passes `.tools/lint-apps.mjs` with 0 errors.

New web digest: `sha256:634c78bab179fcb6d865cbd93c52252762480e670609de41e72e0bcd200a1f7a` (amd64 + arm64). `wg` image unchanged (`1.0.6`).

Thanks again for all the review rounds — happy to adjust anything.

PR #5764, opened 2026-06-18
Update Bolt12 Pay to v0.2.108.1

Alex71btc:
## Type

App update

## App

App ID: bolt12-pay
Upstream project: https://github.com/Alex71btc/lndk-pay
Version: 0.2.108.1

## Summary

Bugfix release for v0.2.108.

Improved reliability of BOLT12 offer creation by automatically retrying transient LNDK TLS connection failures that have been observed on some Umbrel systems after startup.

Also adds a Geyser funding link to the app description for users who would like to support ongoing development.

## Verification

Umbrel testing performed:

Environment tested:
- [x] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [x] amd64
- [ ] arm64

- None

## Notes

Includes all features introduced in v0.2.108:

- Privacy Mode
- Private BIP353 Lightning Addresses
- Offer History BIP353 publishing
- Automatic DNS cleanup

This release only adds a reliability fix and funding metadata.

### DIFF
--- a/bolt12-pay/docker-compose.yml
+++ b/bolt12-pay/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8081
       PROXY_AUTH_WHITELIST: "/,/alias,/alias/*,/.well-known,/.well-known/*,/api/lnurl,/api/lnurl/*,/api/qr,/api/qr/*,/icon.svg,/icon.png,/service-worker.js,/public.webmanifest,/assets,/assets/*"
   web:
-    image: alex71btc/bolt12-pay:0.2.108@sha256:33df5203b894ef102b4e9823ee7005548e7304747b6466347cfc31a2c89a5910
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @Alex71btc. Just checking: you've got the `version` in the manifest as 0.2.108, but the image is tagged 0.2.109. Is this intentional?

Alex71btc:
@nmfretz Yes I made the docker image first then decided, that to be consistent with startos it should be 0.2.108:1 there, because it's just a bug fix and not a feature update.
Therefore I did not bother to push again another docker image and get the digest, but if you prefer it to be consistent, I can push a new docker image 0.2.108.1 and update the PR, no problem.

nmfretz:
No worries @Alex71btc, was just checking in case! Going live.

FYI, Docker actually preferentially uses the digest when included and ignores the tag... so technically you could just write this here with no docker build/published needed:

```yml
image: alex71btc/bolt12-pay:0.2.108.1@sha256:681c1a2af8fb0f3b4f439f915f0ec04f8bf60ccfbc7f0ec761bb3eb5f59effb4
```

Alex71btc:
Ah good to know! Thanks @nmfretz

PR #5770, opened 2026-06-19
Lnd 021 warning clean

Alex71btc:
This PR adds an important compatibility notice for BOLT12 Pay users.

Current BOLT12 functionality relies on LNDK's custom onion messaging integration using custom message type 513. LND v0.21 introduced native onion messaging support and is currently not compatible with the LNDK version used by BOLT12 Pay.

Without this notice, users may upgrade Lightning to LND v0.21 and unexpectedly lose BOLT12 Offer payment functionality.

Changes:

    Added LND v0.21 compatibility warning to the app description.
    Clarified that the protocol.custom-* configuration is only applicable to LND v0.20.x.
    Added release notes warning for users considering a Lightning upgrade.

No functional code changes.
Documentation update only.


Alex71btc:
@nmfretz unfortunately as it looks LND 0.21 seems to break Lndk quite thoroughly. 
Still have to work my way through the protocol changes.

Alex71btc:
Users that want to continue using this app should restrain vom updating lnd. Claude code says a fix in lndk is not trivial but possible altogether. I will see if i get lucky but the lnd update will break my app unfortunately 🫤

PR #5771, opened 2026-06-19
App Submission: EE Gateway

jmarbach:
## Type

New app

## App

App ID: `ee-gateway`
Upstream project: https://github.com/Encrypted-Energy/gateway
Version: 0.10.3

## Summary

Self-hosted Bluetooth gateway for open BLE networks. An always-on Umbrel
is the perfect ground node for Bluetooth-LE devices that need to reach
the cloud. EE Gateway turns it into one.

It scans with the host's Bluetooth radio for BLE devices from supported
open networks (Hubble Network today; more partners coming) and forwards
their packets to the right upstream cloud. Free, open source, runs in
the background, no maintenance.

**Architecture:** Two containers, both running as uid 1000:

- `worker` — scans BLE, queues packets to SQLite, drains to
  `encryptedenergy.com` over outbound HTTPS.
- `ui` — Flask dashboard behind `app_proxy` on port 8083. Read-only on
  the worker's status file, no Bluetooth, no outbound calls.

State lives in `${APP_DATA_DIR}/data` as bind mounts. Already shipping at
`v0.10.3` via the
[encryptedenergy community app store](https://github.com/Encrypted-Energy/umbrel-community-app-store/tree/main/encryptedenergy-ee-gateway).

## Verification

Environment tested:
- [x] Umbrel device (Raspberry Pi 5, arm64)
- [ ] Not runtime tested

Architecture tested:
- [ ] amd64
- [x] arm64

Both images (`encryptedenergy/ee-gateway-worker:0.7.4`,
`encryptedenergy/ee-gateway-ui:0.6.2`) are published multi-arch
(`linux/amd64` + `linux/arm64`); amd64 verified by image manifest,
runtime amd64 testing pending. Install, restart, stop/start, and
reinstall all preserve data across `${APP_DATA_DIR}/data`.

## Notes

**Why the worker needs host network + NET_ADMIN/NET_RAW + /dev:**
Bluetooth scanning is the product. The worker talks to the host's `bluetoothd`
over the D-Bus system bus (mounted read-only), and the BLE stack
requires `NET_ADMIN` + `NET_RAW` to manage the radio and read raw
frames. Host networking is the only path that exposes the host's
Bluetooth radio to the container — Docker doesn't bridge BlueZ.

`/dev:/dev` is mounted so an optional USB GPS dongle (vehicle-mounted
gateways) becomes visible without rebuilding the container. The
`device_cgroup_rules` lock access to char-major 188 (USB serial) and
166 (CDC ACM) only. No other devices are reachable.

Explicitly NOT used: `privileged: true`, Docker socket, named volumes,
public listening ports. The `ui` container is fully unprivileged.

**Dependencies:** None. 
**Default credentials:** None. Entered by the user in the in-app setup.

**Links:** 
- [website](https://encryptedenergy.com) ·
- [source (GPL-3.0)](https://github.com/Encrypted-Energy/gateway) ·
- [support](https://github.com/Encrypted-Energy/gateway/issues)

**Assets** 
Icon + three gallery screenshots attached inline below for
the Umbrel team to seed the official assets.

icon

1-setup
2-dashboard
3-status

### DIFF
--- a/ee-gateway/docker-compose.yml
+++ b/ee-gateway/docker-compose.yml
@@ -35,7 +35,7 @@ services:
   # extra download, and it is the only place root is used. This is NOT privileged
   # mode; the app containers themselves stay non-root.
   init_perms:
-    image: encryptedenergy/ee-gateway-worker:0.7.5@sha256:e1ce654087e6db114ea8bbb6ae195e96b85330bd3108d6e1616c3a9bee4880d3
+    image: encryptedenergy/ee-gateway-worker:0.7.5@sha256:0b58f6e9e1b0ca5c15150b23ac66958a27fac84db5487581495bf9a0a2c99999
     user: "0:0"
     entrypoint: ["sh", "-c", "chown -R 1000:1000 /data"]
     restart: on-failure
@@ -66,7 +66,7 @@ services:
   # Python process runs as uid 1000; gpsd starts as root only to open the
   # serial device, then drops to user nobody.
   worker:
-    image: encryptedenergy/ee-gateway-worker:0.7.5@sha256:e1ce654087e6db114ea8bbb6ae195e96b85330bd3108d6e1616c3a9bee4880d3
+    image: encryptedenergy/ee-gateway-worker:0.7.5@sha256:0b58f6e9e1b0ca5c15150b23ac66958a27fac84db5487581495bf9a0a2c99999
     init: true
     restart: on-failure
     depends_on:
--- a/ee-gateway/docker-compose.yml
+++ b/ee-gateway/docker-compose.yml
@@ -35,10 +35,10 @@ services:
   # extra download, and it is the only place root is used. This is NOT privileged
   # mode; the app containers themselves stay non-root.
   init_perms:
-    image: encryptedenergy/ee-gateway-worker:0.7.4
+    image: encryptedenergy/ee-gateway-worker:0.7.4@sha256:c9487c7d138e871fd49298ba4f2a4ba2e57173a4e24cfd30e57774026dbcc898
     user: "0:0"
     entrypoint: ["sh", "-c", "chown -R 1000:1000 /data"]
-    restart: "no"
+    restart: on-failure
     volumes:
       - ${APP_DATA_DIR}/data:/data
 
@@ -46,8 +46,9 @@ services:
   # no added capabilities, no Bluetooth access. Runs as uid 1000 (set in the
   # image). init: true so SIGTERM reaches Flask for a clean shutdown.
   ui:
-    image: encryptedenergy/ee-gateway-ui:0.6.2
+    image: encryptedenergy/ee-gateway-ui:0.6.2@sha256:c565138959c1498706d4954521ff7630cb41f8fec4b2ec1ef91e6a2d36f15a4e
     init: true
+    restart: on-failure
     depends_on:
       init_perms:
         condition: service_completed_successfully
@@ -65,7 +66,7 @@ services:
   # Python process runs as uid 1000; gpsd starts as root only to open the
   # serial device, then drops to user nobody.
   worker:
-    image: encryptedenergy/ee-gateway-worker:0.7.4
+    image: encryptedenergy/ee-gateway-worker:0.7.4@sha256:c9487c7d138e871fd49298ba4f2a4ba2e57173a4e24cfd30e57774026dbcc898
     init: true
     restart: on-failure
     depends_on:
--- a/ee-gateway/docker-compose.yml
+++ b/ee-gateway/docker-compose.yml
@@ -1,0 +1,111 @@
+# EE Gateway: Umbrel app composition
+# Copyright (C) 2026 encryptedenergy.com
+# SPDX-License-Identifier: GPL-3.0-only
+#
+# Two services share ${APP_DATA_DIR}/data (mounted at /data) for file-based
+# IPC: the UI writes config.json, the worker writes state.json and packets.db.
+# They never communicate over the network, so the worker can run on the host
+# network (which it needs to reach the Bluetooth radio) while the UI stays on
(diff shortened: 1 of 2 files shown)


jmarbach:
Heads up for the reviewer: completing setup requires a free encryptedenergy.com account (the in-app setup wizard intakes an org ID + API token). Sign up at https://encryptedenergy.com/sign-up.

nmfretz (commented):
Thanks for the submission @jmarbach.

I found one host-networking issue that we should tighten before I test this further. The worker needs host networking for Bluetooth, which makes sense, but the worker image currently starts `gpsd` with `-G` when a GPS dongle is present. Because that service runs in `network_mode: host`, this exposes gpsd on the Umbrel host at `0.0.0.0:2947`.

From the upstream worker code, it looks like gpsd is only consumed locally at `127.0.0.1:2947`, so I don't think LAN access is needed here. Can you rebuild the worker image with gpsd loopback-only, then re-pin the digest in this package? Dropping `-G` appears to preserve the local worker path while avoiding the extra host listener.

Once that gpsd listener is tightened up, I can re-test the install path.

ee-gateway/docker-compose.yml
```diff
@@ -0,0 +1,112 @@
...
+    image: encryptedenergy/ee-gateway-worker:0.7.4@sha256:c9487c7d138e871fd49298ba4f2a4ba2e57173a4e24cfd30e57774026dbcc898
+    init: true
+    restart: on-failure
+    depends_on:
+      init_perms:
+        condition: service_completed_successfully
+    # Host networking gives the container direct use of the host Bluetooth
+    # stack. Docker does not bridge BlueZ, so this is the only path that
+    # exposes the host's Bluetooth radio to the container. The worker uses
+    # the host network only for outbound HTTPS; it does not bind any
```
nmfretz: Can we tighten this part? The worker image starts `gpsd` with `-G` when a GPS device is present. Since this service runs with `network_mode: host`, that exposes gpsd on the Umbrel host at `0.0.0.0:2947`, even though the package notes say the worker does not bind any listening port.

From the upstream worker code, gpsd is only consumed locally at `127.0.0.1:2947`, so I think this can be loopback-only. Dropping `-G` appears to keep the worker path working while avoiding the LAN-facing host listener. Can you rebuild/re-pin the worker image with that tightened up, and update this comment/PR notes to mention any remaining localhost gpsd listener?

jmarbach:
Good catch, thanks. You're right on both counts: `gpsd -G` plus `network_mode: host` was binding 2947 to the LAN, and the PR description's "no listening port" line wasn't accurate once the GPS dongle came into play.

Pushed worker 0.7.5 (manifest 0.10.4). The entrypoint drops `-G`, so gpsd now binds 127.0.0.1:2947 only. The worker reads from that same loopback (gps.py), so nothing changes on the functional side.

I also rewrote the docker-compose comment block. It now describes the loopback gpsd honestly instead of claiming there's no listening port at all.

New worker image digest pinned in this commit: `sha256:e1ce654087e6db114ea8bbb6ae195e96b85330bd3108d6e1616c3a9bee4880d3`.

Ready for re-test whenever you've got a minute. Thanks again for the careful read.

jmarbach:
Source change for reference: [ Encrypted-Energy/gateway@57e7e8b#diff-worker/entrypoint.sh
](https://github.com/Encrypted-Energy/gateway/commit/57e7e8b26ea1e19168ffebfcfba56bb2396ff8c2#diff-worker/entrypoint.sh)

nmfretz (commented):
Thanks @jmarbach, and thanks for jumping on the source change.

I re-tested the currently pinned worker image from this PR:

`encryptedenergy/ee-gateway-worker:0.7.5@sha256:e1ce654087e6db114ea8bbb6ae195e96b85330bd3108d6e1616c3a9bee4880d3`

It looks like the published image still has the old worker content. Inside the image, `/entrypoint.sh` still starts gpsd with `-G`:

`if gpsd -n -G $DEVICES; then`

and the installed worker package still reports `0.7.4`.

The upstream source commit you linked looks right, so I think this is just the Docker image/manifest needing another pass. Could you rebuild and push `encryptedenergy/ee-gateway-worker:0.7.5` from the fixed source, then re-pin this package to the new multi-arch manifest-list digest?

Could you also bump `ee-gateway/umbrel-app.yml` to `version: "0.10.4"`? The compose file now sets `EE_GATEWAY_VERSION: "0.10.4"`, but the app manifest is still on `0.10.3`.

Once those are updated I can re-test the package.

jmarbach:
You're right, thanks for catching that. The previous push tagged 0.7.5 but the image had been built before my source patch landed locally, so the published bytes still had `gpsd -G` and worker 0.7.4 inside. Operator error on my end.

Rebuilt from the corrected source (Encrypted-Energy/gateway@57e7e8b) and verified the new image locally before pushing: `/entrypoint.sh` now reads `if gpsd -n $DEVICES; then` (no `-G`) and the worker package reports `0.7.5`.

The new multi-arch index digest is pinned in this commit. Also bumped `umbrel-app.yml` to `0.10.4` to match the compose `EE_GATEWAY_VERSION`.

New worker image digest: `sha256:0b58f6e9e1b0ca5c15150b23ac66958a27fac84db5487581495bf9a0a2c99999`

Ready for another re-test.

nmfretz:
Awesome @jmarbach I'll take a look shortly. In the meantime are you okay with these for gallery assets?
https://github.com/getumbrel/umbrel-apps-gallery/pull/117/changes

nmfretz:
Alrighty @jmarbach thanks for making the change. This all looks good now. 

I haven’t completed the external Encrypted Energy account/token flow or real packet forwarding on my side, so I’ll rely on your app-level testing for that part.

I've pushed some finalization commits and we're good to go live! Thanks again for submitting this 🚀

image

PR #5773, opened 2026-06-20
Add Hestia Control Panel (non-privileged, multi-arch)

ptrinh:
## Hestia Control Panel

Adds [HestiaCP](https://hestiacp.com), an open-source web hosting control panel, as an Umbrel app - the full panel, with every bundled service's ports published and its state persisted (the review's "fully wire and document" option).

### What makes this Umbrel-friendly
HestiaCP normally needs systemd + `privileged: true`. This uses a purpose-built, non-privileged, multi-arch image ([ghcr.io/ptrinh/hestiacp](https://github.com/ptrinh/hestiacp-docker)):

- No privileged mode, no extra capabilities. systemd is replaced with a `systemctl` shim ([docker-systemctl-replacement](https://github.com/gdraheim/docker-systemctl-replacement)); the privilege-requiring features (firewall / fail2ban / disk quota) are disabled.
- Multi-arch: linux/amd64 + linux/arm64, built in CI, pinned by an immutable `build-<run_id>` tag + `@sha256` digest.
- Latest stable HestiaCP, installed from HestiaCP's `release` branch; weekly CI rebuild.
- Full panel, fully wired: web hosting (nginx + Apache with PHP-FPM), databases (MariaDB + PostgreSQL, with phpMyAdmin/phpPgAdmin served through the panel), mail (Exim + Dovecot), DNS (Bind), FTP (vsftpd), File Manager, cron and backups.
- Every service users can enable in the panel is reachable through the package: mail on standard ports 25/465/587 + 110/995/143/993, DNS on 53 tcp+udp, FTP on 21 with its passive range 12000-12100 published 1:1 (all verified unoccupied on umbrelOS 1.7.3). The mail queue is persisted in a dedicated exim-spool volume; mailboxes, DKIM keys, DNS zones and FTP accounts live in the existing persisted volumes, with the /etc configs regenerated from that data on every start.

### How it works
- The panel serves self-signed HTTPS on 8083. Since app_proxy validates upstream TLS, a small socat sidecar tunnels HTTP to TLS to it (PROXY_AUTH_ADD off; HestiaCP has its own login). The browser only ever sees Umbrel's proxy, so there is no certificate warning.
- app_proxy `depends_on` the hestia service becoming healthy, and the healthcheck requires the admin password to be applied AND the login UI to actually render - so Umbrel only reports ready when the panel is usable.
- No port conflict: the panel is reached via app_proxy; HestiaCP does not claim Umbrel's 80/443. Hosted sites are published on 9088 (HTTP) and 9448 (HTTPS).
- Persistence: data lives under ${APP_DATA_DIR}/data/* (MariaDB, PostgreSQL, HestiaCP config, /home, exim mail queue, backups); the image seeds these on first run.
- Login: admin / the app password Umbrel shows. The image entrypoint applies APP_PASSWORD on first start (before the app reports ready).
- sshd runs inside the container (the File Manager lists files over internal SFTP) but its port is not published - no external SSH surface.

### Image source
- Image repo (Dockerfile + CI): https://github.com/ptrinh/hestiacp-docker
- Upstream: https://hestiacp.com

### Testing (x86_64 / amd64 umbrelOS)
A scripted end-to-end suite (`tests/e2e.sh` in the image repo) drives the panel like a browser on a fresh install: login page renders -> login -> create hosting user -> add web domain -> site serves -> create MariaDB + PostgreSQL DBs (phpMyAdmin/phpPgAdmin render) -> File Manager upload -> PHP executes and connects to both DBs -> per-domain self-signed SSL served via SNI on 9448 -> hosting user hosts their own domain -> cron runs -> mail domain + mailbox created, message delivered over SMTP and read back over IMAP -> DNS zone answers SOA/A via dig on 53 -> FTP login + upload served by the site -> scheduled backup completes -> app restart: panel, sites and PHP recover. Container runs Privileged=false, CapAdd=[]. CI also runs an unprivileged smoke test that render-checks /login/.

### Honest caveats
- On the very first domain you add, HestiaCP may print "Restart of nginx failed" although the site serves fine; later adds are clean.
- After an app restart, a hosted site can take roughly a minute to come back while the on-start config rebuild runs (the panel is up immediately).
- arm64: the image builds and CI smoke-tests it, but the full hosting flow was validated on amd64 umbrelOS only.
- No in-panel firewall/fail2ban/quota (handle network security at the Umbrel/router level).
- Receiving internet mail needs port 25 forwarded to the Umbrel and correct MX/rDNS; most residential ISPs block OUTBOUND 25, so sending usually needs a smarthost (configurable in HestiaCP).
- FTP passive mode advertises the container IP; clients with "use control-connection address" fallback (most do) work out of the box. jailed FTP/SFTP chroots need privileges this container doesn't have, so FTP lands in the user's real home.
- For a hosting user's SSH Access shell use "bash" (or sh/dash/rbash); "jailbash" is not supported on this unprivileged build.

### Logo and screenshot
- Logo: https://raw.githubusercontent.com/hestiacp/hestiacp/main/web/images/logo.png

HestiaCP panel

(gallery is [] and icon omitted per new-submission lint guidance, so Umbrel can generate the final assets.)


hestiacp/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+  sslproxy:
+    image: alpine/socat:1.8.0.0@sha256:a6be4c0262b339c53ddad723cdd178a1a13271e1137c65e27f90a08c16de02b8
+    restart: on-failure
+    command:
+      - "TCP-LISTEN:8081,fork,reuseaddr"
+      - "OPENSSL-CONNECT:hestiacp_hestia_1:8083,verify=0"
+
+  hestia:
+    # Non-privileged, multi-arch HestiaCP (built in github.com/ptrinh/hestiacp-docker).
+    image: ghcr.io/ptrinh/hestiacp:c6d176ee763a0cb07bc271e77c7484982b372d86@sha256:e1939cd9d619e89eb845bcf124a04b29d217b7eca0fff576d5cb9705e12a8883
```
nmfretz: Looks like this tag has moved since the package was pinned. `npm run lint:apps -- hestiacp --check-images` warns that the tag now resolves to a different manifest digest.

Can you re-pin this to the exact multi-arch image you want shipped after rebuilding the image?

hestiacp/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+      - ${APP_DATA_DIR}/data/hestia:/usr/local/hestia/data
+      - ${APP_DATA_DIR}/data/hestia-conf:/usr/local/hestia/conf
+      - ${APP_DATA_DIR}/data/backup:/backup
+    # Hosted sites on alternate host ports so HestiaCP never claims Umbrel's
+    # 80/443. The panel itself is reached through app_proxy, not these.
+    ports:
+      - "9088:80"
+      - "9448:443"
+    healthcheck:
+      test: ["CMD-SHELL", "curl -k -sf https://localhost:8083/ >/dev/null || exit 1"]
```
nmfretz: This healthcheck is a bit too shallow for this app.

On my Umbrel test install the app reached `ready`, but `/login/` only returned the Hestia PHP fatal about missing `vendor/autoload.php`. It'd be better to check a path that fails when the panel UI is broken.

hestiacp/hooks/post-start
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+# Set HestiaCP's admin password to Umbrel's app password on first start (once).
+CONTAINER="hestiacp_hestia_1"
```
nmfretz: This should go through Umbrel's app script instead of calling Docker directly.

For example, LibreChat uses `"${UMBREL_ROOT}/scripts/app" compose "${APP_ID}" exec -T ...` in its hook: https://github.com/getumbrel/umbrel-apps/blob/master/librechat/hooks/pre-start#L40

That keeps this tied to Umbrel's app lifecycle and the compose service name instead of direct Docker/runtime details.

Longer-term, since you're maintaining the wrapper image, it may be cleaner for the image entrypoint to accept the desired admin password and only expose Hestia after applying it. Not a blocker for this PR, but the current backgrounded `post-start` hook means Umbrel can mark the app ready before password setup has definitely completed.

hestiacp/umbrel-app.yml
```diff
@@ -0,0 +1,45 @@
...
+
+
+  Notes:
+
+  - The panel uses a self-signed TLS certificate, so your browser shows a
+    one-time certificate warning.
+
+  - Log in with username "admin" and the password shown on this page.
+
+  - HestiaCP does not claim Umbrel's ports 80/443; hosted sites are published on
```
nmfretz: Let's name the hosted-site ports in the description here.

Exposing Hestia's site ports makes sense for this app, but since Umbrel can't give it normal host `80/443`, users should know that hosted sites are published on `9088` and `9448`.

nmfretz (commented):
Thanks for the submission @ptrinh.

I tested this on x86_64 umbrelOS and the current image is not opening cleanly yet. The app reaches `ready`, but the launch URL shows Hestia's `Unable to load required libraries` error instead of the login UI.

I also hit a hosting-path issue after adding a basic web domain: nginx fails to restart because `:80` is already held by Apache. I left the details inline.

Once the image is rebuilt and re-pinned, could you re-test a fresh Umbrel install through login, adding a hosted site, restarting the app, and confirming both the panel and hosted site still work?

ptrinh:
Thanks for the careful review @nmfretz - all fair, and the panel issue was a real miss (I'd been checking HTTP status, not the rendered page). Rebuilt the image and addressed each point:

1. Panel "Unable to load required libraries" (vendor/autoload.php): the installer never ran `v-add-sys-dependencies`, so HestiaCP's composer deps were missing. The image now installs them at build (they land in `web/inc/vendor`), and the login UI renders. This was also the cause of the php-fpm restart failing during domain-add.

2. Healthcheck: now render-checks `/login/` (verifies the login form is present) instead of `/`, so the app only reports ready when the panel actually works. The CI smoke test does the same now.

3. post-start hook: removed. The image entrypoint sets the admin password from `APP_PASSWORD` on first start, before the app is reported ready - no backgrounded hook, no direct docker calls.

4. Moving tag: my CI was re-tagging the same commit SHA on rebuilds. It now also pushes a unique, never-reused `build-<run_id>` tag, and the app is re-pinned to that tag + its `@sha256` digest.

5. nginx restart / `:80` held by Apache: the image now defaults to nginx-only, which makes live domain-add work (serves 200). Apache's restart is unreliable in a container (your `:80` observation), so it's now opt-in (`WITH_APACHE=yes`). Honest caveat: on the very first domain add HestiaCP may still print "Restart of nginx failed" even though the site serves; later adds are clean.

6. Ports: documented in the manifest description - hosted sites are published on 9088 (HTTP) and 9448 (HTTPS).

Re-tested on x86_64 umbrelOS: fresh install -> Open shows the rendered login UI -> log in with the app password -> add a hosted domain (serves 200) -> restart the app -> panel is up immediately and the hosted site recovers after the on-start rebuild (about a minute). Privileged=false, CapAdd=[]. Image re-pinned to `build-28066541490@sha256:46e47e46...`.

Two honest notes: (a) after an app restart a hosted site is briefly down (~1 min) while the on-start rebuild runs; (b) I validated the full hosting flow on amd64 umbrelOS - arm64 is built and CI-smoke-tested but I don't have an arm64 node to run the full flow on.

nmfretz:
Thanks for making the changes @ptrinh! This is definitely getting closer.

I tested a fresh install on x86_64 umbrelOS. The app installs and the Hestia login page now renders, but I can’t log in... after entering the default credentials, it just sends me back to the first login step.

Could you take another look and confirm the full login flow works from a fresh Umbrel install?

ptrinh:
Found and fixed it - thanks for the precise repro @nmfretz.

Two compounding causes (both confirmed by reproducing the bounce-to-login):

1. PHP sessions weren't writable. The panel's session dir (`/usr/local/hestia/data/sessions`) lives on the bind-mounted data volume and came back root-owned, but hestia-php runs as `hestiaweb` -> `session_start(): Permission denied` -> no CSRF token in the session -> every login bounced. The entrypoint now chowns that dir to `hestiaweb` on start.

2. The session cookie was Secure. The bundled php-fpm forced `session.cookie_secure=on` + `SameSite=Strict`; behind Umbrel's plain-HTTP app_proxy the browser drops a Secure cookie. Changed to non-Secure + `SameSite=Lax`.

I also gated the healthcheck on the admin-password step, so the app only reports healthy once login actually works (previously it could report ready a few seconds before the password was applied - a small race I hit while testing).

Re-tested on x86_64 umbrelOS, fresh install, through the real flow:
- App goes healthy only when usable; admin login -> `/list/user/` dashboard.
- Created a second user and logged in as that user -> `/list/web/` dashboard.
- Added a web domain under that user -> serves HTTP 200.
- Also verified per-user SSH works when enabled (SSH is off in the app by default): created a user, set bash shell + key, SSH'd in as them.

Re-pinned to `ghcr.io/ptrinh/hestiacp:build-28230633996@sha256:8f287116...`.

One honest caveat (unchanged): on the very first web domain you add, HestiaCP may print "Restart of nginx failed" even though the site serves fine; later adds are clean.

ptrinh:
I did some more thorough tests. It's ready for review now @nmfretz. Thanks a lot!

hestiacp/docker-compose.yml
```diff
@@ -0,0 +1,52 @@
+version: "3.7"
+
+services:
+  # HestiaCP serves a self-signed HTTPS panel on 8083. Umbrel's app_proxy
+  # validates upstream TLS and can't skip it, so a tiny socat sidecar tunnels
+  # plain HTTP to TLS to the panel. PROXY_AUTH_ADD is off (HestiaCP has its
+  # own login).
+  app_proxy:
```
nmfretz, line 8: Umbrel isn't currently waiting for the `hestia` healthcheck here. On both fresh install and restart, the app reached `ready` while `hestia` was still `health: starting`, and opening it returned `502`.

Can we make `app_proxy` depend on Hestia becoming healthy so Umbrel waits until the panel is actually usable?

```suggestion
  app_proxy:
    depends_on:
      hestia:
        condition: service_healthy
```

hestiacp/umbrel-app.yml
```diff
@@ -0,0 +1,52 @@
...
+  privilege-requiring features (firewall, fail2ban, disk quota) are disabled, so
+  it runs with no privileged mode and no extra Linux capabilities. The full
+  panel is intact: Nginx + Apache, PHP-FPM, MariaDB + PostgreSQL, mail
+  (Exim/Dovecot/Roundcube), DNS, FTP, File Manager, cron, phpMyAdmin and
+  phpPgAdmin.
+
+
+  Notes:
+
+  - The panel uses a self-signed TLS certificate, so your browser shows a
```
nmfretz: The browser never sees Hestia's self-signed certificate here. The socat sidecar handles that upstream TLS connection, while users open the panel through Umbrel's HTTP app proxy. I didn't get a certificate warning in testing, so can we remove this note?

nmfretz (commented):
Sorry for leaving this for a while @ptrinh. I gave the latest build another pass on arm64 umbrelOS, and the previous login issue is fixed now — thanks for working through that.

The main thing that still needs tightening is the scope of the new full-panel build. The image now starts mail, DNS, FTP and SSH, but the package does not publish the ports those services need. I confirmed the daemons are listening inside Hestia and are reachable by sibling containers on Umbrel's shared app network, but external clients can't reach them through the app. FTP would also need its passive data ports, and Exim's queue at `/var/spool/exim4` isn't persisted.

So users can't actually use those features as packaged, even though the description says the full panel is intact. Can we either scope the image and description back to the functionality users can reach through the package — web hosting, databases and File Manager — which seems like the cleanest fit for Umbrel, or fully wire and document the additional services and their persistence requirements?

The PR description also still says nginx-only with Apache opt-in, so that will need updating to match whichever setup we land on.

I've left two smaller lifecycle/copy notes inline. Thanks again for all the work you've put into this.

ptrinh:
Thanks for the detailed pass @nmfretz. All points addressed in d16d9d9:

**Scope** - went with your recommended option and scoped the image back to
what the package can actually deliver. The image is rebuilt without mail,
DNS and FTP (`--exim no --dovecot no --named no --vsftpd no`), so the panel
no longer offers features whose ports the package does not publish. What
remains: web hosting (nginx + Apache with PHP-FPM), MariaDB + PostgreSQL
(phpMyAdmin/phpPgAdmin through the panel), File Manager, cron and backups.
sshd still runs inside the container because the File Manager lists files
over internal SFTP, but its port stays unpublished (no external SSH
surface). New pin: `build-29484723625` (multi-arch, digest-pinned,
CI smoke-tested unprivileged).

**Lifecycle** - added your suggested `depends_on: hestia:
condition: service_healthy` to `app_proxy`, so Umbrel waits for the
healthcheck (password applied + login UI actually rendering) before
reporting ready. No more 502 window on install/restart.

**Copy** - removed the self-signed certificate note from the description
(you are right, the socat sidecar terminates that hop and the browser only
sees the HTTP proxy), and updated the description to the scoped feature
set. The PR description is also updated to drop the stale nginx-only
wording.

`npm run lint:apps -- hestiacp --check-images` passes on the branch.

CI ran its unprivileged smoke test against the new build (login UI
renders, no PHP fatal). I will re-run the full flow on a fresh umbrelOS
install (install -> login -> add web domain -> restart -> panel and site
recover) and report back here. Happy to adjust anything else.

ptrinh:
Following up with the promised re-test. Running the full flow on a fresh
umbrelOS 1.7.3 (amd64) install surfaced two more restart bugs, now fixed
and re-pinned as `build-29504071755` (ccb4a2a):

1. After an app restart the container gets a new IP, and nginx could die
   with its ports half-held by old workers ("bind() ... address already in
   use" on its status listener) - hosted sites went dark while the panel
   stayed up. The systemctl shim now makes sure nginx/apache processes are
   really gone before a restart, and the entrypoint verifies the web server
   answers on the current IP:80 before finishing.

2. Admin-owned domains served the default page after a restart: the
   entrypoint's repoint via `v-change-web-domain-ip` always fails once the
   old IP is deregistered, and `v-rebuild-user` only rebuilds web domains
   when it had to create the system user - admin is baked into the image's
   /etc/passwd, so admin-owned vhosts were never regenerated. The
   entrypoint now rewrites the stale IP in the persisted data before the
   rebuild and runs `v-rebuild-web-domains` explicitly per user.

The healthcheck also now requires the hosted-site web server to answer (in
addition to the password sentinel and the login UI rendering), so Umbrel
will not report the app ready while sites are down.

To keep this verifiable, the image repo now ships a scripted end-to-end
test (`tests/e2e.sh` in ptrinh/hestiacp-docker) that drives the panel like
a browser: login page renders -> login -> create hosting user -> add web
domain -> hosted site serves its own docroot on 9088 -> restart the app ->
panel and site recover. Full run against a fresh install on umbrelOS
1.7.3 amd64 (fresh app data, install -> ready -> e2e with restart):
9 passed, 0 failed. `npm run lint:apps -- hestiacp --check-images` passes.

hestiacp/docker-compose.yml
ptrinh: Added exactly as suggested in d16d9d9 - app_proxy now waits for the hestia healthcheck (which itself requires the admin password applied, the login UI rendering, and as of ccb4a2a the hosted-site web server answering). Verified on a fresh install: no 502 window anymore.

hestiacp/umbrel-app.yml
ptrinh: Removed in d16d9d9 - you are right, the socat sidecar terminates that hop and the browser only ever talks to Umbrel's HTTP proxy. The description now reflects the scoped feature set as well.

ptrinh:
Update: we went with your second option after all - the additional services
are now fully wired and documented rather than removed (c8f17bc, pin
`build-29544877786`).

**What's wired**

- Mail (Exim + Dovecot): standard ports 25/465/587 and 110/995/143/993
  published; the mail queue is persisted in a new exim-spool volume
  (mailboxes and DKIM keys were already inside the persisted home/data
  volumes); per-user mail config is rebuilt from persisted data on every
  start. The description documents that receiving internet mail needs port
  25 forwarded and that most ISPs block outbound 25 (smarthost supported).
- DNS (Bind): 53 tcp+udp published; zones rebuilt from persisted data on
  start. All these standard ports are unoccupied on umbrelOS 1.7.3.
- FTP (vsftpd): 21 plus the passive range 12000-12100 published 1:1.

**Bugs this surfaced (fixed in the image)**

- vsftpd chrooted users into Hestia's SFTP jail, which needs mount
  privileges this container doesn't have - every FTP CWD failed. It now
  chroots to the real home, same as the File Manager fix.
- The installer bakes the BUILD machine's public IP into vsftpd's
  `pasv_address` - dropped at build (runtime `FTP_PASV_ADDRESS` env can
  re-add one).
- Exim's default DNSBLs reject ALL inbound mail behind public DNS
  resolvers (Spamhaus answers "open resolver" and exim treats it as
  listed) - umbrelOS resolves publicly, so this bit immediately. DNSBLs
  are now off by default and re-addable for hosts with a local resolver.

**Verification** - the scripted e2e suite (`tests/e2e.sh` in the image
repo) now covers every shipped service. Fresh install on umbrelOS 1.7.3
amd64: login, hosting user, web domain, hosted site, MariaDB + PostgreSQL
(+ phpMyAdmin/phpPgAdmin through the panel), File Manager upload, PHP
executing and connecting to both DBs, per-domain SSL via SNI on 9448,
user-owned domain, cron, mail (SMTP delivery -> IMAP read-back), DNS (SOA
and A answered by bind on 53), FTP (login + upload served by the site),
scheduled backup, and app-restart recovery of panel/site/PHP:
**31 passed, 0 failed, 0 skipped**. Lint passes (one advisory warning
about the published FTP passive port range).

The PR description is updated to match. Happy to adjust port choices or
scope if you'd prefer different trade-offs.

ptrinh:
Re the `ports.unresolved` lint warning on the FTP passive range: the linter's port parser only resolves single port numbers, so any published range warns regardless of conflicts. The range 12000-12100 is verified unoccupied on umbrelOS 1.7.3, and there is store precedent for published ranges: neko publishes `52000-52100:52000-52100/udp` the same way.

hestiacp/docker-compose.yml
```diff
@@ -0,0 +1,76 @@
...
+      - "21:21"                  # FTP control
+      - "12000-12100:12000-12100"  # FTP passive data range
+    # Only report healthy once the panel really works: the admin password has
+    # been applied (sentinel), the login UI renders (a plain request to /
+    # still 302s even when the PHP UI is broken), AND the hosted-sites web
+    # server answers on the container IP (any HTTP response counts). This
+    # avoids both "ready but can't log in yet" and "ready but hosted sites
+    # are down" after install/restart.
+    healthcheck:
+      test: ["CMD-SHELL", "test -f /usr/local/hestia/data/.admin-password-set && curl -k -fs https://localhost:8083/login/ | grep -q 'name=.user.' && curl -s -o /dev/null --max-time 5 http://$$(hostname -i | awk '{print $$1}'):80/ || exit 1"]
```
nmfretz, line 72: Can we require a successful response here? `curl -s` exits successfully for a `502`, so the app can still be marked healthy while hosted sites are down.

```suggestion
    # server returns a successful HTTP response. This avoids both "ready but
    # can't log in yet" and "ready but hosted sites are down" after
    # install/restart.
    healthcheck:
      test: ["CMD-SHELL", "test -f /usr/local/hestia/data/.admin-password-set && curl -k -fs https://localhost:8083/login/ | grep -q 'name=.user.' && curl -fs -o /dev/null --max-time 5 http://$$(hostname -i | awk '{print $$1}'):80/ || exit 1"]
```

hestiacp/umbrel-app.yml
```diff
@@ -0,0 +1,65 @@
...
+    may require a smarthost (configurable in HestiaCP).
+
+  - DNS (port 53) serves the zones you create in the panel; point your
+    domain's NS records at your public IP only if you really want to host
+    authoritative DNS at home.
+
+  - FTP runs on port 21 with passive ports 12000-12100. Prefer the built-in
+    File Manager or SFTP where possible; if an FTP client fails to list
+    directories, enable its "use control-connection address for passive
+    mode" option.
```
nmfretz, line 45: Can we make the default private-address behavior explicit here? Hestia currently uses its container address in generated DNS records and FTP passive responses, so the existing instructions can still leave external clients unable to connect.

```suggestion
  - DNS (port 53) serves zones created in the panel. Newly created zones use
    Hestia's private container address by default, so replace any generated
    private addresses with the publicly reachable address your records should
    use before delegating the domain to this server. Hosting authoritative DNS
    at home also requires forwarding both TCP and UDP port 53 to your Umbrel.

  - FTP runs on port 21 with passive ports 12000-12100. The server advertises
    its private container address for passive connections, so enable your
    client's "use control-connection address for passive mode" option. For
    access from outside your network, forward port 21 and the full passive
    range to your Umbrel.
```

nmfretz (commented):
Thanks for the thorough follow-up @ptrinh. There’s one remaining lifecycle issue after an Umbrel app restart: the panel returns and the app reports ready/healthy, but existing hosted sites return `502`.

I had Codex dig into the Docker-wrapper-specific state here, and it surfaced a likely cause: the persisted Hestia IP record survives container recreation, while `/etc` does not. Because the record already exists, startup appears to skip regenerating the corresponding Apache listener configuration. The rebuilt nginx vhosts then proxy to the current container IP on port `8080`, while Apache is only listening on `127.0.0.1:8081`.

Could you update the wrapper startup to regenerate that listener configuration after container recreation and confirm existing sites recover after an Umbrel restart?

I’ve left the package-side healthcheck fix and external-network copy changes inline. Thanks again for continuing to work through this.

ptrinh:
Confirmed and fixed - your Codex analysis was exactly right, thanks @nmfretz.

The persisted IP record made the startup IP registration a no-op when the
recreated container came back with the SAME IP, so the per-IP listener
configs (/etc/{nginx,apache2}/conf.d/<ip>.conf - apache's
"Listen <ip>:8080" lives there) were never regenerated: apache stayed on
its defaults while the rebuilt nginx vhosts proxied to <ip>:8080, hence
the 502s. Our earlier restart testing kept passing because each restart
happened to come back on a NEW IP, which takes the repoint path and
regenerates everything.

Fix (1f45203, pinned as `build-29554312819` in 9469254f): on startup, if
the IP record exists but those per-IP configs are missing, the wrapper
drops the stale record so v-add-sys-ip recreates them from templates.

Verification:

- Reproduced locally by recreating the container with a fixed IP and the
  previous image: 502, no <ip>.conf for nginx/apache, apache not listening
  on <ip>:8080. Same recreate on the fixed image: 200, both configs
  regenerated, apache listening.
- On umbrelOS 1.7.3 (amd64), fresh install then the full e2e suite:
  34 passed, 0 failed - including two app restarts. A follow-up restart
  where Docker preserved the container IP (the exact case you hit) now
  recovers: the hosted site serves its own page and the PHP+DB probe
  returns OK after the on-start rebuild (about a minute, as before).
- The e2e restart step now records the system IP across restarts and
  restarts a second time when the first one lands on a new IP, so both
  recreate paths (same-IP and changed-IP) stay covered going forward.

PR #5779, opened 2026-06-22
Update L-Town to 20260621

kaichappel:
Hey @besoeasy, we saw there was an available L Town update. You okay going live with this?


besoeasy:
@kaichappel good to go

PR #5780, opened 2026-06-22
dvb-warppool: Bitcoin solo mining pool (Stratum V1 + V2, operator UI)

dvb-projekt:
## dvb-WarpPool — Bitcoin solo mining pool

Open-source (AGPL-3.0-or-later), from-scratch **Rust** solo Bitcoin mining pool. The block reward flows straight to the operator's wallet via the coinbase — no custody, no accounts, no fees. On umbrelOS it wires up to the Bitcoin node automatically (RPC + ZMQ); a browser **setup wizard** on first run asks only for a payout address, then miners can connect.

- **Repo:** https://github.com/dvb-projekt/dvb-WarpPool
- **License:** AGPL-3.0-or-later (+ commercial) — open source
- **Handbook:** https://dvb-projekt.github.io/dvb-WarpPool/
- **Support:** https://github.com/dvb-projekt/dvb-WarpPool/issues

### Highlights
- Stratum V1 (plain + TLS) **and** a native **Stratum V2** server (NOISE, 2024 spec), enabled out of the box
- Persisted, EMA-smoothed per-worker variable difficulty (survives miner reconnects)
- Live PWA operator cockpit (SSE), 24-hour hashrate chart with best-share peaks, all-time leaderboard, coinbase decoder, full-screen kiosk/wall screensaver
- Per-miner vendor telemetry for 12 miner families (temp / power / fan / hashrate) + mDNS discovery
- **Energy + solar dashboard**: electricity cost (kWh × tariff) and, with a PV source (Anker Solix or any generic HTTP endpoint), how much of your mining runs on solar surplus
- Operator security: Argon2id admin login, optional TOTP 2FA, scoped API tokens, audit log, auto-ban + rate limiting
- Hardware-aware profiles, Prometheus `/metrics`, notifications (ntfy/Telegram/Discord/Slack/email/web-push) + outbound webhooks
- Operator UI in 8 languages; signed, reproducible, multi-arch builds

Honest status: the project is young — **no mainnet block found yet**. The store already has `public-pool` and `gobrrr-pool`; dvb-WarpPool's differentiators are native Stratum V2 + TLS, the multi-language PWA cockpit, and per-miner vendor telemetry.

### Packaging
- **Image:** `ghcr.io/dvb-projekt/dvb-warppool:1.9.0` — multi-arch (linux/amd64 + linux/arm64), digest-pinned (`sha256:956372db…`)
- **Dependencies:** `bitcoin` (RPC `getblocktemplate` / `submitblock` + ZMQ `hashblock`)
- **Ports:** `18334` web UI / app port · `18335` Stratum V1 · `18336` Stratum V2 · `18337` Stratum V1 TLS
- **Persistence:** `${APP_DATA_DIR}/data` + `${APP_DATA_DIR}/config`
- **Default credentials:** none — the admin password is set in the browser setup wizard
- **Host access:** none beyond a read-only Bitcoin RPC cookie mount; no Docker socket, no privileged mode, no broad host mounts
- Passes `node .tools/lint-apps.mjs dvb-warppool --check-images` clean (no issues).

### Screenshots
Dashboard

Hashrate chart with best-share peaks

Per-miner vendor telemetry

### Logo / source logo
dvb-WarpPool logo

---

I'll add the PR URL to the `submission` field after this PR is opened, per the packaging guide. Happy to adjust anything for umbrelOS compatibility — thanks for the review!

### DIFF
--- a/dvb-warppool/docker-compose.yml
+++ b/dvb-warppool/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 18334
 
   web:
-    image: ghcr.io/dvb-projekt/dvb-warppool:1.12.0@sha256:33360466a2f6927f17ea091f937d475fe850125c0724d58a5da308cf8cff9432
+    image: ghcr.io/dvb-projekt/dvb-warppool:1.12.1@sha256:4db092ef0ba7176756aa77a456734d4082ebfd2e42d2a3c810f8a0fee1927d37
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/dvb-warppool/docker-compose.yml
+++ b/dvb-warppool/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 18334
 
   web:
-    image: ghcr.io/dvb-projekt/dvb-warppool:1.12.1@sha256:4db092ef0ba7176756aa77a456734d4082ebfd2e42d2a3c810f8a0fee1927d37
+    image: ghcr.io/dvb-projekt/dvb-warppool:1.12.2@sha256:ebdb88333c04847c23b55d38a077b1a1e408e79adca624029493424501a77a61
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/dvb-warppool/docker-compose.yml
+++ b/dvb-warppool/docker-compose.yml
@@ -12,7 +12,7 @@ services:
   # mountpoints as root:root and the setup wizard can't write config.toml.
   # The container exits immediately after the chown.
   init:
-    image: ghcr.io/dvb-projekt/dvb-warppool:1.9.0@sha256:956372db99ffcb7bd0b47407fb6c46b2f1d85a0c428f8b897038fabae34cd79b
+    image: ghcr.io/dvb-projekt/dvb-warppool:1.10.0@sha256:1619cc88ece78474e98aad252285c2ec05e14e6ae681c34d591c556fe08359e8
     user: "0:0"
     restart: on-failure
     volumes:
@@ -27,7 +27,7 @@ services:
         echo "[init] chown done — /data + /config now owned by warppool:1000"
 
   web:
-    image: ghcr.io/dvb-projekt/dvb-warppool:1.9.0@sha256:956372db99ffcb7bd0b47407fb6c46b2f1d85a0c428f8b897038fabae34cd79b
+    image: ghcr.io/dvb-projekt/dvb-warppool:1.10.0@sha256:1619cc88ece78474e98aad252285c2ec05e14e6ae681c34d591c556fe08359e8
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/dvb-warppool/docker-compose.yml
+++ b/dvb-warppool/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 18334
 
   web:
-    image: ghcr.io/dvb-projekt/dvb-warppool:1.14.1@sha256:87c86310eb1e18643d6356f08df0dec1331115e60055a54ec9e77bd17ed1008a
+    image: ghcr.io/dvb-projekt/dvb-warppool:1.15.0@sha256:0ab305568643b9cb6dc21221895745d1eddbd49eb3e7451009805dad327b0a55
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/dvb-warppool/docker-compose.yml
+++ b/dvb-warppool/docker-compose.yml
@@ -1,0 +1,105 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: dvb-warppool_web_1
+      APP_PORT: 18334
+      PROXY_AUTH_ADD: "false"
+
+  # First-run init: runs as root only to chown the bind-mounted data + config
+  # dirs to the warppool user (uid 1000). Without this, Docker creates the
+  # mountpoints as root:root and the setup wizard can't write config.toml.
+  # The container exits immediately after the chown.
+  init:
+    image: ghcr.io/dvb-projekt/dvb-warppool:1.9.0@sha256:956372db99ffcb7bd0b47407fb6c46b2f1d85a0c428f8b897038fabae34cd79b
+    user: "0:0"
+    restart: on-failure
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${APP_DATA_DIR}/config:/config
+    command:
+      - /bin/sh
+      - -c
+      - |
+        chown -R 1000:1000 /data /config
+        chmod 755 /data /config
+        echo "[init] chown done — /data + /config now owned by warppool:1000"
+
+  web:
+    image: ghcr.io/dvb-projekt/dvb-warppool:1.9.0@sha256:956372db99ffcb7bd0b47407fb6c46b2f1d85a0c428f8b897038fabae34cd79b
+    user: "1000:1000"
+    restart: on-failure
+    stop_grace_period: 1m
+    depends_on:
+      init:
+        condition: service_completed_successfully
+    ports:
+      # Stratum listeners. Host port = container port so the UI's connect
+      # modal shows the correct port. Picked in the app's 188xx range to stay
+      # clear of host ports already taken by other App Store apps.
+      # Stratum V1 (plain).
+      - "18335:18335"
+      # Stratum V2 (NOISE-encrypted). The daemon generates its authority key
+      # on first start and persists it in secrets.toml.
+      - "18336:18336"
+      # Stratum V1 over TLS. The daemon generates a self-signed cert under
+      # /data/tls/ on first start.
+      - "18337:18337"
+    volumes:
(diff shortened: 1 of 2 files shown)


dvb-warppool/docker-compose.yml
```diff
@@ -0,0 +1,105 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: dvb-warppool_web_1
+      APP_PORT: 18334
+      PROXY_AUTH_ADD: "false"
```
nmfretz: Can we remove `PROXY_AUTH_ADD: "false"` and keep Umbrel auth enabled here?

This one is riskier than a normal “app has its own login” case. On a fresh install, the app’s own admin account does not exist yet, and the setup service is served on the app port. That setup flow exposes the Umbrel-provided Bitcoin RPC defaults from `APP_BITCOIN_*` and can write `config.toml` / `secrets.toml`, including the payout config and saved RPC credentials. After that, `/api/auth/bootstrap` is first-come-first-served for creating the app admin account.

With Umbrel auth disabled, another LAN client could reach that flow before the Umbrel owner, configure the pool, and claim the admin account.

Keeping Umbrel auth on should not be painful for users. If they are already signed in to umbrelOS, they can open the app without another Umbrel login prompt. It also gives the setup/bootstrap flow the user’s Umbrel login protection, including Umbrel 2FA if they have it enabled.

If any post-setup routes need to be public, we can whitelist those narrowly with `PROXY_AUTH_WHITELIST`, but setup, config writes, and admin bootstrap should stay behind Umbrel auth by default.

dvb-warppool/docker-compose.yml
```diff
@@ -0,0 +1,105 @@
...
+
+  # First-run init: runs as root only to chown the bind-mounted data + config
+  # dirs to the warppool user (uid 1000). Without this, Docker creates the
+  # mountpoints as root:root and the setup wizard can't write config.toml.
+  # The container exits immediately after the chown.
+  init:
```
nmfretz: Can we try dropping this init service?

With `data/.gitkeep` and `config/.gitkeep` committed, umbrelOS should already create those dirs in app data with the usual `1000:1000` ownership. Since the `web` service also runs as `1000:1000`, the setup wizard should be able to write `config.toml` and `secrets.toml` without a root chown sidecar.

This looks like it may be left over from local `docker compose up` testing, where Docker had to create the bind-mount dirs itself. If it still fails through Umbrel, let me know what breaks, but I think we can keep this package simpler.

dvb-warppool/docker-compose.yml
```diff
@@ -0,0 +1,105 @@
...
+      - "18336:18336"
+      # Stratum V1 over TLS. The daemon generates a self-signed cert under
+      # /data/tls/ on first start.
+      - "18337:18337"
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+      - ${APP_DATA_DIR}/config:/config
+      # Bitcoin Core RPC cookie (read-only). User/Pass via APP_BITCOIN_* env
+      # vars is the default path on umbrelOS; the cookie is a fallback.
+      - ${APP_BITCOIN_DATA_DIR}/.cookie:/bitcoin.cookie:ro
```
nmfretz: Can you confirm whether this cookie mount is needed for the Umbrel package?

From the package, it looks like the normal path is to use the Bitcoin dependency exports via `APP_BITCOIN_RPC_USER` and `APP_BITCOIN_RPC_PASS`. If that is the intended path, we can drop this mount.

The concern is the file bind mount. If `${APP_BITCOIN_DATA_DIR}/.cookie` is missing when Docker starts this app, Docker can create a directory at that path in the Bitcoin app data, which can break Bitcoin Core when it later tries to write the cookie.

If cookie auth is needed as a fallback, can we switch this to an approach that does not bind-mount the cookie file directly?

nmfretz (commented):
Thanks for the submission @dvb-projekt.

I took a look through the packaging and found a few Umbrel-specific things I want to tighten up before I test the install path properly.

The main one is first-run auth. The app having its own login is fine, but this package disables Umbrel auth before the app admin account exists, while the setup flow can write config and secrets. I left the details inline.

dvb-warppool/docker-compose.yml
dvb-projekt: Agreed and fixed in 383e42eb. Removed `PROXY_AUTH_ADD: "false"` so the setup wizard, the config/secrets writes, and `/api/auth/bootstrap` all sit behind Umbrel auth (and its 2FA) on a fresh install.

The Stratum listeners (18335–18337) are their own host ports and don't go through app_proxy, so miners are unaffected. The dashboard widget polls `web:18334/api/umbrel-widget` over the internal Docker network (same as bitcoin/lightning, which keep their widgets with Umbrel auth enabled), so it keeps working without a whitelist — no `PROXY_AUTH_WHITELIST` needed. Please confirm on your install test.

dvb-warppool/docker-compose.yml
dvb-projekt: Done in 383e42eb — dropped the init sidecar. You're right, it was a leftover from local `docker compose` testing. With `config/.gitkeep` + `data/.gitkeep` committed and `web` running as `1000:1000`, the wizard writes `config.toml`/`secrets.toml` fine. Flagging it for your install test in case umbrelOS dir ownership differs in practice.

dvb-warppool/docker-compose.yml
dvb-projekt: Confirmed not needed — removed in 383e42eb. The package uses `APP_BITCOIN_RPC_USER` / `APP_BITCOIN_RPC_PASS` (the Bitcoin dependency exports); the cookie mount was only a fallback. Dropping the bind-mount also avoids the directory-creation hazard you described.

dvb-projekt:
Thanks for the thorough review @nmfretz — all three points addressed in 383e42eb (replied inline). Summary:

1. `PROXY_AUTH_ADD: "false"` removed → first-run setup + admin bootstrap now sit behind Umbrel auth.
2. init chown sidecar dropped (committed `.gitkeep`s + `web` runs as `1000:1000`).
3. `.cookie` bind-mount dropped (uses `APP_BITCOIN_RPC_USER`/`PASS`).

In the same push I also bumped the image + app version to **v1.12.0** (Stratum V2 VarDiff parity + a robustness pass), multi-arch digest pinned. Ready for your install test — happy to adjust if anything trips on the dir ownership or the widget path.

dvb-projekt:
Bumped to **v1.12.1** in 10bf69dd (digest re-pinned to the v1.12.1 multi-arch image).

Besides the version bump, v1.12.1 fixes a Stratum V2 stale-share issue: the Sv2 server was keeping only the single newest job, so a busy-mempool template refresh (~every 30 s) rejected in-flight shares as "stale" on a direct Sv2 miner (~13 % reject rate observed). It now keeps a short job history like the Stratum V1 path — only a new block purges it. No packaging changes from the version you're reviewing; the auth/init/cookie fixes from the last round are unchanged.

dvb-projekt:
Bumped the branch to **v1.12.2** (image digest re-pinned to `sha256:ebdb88333c04847c23b55d38a077b1a1e408e79adca624029493424501a77a61`, multi-arch amd64+arm64).

This release is an **internal performance fix only**: the solar/grid energy-split query no longer scales badly with months of accumulated solar data (a correlated per-bucket lookup replaces a nested-loop join). Results are byte-identical, and there is no behaviour, port, or config change. The three hardening changes from your earlier review — Umbrel auth left on (no `PROXY_AUTH_ADD`), no init/chown service, no `bitcoin .cookie` bind-mount — are unchanged.

Changelog: https://github.com/dvb-projekt/dvb-WarpPool/blob/main/CHANGELOG.md

dvb-projekt:
Re-pinned to **v1.14.0** (multi-arch index digest `sha256:1ee90be9cdf40a3a9f3ac2623762923b79f805c4809d7093424aa695bbff8dd4`). Only the version and the pinned image digest changed; `releaseNotes` stays empty per the new-app convention and the hardening is unchanged. v1.14.0 adds a per-worker reject-reason breakdown to the workers API (additive; share validation unchanged).

dvb-projekt:
Re-pinned to v1.14.1 (image digest `sha256:87c86310…`). This is a small correctness patch: the VarDiff grace window now retains the last several superseded targets instead of just one, which fixes false "low difficulty" rejects for miners with a noisy share rate (small ASICs). Purely additive, no config changes; `releaseNotes` kept empty per the new-app guideline.

dvb-projekt:
Re-pinned to **v1.15.0** (multi-arch index digest `sha256:0ab305568643b9cb6dc21221895745d1eddbd49eb3e7451009805dad327b0a55`). Only the version and the pinned image digest changed in the package; `releaseNotes` stays empty per the new-app convention, and the auth / init / cookie hardening from your earlier review is unchanged. No port, config, or packaging changes. `Lint apps` is green.

v1.15.0 is the result of a full internal security/correctness audit — 25 findings, all fixed and adversarially re-reviewed — plus one small user-facing feature (a "Hide permanently" option for the Bitcoin Core health-banner warnings, from a user request). The most notable fix is on the block-submit path: a block found by a version-rolling (overt ASICBoost) miner was reassembled with the un-rolled header version and would have been rejected by Bitcoin Core; it is now submitted with the exact winning header, for both Stratum V1 and V2.

PR #5781, opened 2026-06-23
Update yuvomi to v0.75.2

nmfretz:
Hi @ulsklyc, while testing Yuvomi 0.75.2 for Umbrel I saw a recurring error on both fresh installs and updates:

`PushScheduler: no such column: r.pushed_at`

It looks like migration 54 adds `reminders.pushed_at`, but migration 57 rebuilds the `reminders` table without preserving that column. The scheduler then queries `r.pushed_at`, so reminder notifications appear broken.

Could you patch the migration/schema and cut a new release? I can retest once it’s available.

### DIFF
--- a/yuvomi/docker-compose.yml
+++ b/yuvomi/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Bumped this to 0.77.2, which includes the upstream fix. Tested the update on Umbrel and it’s looking good.

PR #5783, opened 2026-06-23
App submission: Lightning Mate

opensourceminers:
**Lightning Mate** — profit-aware management & autopilot for an LND node.

- **Source:** https://github.com/opensourceminers/lightningmate (MIT)
- **Image:** `ghcr.io/opensourceminers/lightningmate` (multi-arch amd64 + arm64, digest-pinned)
- **Dependencies:** `lightning` (LND) — auto-discovered, zero-config
- **Default credentials:** none (UI sits behind the `app_proxy` auth)
- **Host access:** none beyond the mounted LND data dir
- Runs as **uid 1000 (non-root)**; a one-shot `init` container fixes the data-dir owner so user settings persist across updates
- Uses the LND **admin macaroon** for fee / rebalance / channel writes (same as Ride The Lightning, ThunderHub, Lightning Terminal). The **autopilot stays OFF** until the user enables it, and every fund-moving action asks for confirmation. Comment out two env lines in the compose to run fully read-only.

### What it does
Channels with live balance bars and source / sink / router roles, a Forwards routing report, a Profit & Loss overview, a node health score, graph-based channel-peer suggestions with one-click open, a **Lightning + on-chain wallet** (send / receive), **profit-aware rebalancing** (manual + adaptive), and an **autopilot** that tunes fees, runs profitable rebalances and opens channels on safe defaults.

### Testing
Tested on **umbrelOS 1.7.3**: installs zero-config against the Lightning (LND) app, all tabs load, the read-only view works, the fee / rebalance / autopilot write paths were verified, and settings persist across an app update.

### Screenshots
Overview
Wallet
Channels
Routing
Autopilot

### Icon
256×256 SVG, square corners: https://cdn.jsdelivr.net/gh/opensourceminers/umbrel-app-store@master/opensourceminers-lightningmate/icon.svg

### DIFF
--- a/lightningmate/data/.gitkeep
+++ b/lightningmate/data/.gitkeep
@@ -1,0 +1,3 @@
+# Bind-mount source for the app's writable data dir (autopilot config + rebalance
+# accounting). Committing it lets umbrelOS pre-create ${APP_DATA_DIR}/data with the
+# right owner (uid 1000), so settings persist across restarts and updates.
--- a/lightningmate/data/.gitkeep
+++ b/lightningmate/data/.gitkeep
@@ -1,3 +1,0 @@
-# Bind-mount source for the app's writable data dir (autopilot config + rebalance
-# accounting). Committing it lets umbrelOS pre-create ${APP_DATA_DIR}/data with the
-# right owner (uid 1000), so settings persist across restarts and updates.
--- a/lightningmate/docker-compose.yml
+++ b/lightningmate/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3001
 
   web:
-    image: ghcr.io/opensourceminers/lightningmate:v0.6.6@sha256:6b1ccb4f38816433cc0c0fddbb7f9d085e5f02e7f9c9b93cde62431f226d8dab
+    image: ghcr.io/opensourceminers/lightningmate:v0.6.10@sha256:199a9df7855714d3b2fe2bc011ebdda88bd58fee2653f283111b35094ca61c1a
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/lightningmate/docker-compose.yml
+++ b/lightningmate/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3001
 
   web:
-    image: ghcr.io/opensourceminers/lightningmate:v0.6.4@sha256:17c0e24aa4b01c34c0e76b37dfd4ac3b87f4a04db72b93cb09e3306dc687a46d
+    image: ghcr.io/opensourceminers/lightningmate:v0.6.6@sha256:6b1ccb4f38816433cc0c0fddbb7f9d085e5f02e7f9c9b93cde62431f226d8dab
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/lightningmate/docker-compose.yml
+++ b/lightningmate/docker-compose.yml
@@ -6,25 +6,11 @@ services:
       APP_HOST: lightningmate_web_1
       APP_PORT: 3001
 
-  # One-shot: make the bind-mounted data dir writable by the non-root app user.
-  # Docker creates the mount source as root, so without this the app (uid 1000)
-  # can't persist its config and settings silently reset on every restart/update.
-  init:
-    image: alpine:3.20
-    user: "0:0"
-    restart: "no"
-    command: sh -c "chown -R 1000:1000 /data"
-    volumes:
-      - ${APP_DATA_DIR}/data:/data
-
   web:
     image: ghcr.io/opensourceminers/lightningmate:v0.6.1@sha256:a1fc9f52c043f14b7b55ce29497ac6874085079abd527c3b9749c9d711100d62
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
-    depends_on:
-      init:
-        condition: service_completed_successfully
     volumes:
       # Read-only access to LND's tls.cert + macaroons.
       - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
--- a/lightningmate/docker-compose.yml
+++ b/lightningmate/docker-compose.yml
@@ -1,0 +1,49 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: lightningmate_web_1
+      APP_PORT: 3001
+
+  # One-shot: make the bind-mounted data dir writable by the non-root app user.
+  # Docker creates the mount source as root, so without this the app (uid 1000)
+  # can't persist its config and settings silently reset on every restart/update.
+  init:
+    image: alpine:3.20
+    user: "0:0"
+    restart: "no"
+    command: sh -c "chown -R 1000:1000 /data"
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
+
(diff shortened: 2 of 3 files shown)


lightningmate/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+      LND_MACAROON_PATH: "/lnd/data/chain/bitcoin/mainnet/readonly.macaroon"
+      DATA_DIR: "/data"
+      PORT: "3001"
+      # Bind all interfaces so app_proxy (which authenticates) can reach us.
+      LM_BIND: "0.0.0.0"
+      # Node management needs write access (like RTL / Thunderhub / Lightning
+      # Terminal). The autopilot itself stays OFF until the user turns it on, and
+      # every fund-moving action asks for confirmation. To run read-only instead,
+      # comment these two lines out.
+      LM_ENABLE_WRITE: "true"
```
nmfretz, line 40: Can we add server-side auth before enabling write mode here?

The `:ro` mount only makes LND's files read-only. With `LM_ENABLE_WRITE=true`, the app still reads `admin.macaroon` and can write to LND over gRPC.

Since the upstream API has no built-in login and the service binds to `0.0.0.0`, `app_proxy` only protects the browser route. Other app containers on Umbrel's shared Docker network could still call `lightningmate_web_1:3001` directly.

Write-capable LND access is fine for this type of app, but the API itself needs an auth check around write endpoints, ideally using an Umbrel-derived per-install secret like `APP_PASSWORD` / `deterministicPassword`. Otherwise write mode should stay off in the official package until that exists.

lightningmate/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+      - ${APP_LIGHTNING_NODE_DATA_DIR}:/lnd:ro
+      # App's own writable data dir (autopilot config + rebalance accounting).
+      - ${APP_DATA_DIR}/data:/data
+    environment:
+      APP_LIGHTNING_NODE_IP: "$APP_LIGHTNING_NODE_IP"
+      APP_LIGHTNING_NODE_GRPC_PORT: "$APP_LIGHTNING_NODE_GRPC_PORT"
+      LND_SOCKET: "$APP_LIGHTNING_NODE_IP:$APP_LIGHTNING_NODE_GRPC_PORT"
+      LND_DIR: "/lnd"
+      LND_CERT_PATH: "/lnd/tls.cert"
+      LND_MACAROON_PATH: "/lnd/data/chain/bitcoin/mainnet/readonly.macaroon"
```
nmfretz: Can we use the selected Bitcoin network here? The `lightning` dependency exports `APP_BITCOIN_NETWORK`, and hardcoding `mainnet` breaks non-mainnet installs.

```suggestion
      LND_MACAROON_PATH: "/lnd/data/chain/bitcoin/$APP_BITCOIN_NETWORK/readonly.macaroon"
```

lightningmate/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+      DATA_DIR: "/data"
+      PORT: "3001"
+      # Bind all interfaces so app_proxy (which authenticates) can reach us.
+      LM_BIND: "0.0.0.0"
+      # Node management needs write access (like RTL / Thunderhub / Lightning
+      # Terminal). The autopilot itself stays OFF until the user turns it on, and
+      # every fund-moving action asks for confirmation. To run read-only instead,
+      # comment these two lines out.
+      LM_ENABLE_WRITE: "true"
+      LND_WRITE_MACAROON_PATH: "/lnd/data/chain/bitcoin/mainnet/admin.macaroon"
```
nmfretz: Same here for the write macaroon.

```suggestion
      LND_WRITE_MACAROON_PATH: "/lnd/data/chain/bitcoin/$APP_BITCOIN_NETWORK/admin.macaroon"
```

lightningmate/umbrel-app.yml
```diff
@@ -0,0 +1,37 @@
...
+
+  See channels with live balance bars and source / sink / router roles, a full
+  Forwards routing report, and a Profit & Loss overview (routing revenue vs
+  channel-open and rebalancing costs). Get scored channel-peer suggestions from
+  the network graph and open channels in a click, plus a node health score.
+
+
+  Flip on the Autopilot and the node tunes its own fees, runs only profitable
+  rebalances, and opens channels to top peers — all on recommended settings with
+  safe caps and cooldowns. Read-only by default; writes are opt-in.
```
nmfretz: Can we update this once the auth issue is handled? The package currently enables write mode on first launch with the admin macaroon, so the app is not read-only by default.

lightningmate/data/.gitkeep
```diff
@@ -0,0 +1,3 @@
+# Bind-mount source for the app's writable data dir (autopilot config + rebalance
```
nmfretz: Can we keep `.gitkeep` empty? The compose comment already explains why the directory is committed, and `.gitkeep` is just the placeholder file for Git.

nmfretz (commented):
Thanks for the submission @opensourceminers.

The packaging is super close! There's one important thing to tighten up around write mode.

The read-only LND data mount is fine, and write-capable LND access makes sense for a node-management app like this. But once the app reads `admin.macaroon`, it can ask LND to move funds over gRPC.

The browser route is behind `app_proxy`, but the app API itself does not appear to have server-side auth. Since Umbrel apps share a Docker network, another app container could bypass `app_proxy` and call the service directly.

I left that inline, plus a couple smaller packaging tweaks.

nmfretz:
Also @opensourceminers are you okay with these tweaked gallery assets?
https://github.com/getumbrel/umbrel-apps-gallery/pull/115/changes

opensourceminers:
Thanks for the thorough review — all four points are addressed:

1. **Server-side auth.** The whole API now sits behind a session, not just the write endpoints. On Umbrel the app requires signing in with the per-install app password — `deterministicPassword: true`, which Umbrel shows the user on open and injects as `$APP_PASSWORD`; `$APP_SEED` signs the session tokens. So another container on the shared Docker network can no longer read the node or move funds. Standalone (bound to `127.0.0.1`) has no cross-container exposure, so auth is a no-op there. With that in place I've kept write mode enabled.
2. **Network parameterised.** Both macaroon paths now use `$APP_BITCOIN_NETWORK` (readonly + admin) instead of hardcoded `mainnet`.
3. **Description updated** to the sign-in model — it no longer claims read-only-by-default.
4. **`data/.gitkeep` emptied.**

Image bumped to `v0.6.4` (multi-arch amd64+arm64, digest-pinned) and tested on umbrelOS. Thanks again for the careful look!

opensourceminers:
Thanks @nmfretz — the tweaked gallery assets look great, happy to go with them 🙏

Heads-up: I've pushed v0.6.6, which adds an Amboss Magma liquidity marketplace (buy & sell channel liquidity) as a new **Market** tab. I put together a screenshot in the same style for the gallery (below).

Does this work as-is, or would you like to rework it to better fit the set? Happy either way — we'll gladly take any suggestion. And sorry for the late addition / the extra back-and-forth!

Market

nmfretz:
Excellent thanks for making all those changes @opensourceminers! This is looking really great.

> Does this work as-is, or would you like to rework it to better fit the set?

Thanks for this. We added it to the set with the same background and text formatting:
https://github.com/getumbrel/umbrel-apps-gallery/pull/115/changes

Welcome to the app store 🚀

image

PR #5785, opened 2026-06-24
Update sv2-ui

### LINKED ISSUE #5782: Package issue: `sv2-ui` install can fail on Umbrel Pro because Docker-in-Docker misses startup health deadline
### App ID

sv2-ui

### umbrelOS version

1.7.3

### Device architecture

amd64

### When did this happen?

Fresh install

### What failed?

Install

### Expected behavior

The `sv2-ui` app should complete the first installation without any issues.

### Actual behavior

`sv2-ui` appears to intermittently fail during install on Umbrel Pro because the Docker-in-Docker service, `sv2-ui_docker_1`, does not become healthy before the current hard-coded startup deadline.

More context can be found on the original issue opened against SRI's `sv2-ui` repo: https://github.com/stratum-mining/sv2-ui/issues/189

# sv2-ui install can fail on Umbrel Pro because Docker-in-Docker misses startup health deadline

## Summary

`sv2-ui` appears to intermittently fail during install/startup on Umbrel Pro because the Docker-in-Docker service, `sv2-ui_docker_1`, does not become healthy before the current hard-coded startup deadline.

When this happens, the main `sv2-ui` container never starts because it depends on the DIND service being healthy, and the app proxy repeatedly logs that it cannot resolve the UI container.

Related upstream issue: https://github.com/stratum-mining/sv2-ui/issues/189  
Umbrel app PR: https://github.com/getumbrel/umbrel-apps/pull/5721

## Environment

- Device: Umbrel Pro / U4XN1
- UmbrelOS: 1.7.3
- App: `sv2-ui`
- Install state: fresh device, Bitcoin Core still syncing
- App package source: `getumbrel/umbrel-apps/tree/master/sv2-ui`

## Observed behavior

Install/startup gets stuck or fails with logs like:

```text
sv2-ui_app_proxy_1 | Error waiting for port: "The address 'sv2-ui_sv2-ui_1' cannot be found"

sv2-ui_docker_1 | Waiting for dockerd... (attempt 28/30)
sv2-ui_docker_1 | Waiting for dockerd... (attempt 29/30)
sv2-ui_docker_1 | Waiting for dockerd... (attempt 30/30)
sv2-ui_docker_1 | Dockerd did not become ready in time.
```

Compose then reports:

```text
dependency failed to start: container sv2-ui_docker_1 is unhealthy
```

After retries, DIND eventually comes up and the app can work, which suggests this is not a permanent incompatibility but a startup timing/lifecycle issue.

## Suspected root cause

The main failure seems to be that `sv2-ui_docker_1` has a strict Docker-in-Docker readiness deadline. If `dockerd` does not answer on `/data/docker.sock` within the current 30-attempt loop, the entrypoint exits.

Because the `sv2-ui` service has:

```yaml
depends_on:
  docker:
    condition: service_healthy
```

the UI service is blocked until DIND becomes healthy. If DIND misses the deadline, the install/startup fails before the UI can run.

The app proxy error appears to be secondary. In PR #5721, the app proxy dependency on `sv2-ui` was removed, so `app_proxy` starts immediately and tries to resolve `sv2-ui_sv2-ui_1` before that container is actually running/attached to the network.

## Relevant PR changes

The most relevant changes appear to be:

- Docker-in-Docker lifecycle hardening:  
  https://github.com/getumbrel/umbrel-apps/commit/ddadac7b67e0a5cc9e4bba2df46a21e635192bcb

- Removal of app proxy dependency:  
  https://github.com/getumbrel/umbrel-apps/commit/3c6aeb655ee8014191ca0aec996e09ad95fbcbf0

- App data bind mount persistence changes:  
  https://github.com/getumbrel/umbrel-apps/commit/6f711db44e08efbe1e6ac12a2e03847f793315d3

The hardening change makes sense, but on slower/cold Umbrel Pro starts it seems to turn a slow DIND boot into a hard app failure.

## Additional observations

The logs also show DIND creating/using `dind0`, while UmbrelOS/NetworkManager notices and manages that bridge. There are also iptables/nftables-related warnings from the DIND startup path.

These do not appear fatal by themselves, since a later startup succeeds, but they may explain why DIND sometimes takes longer than expected on Umbrel Pro.

The Bitcoin Core sync state does not look like the primary cause. The failure happens before SV2 setup/mining behavior matters, at the DIND health dependency layer.

## Suggested fix

A few possible mitigations:

1. Increase the DIND startup deadline in `sv2-ui/entrypoint.sh`, for example from ~30 seconds to 120-180 seconds.
2. Align the Compose healthcheck `start_period` / `retries` with that longer startup window.
3. Avoid killing `dockerd` immediately when it misses the short readiness window; let the Docker healthcheck represent readiness instead.
4. Restore an app proxy dependency on `sv2-ui`, or otherwise avoid starting `app_proxy` before the UI service can exist on the network.
5. Consider making `dind0` unmanaged by NetworkManager, or otherwise reviewing the DIND networking approach on UmbrelOS/Umbrel Pro.

## Expected behavior

`sv2-ui` should tolerate slower cold starts of Docker-in-Docker on Umbrel Pro and eventually become healthy without the install being marked failed prematurely.

### Logs or screenshots

I couldn't paste the full logs here, because of characters limits.

But you can find the full logs here: https://github.com/stratum-mining/sv2-ui/issues/189#issuecomment-4743929119 and https://github.com/stratum-mining/sv2-ui/issues/189#issuecomment-4743929119

nmfretz:
Fixes #5782 and https://github.com/stratum-mining/sv2-ui/issues/189.

I've updated the `sv2-ui` package to make app startup more reliable when the Docker-in-Docker sidecar takes longer than usual to become ready.

@GitGab19 @lucasbalieiro @pavlenex do you guys have an imminent release you'd wanna squeeze in to this update?

### DIFF
--- a/sv2-ui/docker-compose.yml
+++ b/sv2-ui/docker-compose.yml
@@ -22,11 +22,11 @@ services:
         --host unix:///data/docker.sock
         --pidfile /data/docker.pid
     healthcheck:
-      test: ["CMD", "docker", "-H", "unix:///data/docker.sock", "info"]
+      test: ["CMD-SHELL", "docker -H unix:///data/docker.sock volume inspect sv2-config >/dev/null"]
       interval: 5s
       timeout: 5s
       retries: 10
-      start_period: 10s
+      start_period: 180s
     volumes:
       - ${APP_DATA_DIR}/entrypoint.sh:/entrypoint.sh
       - ${APP_DATA_DIR}/data/docker:/data
--- a/sv2-ui/docker-compose.yml
+++ b/sv2-ui/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   docker:
-    image: docker:27.2.0-dind@sha256:d1d6d5265119696d8fa9b883f813953c560c5d87ea128a57681239b576f62997
+    image: docker:28.3.0-dind@sha256:56e360024d7d35dac801b8fc87f869ccf5c89b88399dd266f7cfc1169690a86b
     privileged: true
     network_mode: host
     stop_grace_period: 1m
--- a/sv2-ui/docker-compose.yml
+++ b/sv2-ui/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   docker:
-    image: docker:27.2.0-dind@sha256:f9f72ad901a78f27be922b2d320bbc263174f12919c1b37e6a01f828fa904565
+    image: docker:27.2.0-dind@sha256:d1d6d5265119696d8fa9b883f813953c560c5d87ea128a57681239b576f62997
     privileged: true
     network_mode: host
     stop_grace_period: 1m
--- a/sv2-ui/entrypoint.sh
+++ b/sv2-ui/entrypoint.sh
@@ -78,6 +78,7 @@ fi
 DOCKER_SOCKET="/data/docker.sock"
 DOCKER_VOLUME_NAME="sv2-config"
 CONFIG_SOURCE_PATH="/app/data/config"
+DOCKER_READY_TIMEOUT_SECONDS=180
 
 echo "Starting dockerd in background..."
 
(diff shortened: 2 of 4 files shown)


GitGab19:
Hi @nmfretz, thanks for working on this patch so rapidly!

I tested this PR on my Umbrel's VM, and the installation worked, but I also have to say I never experienced this issue on my VM :)

Do you need any manual tests on Umbrel Pro? 

About new releases, we're still working on the next one, so we're gonna submit a PR in the next weeks to update the app.

nmfretz:
No worries @GitGab19! I managed to reproduce the issue @pavlenex hit on a Pro. The `docker:27.2.0-dind` image could fail during cgroup setup before dockerd ever became ready. The newer `docker:28.3.0-dind` includes an upstream fix for exactly this issue, so I've bumped to that.

I've left in the other startup hardening fixes in as well.

I re-tested on the Pro and the app consistently installs and opens normally.

Huge thanks for reporting!

kaichappel:
@nmfretz I tested this on an N150 Umbrel Home and it works well for me. The newer DIND sidecar starts cleanly and reaches healthy. Restarts work well too.

GitGab19:
Great news @nmfretz, thanks for diving into it! :pray:

PR #5786, opened 2026-06-24
Update syslog-ng

### LINKED ISSUE #5775: Syslog-ng: Logdy viewer shows 0 logs — follow /data/*.log glob fails at startup and never re-scans
### App ID

syslog-ng

### umbrelOS version

1.7.3

### Device architecture

amd64

### When did this happen?

Fresh install

### What failed?

Other

### Expected behavior

Syslog messages received by the app's syslog-ng (UDP 514 / TCP 601) should appear in the Logdy web UI.

### Actual behavior

syslog-ng receives the messages correctly — they land in /var/log/messages — but the Logdy web UI stays at "0 logs". At startup Logdy logs:

    Following file changes failed; stat /data/*.log: no such file or directory

…and never picks up log files that appear afterward, so the UI is permanently empty even though logs are being stored.

**Root cause — startup race in the app's packaging:**

The Logdy viewer is launched with `follow /data/*.log`. That glob is evaluated **once, at container start**. On a fresh boot there are no `*.log` files yet (nothing received), so Logdy errors with `Following file changes failed; stat /data/*.log: no such file or directory` and never re-scans when files appear later.

Meanwhile the bundled syslog-ng config writes its default destination to `/var/log/messages` (no `.log` extension → never matches Logdy's glob) plus `/var/log/messages-kv.log`. `/var/log` (syslog-ng container) and `/data` (Logdy/web container) are the **same shared volume** — both bind `~/umbrel/app-data/syslog-ng/data/logs` — so `messages-kv.log` does eventually match the glob, but only if the web container is restarted *after* that file exists.

Net: the app receives syslog fine, but the UI shows nothing until something forces Logdy to re-scan.

**Reproduce:** fresh install → point a device at UDP 514 / TCP 601 → logs land in `/var/log/messages` but the Logdy UI stays at "0 logs".

**Suggested fix (any one):**
1. Ship a syslog-ng destination that writes a `*.log` file into the shared dir and creates it eagerly at startup, so Logdy's glob always matches.
2. Order the s6 services so syslog-ng creates its first log file before Logdy starts.
3. Make Logdy's `follow` watch the directory for newly-appearing files matching the glob.

**Workaround that fixed it for me:** appended to `/config/syslog-ng.conf` (in `syslog-ng_syslog_1`) a destination `file("/var/log/syslog.log" perm(0644))` fed by the network sources, then `syslog-ng-ctl reload --control=/config/syslog-ng.ctl` and `docker restart syslog-ng_web_1`. Logdy then attaches to `/data/syslog.log` (the same shared file) and displays everything.

### Logs or screenshots

`Starting logdy service with configuration: --no-analytics --no-updates --ui-ip=0.0.0.0 \
  --bulk-window=1000 --port=8080 --ui-pass=<REDACTED> --max-message-count=1000 \
  --api-key=<REDACTED> follow /data/*.log
level=error msg="Following file changes failed" \
  error="stat /data/*.log: no such file or directory" path="/data/*.log"`

nmfretz:
fixes https://github.com/getumbrel/umbrel-apps/issues/5775

### DIFF
--- a/syslog-ng/docker-compose.yml
+++ b/syslog-ng/docker-compose.yml
@@ -21,6 +21,14 @@ services:
     volumes:
       - ${APP_DATA_DIR}/data/config:/config
       - ${APP_DATA_DIR}/data/logs:/var/log
+    # Logdy expands /data/*.log once at startup, so create the first matching
+    # log file before the web service starts on fresh installs.
+    healthcheck:
+      test: ["CMD-SHELL", "test -f /var/log/messages-kv.log || touch /var/log/messages-kv.log"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+      start_period: 5s
 
   web:
     image: rickraven/logdy:0.17.1@sha256:a1438c758ed3735e9b06205168a3c63e39bd0895bc5f0cd1adf4ed2270ae66cb
@@ -33,4 +41,5 @@ services:
     volumes:
       - ${APP_DATA_DIR}/data/logs:/data:ro
     depends_on:
-      - syslog
+      syslog:
+        condition: service_healthy
--- a/syslog-ng/umbrel-app.yml
+++ b/syslog-ng/umbrel-app.yml
@@ -3,7 +3,7 @@ id: syslog-ng
 name: Syslog-ng
 tagline: Centralized syslog server with real-time web UI
 category: networking
-version: "4.10.2"
+version: "4.10.2-1"
 port: 5142
 description: >-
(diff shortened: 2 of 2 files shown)


kaichappel (approved):
Tested this update on an amd64 Umbrel device. syslog-ng installed cleanly, reached ready with the syslog container healthy, Logdy followed /data/messages-kv.log without the previous glob-follow failure, and a UDP syslog marker showed up in both persisted log files and the Logdy UI stream after restart. Looks good.

PR #5787, opened 2026-06-24
App Submission: Joinstr

1440000bytes:
## Joinstr

Decentralized coinjoin over Nostr

- Source: https://gitlab.com/invincible-privacy/joinstr
- Image: https://ghcr.io/1440000bytes/joinstr-web 
- Depends on the Bitcoin Node app

---

## Logo

logo

## Screenshots

Screenshot from 2026-06-24 22-48-18

Screenshot from 2026-06-24 22-48-45

Screenshot from 2026-06-24 22-49-08

Screenshot from 2026-06-24 22-52-37

Screenshot from 2026-06-24 22-53-19

### DIFF
--- a/joinstr/docker-compose.yml
+++ b/joinstr/docker-compose.yml
@@ -1,0 +1,23 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: joinstr_web_1
+      APP_PORT: 5000
+
+  web:
+    # Multi-arch (amd64+arm64) image pinned by digest — required by umbrel-apps.
+    image: ghcr.io/1440000bytes/joinstr-web:0.1.1@sha256:3a3d513c758fb5a5346dc1e26c430c3e35f72062a3fe5dcc064cf75b6fd83d7c
+    restart: on-failure
+    stop_grace_period: 1m
+    environment:
+      JOINSTR_MANAGED: "1"
+      JOINSTR_WALLET: joinstr
+      JOINSTR_DATA_DIR: /app/data
+      BITCOIN_RPC_HOST: $APP_BITCOIN_NODE_IP
+      BITCOIN_RPC_PORT: $APP_BITCOIN_RPC_PORT
+      BITCOIN_RPC_USER: $APP_BITCOIN_RPC_USER
+      BITCOIN_RPC_PASS: $APP_BITCOIN_RPC_PASS
+    volumes:
+      - ${APP_DATA_DIR}/data:/app/data
--- a/joinstr/umbrel-app.yml
+++ b/joinstr/umbrel-app.yml
@@ -22,4 +22,4 @@ port: 5500
 gallery: []
 path: ""
 submitter: 1440000bytes
-submission: ""
+submission: "https://github.com/getumbrel/umbrel-apps/pull/5787"
--- a/joinstr/umbrel-app.yml
+++ b/joinstr/umbrel-app.yml
@@ -19,7 +19,12 @@ dependencies:
 repo: https://gitlab.com/invincible-privacy/joinstr
 support: https://gitlab.com/invincible-privacy/joinstr/-/issues
 port: 5500
-gallery: []
+gallery:
+  - 1.webp
+  - 2.webp
+  - 3.webp
+  - 4.webp
+  - 5.webp
 path: ""
 submitter: 1440000bytes
 submission: "https://github.com/getumbrel/umbrel-apps/pull/5787"
--- a/joinstr/umbrel-app.yml
+++ b/joinstr/umbrel-app.yml
@@ -1,0 +1,25 @@
+manifestVersion: 1
+id: joinstr
+category: bitcoin
+name: Joinstr
+version: "0.1.0"
+tagline: Decentralized coinjoin over Nostr
+description: >-
+  Joinstr is a coordinator-less coinjoin that uses Nostr relays to find peers
+  and collaboratively build a Bitcoin transaction, improving on-chain privacy.
+
+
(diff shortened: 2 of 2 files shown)


nmfretz (commented):
Thanks for the submission @1440000bytes, this looks sweet.

I took a look through the package and found a couple of Umbrel lifecycle issues that need tightening before I can test this properly. The main ones are making sure Joinstr can resume correctly after restarts, and making sure the Bitcoin wallet it uses exists on first run.

I left the details inline. Once those are updated I can re-test the install path through Umbrel.

joinstr/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+      APP_HOST: joinstr_web_1
+      APP_PORT: 5000
+
+  web:
+    # Multi-arch (amd64+arm64) image pinned by digest — required by umbrel-apps.
+    image: ghcr.io/1440000bytes/joinstr-web:0.1.0@sha256:778b4e65f0adc7e98dbe9903fc80409e83ee5f4707e6e5758e27c13a71beb9b7
+    restart: on-failure
+    stop_grace_period: 1m
+    environment:
+      JOINSTR_WALLET: joinstr
```
nmfretz, line 16: How is the `joinstr` Bitcoin Core wallet meant to get created or loaded on a fresh install?

The package sets `JOINSTR_WALLET: joinstr`, and the wrapper passes that to Joinstr so wallet RPC calls go to `/wallet/joinstr`. A fresh Umbrel Bitcoin Node install won't already have a loaded wallet by that name, so the first real workflow looks like it can fail once Joinstr calls wallet RPC methods like `getnewaddress`, `listunspent`, and `walletprocesspsbt`.

joinstr/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+      BITCOIN_RPC_HOST: $APP_BITCOIN_NODE_IP
+      BITCOIN_RPC_PORT: $APP_BITCOIN_RPC_PORT
+      BITCOIN_RPC_USER: $APP_BITCOIN_RPC_USER
+      BITCOIN_RPC_PASS: $APP_BITCOIN_RPC_PASS
```
nmfretz, line 21: Can you confirm how Joinstr handles an in-progress round if the app restarts?

From what I can tell, the web UI keeps its pool/session state inside the container under `/app/gui`. Umbrel recreates containers during restarts and updates, so that state would be lost unless it is bind-mounted under `${APP_DATA_DIR}/data/...`.

joinstr/docker-compose.yml
1440000bytes: Fixed by having the image's entrypoint ensure the wallet exists before the web UI starts serving. It calls Bitcoin Core RPC idempotently:

```
createwallet "$JOINSTR_WALLET"   # on a fresh node
loadwallet   "$JOINSTR_WALLET"   # if it already exists on disk (no-op if already loaded)
```

joinstr/docker-compose.yml
1440000bytes: The web UI stored `pools_<port>.json`, `flask_sessions_<port>/` and `join_session_<port>.json` under gui/, so a container recreate would lose any in-progress round.

Fixed in two parts:

1. Upstream ([invincible-privacy/joinstr!29](https://gitlab.com/invincible-privacy/joinstr/-/merge_requests/29)) added a `JOINSTR_DATA_DIR` env that relocates all three state files out of the source tree. It defaults to the old location, so behavior is unchanged when unset.

2. This package sets `JOINSTR_DATA_DIR` and bind-mounts it under `${APP_DATA_DIR}`:

```
environment:
  JOINSTR_DATA_DIR: /app/data
volumes:
  - ${APP_DATA_DIR}/data:/app/data
```

So the creator's pool (with its key), the joiner's join-session, and the round's session state all live under `${APP_DATA_DIR}/data` and survive restarts.

1440000bytes:
> I took a look through the package and found a couple of Umbrel lifecycle issues that need tightening before I can test this properly. The main ones are making sure Joinstr can resume correctly after restarts, and making sure the Bitcoin wallet it uses exists on first run.
> 
> I left the details inline. Once those are updated I can re-test the install path through Umbrel.

Thanks for the review. I have addressed both the comments in last push.

nmfretz (commented):
Sorry I left you waiting so long on this, @1440000bytes. I re-tested the latest changes through a fresh Umbrel install, and both of my original concerns are resolved. The `joinstr` wallet was created correctly, and the pool/session state survived an Umbrel restart and could be resumed.

During the re-test, Codex surfaced one additional issue in the web UI, which I then confirmed in the source. The Settings panel saves the RPC URL, wallet name, RPC username/password, and Nostr relay in browser cookies. However, the backend gets its configuration from the environment/Flask session and never reads those cookies. This means changing the settings has no effect, while any RPC credentials entered there are still stored client-side unnecessarily.

This looks like leftover settings UI rather than intentional behavior for the Umbrel package. Since Umbrel already supplies the RPC configuration, could you remove the non-functional settings and cookie handling from the Umbrel image, or wire the settings to the backend without storing the RPC password in JavaScript-readable cookies?

Once that’s addressed, I’ll give it another quick test.

1440000bytes:
Settings are now sent to a new `/settings` endpoint and stored in the server side Flask session (HttpOnly signed cookie, filesystem storage), `document.cookie `handling is removed and the RPC password is never returned to the browser. The Umbrel package now runs in a managed mode (JOINSTR_MANAGED=1) where the RPC fields are hidden and the env supplied RPC config is authoritative.

Only the nostr relay remains configurable.

nmfretz:
Excellent, great work @1440000bytes! Thanks again for getting this on the App Store. We're good to go live 🚀 

image

PR #5788, opened 2026-06-24
Update Threema to v2.6.4

dennysubke:
Tested update on Umbrel VM instance.

Changes:

- Updated the beta app hint banner
- Upgraded dependencies
- Bumped NodeJS from version 20 to version 24

Full changelog can be viewed here:
https://github.com/threema-ch/threema-web/blob/master/CHANGELOG.md


nmfretz:
Welcome back @dennysubke!!

We actually have some internal tracking tools that notify us when new images are available for apps, so @kaichappel updated Threema here: https://github.com/getumbrel/umbrel-apps/commit/5b2d21d1e7c0b02d20cb49466b702d876f3a79ec

PR #5795, opened 2026-06-26
Fix sv2-ui DIND networking on umbrelOS

nmfretz:


GitGab19:
Tested on my end and everything is working fine now, thanks for patching it so quickly!

PR #5796, opened 2026-06-26
Add Freeport - self-hosted P2P marketplace node (web app + MCP + notifier + relay)

ptrinh:
# App Submission

### App name
Freeport

### 256x256 SVG icon
https://raw.githubusercontent.com/ptrinh/freeport/main/docs/icon.svg

### Description
Self-hosted node for [Freeport](https://freeport.network), the decentralized peer-to-peer marketplace for rides, services and goods built on Nostr - MIT-licensed, [source on GitHub](https://github.com/ptrinh/freeport).

One app, five jobs (one container, port 1988):

1. **The Freeport web app itself** - open the app from the Umbrel dashboard and you're on Freeport, hosted from your own box.
2. **Read-only MCP endpoint** (`/mcp`) - AI agents can search the marketplace (rides/services/goods, reputation, profiles) by kind, tag and geohash radius. Never signs or publishes.
3. **Content-blind push notifier** - Web Push / Expo push for new matching activity; messages stay end-to-end encrypted.
4. **Embedded NIP-01 Nostr relay** (port 1989) - nearby users can add `ws://<umbrel>:1989` to their relay list.
5. **Optional Telegram bridge** - market feed into groups + personal pings (set a bot token to enable).

VAPID keys are generated and persisted automatically on first run. Multi-arch image (amd64 + arm64) published to GHCR from [this workflow](https://github.com/ptrinh/freeport/blob/main/.github/workflows/self-hosted-image.yml), digest-pinned in the compose file.

Note: this submission was refreshed from the original "Freeport Nostr MCP" (MCP + notifier only) to the full self-hosted Freeport node described above - the app id is now `freeport` and the default port is 1988.


nmfretz (commented):
Thanks for another submission @ptrinh! And sorry for the delay.

I gave this a fresh-install and restart pass on umbrelOS 1.7.4 (arm64). The web UI opens and the account flow starts correctly. The main thing I want to tighten up is how the bundled services are exposed and described: the MCP/push routes are currently behind Umbrel auth, while the bundled client appears to keep using the public relays and hosted notifier by default.

I left a few questions and small manifest fixes inline. Once those are clarified and updated, I can re-test the client routes and restart behavior.

freeport/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: freeport_server_1
+      APP_PORT: 1988
```
nmfretz, line 7: The MCP and push routes are behind Umbrel auth right now. On a fresh install, unauthenticated requests to `/mcp`, `/health`, `/vapidPublicKey`, `/subscribe`, and `/unsubscribe` redirect to Umbrel login, so standard MCP clients and native/mobile push registration cannot use them.

Can we add a narrow `PROXY_AUTH_WHITELIST` for only the routes external clients need, while keeping `/` protected? Before `/subscribe` is exposed, the pinned image should include the current upstream subscription hardening, and this fresh package can enable `REQUIRE_SUBSCRIBE_AUTH=1`.

freeport/docker-compose.yml
```diff
@@ -0,0 +1,33 @@
...
+
+  server:
+    image: ghcr.io/ptrinh/freeport-self-hosted:0.7.0@sha256:24200894eeab021d88688baac9383f41adf3911222ecc6290705199059262943
```
nmfretz: If we’re exposing the push routes, can you cut and pin a release containing the current upstream `/subscribe` hardening? The pinned `0.7.0` image predates those changes.

It also reports `0.6.9` from `/health`, while the manifest declares `0.7.0`, so it would be good to align those versions in the new build.

freeport/umbrel-app.yml
```diff
@@ -0,0 +1,53 @@
+manifestVersion: 1
+id: freeport
+category: networking
+name: Freeport
+version: "0.7.0"
+tagline: Self-hosted Freeport — the P2P marketplace, on your own box
+description: >-
+  Run all of Freeport, the decentralized peer-to-peer marketplace for rides,
```
nmfretz, line 8: Can you confirm the expected first-run setup here? The bundled client appears to use the public `DEFAULT_RELAYS`, while its notification endpoint defaults to `https://mcp.freeport.network`.

If users need to manually point the client at the services running on their Umbrel, that setup should be included in the description. If it is meant to happen automatically, we should pin a build where that wiring is in place.

freeport/umbrel-app.yml
```diff
@@ -0,0 +1,53 @@
...
+  search Freeport's Nostr events — ride/service/goods requests and offers,
+  reputation, and profiles — filtered by kind, tag, and geohash radius.
+
+
+  3. A push notifier that watches the relays and pushes new matching intents
+  and direct messages to subscribers over Web Push (browsers / PWA) and Expo
+  Push (native iOS/Android).
+
+
+  4. An embedded NIP-01 Nostr relay (port 1989) that nearby users can add to
```
nmfretz, line 27: Just checking the intended behavior here: the embedded relay is currently in-memory. I confirmed that an event disappears after restarting the app through Umbrel.

If that is intentional, can we make the ephemeral behavior clear in the description? The current wording reads more like durable local relay infrastructure.

freeport/umbrel-app.yml
```diff
@@ -0,0 +1,53 @@
...
+  their relay list, so your community's traffic doesn't depend on public
+  relays.
+
+
+  5. An optional Telegram bridge: relay a market feed into Telegram groups,
```
nmfretz: Could we remove the Telegram bridge from the App Store description for now? In this package it can only be enabled by editing the commented Compose environment variables, so Umbrel users do not have a supported setup path. If there is an in-app configuration flow I missed, please point me to it.

freeport/umbrel-app.yml
```diff
@@ -0,0 +1,53 @@
...
+  server. VAPID keys are generated and persisted automatically on first run.
+developer: Freeport
+website: https://freeport.network
+dependencies: []
+repo: https://github.com/ptrinh/freeport
+support: https://github.com/ptrinh/freeport/issues
+port: 1988
+gallery:
+  - 1.jpg
+  - 2.jpg
```
nmfretz: These gallery URLs currently return 404. For new submissions we keep the gallery empty until the App Store assets are added.

```suggestion
gallery: []
```

freeport/umbrel-app.yml
```diff
@@ -0,0 +1,53 @@
...
+releaseNotes: ""
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+submitter: ptrinh
+submission: https://github.com/ptrinh/freeport
```
nmfretz: Can we point this at the App Store submission?

```suggestion
submission: https://github.com/getumbrel/umbrel-apps/pull/5796
```

ptrinh:
Thanks for the thorough test! Addressed all points in the latest push:

- **Auth**: added `PROXY_AUTH_WHITELIST` for `/mcp`, `/health`, `/vapidPublicKey`, `/subscribe`, `/unsubscribe` — the web app (`/`) stays behind Umbrel auth. `/subscribe` is now gated by `REQUIRE_SUBSCRIBE_AUTH=1`: a DM-watch subscribe must carry a signed kind-27235 proof of the watched pubkey (filter-only subscriptions, which only read public intents, need none), so whitelisting it doesn't expose DM metadata.
- **Image/version**: re-pinned a freshly built multi-arch (amd64 + arm64) `0.7.0` image that includes the upstream `/subscribe` hardening; `/health` now reports `0.7.0` (it was a stale hardcoded `0.6.9`).
- **First-run**: documented in the description — the client ships with public defaults and works immediately; to route it through this box, set the notification service URL and add `ws://<umbrel>:1989` in Settings.
- **Relay**: the description now states the embedded relay is ephemeral / in-memory (cleared on restart), a low-latency local hop rather than durable storage.
- **Telegram**: removed from the description for now — there's no supported in-app setup path in this package.
- **Gallery**: set to `[]`. **Submission**: pointed at this PR.

PR #5797, opened 2026-06-26
Bump HermitStash to 1.13.0

dotCooCoo:
Updates HermitStash from 1.12.11 to 1.13.0.

**Image:** pinned to the multi-arch (linux/amd64 + linux/arm64) OCI index digest `ghcr.io/dotcoocoo/hermitstash:1.13.0@sha256:e36518ec236d7576f835ec804723626bf3b009125e8e14d4683911a4d414fe98`.

### What changed (1.12.12 → 1.13.0)

**Sign-in and accounts**
- Two-factor authentication is now required on Google and passkey sign-in, not only password login, and a code-entry page completes the flow.
- Email changes are verified before they take effect: the current address stays the login identity until the new one is confirmed, so a mistyped address can't lock the user out.
- Passkey assertions and two-factor backup codes are strictly single-use, even when requests arrive at the same time.

**Encryption and key safety**
- Vault key rotation can no longer strand a file under a discarded key, and a lost or mismatched vault key no longer overwrites the database key — the server refuses to start instead, keeping data recoverable.
- The file paths of skipped uploads are now encrypted at rest like every other field, and values inserted into HTML email templates are escaped.
- The first-response session-key exchange requires a verified client certificate.

**Sync, uploads, and access control**
- Revoking a sync API key immediately disconnects its live connection, and the sync channel enforces the IP blocklist and per-IP rate limits.
- API-key scopes are enforced on read and mutating routes, so a narrowly-scoped key can't act with full administrator authority.
- Uploads that fail content validation are rejected, and the global and per-account storage quotas hold under concurrent uploads.

**Reliability**
- Scheduled backups run at the configured time of day and survive restarts; automatic client-certificate renewal no longer locks out an offline sync client.
- Large sync catch-up requests are paged, and corrupted stored data fails with a clear error instead of crashing.

### Description correction

The app `description` previously stated the wire is "end-to-end post-quantum," which isn't accurate on Umbrel's plain-HTTP local-network origin (there's no TLS on that hop). The text now scopes the post-quantum TLS key exchange to HTTPS access (custom domain / Tor / Tailscale) while noting that the at-rest encryption applies in all cases.

`releaseNotes` in `umbrel-app.yml` carries the same summary for the App Store update card.

### DIFF
--- a/hermitstash/docker-compose.yml
+++ b/hermitstash/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
-    image: ghcr.io/dotcoocoo/hermitstash:1.12.11@sha256:1c0c7b52e6a09eb600f9fc0dca683d273ebe3ac3c523d756daad7b270eefb843
+    image: ghcr.io/dotcoocoo/hermitstash:1.13.0@sha256:e36518ec236d7576f835ec804723626bf3b009125e8e14d4683911a4d414fe98
     user: "1000:1000"
     init: true
     restart: on-failure
--- a/hermitstash/umbrel-app.yml
+++ b/hermitstash/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: hermitstash
 category: files
 name: HermitStash
-version: "1.12.11"
+version: "1.13.0"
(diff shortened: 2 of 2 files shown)


nmfretz:
Wonderful, thanks for this update and the wording tweaks @dotCooCoo.

PR #5801, opened 2026-06-27
Add LaWallet NWC

agustinkassis:
## Type

New app

## App

App ID: lawallet-nwc
Upstream project: https://github.com/lawalletio/lawallet-nwc
Version: 1.0.9

## Summary

Adds LaWallet NWC to the official Umbrel App Store as a standalone Bitcoin app.

LaWallet NWC is an open-source Lightning Address platform and Nostr CRM with an admin dashboard, user wallet, BoltCard management, Lightning Address routing, and Nostr-based login.

This package uses the upstream published Docker Hub image and an internal Postgres service with app data persisted under `${APP_DATA_DIR}/data/postgres`.

## Verification

Umbrel testing performed:

Environment tested:
- [x] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [x] amd64
- [x] arm64

Static checks performed:
- `npm install`
- `npm run lint:apps -- lawallet-nwc --check-images`
- `git diff --check`
- `docker buildx imagetools inspect masize/lawallet-nwc:1.0.9`
- `docker buildx imagetools inspect postgres:15-alpine`

## Notes

- LaWallet NWC is pre-alpha software; the manifest description warns users not to use it with real funds until they are comfortable with the risks.
- The app is packaged as standalone with `dependencies: []`. Users can connect NWC wallets manually after install.
- The package sets `PROXY_AUTH_ADD: "false"` because LaWallet exposes public Lightning Address, Nostr discovery, BoltCard, wallet activation, and API endpoints. Sensitive routes use LaWallet's own Nostr/JWT authorization.
- No default Umbrel credentials are displayed; users authenticate through the app's Nostr/login flow.
- Images are pinned by tag plus manifest digest:
  - `masize/lawallet-nwc:1.0.9@sha256:cb1a8b87154cb062eb902786ae92b817c866da913e0f8a1df5c31833e64ca4d6`
  - `postgres:15-alpine@sha256:cd17e2ac98240fce1541ad2a803b34009b4eea5aec8a832363cdc7eca62e722e`
- Logo reference: https://raw.githubusercontent.com/lawalletio/lawallet-nwc/main/apps/web/public/logos/lawallet.svg
- Screenshot/source image references:
  - https://raw.githubusercontent.com/lawalletio/umbrel-app-store/master/lawallet-nwc/gallery/1-lightning-address.jpg
  - https://raw.githubusercontent.com/lawalletio/umbrel-app-store/master/lawallet-nwc/gallery/2-all-in-one.jpg
  - https://raw.githubusercontent.com/lawalletio/umbrel-app-store/master/lawallet-nwc/gallery/3-users.jpg
  - https://raw.githubusercontent.com/lawalletio/umbrel-app-store/master/lawallet-nwc/gallery/4-card-designs.jpg
  - https://raw.githubusercontent.com/lawalletio/umbrel-app-store/master/lawallet-nwc/gallery/5-branding.jpg


agustinkassis:
@nmfretz @lukechilds @kaichappel please check it out 🙏
This is serious work, funded by [OpenSats](https://opensats.org/blog/fifteenth-wave-of-bitcoin-grants#lawallet).

nmfretz:
Huge apologies for the wait here @agustinkassis! I had my head down on other work. Thanks a ton for this submission.

I’ve just pushed a few updates to bring the package up to `v1.4.0` and wire in the new listener service. The app and listener are both starting cleanly and coming back after a restart as expected.

The one thing I want to check with you is `PROXY_AUTH_ADD: "false"`. Are you happy for us to leave Umbrel auth enabled by default and only whitelist the routes that need public access?

If the user is already logged into Umbrel, they won’t see another Umbrel prompt when opening the app. This way they also get their Umbrel login protection and 2FA in front of it, if they have that enabled.

We’re also about to ship a setting that lets users turn Umbrel auth off per app, so anyone who wants direct access can still choose that.

It would also protect the initial setup, so another device that can reach the app can’t claim the root admin before the Umbrel owner gets there.

If you’re good with that, I’ll take care of the whitelist on our side.

agustinkassis:
Hello @nmfretz, Thank you very much for taking the time to review it. I really appreciate it.
I'm not entirely sure about `PROXY_AUTH_ADD`. The whole app is meant to be publicly accessible, either by configuring nginx for SSL termination or by using Cloudflare for tunneling. This service is designed for communities that provide lightning addresses management to their users, they need access to the dashboard and wallet. If we set `PROXY_AUTH_ADD` to `true` I would need to whitelist the entire app paths anyway.

Regarding the initial root installation/onboarding process, for someone to front-run the root admin (first run - initial setup), the server would need to have all ports publicly open — something that is quite rare in an Umbrel environment.
I know of one similar case that happened with Alby Hub, where the port was left publicly exposed, the user forgot to make the initial root setup, someone else gained root access during the first run, initial setup and took funds from the lnd.
If that were to happen here, it shouldn’t be a problem, since LaWallet doesn’t have access to any sensitive data — only to previously configured NWC connections from users, not to internal node operations.

I prefer to leave `PROXY_AUTH_ADD` to `false` until you roll the user switch setting, that would give nice versatility.

Thank you again for your time.

agustinkassis:
Updated this submission from 1.4.0 to **2.5.1**, matching the current package in our store ([lawalletio/umbrel-app-store](https://github.com/lawalletio/umbrel-app-store/tree/master/lawallet-nwc)).

**The reason this needed updating is not the version bump.** Since 2.1.0 the app requires `NWC_VAULT_SECRET` — it encrypts stored NWC connection strings, the LUD-16 proxy credentials and the NIP-57 receipt signer, and both `web` and `listener` refuse to boot without it once any remote wallet exists. This submission never declared it, so merging it against a current image would have produced a crash-looping install. It is now generated in `exports.sh` via `derive_entropy`, together with `KEY_VAULT_SECRET` (passkey signup) and `LISTENER_REQUEST_AUTH_SECRET`.

Both vault secrets are one-way doors — envelopes are AES-256-GCM and only the exact secret decrypts them — so their `derive_entropy` identifiers must stay stable once an install has run. That's called out in `exports.sh` so a future rename doesn't silently destroy user data.

Other changes:
- both app images pinned to 2.5.1 by digest; the postgres pin is unchanged
- `port` 9877 → 2288 to match the upstream package — checked it against the other 380 apps in this store and it's unused
- adds the 5 gallery screenshots, plus the current description and `path`

**Verified locally** with exactly the compose in this PR and `derive_entropy`-shaped secret values: 3/3 containers healthy, `/api/health` 200, listener `/status` reporting 2.5.1, 44 migrations applied on a clean volume, and `NWC_VAULT_SECRET` resolving identically in web and listener.

Happy to adjust anything that doesn't fit the store's conventions.

claudiomolt:
@nmfretz Umbrel login is now on by default (`PROXY_AUTH_ADD: "true"`).

umbrelOS 2.0 shipped the per-app **Require Umbrel login** setting, so an operator who wants the member wallet, signup, and activation links reachable without an Umbrel account can turn that off for this app. It applies without a restart and survives updates. `/admin`, `/wallet`, setup, and the passkey/JWT routes stay behind Umbrel login until they do.

Protocol clients have no Umbrel session, so these paths are whitelisted:

- `/.well-known/lnurlp/*` — LUD-16 lookup (the app rewrites this to `/api/lud16/:username`; the gateway sees the original path)
- `/api/lud16/*` — pay callback and LUD-21 verify
- `/.well-known/nostr.json` — NIP-05
- `/.well-known/lawallet.json` — domain probe
- `/api/cards/*` — BoltCard `scan` / `scan/cb` and the public OTC lookup

The gateway only matches exact paths and trailing wildcards, so BoltCard cannot be listed without the rest of `/api/cards`. Mutating card routes still require LaWallet auth. `/api/health` is unlisted; the healthcheck runs inside the container.

The same `app_proxy` block is in [lawalletio/umbrel-app-store](https://github.com/lawalletio/umbrel-app-store/blob/master/lawallet-nwc/docker-compose.yml). Images, secrets, and ports are unchanged.

claudiomolt:
@nmfretz This submission is now on **2.7.1**, the newest published release (21 Sep 2026). Both app images are pinned by multi-arch manifest digest, with `linux/amd64` and `linux/arm64` in the index:

- `masize/lawallet-nwc:2.7.1@sha256:37455a5beeb0b161b314324d9589291f70abde1efeb9949e129c6e1e25a7bf1a`
- `masize/lawallet-nwc-listener:2.7.1@sha256:c4652268b3ead934148d279e614be702d00ff2e3acb710cba9bd3d96957e15d9`

The Postgres pin and the Umbrel auth whitelist are unchanged. The community store package was already on 2.7.1.

PR #5806, opened 2026-06-28
fix(dockge): remove stale dockerd pidfile at the correct path so the app recovers after a bad shutdown

ptrinh:
## Problem

The Dockge app fails to start after an unclean shutdown (power loss, hard reboot, OOM, `docker kill`) with:

```
failed to start daemon: pid file found, ensure docker is not running or delete /data/docker.pid
```

(on the host: `app-data/dockge/data/docker/docker.pid`). The app's Docker-in-Docker daemon leaves a stale pidfile behind, and on the next boot `dockerd` refuses to start because the pidfile already exists. Recovery currently requires manually deleting the file over SSH.

## Root cause

`dockge/entrypoint.sh` already tries to handle this, but cleans the **wrong path**:

```sh
pidfile="/var/run/docker.pid"      # cleaned here
...
rm -f "${pidfile}"
```

while `docker-compose.yml` starts the daemon with an explicit, different pidfile path:

```yaml
command: >
  dockerd
    ...
    --pidfile /data/docker.pid       # dockerd actually writes here
volumes:
  - ${APP_DATA_DIR}/data/docker:/data
```

Inside the dind `docker` container `/var/run` is just the container's ephemeral tmpfs (not the mounted data volume), so `/var/run/docker.pid` is essentially always absent and the cleanup is a **no-op**. The real stale pidfile at `/data/docker.pid` is never removed.

## Fix

Remove the stale pidfile at the path `dockerd` is actually configured to use (`/data/docker.pid`), unconditionally with `rm -f` (a no-op when absent — also avoids relying on the `[[ ]]` bashism under the image's `/bin/sh`):

```sh
rm -f /data/docker.pid
```

One-line behavior change in `dockge/entrypoint.sh`; no compose/data changes. After this, Dockge recovers on its own after a bad shutdown instead of getting stuck.

### DIFF
--- a/dockge/entrypoint.sh
+++ b/dockge/entrypoint.sh
@@ -2,12 +2,13 @@
 
 # This hack can be removed if https://github.com/docker-library/docker/pull/444 gets merged.
 
-# Remove docker pidfile if it exists to ensure Docker can start up after a bad shutdown
-pidfile="/var/run/docker.pid"
-if [[ -f "${pidfile}" ]]
-then
-    rm -f "${pidfile}"
-fi
+# Remove a stale docker pidfile left behind by an unclean shutdown (power loss,
+# hard stop, OOM) so dockerd can start again. This must match the --pidfile path
+# set in docker-compose.yml (/data/docker.pid). The previous code removed
+# /var/run/docker.pid, which is not the file dockerd actually writes here, so the
+# cleanup was a no-op and Dockge failed to start with "pid file found" after a
+# bad shutdown. `rm -f` is a no-op when the file is absent.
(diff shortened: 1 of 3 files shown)


nmfretz:
Thanks @ptrinh, great catch!

I pushed a few small follow-up commits to make this work properly as an Umbrel app update too. Since existing installs don't automatically receive arbitrary top-level files like `entrypoint.sh`, the updated `pre-start` hook now refreshes the installed entrypoint before the DinD sidecar starts. I also bumped the  version in the manifest to `1.5.0-2` so existing users see the update.

PR #5814, opened 2026-06-29
Fix Tailscale subnet routing for kernel TUN mode

lukechilds:
Tailscale was defaulting to iptables mode inside the container, which installed subnet-router forwarding and masquerade rules into iptables-legacy on umbrelOS. Docker and umbrelOS use the nftables-backed firewall path, so forwarded subnet traffic could still hit the nftables FORWARD policy and fail before NAT.

Forcing nftables makes Tailscale install its ts-forward and ts-postrouting rules into the same firewall backend used by Docker.

This was working fine previously since we used userspace networking mode.

### DIFF
--- a/tailscale/umbrel-app.yml
+++ b/tailscale/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: tailscale
 category: networking
 name: Tailscale
-version: "v1.98.3-1"
+version: "v1.98.3-2"
 tagline: Zero config VPN to access your Umbrel from anywhere
 description: >-
   Tailscale is zero config VPN that creates a secure network between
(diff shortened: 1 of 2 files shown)


tailscale/umbrel-app.yml
```diff
@@ -30,10 +30,10 @@ torOnly: false
 submitter: Umbrel
 submission: https://github.com/getumbrel/umbrel/pull/1248
 releaseNotes: >-
-  This update lets umbrelOS use the tailnet routes created by the Tailscale app.
+  This update fixes subnet routing when the Tailscale app is running in kernel networking mode.
 
 
-  Good news for umbrelOS Backups: you can now back up one Umbrel to another over Tailscale. When adding the destination, use its Tailscale IP address shown in the Tailscale app; Tailscale device names are not supported.
+  The Tailscale app now installs its firewall rules using nftables, matching the firewall backend used by umbrelOS and Docker.
```
lukechilds:
```suggestion
  This update lets umbrelOS use the tailnet routes created by the Tailscale app. It also fixes a regression that prevented subnet routing from working properly in the last version.


  Good news for umbrelOS Backups: you can now back up one Umbrel to another over Tailscale. When adding the destination, use its Tailscale IP address shown in the Tailscale app; Tailscale device names are not supported.
```

PR #5815, opened 2026-06-30
tela: new app

zcag:
## tela — markdown-native team wiki with first-class AI agents

Adds a new app package: **tela**, a self-hostable, markdown-native team wiki.

tela logo

- **App / version:** tela `0.7.0`
- **Category:** files
- **Upstream project:** https://github.com/zcag/tela (AGPL-3.0) · website https://telawiki.com
- **Manifest port:** 8780 (fronted by `app_proxy`)

### What it is
Go + PostgreSQL backend with a React editor featuring **live Yjs collaborative editing** — teammates write together in real time, and every page is canonical markdown. Features: ranked full-text search (built in) plus optional semantic search / "ask your docs", a **built-in MCP server so AI agents are first-class** editors and readers, WebDAV sync (edit pages as files), public/blog spaces, Slidev decks, and PDF export.

### Architecture (services)
- `proxy` — Caddy, the single web entrypoint (serves the SPA, reverse-proxies `/api` to the backend). `app_proxy` → `tela_proxy_1:80`.
- `backend` — Go API (:8080); runs embedded Postgres migrations on first connect.
- `frontend` — nginx serving the built React assets.
- `postgres` — `pgvector/pgvector:pg17` (the `vector` extension is required for semantic search). **All user content and uploaded attachments live in Postgres** — tela has no on-disk blob store, so `/data/db` holds everything.
- `gotenberg` — headless-Chromium HTML→PDF for export (internal only).
- `deck` — Slidev render sidecar; `/data` cache is regenerable (`backupIgnore`).

### First-run experience
After install, open the app and complete the one-time **`/setup`** wizard to create the admin account. No SSH, CLI, default credentials, or config-file edits. `PROXY_AUTH_ADD: "false"` because tela ships its own email-first auth and exposes endpoints that can't carry an Umbrel session cookie (MCP `/api/mcp` bearer auth, WebDAV `/dav` PAT-as-Basic, public spaces, password-gated share links).

### Images
- Backend / frontend / proxy / deck: `ghcr.io/zcag/tela-*:0.7.0`, public, pinned by multi-arch manifest-list digest. Verified `linux/amd64` + `linux/arm64`.
- `pgvector/pgvector:pg17` and `gotenberg/gotenberg:8` — official, multi-arch, pinned by digest.

### Secrets / persistence
Postgres password, `TELA_SHARE_SECRET`, and `TELA_API_KEY_SECRET` are derived per-install via `derive_entropy` in `exports.sh` (stable across restarts/updates — rotating them would invalidate PATs and share links). Durable state is the Postgres bind mount under `${APP_DATA_DIR}/data/db`.

### Host access
None. No Docker socket, no privileged mode, no host networking, no device mounts, no extra capabilities. Only `app_proxy` publishes a port.

### Testing
- `npm run lint:apps -- tela --check-images` → **No issues found** (manifest, port uniqueness, multi-arch image pinning/pullability, app_proxy wiring, persistence paths all pass).
- The image set was smoke-tested via `docker compose`: proxy serves the SPA and routes `/api` to the backend; backend migrates and boots against `pgvector/pgvector:pg17`.

### Screenshots
| Page (light) | Search (light) |
|---|---|
|  |  |

| Page (dark) | Search (dark) |
|---|---|
|  |  |

### DIFF
--- a/tela/docker-compose.yml
+++ b/tela/docker-compose.yml
@@ -5,13 +5,19 @@ services:
     environment:
       APP_HOST: tela_proxy_1
       APP_PORT: 80
-      # tela ships its own email-first auth (the /setup wizard creates the admin
-      # on first run) and exposes endpoints that cannot carry an Umbrel session
-      # cookie: the built-in MCP server (/api/mcp, bearer-authed), WebDAV sync
-      # (/dav, PAT-as-Basic), public/blog spaces, and password-gated share links.
-      # Fronting all of that with Umbrel auth would break those clients, so the
-      # app handles its own access control.
-      PROXY_AUTH_ADD: "false"
+      # tela is a multi-user team wiki that ships its own email-first auth, and
+      # exposes endpoints that cannot carry an Umbrel session cookie: the MCP
+      # server (/api/mcp, bearer-authed), WebDAV sync (/dav, PAT-as-Basic),
+      # public/blog spaces, and password-gated share links. Fronting the whole
+      # app with Umbrel's single-owner auth would lock out non-owner teammates
+      # and break those clients — so the app owns access control (WHITELIST "*").
+      # The one exception is first-run admin setup: gate /setup + /api/setup
+      # behind Umbrel auth so only the box owner can claim the first admin
+      # account (a LAN device can't race to it). Blacklist wins over whitelist;
+      # bare rules match that exact path, so /api/setup/status stays public.
+      PROXY_AUTH_ADD: "true"
+      PROXY_AUTH_WHITELIST: "*"
+      PROXY_AUTH_BLACKLIST: "/setup,/api/setup"
 
   # Caddy — the single web entrypoint. Serves the SPA and reverse-proxies
   # /api to the backend. Listens on :80 internally; app_proxy fronts it.
--- a/tela/docker-compose.yml
+++ b/tela/docker-compose.yml
@@ -1,0 +1,94 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: tela_proxy_1
+      APP_PORT: 80
+      # tela ships its own email-first auth (the /setup wizard creates the admin
+      # on first run) and exposes endpoints that cannot carry an Umbrel session
+      # cookie: the built-in MCP server (/api/mcp, bearer-authed), WebDAV sync
+      # (/dav, PAT-as-Basic), public/blog spaces, and password-gated share links.
+      # Fronting all of that with Umbrel auth would break those clients, so the
+      # app handles its own access control.
+      PROXY_AUTH_ADD: "false"
+
+  # Caddy — the single web entrypoint. Serves the SPA and reverse-proxies
+  # /api to the backend. Listens on :80 internally; app_proxy fronts it.
+  proxy:
+    image: ghcr.io/zcag/tela-proxy:0.7.0@sha256:e320276641781f38f48c97ec10dc3432cc4ecf933ee63c5946d155ded42f81f7
+    restart: on-failure
+    environment:
+      TELA_SITE_ADDRESS: ":80"
+      TELA_UPSTREAM_BACKEND: backend:8080
+      TELA_UPSTREAM_FRONTEND: frontend:80
+    depends_on:
+      - backend
+      - frontend
+
+  # Go API server. Runs its embedded Postgres migrations on first connect.
+  backend:
+    image: ghcr.io/zcag/tela-backend:0.7.0@sha256:fe0f23fcd53fb7903889d8659844edf11d0f7f8aef9572251ab4f336934e2a9f
+    restart: on-failure
+    environment:
+      TELA_PUBLIC_BASE_URL: "http://${DEVICE_DOMAIN_NAME}:${APP_TELA_PORT}"
+      TELA_DATABASE_URL: "postgres://tela:${APP_TELA_POSTGRES_PASSWORD}@postgres:5432/tela?sslmode=disable"
+      TELA_SHARE_SECRET: ${APP_TELA_SHARE_SECRET}
+      TELA_API_KEY_SECRET: ${APP_TELA_API_KEY_SECRET}
+      # PDF export: the backend POSTs to gotenberg, which renders the reader via
+      # the internal proxy origin (never a public URL).
+      TELA_GOTENBERG_URL: http://gotenberg:3000
+      TELA_PDF_RENDER_BASE_URL: http://proxy
+      # Slidev deck render sidecar.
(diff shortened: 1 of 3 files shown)


tela/umbrel-app.yml
```diff
@@ -0,0 +1,51 @@
...
+  All data stays on your server in PostgreSQL. tela is open source (AGPL-3.0).
+releaseNotes: ""
+developer: zcag
+website: https://telawiki.com
+dependencies: []
+repo: https://github.com/zcag/tela
+support: https://github.com/zcag/tela
+port: 8780
+gallery: []
+path: ""
```
nmfretz: This currently opens Tela’s hosted landing page, where the login and get-started links lead to `telawiki.com` instead of the user’s local instance.

Can we launch the local login route instead? It already sends a fresh install to `/setup` and works for returning users too.

```suggestion
path: "/login"
```

tela/docker-compose.yml
```diff
@@ -0,0 +1,94 @@
...
+    environment:
+      APP_HOST: tela_proxy_1
+      APP_PORT: 80
+      # tela ships its own email-first auth (the /setup wizard creates the admin
+      # on first run) and exposes endpoints that cannot carry an Umbrel session
+      # cookie: the built-in MCP server (/api/mcp, bearer-authed), WebDAV sync
+      # (/dav, PAT-as-Basic), public/blog spaces, and password-gated share links.
+      # Fronting all of that with Umbrel auth would break those clients, so the
+      # app handles its own access control.
+      PROXY_AUTH_ADD: "false"
```
nmfretz: One thing I wanted to check with you is `PROXY_AUTH_ADD: "false"`. Are you happy for us to leave Umbrel auth enabled by default and only whitelist the routes that need public or client access?

If the user is already logged into Umbrel, they won’t see another Umbrel login prompt when opening Tela. This way they also get their Umbrel login protection and 2FA in front of it, if they have that enabled.

We’re also about to ship a setting that lets users turn Umbrel auth off per app, so anyone who wants direct access can still choose that.

It would also protect the initial setup, so another device that can reach the app can’t claim the first admin account before the Umbrel owner gets there.

If you’re good with that, can you let me know which routes need to stay public? I can handle the whitelist on our side.

tela/umbrel-app.yml
```diff
@@ -0,0 +1,51 @@
...
+  - Ranked full-text search built in, plus optional semantic search and
+  "ask your docs" when you point tela at an embedding/LLM endpoint.
+
+  - A built-in MCP server so AI agents are first-class editors and readers of
+  your wiki, alongside Atlas agent workflows.
+
+  - WebDAV sync to edit pages as files from your machine, public/blog spaces,
+  Slidev decks, and PDF export.
+
+  - Email-first accounts with self-registration, granular space permissions,
```
nmfretz, line 32: One other thing I noticed: this package doesn’t configure SMTP, so account verification and password-reset links are written to the container logs instead of being delivered. That means the self-registration flow described here won’t work normally on a fresh install.

Is there an SMTP setup path I missed, or should we default registration to closed until SMTP is configured? Admin-created users already work, so that may be the simpler default.

tela/docker-compose.yml
```diff
@@ -0,0 +1,94 @@
...
+  frontend:
+    image: ghcr.io/zcag/tela-frontend:0.7.0@sha256:916648383a9856bc80647ebc767f395c6292514ffcba23db1866e19c86189731
+    restart: on-failure
+
+  # Primary datastore. pgvector/pgvector ships the `vector` extension on top of
+  # the official postgres:17 image — tela's semantic search needs CREATE
+  # EXTENSION vector. All user content, pages, and uploaded attachments live
+  # here (tela has no on-disk blob store), so this volume holds everything.
+  postgres:
+    image: pgvector/pgvector:pg17@sha256:be400b50812ab2cc908ed78593fda2e51e3b45fe774fa637f1c7b16e68531d95
```
nmfretz: Can we use the exact version tag that matches this digest? `pg17` has moved since the PR was opened, while `0.8.3-pg17` still resolves to the pinned image.

```suggestion
    image: pgvector/pgvector:0.8.3-pg17@sha256:be400b50812ab2cc908ed78593fda2e51e3b45fe774fa637f1c7b16e68531d95
```

tela/docker-compose.yml
```diff
@@ -0,0 +1,94 @@
...
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U tela -d tela"]
+      interval: 5s
+      timeout: 3s
+      retries: 20
+      start_period: 10s
+
+  # PDF export engine (headless-Chromium HTML→PDF). Internal only.
+  gotenberg:
+    image: gotenberg/gotenberg:8@sha256:67097317623a503ba2a6a7e9ae8db6929a1f7e1bbd88077bacf2d325fbdab923
```
nmfretz: Can we use the exact Gotenberg version tag here too?

```suggestion
    image: gotenberg/gotenberg:8.34.0@sha256:67097317623a503ba2a6a7e9ae8db6929a1f7e1bbd88077bacf2d325fbdab923
```

nmfretz (commented):
Huge apologies for the wait here @zcag! Thanks a ton for the submission.

I gave it a fresh install and restart test. Everything starts cleanly and content comes back as expected.

I left a few things inline around the launch path, Umbrel auth, and the first-run account flow. v0.8.0 landed while this was waiting on us, so I’ll take care of bringing the package up to the current release on our side once we sort those.

tela/umbrel-app.yml
zcag: Updated, hosted landing page is a bit or tech debt on my side that will be patched in a proper self-host package, updating the path fixes the issue for now, thanks

tela/umbrel-app.yml
zcag: You’re right, i was assuming there would be a way for users to have their env’s configured for apps in umbrel, but makes sense to keep things web-ui setup only per umbrel design so its more straightforward. Added an option to disable registration by default on start, kept it off in manifest, and bumped the tela version to latest

tela/docker-compose.yml
zcag: updated

tela/docker-compose.yml
zcag: updated

tela/docker-compose.yml
zcag: for general use i’m not sure how enabling proxy auth would work with different users etc, but good point about the initial setup flow. found this `PROXY_AUTH_BLACKLIST` options and set it to tela’s setup paths, while having `*` for the whitelist. What do you think about this way of handling it?

PR #5818, opened 2026-06-30
Remove unneeded shebang and exec bit from exports.sh files

mrv777:
## Type

Metadata only

## App

App ID: albyhub, nostr-vpn, onlyoffice-nextcloud, paperclip, paperless
Upstream project: n/a (packaging cleanup)
Version: unchanged

## Summary

umbrelOS sources `exports.sh`, so a leading shebang line and the executable bit don't do
anything. The repo linter flags both (`exports.shebang`, `exports.executable`). This clears
those warnings without changing behavior:

- albyhub, nostr-vpn, onlyoffice-nextcloud: dropped the leading shebang line
- paperclip, paperless: changed the file mode from 755 to 644

The files still get sourced exactly the same way.

## Verification

Linter: ran on the 5 apps before and after — the targeted warnings are gone and nothing else changed.

Runtime smoke test on a real Umbrel (umbrelOS 1.7.3, amd64): installed paperclip with the
non-executable `exports.sh` (one of the two mode changes). It installs and reaches ready, the
app serves (HTTP 200 through the app_proxy), and `exports.sh` lands non-executable on the device —
confirming umbrelOS sources it fine without the exec bit. (Tested via a throwaway community-store
clone rather than touching the device's managed official source.)

Environment tested:
- [x] Umbrel device (umbrelOS 1.7.3)

Architecture tested:
- [x] amd64
- [ ] arm64

Known lint warnings or caveats: a few pre-existing, unrelated warnings on these apps (nostr-vpn
host networking/devices/cap_add, paperless exports.lifecycle) are left as-is.

## Notes

Cosmetic packaging change only. No permissions, dependencies, migrations, or credentials affected.


nmfretz:
Thanks @mrv777! Appreciate you picking up these new lint warnings so quickly. I missed those when I created the linter the other day.

Just FYI, this won't update existing apps or change anything behaviourally: none of the manifest versions are changing, so installed apps won't see an app update from this PR. And when umbrelOS loads `exports.sh`, it sources the file into the app lifecycle shell rather than executing it directly, so the executable bit isn't used and the shebang is ignored. So it's just satisfying the new linter.

Thanks again for the cleanup!

PR #5819, opened 2026-06-30
Fix blockstream-blind-oracle shared server key (generate a unique per-install keypair)

mrv777:
## Type

App update

## App

App ID: blockstream-blind-oracle
Upstream project: https://github.com/Blockstream/blind_pin_server
Version: 0.1.2 → 0.1.3

## Summary

The package commits a real server keypair (`data/server_private_key.key` and
`server_public_key.pub`), and the compose file bind-mounts those files into both the `node` and
`web` containers. So every install ships the same oracle keypair, and the private key is readable
in this repo and its git history. Since the blind oracle holds the key that unlocks a user's Jade
PIN, a shared key defeats the point of the app.

Upstream doesn't auto-generate keys; it ships a `generateserverkey` module you're meant to run
manually, which is presumably why a static key got committed.

What this PR does:

- Replaces the committed keypair with empty placeholder files. umbrelOS (and the linter) require
  committed sources for bind mounts, so the files have to exist.
- Generates the keypair on first start in `hooks/pre-start`: if the key is still the empty
  placeholder, it removes the placeholders and runs the node image's own `generateserverkey` to
  write a unique key into `${APP_DATA_DIR}/data/`. It's guarded so it never overwrites a real key
  on later starts, and it bails out (clearing partials) if generation fails, so the next start retries.
- Drops the unneeded executable bit on the sourced `exports.sh`.

## Verification

Tested on a real Umbrel (umbrelOS 1.7.3, amd64):

- **Fresh install** turns the empty placeholders into a real 32-byte private / 33-byte public keypair, and the app comes up (node + web + tor all running).
- **Per-install uniqueness** (the whole point): installed, recorded the private key's SHA256, uninstalled, reinstalled — the second install produced a *different* key. So two installs no longer share a keypair.
- **Idempotent**: restarting the app leaves the key untouched (same hash before/after) — the guard correctly skips regeneration when a real key exists.
- node and web share the one keypair via the bind mounts, as intended.
- Linter passes (one harmless `package.runtime_artifact` warning on the empty placeholder).

(Tested via a throwaway community-store clone of this package rather than overwriting the device's managed official source. The clone needed the `node` container name updated in its `torrc` because it's renamed — the official package is unaffected.)

Environment tested:
- [x] Umbrel device (umbrelOS 1.7.3)

Architecture tested:
- [x] amd64
- [ ] arm64

## Notes / open questions

1. Existing installs keep working and are not migrated. On update umbrelOS only copies
   `docker-compose.yml`, top-level templates, `exports.sh`, `torrc`, `hooks`, and `umbrel-app.yml`
   into app data, not the `data/` directory, so an existing install keeps its current key file and
   the pre-start guard skips generation for it. Nothing breaks, but those users stay on the old
   shared key. That key is also in this repo's git history, so it should be treated as compromised.
   Rotating it automatically isn't safe, since it would break an already-paired Jade (its data is
   tied to the old key), so moving to a unique key has to be a deliberate reset and re-pair by the
   user. The release notes say this. I can add an opt-in rotation path if you want one.
2. Worth coordinating with Blockstream (the original submitter) on the key handling.
3. The empty placeholder files only exist to satisfy the bind-mount source requirement. Happy to
   switch to a cleaner approach if you'd prefer one.


nmfretz:
Hey @mrv777, the shared committed keypair definitely looks concerning at first glance, and I remember checking this pretty carefully when the app was first submitted by the Blockstream devs.

I double-checked and fresh installs already generate a unique keypair. The `web` container has startup logic that checks for the bundled public key (`0332b360...`) and, when it sees that sentinel key, generates a new private/public keypair and writes it back to the bind-mounted key files. The existing `pre-start` hook intentionally starts `web` before `node`, so the node then starts with the generated per-install key.

Closing.

PR #5822, opened 2026-06-30
portainer: pin bundled DinD back to 27.2.0

### LINKED ISSUE #5802: Portainer v2.39.4: DinD bump to docker:29.6.0 breaks host iptables and all container networking
### App ID

Portainer

### umbrelOS version

1.7.3

### Device architecture

I don't know

### When did this happen?

App update

### What failed?

Other

### Expected behavior

Portainer's DinD sidecar should not interfere with the host Docker daemon's iptables rules.

### Actual behavior

After the Portainer v2.39.4 update, the DinD container (`docker:29.6.0-dind`) starts its own Docker daemon on the host network. This daemon overwrites the host's `DOCKER-FORWARD` iptables chain, inserting ACCEPT rules only for its own bridges (`br-1044f58e916e`, `dind0`) while leaving the chain's default policy at DROP. The host's `umbrel_main_network` bridge (`br-d15187fee91b`) is not included, so all outbound traffic from every other Umbrel app container is silently dropped. In my case, cloudflared lost its tunnel connection and could not re-establish it, effectively taking all Cloudflare-tunneled services offline. DNS resolution, QUIC, TCP, and Cloudflare API connectivity all failed from within containers, while the host itself retained full internet access.

After updating Portainer from v2.39.3 to v2.39.4, all containers on the `umbrel_main_network` bridge lose outbound internet connectivity. This breaks cloudflared tunnels, Immich, and any other app that needs to reach the internet.

## Root Cause

The v2.39.4 update bumped the bundled DinD sidecar from `docker:27.2.0-dind` to `docker:29.6.0-dind`. Because this container runs with `privileged: true` and `network_mode: host`, the DinD's Docker daemon takes control of the host's iptables chains (`DOCKER-FORWARD`, `DOCKER-CT`, `DOCKER-BRIDGE`).

Docker 28+ introduced a new firewall management system. The DinD daemon only adds ACCEPT rules for its own networks (`br-1044f58e916e`, `dind0`) while the default policy for `DOCKER-FORWARD` is DROP. This effectively blocks all traffic for `umbrel_main_network` (`br-d15187fee91b`), killing outbound connectivity for every other Umbrel app.

## Steps to Reproduce

1. Install Portainer on UmbrelOS 1.7.3
2. Update to v2.39.4 (or install fresh)
3. Wait ~1 minute for the DinD container to start and overwrite iptables
4. Observe that all other apps lose internet access (e.g. cloudflared tunnel disconnects)

## Expected Behavior

Portainer's DinD sidecar should not interfere with the host Docker daemon's iptables rules.

## Workaround

Manually edit `/home/umbrel/umbrel/app-data/portainer/docker-compose.yml` and change the DinD image back to:

```
docker:27.2.0-dind@sha256:f9f72ad901a78f27be922b2d320bbc263174f12919c1b37e6a01f828fa904565
```

Then reboot Umbrel (or restart Docker).

### Logs or screenshots

_No response_

mrv777:
The v2.39.4 update bumped the bundled DinD sidecar from `docker:27.2.0-dind` to `docker:29.6.0-dind` alongside the Portainer CE bump. The DinD bump is what's causing trouble.

Docker 28+ rewrites the host `DOCKER-FORWARD` iptables chain when its daemon starts. Because this sidecar runs `privileged: true` with `network_mode: host`, the DinD daemon only inserts ACCEPT rules for its own bridges and leaves `umbrel_main_network` (`br-…`) under the default DROP policy. The result is that every other app loses outbound connectivity a minute or so after Portainer starts — cloudflared tunnels drop, Immich can't reach the internet, etc. The host itself is unaffected, which makes it easy to misdiagnose.

This reverts only the DinD image to the `27.2.0-dind` digest that shipped through 2.39.1–2.39.3 without this problem. Portainer CE stays on 2.39.4, and 27.x is fully compatible with the Portainer CE daemon API. Pre-28 DinD doesn't touch the new firewall manager, so it no longer clobbers the host chain.

Closes #5802.

One thing for maintainers to decide: the manifest version stays `2.39.4`, so let me know if you'd rather re-release to push this to users who already updated.

### DIFF
--- a/portainer/docker-compose.yml
+++ b/portainer/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_ADD: "false"
 
   docker:
-    image: docker:29.6.0-dind@sha256:7bb861a04bb42bda1d237fc2cb539f9823c9b666ecfbfdbd3e534ab74c8cb976
+    image: docker:27.2.0-dind@sha256:f9f72ad901a78f27be922b2d320bbc263174f12919c1b37e6a01f828fa904565
     privileged: true
     network_mode: host
     stop_grace_period: 1m
--- a/portainer/umbrel-app.yml
+++ b/portainer/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1.1
 id: portainer
 category: developer
 name: Portainer
-version: "2.39.4"
+version: "2.39.4-1"
 tagline: Run custom Docker containers on your Umbrel
 description: >-
   ⚠️ Make sure to only use named Docker volumes for your stacks and containers. Data in bind-mounted volumes
@@ -51,7 +51,10 @@ path: ""
 defaultUsername: "admin"
 defaultPassword: "changeme"
 releaseNotes: >-
-  Portainer 2.39.4 LTS includes security fixes, stack management fixes, and reliability improvements. This update also refreshes the bundled Docker runtime used by Portainer.
+  This release keeps Portainer at version 2.39.4 but reverts the bundled Docker runtime to a known-good version after the previous update caused containers to fail to start on some setups.
+
+
+  Release notes for version 2.39.4 remain the same. Portainer 2.39.4 LTS includes security fixes, stack management fixes, and reliability improvements.
 
 
   Highlights include:


nmfretz:
Thanks for going through Issues and getting fixes together @mrv777 🤝. I ended up doing this same change last week https://github.com/getumbrel/umbrel-apps/commit/77dc9baae496dc1c55abb54ef47d970ac9a22fdf

PR #5823, opened 2026-06-30
Update Public Pool's Web to 0.2.0

martinbarilik:
Hi, thanks in advance for merge


mrv777:
I spotted two things that should be checked:

1. The `public-pool-web` image refs are malformed in both `rails-app` and `sidekiq`:

```yaml
martinbarilik/public-pool-web:0.2.0sha256:f708bdea807986de548df4b2011cee47a827034786293bb7305d53a0a99f86a7
```

That should be:

```yaml
martinbarilik/public-pool-web:0.2.0@sha256:f708bdea807986de548df4b2011cee47a827034786293bb7305d53a0a99f86a7
```

The app linter catches this as an `image.pinned` error, and Docker rejects the current ref as `invalid reference format`.

2. The Postgres change looks like it will break the normal Umbrel update path. The old package stored Postgres 17 data under `${APP_DATA_DIR}/data/postgres` mounted at `/var/lib/postgresql/data`, while Postgres 18.4 defaults `PGDATA` to `/var/lib/postgresql/18/docker`. With the new mount, an existing install will initialize a fresh database under `${APP_DATA_DIR}/data/postgres/18/docker` and leave the old data unused.

The release notes say to uninstall before updating, but Umbrel won’t enforce that during a regular update. I think we either need a real migration/dump-restore path, avoid the Postgres major/storage-layout change in this update, or be explicit that this is intentionally accepted as a destructive update before merging.

martinbarilik:
Hi, sorry for the typo in image. 

About 2.

yes, it is intentional breaking change, i am willing to do that now rather then later without really complicating the code, pg 17 is old enough and the app is not used that much at the moment

nmfretz:
Hey @martinbarilik! Sorry for the delay here, and thanks for putting this update together and continuing to maintain Public Pool’s Web.

I took a closer look at the PostgreSQL change. I understand why upgrading now, while the app is still relatively new, feels simpler than dealing with it later. The issue is that we can’t use uninstalling and reinstalling as the update path for an existing Umbrel app.

Umbrel will present 0.2.0 as a normal update, and some users may simply click **Update All** without reading the release notes. Since the existing PostgreSQL 17 database isn’t migrated, the app could return broken or empty. From the user’s perspective, a routine Umbrel update lost their accounts, workers, configuration, and history. They’ll reasonably conclude that the app or Umbrel’s update system is unreliable.

Fortunately, PostgreSQL 18 doesn’t appear necessary here. PostgreSQL 17 is supported until November 2029, and I tested the 0.2.0 migrations against a database created by the currently shipped 0.1.0 package on PostgreSQL 17.6. The migrations completed successfully and the existing data was preserved.

Would you be happy to keep PostgreSQL 17 and the existing `${APP_DATA_DIR}/data/postgres:/var/lib/postgresql/data` mount for this update, and remove the uninstall/breaking-change instruction? The image can still move to a newer 17.x minor release if desired.

martinbarilik:
Hey @nmfretz. I can't go back to pg 17. The app has already been released for start os with pg 18 and i want to keep dependencies in sync. 

For the sake of the functionality i commited the auto upgrade script and changed the image. Little bit of hustle, but it will work for all users ( fresh install and already installed ).

Hope this is ok now. Tested on raspi 5.

nmfretz:
Hey @martinbarilik, thanks a ton for sorting that out. This is much better UX for users. Great work.

I've pushed a small simplification so we don't need the pre-start hook. The pre-start was only needed because the database service was forced to run as UID/GID 1000. I removed both and let pgautoupgrade follow its normal startup (briefly use root to fix ownership, drop to its postgres user, migrate, and keep running. PostgreSQL itself does not run as root). I also tightened the health check and cleaned up the release notes.

I tested the full 0.1.0 to 0.2.0 update on an Umbrel with existing pool, user, and worker data, then restarted the app. Everything migrated cleanly and remained intact.

Going live!

PR #4843, opened 2026-02-22
Update obsidian to v1.11.7

Mwpereira:
Obsidian for umbrel is currently on v1.8.10, looking to update to the latest release, v1.11.7, since there have been many improvements and bug fixes.

cc: @al-lac

PR #5824, opened 2026-06-30
Update Obsidian to 1.12.7

mrv777:
Obsidian has been pinned at `linuxserver/obsidian:1.8.10` since #2679, and upstream has shipped the whole 1.9–1.12 series since then. This bumps it to 1.12.7, the current stable.

The headline change across those releases is Bases, the new core plugin that turns a set of notes into a database with table, list, and map views. There's also the new Footnotes view sidebar, smoother mobile editing, and a long tail of performance and bug fixes. One bigger under-the-hood change: 1.12.x swaps the remote-desktop backend in the LinuxServer image from KasmVNC to Selkies.

One thing worth flagging in the release notes (and I did): 1.9 dropped the singular `tag`/`alias`/`cssclass` frontmatter properties in favor of the plural `tags`/`aliases`/`cssclasses` as lists. Anyone still using the old singular form will need to update their notes.

Nothing changed on the packaging side. It's still the single `server` service persisting to `/config`, and I pinned the new tag to its multi-arch manifest-list digest (`sha256:bbc7d7fc…`) with amd64 + arm64 both present.

Tested on a real Umbrel (umbrelOS 1.7.3, amd64):
- Fresh install of 1.12.7 comes up clean — Selkies starts, the Wayland desktop and websocket server come up, app_proxy reports ready, no s6-overlay errors.
- Update path: started 1.12.7 on top of an existing 1.8.10 `/config` directory (with a marker file dropped in beforehand). It booted cleanly, `[migrations] no migrations found`, and the marker survived — so existing vaults/config carry over despite the KasmVNC→Selkies switch.
- `npm run lint:apps -- obsidian --check-images` clean.

(I tested by mirroring this exact package into a throwaway community-store app rather than overwriting the managed official source on the device; same images and compose, just a different app id and port.)

Changelog: https://obsidian.md/changelog/

### DIFF
--- a/obsidian/docker-compose.yml
+++ b/obsidian/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   server:
-    image: linuxserver/obsidian:1.8.10@sha256:b85e75e572e75f5e00062fb564102eaad8e19970938ee8ec7bae0a96566c1400
+    image: linuxserver/obsidian:1.12.7@sha256:bbc7d7fc7ad76e7fee9f90055f09b81d6bb44631fbf4fb239cbc807a5a7f26b0
     restart: on-failure
     environment:
       - PUID=1000
--- a/obsidian/umbrel-app.yml
+++ b/obsidian/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: obsidian
 category: files
 name: Obsidian
-version: "1.8.10"
+version: "1.12.7"
 tagline: The private and flexible writing app that adapts to the way you think
 description: >-
   **Your thoughts are yours.**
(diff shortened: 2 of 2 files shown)


nmfretz:
Hey @mrv777 we're actually blocked on Obsidian updates until umbrelOS has built-in https (coming really really soon!). There's already a PR opened (#4843) that is tagged as "pending umbrelOS feature" that I will make sure gets updated to latest as soon as we go live with the umbrelOS update.


> Fresh install of 1.12.7 comes up clean — Selkies starts, the Wayland desktop and websocket server come up, app_proxy reports ready, no s6-overlay errors.

This would not have caught the issue, because the newer LinuxServer Obsidian/Selkies images require HTTPS for full browser functionality. I'll close this and continue to track in #4843.

PR #5825, opened 2026-06-30
fix(core-lightning): wait for Tor before starting lightningd

### LINKED ISSUE #5529: core-lightning: lightningd crashes on startup due to Tor race condition — missing healthcheck on tor service
## Summary

The `core-lightning` app ships without a healthcheck on the `tor` service and without `condition: service_healthy` in `lightningd`'s `depends_on`. This causes lightningd to crash on every fresh start or OS update because it tries to connect to the Tor control port before Tor is ready to accept connections.

## Environment

- Umbrel OS: 1.7.1 (Debian 13 / trixie)
- Core Lightning: v25.09.3-modded
- Tor image: `getumbrel/tor:0.4.7.8`
- Hardware: Umbrel PRO
- Triggered by: OS update causing full container recreation

## Symptom

lightningd log shows:

```
**BROKEN** plugin-bcli: Could not connect to bitcoind using bitcoin-cli. Is bitcoind running?
```

This error is **misleading** — Bitcoin is healthy. The real error follows shortly after:

```
connectd: STATUS_FAIL_INTERNAL_ERROR: Connecting stream socket to Tor service
```

lightningd crashes and does not recover. The root cause is that `connectd` calls `status_failed()` — a fatal, non-retrying crash — when it cannot reach the Tor control port at startup.

This is a known upstream CLN bug (issues [#5499](https://github.com/ElementsProject/lightning/issues/5499), [#5617](https://github.com/ElementsProject/lightning/issues/5617), confirmed by Rusty Russell), fixed upstream in CLN v22.11 via [PR #5924](https://github.com/ElementsProject/lightning/pull/5924). However, the Umbrel build (`v25.09.3-modded`) still exhibits the crash, and the stock `docker-compose.yml` does not include the Docker-level guard that prevents the race condition.

## Root Cause

Docker's default `depends_on: tor` only waits for the tor container to **start**, not for Tor to actually be **listening** on its control port. On slower hardware or after bulk container recreation (e.g. OS update), lightningd starts before Tor is ready.

## Fix

Add a healthcheck to the `tor` service and change `lightningd`'s `depends_on` to use `condition: service_healthy`.

### tor service — add healthcheck:

```yaml
    healthcheck:
      test: ["CMD", "bash", "-c", "echo > /dev/tcp/127.0.0.1/29051"]
      interval: 10s
      timeout: 5s
      retries: 12
      start_period: 30s
```

> **Note:** The `getumbrel/tor` image does **not** have `nc` (netcat) installed. The healthcheck must use `bash` with `/dev/tcp`. Use `CMD` (not `CMD-SHELL`) because `/bin/sh` in this image does not support the `/dev/tcp` built-in — only `bash` does.
>
> **Port note:** Umbrel PRO uses control port **29051**. Umbrel Home uses **9051**. Adjust accordingly.

### lightningd service — update depends_on:

```yaml
    depends_on:
      tor:
        condition: service_healthy
```

## Impact

Every Umbrel Core Lightning user is affected on:
- First install (if Tor is slow to start)
- Any Umbrel OS update that recreates containers
- Any manual `umbreld` restart

The node silently fails to start and requires manual intervention: restart the tor container, wait ~15 seconds, then restart lightningd.

## Confirmed Working

After applying this fix and restarting `umbreld`, the node came up cleanly:

```
connectd: Static Tor service onion address: <x>.onion:9735
lightningd: Server started with public key <pubkey>
```

Tor container status: **healthy**. lightningd: **running**. Fix survives subsequent restarts and OS updates.

## References

- CLN issue [#5499](https://github.com/ElementsProject/lightning/issues/5499) — original race condition report
- CLN issue [#5617](https://github.com/ElementsProject/lightning/issues/5617) — confirmed by Rusty Russell
- CLN PR [#5924](https://github.com/ElementsProject/lightning/pull/5924) — upstream fix in v22.11

mrv777:
Fixes #5529. On a fresh start — and especially after an umbrelOS update recreates every container at once — lightningd can come up before Tor is ready. When connectd can't reach the Tor control port it calls `status_failed()`, which is fatal and doesn't retry, so the node just dies. The log makes it look like a bitcoind problem (`plugin-bcli: Could not connect to bitcoind`) but the real line is `connectd: STATUS_FAIL_INTERNAL_ERROR: Connecting stream socket to Tor service`.

There's currently no `depends_on` or healthcheck anywhere in the compose file, so nothing holds lightningd back.

The part that took a bit of digging: lightningd doesn't talk to the bundled `tor` service in this app — that one only publishes the CLNRest hidden service (its torrc is just a `HiddenServiceDir`). The control port lightningd actually uses is `${TOR_PROXY_IP}:29051`, which is umbrelOS's system Tor (`TOR_PROXY_IP` is a framework-global, same one the LND app uses). Since `depends_on` can only reference services inside this compose project, I use the bundled `tor` service as the gate: it gets a healthcheck that probes `${TOR_PROXY_IP}:29051`, and lightningd waits for it to go healthy.

The healthcheck has to use bash's `/dev/tcp` — `getumbrel/tor:0.4.7.8` has no `nc`/`ncat`/`curl`, and its `/bin/sh` doesn't support the `/dev/tcp` builtin (it does have `/bin/bash`), which is why it's `CMD` and not `CMD-SHELL`.

Bumped the manifest to `26.06.1-hotfix.1` so existing installs actually get this — the CLN version itself is unchanged. Lint is clean.

Tested on a real Umbrel (umbrelOS 1.7.3, amd64, with the Bitcoin app as the dependency):
- The `tor` service's healthcheck resolves `${TOR_PROXY_IP}` to the system Tor IP and goes `healthy` (`exec 3<>/dev/tcp/<system-tor-ip>/29051`, failing streak 0) — so the gate clears at the right endpoint instead of deadlocking.
- lightningd waits for that, then starts: connectd binds its static Tor onion successfully (the exact step that crashes in this bug), bitcoind connects fine, and "Server started" — with 0 restarts. No `STATUS_FAIL_INTERNAL_ERROR`.

What I could not do is force the actual race on a healthy device — system Tor is always already up, so I couldn't reproduce the original crash to watch the fix prevent it. So this verifies the gate works and doesn't deadlock or change normal startup; it doesn't demonstrate the failure-then-fixed sequence. If a maintainer can reproduce the OS-update race on real hardware, that'd be the last piece. If you'd rather gate the `app` service too, or check a different port, happy to adjust.

(Tested via a throwaway community-store clone of this package — same compose/exports/hooks, different app id — rather than overwriting the managed official app-store source on the device.)

Upstream context for the non-retrying connectd behavior: ElementsProject/lightning#5499 and #5617.

### DIFF
--- a/core-lightning/docker-compose.yml
+++ b/core-lightning/docker-compose.yml
@@ -51,6 +51,9 @@ services:
   lightningd:
     image: elementsproject/lightningd:v26.06.1@sha256:abc26f1ece2ea2986f4cb5dc0117c81aed2c1662f6fb266838c4208c2f99468d
     restart: on-failure
+    depends_on:
+      tor:
+        condition: service_healthy
     ports:
       - ${APP_CORE_LIGHTNING_DAEMON_PORT}:9735
       - ${APP_CORE_LIGHTNING_WEBSOCKET_PORT}:${APP_CORE_LIGHTNING_WEBSOCKET_PORT}
@@ -89,3 +92,16 @@ services:
       - ${TOR_DATA_DIR}:/data
     environment:
       HOME: "/tmp"
+    # Gate lightningd on the Tor control port being reachable. lightningd's connectd
+    # calls a fatal, non-retrying status_failed() if it can't reach Tor at startup, so
+    # on a fresh start or OS update it crashes when it races ahead of Tor. The control
+    # port lives on umbrelOS's system Tor at ${TOR_PROXY_IP}:29051 (same endpoint
+    # lightningd connects to via --addr=statictor). The getumbrel/tor image has no nc
+    # and its /bin/sh lacks the /dev/tcp builtin, so probe with bash via CMD (not
+    # CMD-SHELL).
+    healthcheck:
+      test: ["CMD", "bash", "-c", "exec 3<>/dev/tcp/${TOR_PROXY_IP}/29051"]
+      interval: 5s
+      timeout: 5s
+      retries: 30
+      start_period: 10s
--- a/core-lightning/umbrel-app.yml
+++ b/core-lightning/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1.1
 id: core-lightning
 category: bitcoin
 name: Core Lightning
-version: "26.06.1"
+version: "26.06.1-hotfix.1"
 tagline: Run your personal Core Lightning node
 description: >-
   Get started with the Lightning network today with Core Lightning - a
@@ -32,11 +32,12 @@ defaultPassword: ""
 submitter: Blockstream
 submission: https://github.com/getumbrel/umbrel-apps/pull/475
 releaseNotes: >-
-  This updates Core Lightning to v26.06.1. CLN Application remains on v26.04.
+  This is a packaging fix and does not change the Core Lightning version (still v26.06.1).
 
 
-  **Core Lightning v26.06.1:**
-    - Fixes the `bwatch` plugin failing to register on startup after `make install`.
+  It stops lightningd from crashing on startup when it races ahead of Tor — a problem
+  that showed up most often after an umbrelOS update recreated all the containers at once.
+  The node now waits for the Tor control port to be reachable before lightningd starts.
 
 backupIgnore:
   - data/lightningd/bitcoin/lightningd.sqlite3


nmfretz:
Hey @mrv777, I took a look and I don’t think this is actually a fix for #5529. The reported `plugin-bcli` error is Core Lightning’s built-in Bitcoin backend plugin failing to reach Bitcoin Core RPC, not an Umbrel app plugin or Tor-specific failure.

That points to Bitcoin RPC readiness after restart/update (like a full umbrel restart or a Bitcoin Node restart), not the Tor control port. The report was also against Core Lightning `v25.09.3`, and we’ve since updated  from that, and now CLN waits on Bitcoin RPC with `bitcoin-cli -rpcwait -rpcwaittimeout=30`, so short readiness races should be handled better there.

I see that your testing shows the proposed healthcheck works on a healthy Umbrel, but it doesn’t reproduce the original failure or show that `connectd` was the terminal failure in #5529. I’m also not seeing other user reports of this Tor failure, and `lightningd` already runs with `restart: on-failure`, so a transient Tor readiness race should recover once Tor is available.

Have you since been able to reproduce exactly?

mrv777:
no, i have not been able to reproduce exactly.  can probably just close this for now and in case it can be reproduced later can take a look at this PR again

nmfretz:
sounds good @mrv777 🤝

PR #5829, opened 2026-07-01
Update Urbit to v4.5

yapishu:
- Updated Urbit to official v4.5 binary: https://github.com/urbit/vere/releases


nmfretz:
Thanks for this @yapishu, and sorry for the delay while I was away.

PR #5830, opened 2026-07-01
lightning-terminal: update to `v0.17.0-alpha`

ViktorT-11:
In this PR we bump Litd to `v0.17.0-alpha`.

--------

NOTE!! There's an important change with this release which potentially effects how you want host `lightning-terminal` in Umbrel:

With this release, we added support for SQL-based database backends, and by default, user which have run `litd` prior to this release and has an existing database will therefore be prompted to migrate their old `bbolt` databases to SQL.

For users which runs the `litd` as a standalone daemon, this prompt is shown to the user when the daemon in started, and the user thereafter needs to manually confirm the migration by passing `yes` via `stdin`. I would assume that this user flow is not supported in Umbrel though?

For environments where that user flow does not work, we have provided an alternative approach:
Instead of the approach where the migration prompt needs to manually be confirmed with `yes` through `stdin`, you can instead either set a config flag `auto-migrate-to-sql=true` for `litd`, or set an env variable `LIT_AUTO_MIGRATE_TO_SQL=true`. Once those are set, the prompt will not show on startup and the SQL migration will be started automatically. I'd assume that the env variable approach is likely the most appropriate for Umbrel users, but you'd know this best.

Alternatively, the migration can also be deferred by setting the following config option: `databasebackend=bbolt`. Note though that deferring the migration will only work temporarily, as we will make the migration mandatory in future `litd` releases.

-------

See these release notes for more information:
https://github.com/lightninglabs/lightning-terminal/releases/tag/v0.17.0-alpha

Additionally, you can check these release notes:
https://github.com/lightninglabs/lightning-terminal/blob/master/docs/release-notes/release-notes-0.17.0.md

### DIFF
--- a/lightning-terminal/docker-compose.yml
+++ b/lightning-terminal/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3004
 
   web:
-    image: lightninglabs/lightning-terminal:v0.16.1-alpha@sha256:7f8e16d940ee350ff38b2d8e71bda0c512a89822fa62a60097ac7ecc71d660ae
+    image: lightninglabs/lightning-terminal:v0.17.0-alpha@sha256:fd6d6040f78e0a342437312f3a3c154f4b63d79172c6b466e2dfd242b06b4a1c
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/lightning-terminal/docker-compose.yml
+++ b/lightning-terminal/docker-compose.yml
@@ -17,6 +17,8 @@ services:
     environment:
       HOME: "/data"
       APP_PASSWORD: "$APP_PASSWORD"
+      # Auto-approve LiT's one-time bbolt-to-SQL migration for non-interactive Umbrel updates.
+      LIT_AUTO_MIGRATE_TO_SQL: "true"
     command:
         - --uipassword_env=APP_PASSWORD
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for calling out the SQL migration behaviour @ViktorT-11!

> I would assume that this user flow is not supported in Umbrel though?

Yep that's right, but your env var method is perfect. I added

  ```yaml
  LIT_AUTO_MIGRATE_TO_SQL: "true"
  ```

I also added a short release note for users explaining that Lightning Terminal will automatically migrate its internal database on first startup after updating, and that the first startup may take a little longer while that runs (it's super quick in my testing, but I figure there's no harm in underselling.

I tested on my mainnet node and it worked well, including migration.

```text
2026-07-06 22:27:15.409 [INF] LITD: LiT version: 0.17.0-alpha commit=v0.17.0-alpha commit_hash=08934deeefe306d99e402c6c683bac3171fff88f
2026-07-06 22:27:15.423 [INF] LITD: Basic lnd client connected
2026-07-06 22:27:15.531 [INF] MIGS: Succesfully migrated all KVDB stores to SQL in: 13.02902ms
2026-07-06 22:27:15.547 [INF] SQL2: Database version after migration: 6
2026-07-06 22:27:18.330 [INF] LITD: Internal sub server startup complete
```

PR #5833, opened 2026-07-02
Add Blisspoint v0.5.4

heatpunk:
## App details

**Name:** Hashboard
**Version:** 0.5.1
**Category:** bitcoin
**Port:** 3847

## Description

Hashboard turns a Bitcoin ASIC miner into an ordinary household appliance: open it on your phone, see how much heat it is making, and slide the power up or down. Built for the people you share a home with — no mining knowledge required.

One slider sets the miner's power (and heat). The maximum is captured from the miner's own power target on first connect and locked, so nobody can push the machine harder than it was set up for. Pause and resume with one tap, watch live hashrate, wattage, chip temperature and fan speed, and find miners on your network automatically with the built-in LAN scan.

Miner communication is handled by the 256 Foundation's open asic-rs library, so monitoring works across Antminer (stock), Whatsminer, Avalon, BraiinsOS+, LuxOS, Vnish, ePIC, Marathon, Bitaxe and more, with power control wherever the firmware exposes it.

## Package details

- Image: `ghcr.io/heatpunk/hashboard:0.5.1` pinned to sha256 digest
- Architectures: linux/amd64 and linux/arm64
- No dependencies on other Umbrel apps
- Stateless app — no persistent data needed
- Port 3847 (checked against store for uniqueness)


hashboard/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: hashboard_web_1
+      APP_PORT: 80
+
+  web:
+    image: ghcr.io/heatpunk/hashboard:0.5.2@sha256:0976041ef019b099c34a9a1924ef9ec2bf5cf1e682addb35ecb85319789bab46
+    restart: on-failure
```
nmfretz: Can we add an init process here? The image launches Node and `proxy-rs` as separate processes, and without this the web container consumes the full one-minute grace period on every Umbrel restart. I tested this change locally and restart time dropped from 67 seconds to 2 seconds.

```suggestion
    init: true
    restart: on-failure
```

nmfretz (changes requested):
Thanks for the submission @heatpunk.

I tested the package on a fresh ARM64 umbrelOS 1.7.4 install. It installs and opens through Umbrel, but I found two blockers in the 0.5.2 app image:

- First-run discovery is hard-coded to `192.168.1.0/24`, and the empty state has no way to enter a miner IP or choose another subnet. Users on `192.168.0.x`, `10.x`, `172.16–31.x`, or any other subnet are left with no way to add their first miner.
- The power, pause, and resume routes accept a miner password but discard it before calling `asic-rs`; `miner.set_auth(...)` is never called. This means a custom-password miner can never authenticate, even after the UI asks for its password.

Can you fix these upstream, verify a fresh setup outside `192.168.1.0/24` and control against a custom-password miner, then cut a new release/image and update the version and digest here? I can re-test once that is updated.

I also noticed [heatpunk/hashboard#71](https://github.com/heatpunk/hashboard/issues/71) about a possible rename to Blisspoint. Is that planned before this submission lands? If Blisspoint is the decided launch name, now is the right time to update the submission and store assets; once users install `hashboard`, the package `id` needs to remain stable.

I left one small lifecycle fix inline as well.

heatpunk:
Hi @nmfretz,

Thank you for the detailed review! We have addressed all feedback in this update for release **v0.5.3**:

1. **Subnet Discovery & IP Entry:** Users can now scan subnets beyond `192.168.1.0/24` and manually enter miner IP addresses on setup.
2. **Custom Password Auth:** Fixed `miner.set_auth(...)` handling across power, pause, and resume routes in `proxy-rs` so custom-password miners authenticate correctly.
3. **Container Lifecycle (`init: true`):** Added `init: true` to `docker-compose.yml` for clean signal handling during restarts.
4. **App Rename & Version Bump:** Renamed package/folder to `blisspoint`, updated `umbrel-app.yml` manifest to v0.5.3, and updated the Docker image target to `ghcr.io/heatpunk/blisspoint:0.5.3`.

Ready for re-testing!

PR #5838, opened 2026-07-04
GUPT updated

besoeasy:
Updated to use proper versions

PR #5839, opened 2026-07-04
UPDATE GUPT

besoeasy:
UPDATE GUPT

### DIFF
--- a/gupt/docker-compose.yml
+++ b/gupt/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8000
 
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @besoeasy! I cherry-picked your Docker image update commit from #5838 (`ce8292a2`) into this PR, since it looked like that was accidentally opened separately.

I also did some small App Store metadata polish by shortening the tagline so it renders better in the App Store card layout (we try to keep these pretty short), fixing a small typo/casing issue in the feature list, and adding release notes.

Going live now.

PR #5841, opened 2026-07-04
Remove Oak Node from app store

ok300:
The Oak Node project is unmaintained and is now non-functional: https://oak-node.net

This PR:

- adds `disabled: true` to the app manifest which results in Oak Node not being shown in the app store on umbrelOS >= 1.0
- adds a removal notice to the app description so that umbrelOS 0.5 users are aware that it is no longer functional

For the exact PR diff, I took inspiration from similar "app removal" PRs like #1741. Let me know if this is ok or if I missed anything.

### DIFF
--- a/oak-node/umbrel-app.yml
+++ b/oak-node/umbrel-app.yml
@@ -1,4 +1,4 @@
-manifestVersion: 1
+manifestVersion: 1.2
 id: oak-node
 category: bitcoin
 name: Oak Node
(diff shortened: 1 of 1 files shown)


nmfretz:
Hey @ok300 thanks for the PR. Sad to see Oak Node go, it was one of the early apps on Umbrel 🤝.

I've made one small change to make sure it won't install on the pre-1.0 umbrelOS app store as well.

PR #5842, opened 2026-07-04
App Submission: Foldergram

dennysubke:
# App Submission

### Foldergram
...

### Icon
logo

...

### I have tested my app on:
- [X] umbrelOS on Linux VM


nmfretz:
Thanks for another excellent app submission, @dennysubke! I made a few small packaging tweaks to get this finalized:

- Updated the gallery image reference to WebP
- Kept Umbrel authentication enabled by default, as with LeafWiki
- Moved the photo library bind mount into Foldergram’s app data directory

App data is accessible through Files in umbrelOS, so users can navigate to Foldergram’s gallery directory and add their photo folders there, much like using a `Photos` directory in Home. Keeping the default mount inside the app data directory also avoids relying on a shared directory that may not exist yet and could otherwise be created by Docker with root ownership.

The next umbrelOS release will allow users to customize app bind mounts, so they’ll also be able to point Foldergram at any existing photo directory they prefer.

I tested the install and photo library workflow on Umbrel, including restart and persistence, and everything is working well.

image

PR #5843, opened 2026-07-05
App Submission: LeafWiki

dennysubke:
# App Submission

### LeafWiki
...

### Icon
logo

...

### I have tested my app on:
- [X] umbrelOS on Linux VM


nmfretz:
Thanks for another excellent app submission, @dennysubke! I made a few small packaging tweaks to get this finalized:

- Updated LeafWiki to v0.11.4
- Switched the gallery references to WebP
- Removed the `PROXY_AUTH_ADD` override so Umbrel authentication stays enabled by default
- Replaced the shared default password with a unique per-install password

For apps in the official App Store, we want Umbrel authentication enabled by default wherever possible, unless doing so reduces the app’s functionality. The next umbrelOS release will add a per-app toggle, so users who prefer to rely only on LeafWiki’s built-in authentication can disable Umbrel auth themselves.

image

PR #5844, opened 2026-07-05
App Submission: Transmute

dennysubke:
# App Submission

### Transmute
...

### Icon
logo

...

### I have tested my app on:
- [X] umbrelOS on Linux VM


nmfretz:
Thanks for this as well @dennysubke 🤝. I pushed a few Umbrel-specific packaging tweaks to get this finalized:

- Allowed `/api/*` through `app_proxy` so scripts and automated workflows can use Transmute's documented API with API keys, without requiring an Umbrel browser session. `/api/users` remains behind Umbrel auth because it creates the first administrator during setup; other protected API routes still require Transmute authentication.
- Set `AUTH_SECRET_KEY` from Umbrel's persistent `${APP_SEED}` so JWT signing remains stable across restarts. Without this, existing login tokens become invalid whenever the container restarts.
- Updated the gallery filenames from `.jpg` to `.webp`.

I tested the install, account setup, file conversion, external API access, and restart persistence on umbrelOS. Everything is working well, so we're good to go. Going live!

image

PR #5847, opened 2026-07-06
Add LibreDB Studio

yusuf-gundogdu:
## Type

New app

## App

App ID: libredb-studio
Upstream project: https://github.com/libredb/libredb-studio
Version: 0.14.1

## Summary

MIT-licensed self-hosted SQL and NoSQL IDE. Connects to 16 engines, including PostgreSQL, MySQL, Oracle, SQL Server, SQLite, libSQL, DuckDB, MongoDB, Redis, Couchbase, ClickHouse, Druid, Elasticsearch, OpenSearch, Trino and Cassandra, from the browser. It stores its own data in an embedded SQLite file, so there is no external database to run just to use it.

## Icon

256x256: https://raw.githubusercontent.com/libredb/unraid-templates/main/icons/libredb-studio.png

## Gallery images

All 2160x1350, captured from a real container running the image pinned in this PR, using the sample database that ships inside it.

Query editor with results, connection list and schema tree:
Query editor

Charts built from the same result set:
Charts

Table columns and types next to query output:
Schema

Admin dashboard with connection health and metrics:
Admin dashboard

## Verification

Environment tested:
- [ ] Umbrel device
- [x] Container tested directly with the exact settings in this PR

Architecture tested:
- [x] amd64
- [x] arm64

What was actually checked, using the image, environment variables, volume, uid and port from `docker-compose.yml` in this PR:

- Starts and serves on port 3000.
- `GET /api/storage/config` returns `{"provider":"sqlite","serverMode":true}`, so data is kept server side in the mounted volume rather than in the browser.
- Login works with `ADMIN_EMAIL` and the generated password.
- Data written through the app is still there after a container restart.
- The pinned digest resolves to tag 0.14.1 and the index carries linux/amd64 and linux/arm64.

The July login problem was retested specifically. The app was put behind a proxy that mimics the app_proxy setup, plain HTTP, Host header passed through unchanged and `x-forwarded-host` set to the browser host. Login succeeded, the `auth-token` cookie came back without the `Secure` flag, and `GET /api/auth/me` returned an authenticated session on the next request. That is the loop back to `/login` that you hit, and it does not happen with `AUTH_COOKIE_SECURE=false` set in the compose file.

I have not installed this on umbrelOS itself, so a real install test on your side is still worth doing.

Known lint warnings or caveats: none. `lint-apps.mjs` was run locally against this branch with `--check-images` and reported no issues.

## Notes

Storage is SQLite by default, so no external database is required. The admin password uses `$APP_PASSWORD` and the JWT secret comes from `derive_entropy` in `exports.sh`. `AUTH_COOKIE_SECURE` is set to `false` because umbrelOS serves apps over plain HTTP; the app still defaults to `Secure` everywhere else. AI features are optional and stay off unless a provider key is configured.

Links: GitHub https://github.com/libredb/libredb-studio, Website https://libredb.org/


nmfretz (commented):
Thanks for the submission @yusuf-gundogdu.

I gave this a fresh install test on arm64 with umbrelOS 1.7.4. Right now the app can't stay logged in because LibreDB marks its `auth-token` cookie as `Secure`, while umbrelOS currently opens the local app URL over HTTP. The login request succeeds, but the browser can't send the cookie back and is redirected straight back to `/login`.

We're shipping HTTPS-by-default app access in an upcoming umbrelOS release, so one option is to leave this PR open and I'll re-test it once that lands.

Alternatively, if you'd prefer to support HTTP deployments upstream and move this forward sooner, you could keep the secure behavior as the default but add a runtime override such as `AUTH_COOKIE_SECURE=false`, publish a new image, and wire that option into this package.

Do you have a preference between waiting for the HTTPS rollout or adding the upstream override?

Once I manually supplied the session cookie, SQLite queries and restart persistence worked as expected. I've left two smaller package-level comments inline.

libredb-studio/umbrel-app.yml
```diff
@@ -0,0 +1,31 @@
...
+website: https://libredb.org
+dependencies: []
+repo: https://github.com/libredb/libredb-studio
+support: https://github.com/libredb/libredb-studio/issues
+port: 45832
+gallery: []
+path: ""
+defaultUsername: ""
+defaultPassword: ""
+deterministicPassword: false
```
nmfretz: Can we surface the generated admin login here? umbrelOS currently reports both credential fields as empty, so even after the cookie issue is fixed users won't know how to log in. Please set `defaultUsername: admin@libredb.org`, enable `deterministicPassword`, and drop the empty `defaultPassword` field.

libredb-studio/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+    restart: on-failure
+    user: "1000:1000"
+    environment:
+      PORT: "3000"
+      STORAGE_PROVIDER: sqlite
+      STORAGE_SQLITE_PATH: /app/data/libredb-storage.db
+      ADMIN_EMAIL: admin@libredb.org
+      ADMIN_PASSWORD: $APP_PASSWORD
+      USER_EMAIL: user@libredb.org
+      USER_PASSWORD: $APP_LIBREDB_STUDIO_USER_PASSWORD
```
nmfretz: Can we drop the optional `USER_EMAIL` / `USER_PASSWORD` account and the matching value from `exports.sh`? Umbrel only has a place to surface one credential set, so this leaves a second active login that the user can't retrieve through the UI.

yusuf-gundogdu:
Yeah, that's the Secure cookie flag. umbrelOS serves over plain HTTP, so the browser drops it and you loop back to /login.

I added an opt-in `AUTH_COOKIE_SECURE` flag in the app to turn that off for HTTP setups like this: libredb/libredb-studio#236 (CI's green). Default stays Secure, so nothing changes for normal HTTPS installs. Once it merges and a new image is out I'll bump the tag here and set `AUTH_COOKIE_SECURE=false` in the compose.

For the creds, I'll set `defaultUsername: admin@libredb.org`, `deterministicPassword: true` and drop the empty `defaultPassword`. The app already takes Umbrel's generated app password (`$APP_PASSWORD`) as the admin password, so no code change there, it'll just show up in the UI.

And I'll drop `USER_EMAIL`/`USER_PASSWORD` and ship only admin. It runs single-user fine without them.

I'll do the yaml/compose edits in one go once the image is published.

cevheri:
Hi is there any update

yusuf-gundogdu:
The cookie fix from July is in. `AUTH_COOKIE_SECURE` is set to `false` in the compose file, so the `auth-token` cookie no longer carries `Secure` and the session holds over plain HTTP. I retested that behind a proxy set up like app_proxy, with the Host header passed through and `x-forwarded-host` set to the browser host, and the session stayed alive on the next request.

Lint is green now, that was a `releaseNotes` value I should have left empty. I also bumped the app to 0.14.1 and added the icon and gallery images to the description. Ready for that re-test whenever you have time.

yusuf-gundogdu:
Refreshed the pinned image and digest to 0.16.0, which is the current release.

yusuf-gundogdu:
Refreshed the pinned image and digest to 0.16.1, which is the current release.

0.16.1 fixes an inline cell edit being written to the wrong row after a sort or a refresh, and it makes a container exit at boot when it cannot serve a login, instead of passing its liveness probe and answering every login with 503.

yusuf-gundogdu:
Refreshed the pinned image and digest to 0.16.2.

0.16.2 closes a cross-account authorization defect (GHSA-3wh2-8x78-jfw4, scored 8.8): a connection id supplied in a request could select another session's already-authenticated database provider, so a `user` account could reach a connection restricted to `admin`. It matters on any instance more than one person signs in to, which on umbrelOS is the normal case.

Repo linter run locally against upstream master: "Linting libredb-studio / No issues found", with --check-images so the digest is verified.

PR #5852, opened 2026-07-07
Add Lively

ryanwaits:
## Type

New app

## App

- **App ID:** `lively`
- **Upstream project:** https://github.com/ryanwaits/lively
- **Version:** `0.1.1`

## Summary

I built Lively because I wanted a shared whiteboard and a place to keep notes and
todos that lived on my own hardware, instead of reaching for another SaaS account
every time. It started as a small open-source realtime-collaboration framework — a
self-hosted alternative to Liveblocks — and Umbrel felt like the natural home for
it.

This app bundles four tools built on that framework into a single container: a
shared whiteboard, a Notion-style block editor, a live markdown editor, and a
shared todo list. Live cursors, presence, undo/redo, and conflict-free (CRDT)
editing. No accounts, no cloud, no third-party API calls — everything runs on the
Umbrel and all data stays under `${APP_DATA_DIR}/data`.

**How it's built:** one container, one service. A single Bun server handles the
WebSocket rooms (`/rooms/*`), a small JSON API (`/api/boards`, `/api/config`) and
`/health`, and serves the four apps as static Next.js exports plus a landing page.
Persistence is plain files on the `/data` volume (JSON snapshots + binary Yjs) —
no database, no sidecar. The web UI is only ever exposed through `app_proxy`.

**Access:** works on your LAN or over Tailscale; v1 rides Umbrel's own auth, so
collaboration is among people who share access to the Umbrel. The full
access-model and roadmap (including scoped room tokens to invite guests without
sharing your login) is in the project README.

- **Image:** `ghcr.io/ryanwaits/lively:0.1.0@sha256:72bfc538a9a13834da6cc33444661f98ecbee1bde756b6d5f68ab91646c139f4`
  — multi-arch (linux/amd64 + linux/arm64), public on GHCR, built in CI from the
  tagged release.

## Verification

**Umbrel testing performed:** Installed on a real umbrelOS 1.7.3 instance from a
local app store. Verified: the app opens through `app_proxy` behind Umbrel auth,
all four apps load, real-time WebSocket collaboration works across two browsers,
data persists to the `/data` volume and **survives a reboot** (room files came
back byte-identical), and **uninstall is clean** (no orphaned containers or data).
Separately load-tested the WebSocket path against the real `getumbrel/app-proxy`
image: a fully idle socket held open for 150s (3× the client heartbeat) without
being dropped.

**Environment tested:**
- [ ] Umbrel device
- [ ] Not runtime tested

**Architecture tested:**
- [x] amd64
- [ ] arm64

_arm64 was not tested on a device; the arm64 image is built by the same CI
pipeline from the same commit and was smoke-tested natively (server boot + full
collaboration + reboot persistence) on Apple Silicon, but I don't have an arm64
Umbrel to confirm on-device._

**Known lint warnings or caveats:**
- `npm run lint:apps -- lively --check-images` passes clean.
- The bundled Bun runtime needs AVX2. On real amd64/arm64 hardware it runs
  natively; only under pure CPU *emulation* did the test VM need `-cpu max` to
  expose it. Bun ships a baseline (non-AVX2) build if an ancient CPU ever needs
  it — happy to switch the base image if you'd prefer maximum compatibility.
- The multi-arch manifest includes a buildx provenance-attestation entry (shows
  as `unknown/unknown`). I can re-release with `provenance: false` if you'd rather
  it not be there.

## Notes

- **Permissions / dependencies:** none. `dependencies: []`, no privileged access,
  no Docker socket, no host network. Runs as `user: "1000:1000"`.
- **Default credentials:** none — there are no app-level accounts; access is
  gated by Umbrel's own auth.
- **Data:** single bind-mount at `${APP_DATA_DIR}/data`; `stop_grace_period: 30s`
  lets the server flush pending writes on shutdown.
- **Privacy:** no runtime CDN calls — fonts and JS are self-hosted/vendored
  (verified by grepping the built static output). No telemetry, no external API
  calls.
- **Assets:** `gallery: []` in the manifest — happy for the Umbrel team to
  produce the final store assets. A 256×256 SVG icon and gallery screenshots of
  all four apps are attached to this PR.

## Screenshots & icon

**Icon** (256×256 SVG): [`icon.svg`](https://raw.githubusercontent.com/ryanwaits/lively/main/apps/umbrel/store/icon.svg) — happy for the Umbrel team to produce the final store assets.

**Whiteboard** — infinite canvas, sticky notes, live cursors & presence
Whiteboard

**Notes** — Notion-style block editor
Notes

**Markdown** — collaborative markdown editor
Markdown

**Todo** — shared realtime todo list
Todo

### DIFF
--- a/lively/docker-compose.yml
+++ b/lively/docker-compose.yml
@@ -1,0 +1,19 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: lively_web_1
+      APP_PORT: 8080
+      # v1.1 (external WS clients): PROXY_AUTH_WHITELIST: "/rooms/*"
+
+  web:
+    image: ghcr.io/ryanwaits/lively:0.1.0@sha256:72bfc538a9a13834da6cc33444661f98ecbee1bde756b6d5f68ab91646c139f4
+    restart: on-failure
+    stop_grace_period: 30s
+    user: "1000:1000"
+    environment:
+      PORT: 8080
+      DATA_DIR: /data
+    volumes:
+      - ${APP_DATA_DIR}/data:/data
--- a/lively/docker-compose.yml
+++ b/lively/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       # v1.1 (external WS clients): PROXY_AUTH_WHITELIST: "/rooms/*"
 
   web:
-    image: ghcr.io/ryanwaits/lively:0.1.1@sha256:4480667080b5b6da22e57a94a17efacd2a6fc17750dc94a6b2b9b08e00569aa4
+    image: ghcr.io/ryanwaits/lively:0.2.0@sha256:bbd7a5f3c664c7507ac794f1a4f2fdd5b739775ed81982a3abd8d4f3196c487b
     restart: on-failure
     stop_grace_period: 30s
     user: "1000:1000"
--- a/lively/docker-compose.yml
+++ b/lively/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       # v1.1 (external WS clients): PROXY_AUTH_WHITELIST: "/rooms/*"
 
   web:
-    image: ghcr.io/ryanwaits/lively:0.1.0@sha256:72bfc538a9a13834da6cc33444661f98ecbee1bde756b6d5f68ab91646c139f4
+    image: ghcr.io/ryanwaits/lively:0.1.1@sha256:4480667080b5b6da22e57a94a17efacd2a6fc17750dc94a6b2b9b08e00569aa4
     restart: on-failure
     stop_grace_period: 30s
     user: "1000:1000"
--- a/lively/umbrel-app.yml
+++ b/lively/umbrel-app.yml
@@ -1,0 +1,26 @@
+manifestVersion: 1
+id: lively
+category: files
+name: Lively
+version: "0.1.0"
+tagline: Self-hosted realtime collaboration — whiteboard, Notion-style docs, markdown, and todos
+description: >-
+  Lively is an open-source realtime collaboration framework — a self-hosted
+  alternative to Liveblocks. This app bundles four collaborative tools that run
+  entirely on your Umbrel: a shared whiteboard, a Notion-style block editor, a
+  live markdown editor, and a shared todo list. Live cursors, presence,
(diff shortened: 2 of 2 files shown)


nmfretz (changes requested):
Hey @ryanwaits, thanks for the submission and for the detailed testing notes here!

I gave this a full test on umbrelOS 1.7.4 on arm64. The package is in good shape overall, but there are two things I want to check before we move forward.

As you called out in the PR, the current amd64 image requires AVX2. umbrelOS can be installed on a wide range of users' own amd64 hardware, and AVX2 isn't part of our minimum CPU requirements.

Would you be okay shipping an amd64 image that also works without AVX2? Bun's baseline build may be one option, but I'll leave the implementation up to you. Once that's available, the package version and pinned image digest can be updated here.

I also ran into something in the Todo app that I wanted to check with you. With two browser sessions connected, creating a task synced correctly between them. When I marked the task as complete in one session, the other session didn't update, and refreshing reset the task to incomplete.

Can you confirm whether you're able to reproduce that, or whether I may be missing something about the expected flow? Since collaboration and persistence are central to the app, I wanted to flag it before the new image is published.

Once we've clarified that and tightened up the amd64 compatibility, I can re-test.

nmfretz:
oh and thanks for the screenshots. Are you okay with these gallery assets:

https://github.com/getumbrel/umbrel-apps-gallery/pull/133/changes

ryanwaits:
hey @nmfretz ! appreciate the look and feedback, was able to reproduce that todo sync bug so good catch

before

https://github.com/user-attachments/assets/ed9dc25b-7752-49b8-8d7e-5a8ecaf69da6

after

https://github.com/user-attachments/assets/b226d75c-8c83-4751-8245-2d311f2d9afe

re: **amd64**, i went with your suggestion: the amd64 image now ships Bun's baseline build, so AVX2 is no longer required (and ci now boots every release on a non-AVX2 CPU before publishing)

the version and pinned digest are updated on this PR now (and the gallery assets look great, thanks for that!)

ryanwaits:
hey @nmfretz circling back here: both items from your review are live: todo sync bug's fixed (repro'd + verified, screenshots above) and amd64 now ships buns baseline build so AVX2 isn't required. also bumped to 0.2.0 since then with an updated pinned digest

let me know if you get a chance to re-test, happy to sort out anything else that comes up!

PR #5857, opened 2026-07-09
Update Hashrate Autopilot to 1.17.0

rdouma:
## Type

App update

## App

App ID: hashrate-autopilot
Upstream project: https://github.com/rdouma/hashrate-autopilot
Version: 1.17.0

## Summary

Security release. Credentials the app stores in its SQLite database (Braiins API tokens, bitcoind RPC password, Telegram token, DDNS credential) are now AES-256-GCM encrypted at rest instead of plaintext, and the dashboard password is stored as a one-way scrypt hash. The config-change audit log no longer records credential values (and scrubs any previously logged on upgrade). Users can now change the dashboard password and rotate their Braiins tokens from the dashboard itself, and Profit & Loss counts Lightning payouts via Ocean's earnpay ledger. Existing installs upgrade their stored data in place on first start.

This bump also adds `BHA_SECRET_KEY: "${APP_SEED}"` to the compose so the at-rest encryption key is the device-derived per-app secret, which lives outside the app's data directory. Without it the app falls back to a keyfile next to the database (still encrypted, but inside app-data); using `APP_SEED` keeps an app-data backup from ever carrying the key.

Full release notes: https://github.com/rdouma/hashrate-autopilot/releases/tag/v1.17.0

## Verification

Umbrel testing performed: installed 1.17.0 as an in-place update over the prior version on a real Umbrel device and verified the new in-app password change end to end. The image is pinned by digest.

Environment tested:
- [x] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [ ] amd64
- [ ] arm64

The runtime test above was on the maintainer's own Umbrel device (its native architecture). The published multi-arch image is built and CI-verified for both amd64 and arm64; the digest pinned here is the multi-arch index.

## Notes

- New env var: `BHA_SECRET_KEY: "${APP_SEED}"` (rationale in Summary).
- No new dependencies, no default credentials, no port change (still 3018 / internal 3010).
- Migration: on first start the app transparently encrypts its existing plaintext secrets and hashes the stored password. If the key is ever unavailable it treats the affected secret as unset and prompts re-entry rather than failing to boot.

### DIFF
--- a/hashrate-autopilot/docker-compose.yml
+++ b/hashrate-autopilot/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3010
 
   web:
-    image: ghcr.io/rdouma/hashrate-autopilot:1.16.0@sha256:d44526468eca52d24ac592869ea552c07f3dbdb7acdc3e44df2ad31518a69a46
+    image: ghcr.io/rdouma/hashrate-autopilot:1.17.0@sha256:8f952ab293975694b2c8169b7c5db18d0bc32636336d884712f3d89753bc665a
     restart: on-failure
     stop_grace_period: 1m
     volumes:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @rdouma and sorry for the delay!

PR #5861, opened 2026-07-10
gobrrr-pool: update to 1.2.1

gobrrrme:
## Type
App update

## App
App ID: gobrrr-pool
Upstream project: https://github.com/gobrrrme/GoBrrr-Pool
Version: 1.2.1

## Summary
- ckpool updated to latest upstream (yyjson stratum engine, memory-leak and deadlock fixes)
- exports.sh: follow the app-selected Bitcoin provider (Knots support), courtesy of @BuffaloDyl (gobrrrme/GoBrrr-Pool#5)
- webui shows Bitcoin Core vs Knots, leaderboard color tweak (@rafxo)
- images pinned to v1.2.1 digests

## Verification
Umbrel testing performed: fresh install via community store on umbrelOS, verified dashboard, homescreen widget, node-implementation detection, and both stratum ports (21420 / high-diff 21422) with live miners against Bitcoin Core.

Environment tested:
- [x] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [x] amd64
- [ ] arm64

Known lint warnings or caveats: `submission` field still points to the original submission PR (#5449), happy to update it to this PR's URL if required.

## Notes
No new permissions, dependencies unchanged (bitcoin, mempool), no data migrations, existing installs upgrade in place. Multi-arch images (amd64 + arm64) are built in CI.

@nmfretz please push the required updates to make mempool optional, as discussed [here](https://github.com/getumbrel/umbrel-apps/pull/5449#issuecomment-4656874527).

Thanks again for your help!

### DIFF
--- a/gobrrr-pool/docker-compose.yml
+++ b/gobrrr-pool/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api/*"
 
   ckpool:
-    image: ghcr.io/gobrrrme/gobrrr-ckpool:v1.1.0@sha256:9e831a5100039d9c8246dafabacf9101f60a4a7828bdc94ce0a14a389cc55188
+    image: ghcr.io/gobrrrme/gobrrr-ckpool:v1.2.1@sha256:7f255e43f4b8f27c1ad0aee1247e9534828f372929a2af9f8735418c5e44e86c
     restart: on-failure
     stop_grace_period: 1m
     healthcheck:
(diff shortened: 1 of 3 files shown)


nmfretz:
Thanks @gobrrrme! I pushed the Mempool changes we discussed, and am doing some proper testing now.

Mempool is no longer a required dependency. Users now only select the Bitcoin provider for Go Brrr Pool. The app uses the local Mempool API when Mempool is installed and otherwise falls back to the public mempool.space API.

Removing Mempool from as a dependency also means that you don't need the complex exports.sh resourcing that was added for this PR. Instead, we get the super simple Mempool API selector logic I added.

I also updated the description and release notes to explain the new behavior.

nmfretz:
Alrighty, I tested the update from `1.1.0` with Mempool installed, as well as a fresh install without Mempool. Dashboard, widget, and both stratum ports are all good.

I also verified the public `mempool.space` fallback used when the Mempool app isn’t installed.

Going live! 🚀

gobrrrme:
Thank you so much!

PR #5862, opened 2026-07-10
App Submission: Appwrite

dennysubke:
# App Submission

### Appwrite
...

### Icon
logo

...

### I have tested my app on:
- [X] umbrelOS on Linux VM


appwrite/docker-compose.yml
```diff
@@ -0,0 +1,641 @@
...
+  _APP_SYSTEM_EMAIL_NAME: "Appwrite"
+  _APP_SYSTEM_EMAIL_ADDRESS: "team@appwrite.io"
+  _APP_SYSTEM_SECURITY_EMAIL_ADDRESS: "certs@appwrite.io"
+  _APP_EMAIL_SECURITY: ""
+  _APP_SYSTEM_RESPONSE_FORMAT: ""
+  _APP_OPTIONS_ABUSE: "enabled"
+  _APP_OPTIONS_ROUTER_PROTECTION: "disabled"
+  _APP_OPTIONS_FORCE_HTTPS: "disabled"
+  _APP_OPTIONS_ROUTER_FORCE_HTTPS: "disabled"
+  _APP_OPENSSL_KEY_V1: "b9f7a6d84e3c4a12a6c2f58e91d73b0f4a8e39d76c1b42e7a5f2c8d90e6b3a41"
```
nmfretz, line 29: Can we generate the Appwrite encryption key, Executor secret, database passwords, and Redis password per install using separate `derive_entropy` values?

These values are currently fixed and public, so other app containers on Umbrel’s shared network could authenticate to Appwrite’s internal services.

appwrite/docker-compose.yml
```diff
@@ -0,0 +1,641 @@
...
+    <<: *logging
+    restart: on-failure
+    depends_on:
+      mongodb:
+        condition: service_healthy
+      redis:
+        condition: service_healthy
+    environment: *appwrite-env
+
+  assistant:
```
nmfretz, line 531: The upstream deployment includes PostgreSQL and `appwrite-embedding` for Vector Databases and embeddings. Both are missing here, so those features will not work.

Do you think we need them? If so, can you look into carrying then over here?

nmfretz (changes requested):
Thanks for the submission @dennysubke.

The main blocker is the Functions and Sites runtime setup. `openruntimes-executor` is started by Umbrel’s Docker daemon, but it connects to the nested Docker daemon. Executor expects that daemon to also own its container; here it does not, so Executor exits during startup.

Can you take another look at this setup and then test an actual Function build/execution and Site deployment through Umbrel? The console loading on its own will not catch this.

I’ve left a couple of other issues inline.

PR #5863, opened 2026-07-10
Update Monero daemon to v0.18.5.1

colder-is-better:
Update to the v0.18.5.1 release of the Monero software. This release includes a large number of bug fixes.

### DIFF
--- a/monero/docker-compose.yml
+++ b/monero/docker-compose.yml
@@ -44,7 +44,7 @@ services:
     restart: unless-stopped
     stop_grace_period: 15m
     command: "${APP_MONERO_COMMAND}"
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @colder-is-better! I made a small adjustment to keep the release notes focused on what Umbrel users receive in this update. The upstream changelog also includes wallet, Wallet RPC, and Windows-specific changes, while this package only updates `monerod`, so I condensed the notes to the relevant daemon improvements and kept the link to the full upstream release. Really appreciate you staying on top of these updates as usual 🤝

colder-is-better:
> Thanks @colder-is-better! I made a small adjustment to keep the release notes focused on what Umbrel users receive in this update.

Ah, makes sense. I will keep this in mind for future updates.

PR #5868, opened 2026-07-12
pogolo: update to 1.1.4-hotfix

0xf0xx0:
## Type

App update

## App

App ID: pogolo
Upstream project: github.com/0xf0xx0/pogolo
Version: 1.1.4-hotfix

## Summary

Fixes an issue during share validation

## Verification

Umbrel testing performed:

Environment tested:
- [ ] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [x] amd64
- [ ] arm64

## Notes

### DIFF
--- a/pogolo/docker-compose.yml
+++ b/pogolo/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       - ${APP_DATA_DIR}/data:/data
 
   pogolo:
-    image: 0xf0xx0/pogolo:v1.1.4-hotfix@sha256:8ff8f4e2d269fcd43b02011792063a5ef10815cdb46202280ab2a2c8a3997bb2
+    image: 0xf0xx0/pogolo:v1.1.4-hotfix@sha256:4a2b1d4bf1a7d9c6f8f2adabd0ce2129e6be504dbd62e8df5053b8ad01973b68
     restart: on-failure
     user: "1000:1000"
     stop_grace_period: 30s
--- a/pogolo/docker-compose.yml
+++ b/pogolo/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       - ${APP_DATA_DIR}/data:/data
 
   pogolo:
-    image: 0xf0xx0/pogolo:v1.1.4-hotfix@sha256:4a2b1d4bf1a7d9c6f8f2adabd0ce2129e6be504dbd62e8df5053b8ad01973b68
+    image: 0xf0xx0/pogolo:v1.1.4-hotfix@sha256:5b828e603d47d9f4de17bc0d36424b4f9aa45c193496c196d1b989620abf1411
     restart: on-failure
     user: "1000:1000"
     stop_grace_period: 30s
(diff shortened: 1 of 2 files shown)


nmfretz (commented):
Thanks @0xf0xx0. While reviewing and testing this update with Codex, I found what looks like a subtle edge case in the share-version validation this hotfix is addressing.

[BIP 310](https://github.com/bitcoin/bips/blob/master/bip-0310.mediawiki#changes-in-request-miningsubmit) defines the submitted `nVersion` as:

```text
(job_version & ~last_mask) | (version_bits & last_mask)
```

Pogolo currently uses `job_version + version_bits`, then subtracts `job_version` again when validating the version mask. These approaches produce different headers when a permitted rolling bit is already set in `job_version`.

I reproduced this against the exact hotfix source with `job_version = 0x20000020` and `version_bits = 00000000`. BIP 310 produces `0x20000000`, but Pogolo validates `0x20000020` and rejects the otherwise-valid share as too low difficulty.

In the worst case, if the miner’s header met the Bitcoin network target, Pogolo could validate a different header and never submit the original block, and lose the block subsidy and transaction fees.

Can you double-check this, fix it if necessary, and test the case where a permitted rolling bit is already set in `job_version`?

Ping me after and I'll test again!

0xf0xx0:
@nmfretz i may have fixed the issue, test now?

0xf0xx0:
testing an additional fix...

0xf0xx0:
testable

nmfretz:
Thanks for the quick turnaround @0xf0xx0. I pulled the latest image and gave that exact case another go. Looks like there’s still one small edge here.

The control with `job_version = 0x20000000` and `version_bits = 00000000` passes, but with `job_version = 0x20000020`, the share still comes back with `414: Difficulty too low`.

I think it’s the `if share.VersionMask > 0` check. `00000000` is still a valid sixth parameter under BIP 310, and in this case it needs to clear the permitted bit already set in `job_version`. I confirmed locally that checking whether the sixth parameter was actually present, instead of whether its value is greater than zero, makes the reproduction pass.

I know this is upstream app code and I’m really just reviewing the App Store packaging here, so I don’t want to get too deep into prescribing the fix. But since this update is specifically shipping the share-validation hotfix, I want to make sure this case is covered before I can finish the review.

Can you have another look and add a `mining.submit` test for `job_version = 0x20000020` with `version_bits = 00000000`? Ping me after and I’ll test again!

0xf0xx0:
hmm yeah, can you give me the expected share difficulty and the job (unless youre using the mock job template in the repo)?

nmfretz:
Yep, I’m using the existing mock job template and known-valid share. Expected share difficulty is `0.55803` against the `0.16` target; the only change is `job_version = 0x20000020`, with `version_bits = 00000000` explicitly included as the sixth parameter (`ToRequest()` omits it when zero).

0xf0xx0:
ok i wrote a test case, it *should* be fully patched now uwu

nmfretz:
Nice, thanks @0xf0xx0! I retested the exact case and it passes now. I also ran Umbrel update and restart checks. The packaging looks good from the App Store side.

Given how high-stakes solo mining can be, I’ll leave the broader end-to-end mining and payout validation with you as the app dev. Definitely worth being extra careful there so anyone lucky enough to find a block gets paid as expected 🙏.

Going live

PR #5873, opened 2026-07-14
Update Hashrate Autopilot to 1.17.1

rdouma:
## Type

App update

## App

App ID: hashrate-autopilot
Upstream project: https://github.com/rdouma/hashrate-autopilot
Version: 1.17.1

## Summary

Feature + polish release. New: a current-block-height stats tile that names the pool/miner that found each block and crowns your own; the Timeline's bid-id filter is now a full-text search across bid id, reason, personal notes, and per-row identifiers (block heights, IPs, deposit tx ids, config values); personal notes can be attached to any Timeline event (saved, searchable, and included in the Excel export); and Profit & Loss gains a hard-reset control next to rebuild that wipes and rebuilds the Ocean payout store and the Braiins spend cache. Improved: the P&L "collected" figure self-heals from a partial payout backfill (re-fetched on every restart and periodically), and the alert-condition detail panel breaks an outage into threshold / fired / recovered / duration / total, with durations rounded to the nearest minute. Fix: the Dynamic DNS test push uses the daemon's own detected public IP. Also includes dependency bumps (@fastify/static v10, @fastify/cors).

This is an image-digest-and-release-notes bump only - no structural compose changes, no new env vars, no new dependencies, no port change, relative to the previously accepted 1.17.0.

Full release notes: https://github.com/rdouma/hashrate-autopilot/releases/tag/v1.17.1

## Verification

The published image is multi-arch (linux/amd64 + linux/arm64), CI-built from the tagged commit, and pinned here by its multi-arch index digest (`sha256:01b2541d5ecdf688dd459ad51b0cb8ec55a03f5beb4fba8abbada7737bd64a1b`). The compose is unchanged from the already-Umbrel-tested 1.17.0 apart from the image reference, so the install/upgrade shape is identical.

Environment tested:
- [x] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [x] amd64
- [x] arm64

## Notes

- No new env vars, no new dependencies, no port change.
- In-place update over 1.17.0; the app carries its own SQLite migrations and upgrades existing data on first start.

### DIFF
--- a/hashrate-autopilot/docker-compose.yml
+++ b/hashrate-autopilot/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3010
 
   web:
-    image: ghcr.io/rdouma/hashrate-autopilot:1.17.0@sha256:8f952ab293975694b2c8169b7c5db18d0bc32636336d884712f3d89753bc665a
+    image: ghcr.io/rdouma/hashrate-autopilot:1.17.1@sha256:01b2541d5ecdf688dd459ad51b0cb8ec55a03f5beb4fba8abbada7737bd64a1b
     restart: on-failure
     stop_grace_period: 1m
     volumes:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @rdouma 🫡

PR #5883, opened 2026-07-15
Update Urbit to v4.6

yapishu:
- Updated Urbit to official v4.6 binary: https://github.com/urbit/vere/releases


nmfretz:
Excellent, thanks as usual @yapishu 🤝

PR #5885, opened 2026-07-15
Update Gupt to v2.11.0

nmfretz:
## Summary

Updates Gupt from 1.1.4 to 2.11.0.

Release: https://github.com/besoeasy/gupt/releases/tag/2.11.0  
Image: `ghcr.io/besoeasy/gupt:2.11.0@sha256:c852f064ff08252902a4c2815d81676025868f8a98ebc08ce6752778462fd762`

This is a major, backwards-incompatible upgrade. The user's identity remains, but conversations, Vault entries, and groups created in Gupt v1 are not available in Gupt v2. Gupt v1 and v2 clients also cannot communicate with each other. The release notes warn users to copy anything they need and back up their private key before updating.

## Testing

- `npm run lint:apps -- gupt --check-images`
- Confirmed the image index contains `linux/amd64` and `linux/arm64` images.
- Tested an update from 1.1.4 to 2.11.0 on an amd64 Umbrel device over its normal LAN HTTP app URL.
- Confirmed the existing identity was preserved.
- Confirmed direct messages, private groups, and Vault creation/read work over plain HTTP without `crypto.subtle` or `crypto.randomUUID`.
- Restarted the app through Umbrel and confirmed the Vault item, direct message, and group message persisted.
- Voice/video calls, screen sharing, and offline notifications still require HTTPS.

### DIFF
--- a/gupt/docker-compose.yml
+++ b/gupt/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8000
 
   web:
-    image: ghcr.io/besoeasy/gupt:2.1.15@sha256:65d0b5a3c7eee7a03ccb8756630363d237eae826111e022bc02f4eaad6ef82dd
+    image: ghcr.io/besoeasy/gupt:2.4.4@sha256:62f75f750f8bb6ce2cd10a6dffbc734cb763923b4d609ada8d2c1ace82a146b3
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/gupt/docker-compose.yml
+++ b/gupt/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8000
 
   web:
-    image: ghcr.io/besoeasy/gupt:1.1.4@sha256:9771510a51c3e86f4699b927a9cc86e2dcaa86f7e6951b22468e21e9bda75760
+    image: ghcr.io/besoeasy/gupt:2.1.15@sha256:65d0b5a3c7eee7a03ccb8756630363d237eae826111e022bc02f4eaad6ef82dd
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/gupt/docker-compose.yml
+++ b/gupt/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8000
 
   web:
-    image: ghcr.io/besoeasy/gupt:2.4.4@sha256:62f75f750f8bb6ce2cd10a6dffbc734cb763923b4d609ada8d2c1ace82a146b3
+    image: ghcr.io/besoeasy/gupt:2.10.2@sha256:c6531724bdb5143eab952f691b7f8a07e531a67355063c1097e2ba850ce06785
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/gupt/docker-compose.yml
+++ b/gupt/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 8000
 
   web:
-    image: ghcr.io/besoeasy/gupt:2.10.2@sha256:c6531724bdb5143eab952f691b7f8a07e531a67355063c1097e2ba850ce06785
+    image: ghcr.io/besoeasy/gupt:2.11.0@sha256:c852f064ff08252902a4c2815d81676025868f8a98ebc08ce6752778462fd762
     user: "1000:1000"
     restart: on-failure
     stop_grace_period: 1m
--- a/gupt/umbrel-app.yml
+++ b/gupt/umbrel-app.yml
@@ -2,12 +2,15 @@ manifestVersion: 1
 id: gupt
 category: social
 name: Gupt
-version: "2.1.15"
+version: "2.4.4"
 releaseNotes: >-
-  This update adds last-seen status, read receipts, typing indicators, improved WebRTC calls and file transfers, smarter relay selection and queued message delivery, plus vault item expiry.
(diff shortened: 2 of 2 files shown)


nmfretz:
Hey @besoeasy, I’m trying to update Gupt from 1.1.4 to 2.1.15 on Umbrel and noticed a few update-path changes that I wanted to check with you:

- Gupt now opens `gupt_app_cache_v3` without importing the existing v2 database. In my update test, a conversation visible on 1.1.4 was no longer visible after updating.
- Vault storage changed from kind-1 events containing a `gupt_vault` payload to kind-4 self-DMs. It looks like existing Vault entries are no longer read by 2.1.15.
- Groups changed from the NIP-59 `gupt-group/v2` format to the new group-private-key/kind-4 format. It looks like existing groups are no longer discovered after updating.

Are you already aware of these compatibility changes, and are you planning any migration or backward-compatibility handling in an upcoming release? Thanks!

besoeasy:
The jump from **1.x.x to 2.x.x** is intentional because GUPT v2 is **not compatible with older GUPT clients**.

A major reason is that the underlying architecture has changed significantly. For example, the server and relay infrastructure has evolved considerably:

https://github.com/besoeasy/gupt/blob/dc767747f90c4cff2863e5bfddc4c9d54419d0f3/src/config/servers.js#L1

When GUPT v1 was designed, it primarily worked best with Strfry-based relays. Today, the network consists of **300+ relays running a variety of different software stacks**, which required substantial protocol changes.

We also introduced a **trust system** to help mitigate WebRTC leaks and other privacy concerns.

**Most importantly, GUPT is no longer a Nostr application.** While GUPT originally started within the Nostr ecosystem, the requirements for privacy, anonymity, relay diversity, retention policies, and group messaging pushed us beyond what existing NIPs were designed to support. GUPT v2 uses its own protocol and architecture, and we no longer depend on `nostr-tools`.

Another significant change is that we no longer support Blossom servers. While Blossom is an interesting approach, reliability has been a challenge for us in practice. Despite the size of the Nostr ecosystem, we found that very few Blossom servers consistently accept encrypted blobs, and at times only a single reliable server was available. Building a privacy-focused system around such a small set of storage providers did not align with our goals for resilience and decentralization.

Data retention has also changed. Nothing in GUPT is retained for more than 100 days. Even if relays continue serving older data, the client intentionally ignores it.

Users are welcome to continue using **GUPT v1 for as long as they like**. However, from a development perspective, I prefer shipping anonymity and privacy improvements as soon as they are ready rather than delaying them to preserve compatibility with older protocols and clients.

It was a difficult decision, but maintaining backward compatibility would have required us to keep many of the limitations imposed by existing Nostr NIPs. GUPT v2 was built from the ground up to move beyond those constraints, and a clean break from v1 was ultimately the right choice.

**Bottom line:** for most users, everything should continue to work fine after upgrading. The main exception is **groups**, which have been completely redesigned in GUPT v2 and are not compatible with the old implementation.

If someone needs to access content from older chats or groups, they can always run an older GUPT client and browse that data there. That said, this is only relevant for relatively recent content, since **both GUPT v1 and GUPT v2 automatically expire chat data after 100 days**. Once content is older than that retention window, it is intentionally ignored by the client regardless of which version is being used.

besoeasy:
Hey, Gupt has been updated to **v2.2.0** since this PR was opened. Please use the latest release instead:

Release: https://github.com/besoeasy/gupt/releases/tag/2.2.0
Image: `ghcr.io/besoeasy/gupt:2.2.0@sha256:1047a583996320be941d32a23e4ba4c4c54eb3b925a29c9813473d5eb18cc3ba`

Key changes since 2.1.15:
- Call-request-as-message feature (calls now go through chat first)
- DNS TXT domain-based user discovery via Cloudflare DoH
- Relay standardization and dead code cleanup

nmfretz:
Thanks for clarifying @besoeasy. Totally understand wanting to move v2 forward rather than carry the limitations of v1. I think the clean break is fine as long as we’re very clear in the release notes, which I’ve updated along with bumping this to 2.4.4.

I tested the update on Umbrel and it works, but v2 can’t currently send messages because Umbrel app URLs are HTTP and GUPT now relies on `crypto.subtle`, which requires a secure context. I confirmed it works once served from a secure context.

Built-in HTTPS support is coming to umbrelOS very soon, so we’ll hold this PR until then and retest 🤝

besoeasy:
Hey @nmfretz — we fixed the HTTP secure context issue on the GUPT side. 

The problem was that DM encryption used `crypto.subtle` (AES-GCM), which requires HTTPS or localhost. On Umbrel's LAN HTTP URLs (e.g. `http://192.168.x.x:PORT`), it's undefined and message send/receive fails.

We replaced it with `@noble/ciphers/aes` `gcm()` — a pure JS AES-GCM implementation that's already in our dependencies. No secure context needed, wire format is unchanged, existing encrypted messages still decrypt.

Tested working on plain HTTP. This ships in the next GUPT release, so you no longer need to block on Umbrel's built-in HTTPS for GUPT specifically.

nmfretz:
Thanks @besoeasy, I just tested 2.10.2. DMs and groups are working now, but there's still one secure-context issue in the Vault. `crypto.randomUUID()` is undefined over HTTP, so creating an item fails with `crypto.randomUUID is not a function`.

Can you add a fallback there too? Once that lands I’ll retest and we should be good.

besoeasy:
@nmfretz pull the latest image

nmfretz:
Thanks @besoeasy, pulled 2.11.0 and retested on Umbrel. Vault creation now works over plain HTTP, and DMs, groups, and Vault data all persisted after restart. We’re good to go live 🤝

PR #5889, opened 2026-07-18
Update EE Gateway to 0.10.6

jmarbach:
Setup gets much easier: address geocoding on both location forms (type a street address or landmark and the gateway looks up the coordinates via OpenStreetMap), coordinate entry accepts pasted Google Maps "lat, lon" pairs, degree symbols, N/S/E/W letters, and European decimal commas, and the organization ID field is gone (the API token alone connects the gateway; existing installs unaffected). Also a worker reliability fix: HTTP 408/429 responses from the EE cloud are retried instead of the packet being dropped.

No permission or manifest-structure changes — image digest bumps + release notes only.

Full changelog: https://github.com/Encrypted-Energy/gateway/releases/tag/v0.10.6

### DIFF
--- a/ee-gateway/docker-compose.yml
+++ b/ee-gateway/docker-compose.yml
@@ -8,10 +8,9 @@
 # network (which it needs to reach the Bluetooth radio) while the UI stays on
 # Umbrel's app network behind app_proxy.
 #
-# Credentials are NOT passed as environment variables here. The user enters the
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @jmarbach! I made a small copy cleanup to align the App Store description and Compose comments with the new API-token-only setup. Everything tested and is working well. Going live!

PR #5891, opened 2026-07-18
Add PearCal Seeder

peerloomllc:
### App submission: PearCal Seeder

**PearCal Seeder** runs a *blind seeder* for [PearCal](https://github.com/peerloomllc/pearcal-native), a peer-to-peer calendar app with no accounts and no servers.

PearCal groups sync directly between members' devices, so a group's calendar is only reachable while at least one member's phone is online. This app runs the PearCal seed-mode worklet as an always-on background service on the user's Umbrel, replicating each enrolled group's **encrypted** blocks so members stay caught up regardless of who's online.

It is *blind*: the blocks it stores and serves stay encrypted, so the Umbrel keeps a group's data available without ever being able to read the events, notes, or members inside it. Group members admit the seeder when it enrolls and can revoke it group-wide at any time.

**To use it:** open the seeder dashboard, then in the PearCal phone app mint a seed invite for a group (Profile → Advanced → Blind peer) and paste it in, or scan the dashboard's QR code.

### Notes for review

- **Image:** `ghcr.io/peerloomllc/pearcal-seeder` — public, multi-arch (amd64 + arm64), pinned by manifest-list digest.
- **`security_opt: no-new-privileges:true`** (lint warning): a hardening measure — the seeder never needs elevated privileges; it only reads/writes its own data volume and speaks P2P over the network.
- **No inbound auth inside the container** (`SEEDER_NO_AUTH`): access is gated by Umbrel's `app_proxy`, matching the existing PearCircle Seeder app.
- The in-app update checker is disabled (`SEEDER_NO_UPDATE_CHECK`) since updates come through the App Store.
- Developer: PeerLoom. Same submitter as the existing PearCircle Seeder app.

Gallery images will be provided as requested.

### DIFF
--- a/pearcal-seeder/docker-compose.yml
+++ b/pearcal-seeder/docker-compose.yml
@@ -8,7 +8,7 @@ services:
 
   app:
     # Published, public, multi-arch (amd64 + arm64), pinned by manifest-list digest.
-    image: ghcr.io/peerloomllc/pearcal-seeder:1.0.33@sha256:36b9e9407b053c71b840f29a5f0423a67f15c2806827687b2cbec9dd69586c45
+    image: ghcr.io/peerloomllc/pearcal-seeder:1.0.34@sha256:6273ea9b8bb34a841768598587b601ef02a7b066625952633f08b24bb76111c6
     restart: on-failure
     stop_grace_period: 1m
     security_opt:
--- a/pearcal-seeder/docker-compose.yml
+++ b/pearcal-seeder/docker-compose.yml
@@ -1,0 +1,23 @@
+version: "3.7"
+
+services:
+  app_proxy:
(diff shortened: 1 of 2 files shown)


peerloomllc:
Updated to 1.0.34 (public multi-arch manifest-list digest `sha256:6273ea…`, linux/amd64 + linux/arm64), and cleared `releaseNotes`/`gallery` per the linter for new-app submissions.

Re: the `security_opt: no-new-privileges:true` warning — this is intentional. The seeder runs an always-on background worklet that only replicates encrypted P2P blocks; it needs no privilege escalation, so we set `no-new-privileges` to harden the container. Happy to remove it if you'd prefer it omitted for new submissions.

PR #5912, opened 2026-07-22
Update Go Brrr Pool to 1.2.1-patch.1

lukechilds:
Point Go Brrr Pool directly at the Mempool `web` service instead of the historical `mempool_app_proxy_1` sidecar hostname.


nmfretz:
Hey @gobrrrme, just a quick heads-up. We made a small Umbrel-side maintenance change here to future-proof Go Brrr Pool’s optional local Mempool integration. It now connects directly to Mempool’s web service instead of going through the app proxy.

Nothing needed on your end, I just wanted to keep you in the loop!

gobrrrme:
Thanks!

PR #5914, opened 2026-07-23
bitcoin-cash-node 29.1.36: fix vanishing onion-peers list

danhaus93-ops:
Dashboard fix, field-diagnosed on a live node yesterday.

`getnodeaddresses` excludes addresses the address manager has marked terrible after repeated failed attempts. A stretch of automatic retries against two offline onion services buried 21 of 23 known onion addresses -- including all seven of BCHN's own fixed seeds -- leaving the panel showing only dead entries while Tor itself was healthy end to end. A manual `addnode` to one seed connected in under 90 seconds and the panel healed as its gossip refreshed the address manager.

Two changes so it cannot recur: the seven fixed onion seeds (contrib/seeds/nodes_main.txt) are always merged into the list, marked and sorted first, immune to address-manager decay; and the keep-connected option now retires a pin that has not landed in 20 minutes with a one-hour cooldown instead of retrying dead services forever (which was itself deepening the decay that hid good addresses).

Web image pin updated to the release containing the fix; regression tests cover the decayed-address-manager scenario.

### DIFF
--- a/bitcoin-cash-node/docker-compose.yml
+++ b/bitcoin-cash-node/docker-compose.yml
@@ -55,7 +55,7 @@ services:
         ipv4_address: ${APP_BITCOIN_CASH_NODE_NODE_IP}
 
   web:
-    image: ghcr.io/danhaus93-ops/bchn-dashboard:e9b9860b8b9a87178a9bca1985b4793b694ceb8d@sha256:cbb3fc4bd5bf43906993165d9fdcfc60a4978bdbfc33ce8994ea1f4449ac6312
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks @danhaus93-ops. Going live.

As a quick FYI, while testing I noticed a small edge case where a pinned peer that connects and later goes offline may not rotate out, and a mix of connected and timed-out peers can leave fewer than four pinned peers. Nothing that needs to hold this update up, but it may be worth a look for the next one.

PR #5915, opened 2026-07-23
Update Hashrate Autopilot to 1.17.4

rdouma:
### Type
App update

### App
Hashrate Autopilot (`hashrate-autopilot`) - 1.17.2 → 1.17.4

### Summary
Stability + docs update. Notable since the last accepted version (1.17.2):

- **Fixed:** the daemon no longer crash-loops when the Electrum server (electrs / Fulcrum / ElectrumX) restarts or drops an idle connection mid-request. Previously that surfaced as an uncaught `write EPIPE` that crashed the daemon and restarted it in a loop; it now handles the dropped connection cleanly and retries.
- **Fixed:** removed a misleading "Telegram 2FA tap required" note from bid creation. The app places bids via Braiins' API using the account owner token, which needs no Telegram confirmation (that step exists only in Braiins' web interface).
- **Fixed (1.17.3):** deduced Lightning payout amounts no longer overcount a partial sweep - the recorded amount is the actual unpaid-balance decrease, not the full pre-drop reading.
- **Fixed (1.17.3):** the Bitaxe miners card respects the number-format locale instead of a hard-coded decimal point.
- **New:** a plain-language FAQ (setup/topology, payouts & P&L, bidding, requirements).

Full release notes: https://github.com/rdouma/hashrate-autopilot/releases/tag/v1.17.4

### Verification
- Image is **multi-arch** (`linux/amd64`, `linux/arm64`), built by CI from the tagged release and **digest-pinned** in `docker-compose.yml` (`1.17.4@sha256:d43ce6ae…`). It is the same build already live on the community store.
- `docker-compose.yml` is unchanged from the last accepted version except the image tag + digest.
- I have not re-run a full clean Umbrel-device install for this exact build; leaving the tested-environment checkboxes for a maintainer to confirm.

### DIFF
--- a/hashrate-autopilot/docker-compose.yml
+++ b/hashrate-autopilot/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3010
 
   web:
-    image: ghcr.io/rdouma/hashrate-autopilot:1.17.2@sha256:7346288b77e979ff654a31337d2028b870e7076accdc8a86518764e87ddd2632
+    image: ghcr.io/rdouma/hashrate-autopilot:1.17.4@sha256:d43ce6aedf9166c73687c80aaf5b863bd650328fdd8eb6ad76c4ce25012936fd
     restart: on-failure
     stop_grace_period: 1m
     volumes:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks as usual @rdouma

PR #5916, opened 2026-07-23
bitcoin-cash-node 29.1.37: review-note fixes for the onion peers panel

danhaus93-ops:
Follow-up to #5914, addressing @nmfretz's review note directly:

1. A pinned peer that connected and later went offline could hold its pin slot forever (tracking was dropped on connect). The retirement clock now keeps running while the peer is live, so a peer that dies rotates out 20 minutes after it was last seen connected.
2. The pin budget double-counted connected pinned peers (present in both the live set and the pinned set), which could leave fewer than four pins. It now counts the union.
3. Pinned-but-untracked addresses are re-armed each reconcile pass.

Also fixes a field report from the same install: the known-onion count could disagree with the rendered list (addrman can return one address under two ports). The list is now deduped by address, so the count tracks reality.

Regression tests cover all four behaviors.

### DIFF
--- a/bitcoin-cash-node/docker-compose.yml
+++ b/bitcoin-cash-node/docker-compose.yml
@@ -55,7 +55,7 @@ services:
         ipv4_address: ${APP_BITCOIN_CASH_NODE_NODE_IP}
 
   web:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @danhaus93-ops, going live.

PR #5927, opened 2026-07-27
pearcircle-seeder: update to 1.1.0

peerloomllc:
Bump PearCircle Seeder to 1.1.0 (image pinned to the multi-arch manifest-list digest).


nmfretz:
Thanks @peerloomllc, and sorry for the delay. Added release notes.

PR #5933, opened 2026-07-28
Update Stratum V2 UI to v0.6.0

GitGab19:
## Type
App update

## App

App ID: sv2-ui
Upstream project: https://github.com/stratum-mining/sv2-ui
Version: 0.6.0

## Summary

In `sv2-ui` v0.6.0 we added NexusPool and CK Pool (solo mining), simplified node setup, unified pool configuration flows, and a streamlined dashboard with per-worker best difficulty and more accurate hashrate reporting based on new miner telemetry data. Now coinbase verification and minimum individual worker hashrate can be configured from the UI, custom pools in Job Declaration mode can specify a JDS port, and logs feature smart color coding for easier reading. More details in the release notes at https://github.com/stratum-mining/sv2-ui/releases/tag/v0.6.0.

## Verification

Umbrel testing performed:
Tested on my Umbrel VM and everything works properly.

Environment tested:
- [ ] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [x] amd64
- [ ] arm64

### DIFF
--- a/sv2-ui/docker-compose.yml
+++ b/sv2-ui/docker-compose.yml
@@ -37,7 +37,7 @@ services:
       - ${APP_DATA_DIR}/data/config:/app/data/config
 
   sv2-ui:
-    image: stratumv2/sv2-ui:v0.5.0@sha256:e02bb0ec9588dda71817ef70bad9ff6c41f800d5d2c770ca4e39227d7d48381d
+    image: stratumv2/sv2-ui:v0.6.0@sha256:e9fc51df9c0fc02d367b4fe1a72f8216d7aaa5ce8a502982965610b95f244305
     stop_grace_period: 1m
     restart: on-failure
     # required: sv2-ui expects docker DNS to resolve container names in production.
--- a/sv2-ui/umbrel-app.yml
+++ b/sv2-ui/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1.1
 id: sv2-ui
 category: bitcoin
 name: Stratum V2 UI
-version: "0.5.0"
+version: "0.6.0"
 tagline: Setup and monitor Stratum V2 mining
 description: >-
   Stratum V2 UI is a setup wizard and monitoring dashboard for running a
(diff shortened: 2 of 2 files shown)


lucasbalieiro (approved):
confirmed that the sha256:05cb175c3bb470f9c356410aaadbe5a4e7b606d511cee30647350ffd300d65d4

is from known multi-arch image.

Image

I've sent this update to my Umbrel and checked that it is working. Including the ASIC telemetry

nmfretz:
Hey @GitGab19 and @lucasbalieiro, thanks for this! I hit a snag when updating an app (fresh install works fine) that we need to fix before going live.

It’s a fairly narrow case and easy to miss with a fresh setup.

The outer app updated correctly, but the mining services would not remain running. The nested `sv2-jdc` v0.7.0 container exits immediately because the existing `jdc.toml` is missing the required `version` field under `[template_provider_type.BitcoinCoreIpc]`:

```text
Job Declarator Client config error: missing field `version`
```

The Translator starts afterward, but because JDC has already exited, it cannot resolve `sv2-jdc` and also shuts down.

Codex helped me confirm that this configuration was generated by sv2-ui rather than edited manually. The same legacy format was shipped with the earlier Umbrel 0.1.0 and 0.4.0 packages. The v0.5.0 generator includes the new field, but app updates do not regenerate persisted TOML files, so users carrying an older JD configuration forward can still hit this when v0.6.0 upgrades the nested JDC image to v0.7.0.

Can you guys look into this and see if you can add an update-safe migration that patches or regenerates legacy JD configs before auto-starting the mining services? That way, existing users don’t end up with a broken mining setup after updating.

--------

I also pushed a small release notes change to make them easier to scan:wording-only follow-up to make the release notes easier to scan:

image

GitGab19:
Hey @nmfretz, thanks for reporting the issue you faced, and for adjusting release notes.

We just merged https://github.com/stratum-mining/sv2-ui/pull/234 which introduced safe mechanisms for updates, and overwritten the `0.6.0` release.

I just updated the PR here with the new digest.

nmfretz:
Awesome, thanks for fixing so quickly @GitGab19! I retested the update on the same Umbrel with the legacy JD configuration that originally failed. Totally resolved. We’re good to go 🚀

PR #5938, opened 2026-07-30
Update Bassin to 2.1.2

blockdyor:
## Type

App update

## App

App ID: bassin
Upstream project: https://github.com/blockdyor/bassin
Version: 2.1.2

## Summary

This update transitions ownership and maintenance of the app to blockdyor, fixes UI security issues, adds usability enhancements, and updates Docker image targets:

* **Maintainer & Repository Updates**: Updated metadata (`submitter`, `developer`, `website`, `repo`, `support`) and icon/gallery asset URLs to point to `blockdyor/bassin`.
* **Image Registry Migrations**: Updated `ckpool`, `ui`, and `widget-server` Docker container images in `docker-compose.yml` to point to active builds under `ghcr.io/blockdyor/`.
* **UI Copy-to-Clipboard Feature**: Added a copy button next to the stratum host URL on the onboarding screen for easy clipboard copying.
* **UI Hardening Fix**: Replaced an unsafe HTML-injection pattern in the UI's number formatting utility with safe React rendering.
* **Docker Compose Fixes**: Refactored entrypoints and environment config handling across services (`ckpool`, `ui`, `widget-server`).

## Verification

Umbrel testing performed:

Environment tested:
- [ ] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [ ] amd64
- [x] arm64

Known lint warnings or caveats: None

## Notes

* **Maintainer Handover**: The original author (`duckaxe`) has handed over project maintenance to `blockdyor`.
* **Port & ID Integrity**: The App ID remains `bassin` and port remains `5008` for compatibility with existing installations.


bassin/docker-compose.yml
```diff
@@ -12,10 +12,8 @@ services:
     entrypoint: >-
       /bin/sh -c '
         apk add --no-cache envsubst &&
-        if [ ! -f /config/ckpool.conf ]; then
-          envsubst < /config/ckpool.conf.template > /config/ckpool.conf &&
-          chown 1000:1000 /config/ckpool.conf
-        fi
+        envsubst < /config/ckpool.conf.template > /config/ckpool.conf &&
```
nmfretz: Can you clarify what prompted the removal of this guard?

Its purpose is to generate `/config/ckpool.conf` only when it doesn’t already exist, preserving any settings the user has changed. Without it, the init service regenerates the file from the default template whenever it runs, silently overwriting those changes.

Some of these settings control how ckpool connects to Bitcoin Core and handles connected miners. If a user’s setup depends on custom values, reverting them could change the pool’s behavior or interrupt mining until the user notices and restores their configuration.

bassin/docker-compose.yml
```diff
@@ -25,10 +23,11 @@ services:
       - APP_BITCOIN_RPC_USER=${APP_BITCOIN_RPC_USER}
       - APP_BITCOIN_RPC_PASS=${APP_BITCOIN_RPC_PASS}
       - APP_BITCOIN_ZMQ_HASHBLOCK_PORT=${APP_BITCOIN_ZMQ_HASHBLOCK_PORT}
+    restart: on-failure
```
nmfretz: Can we remove the restart policy from this one-shot init service? `ckpool` already waits for it to complete successfully, and an unbounded `on-failure` policy could leave the app stuck repeatedly retrying a persistent init error.

If there’s a particular transient failure this is intended to handle, can you clarify that?

bassin/docker-compose.yml
```diff
@@ -25,10 +23,11 @@ services:
...
 
   ckpool:
-    image: ghcr.io/getumbrel/docker-ckpool-solo:v1.1.1@sha256:499f0d4f72b93130bde06463b6aab25fa80f1f2349bab0e28f1c9a59b02f745b
-    command: '-k -B -c /config/ckpool.conf'
+    image: ghcr.io/blockdyor/docker-ckpool-solo:v1.1.1@sha256:faf0c1a21017915949e946141bad38aee9bfdcfffe0fc3952b2f8fb283965b1e
```
nmfretz: Are you mainly looking to take ownership of the Bassin-specific UI and widget images, or do you also want to take ongoing ownership of the ckpool image itself?

We’ve kept this image under `getumbrel` for a specific reason. A previous community-provided ckpool image used by Bassin was [found generating incorrect coinbase transactions](https://github.com/duckaxe/bassin/issues/23), with most of the block reward going to an unexpected address. We replaced it with a transparent build from the official ckpool source under our namespace so we could control and verify this particularly sensitive part of the stack.

This isn’t a concern with the code in your fork, which currently matches ours. It’s just important that we’re clear about who is taking responsibility for maintaining and securing this image going forward. We’re more than happy to continue owning and publishing the ckpool image while you maintain the Bassin-specific parts.

bassin/docker-compose.yml
```diff
@@ -25,10 +23,11 @@ services:
...
+    entrypoint: /bin/sh -c './bin/ckpool -k -B -c /config/ckpool.conf'
```
nmfretz: Can we keep passing these as `command` arguments? The image already defines `ckpool` as its entrypoint, so overriding it with `/bin/sh -c` isn’t needed. It also hardcodes the binary’s internal path and may prevent ckpool from receiving shutdown signals directly.

```suggestion
    command: '-k -B -c /config/ckpool.conf'
```

bassin/docker-compose.yml
```diff
@@ -60,10 +59,12 @@ services:
     depends_on:
       - ui
     restart: on-failure
+    # container_name removed: Umbrel injects container names automatically
```
nmfretz: Can we remove this comment? Umbrel already handles container naming, so we don’t need to document it inside the app’s Compose file.

bassin/umbrel-app.yml
```diff
@@ -1,47 +1,45 @@
...
-    - Better difficulty adjustment for bursty and higher-hashrate miners.
-    - Added compatibility for upcoming Bitcoin consensus changes, along with
-      stability and miner-protocol fixes.
+  Key highlights:
+    - Replaced an unsafe HTML-injection pattern in the UI's number
+  formatting with safe React rendering. No visual or functional
+  change, purely a hardening fix.
+    - Added a copy-to-clipboard button next to the stratum host on the
+  onboarding screen, so you can grab it without selecting the text
+  by hand.
```
nmfretz: Can we make these a little more concise and user-facing? Existing users see them in the update dialog, so they don’t need the React implementation details.

```suggestion
  Improves security when displaying mining statistics and adds a button for
  copying the Stratum server address from the onboarding screen.
```

bassin/umbrel-app.yml
```diff
@@ -1,47 +1,45 @@
...
-website: https://github.com/duckaxe
-repo: https://github.com/duckaxe/bassin
-support: https://github.com/duckaxe/bassin/issues
-submission: https://github.com/getumbrel/umbrel-apps/pull/3375
+submitter: blockdyor
+developer: blockdyor
+website: https://github.com/blockdyor
+repo: https://github.com/blockdyor/bassin
+support: https://github.com/blockdyor/bassin/issues
+submission: https://github.com/getumbrel/umbrel-apps/pull/5938
```
nmfretz: Can we keep `submitter: duckaxe` and `submission: https://github.com/getumbrel/umbrel-apps/pull/3375`?

These fields always describe the original App Store submission. The App Store uses `submission` as the hyperlink on the submitter’s name, so changing them to this update PR would replace the original attribution and link away from the actual submission.

We’re very happy for you to take over ongoing maintenance. The `developer`, `website`, `repo`, and `support` fields can all be updated to reflect that.

bassin/umbrel-app.yml
```diff
@@ -1,47 +1,45 @@
...
+icon: https://raw.githubusercontent.com/blockdyor/bassin/main/duckaxe-bassin/icon.svg
 gallery:
-  - 1.jpg
-  - 2.jpg
-  - 3.jpg
+  - https://raw.githubusercontent.com/blockdyor/bassin/main/duckaxe-bassin/gallery_1.jpg
+  - https://raw.githubusercontent.com/blockdyor/bassin/main/duckaxe-bassin/gallery_2.jpg
+  - https://raw.githubusercontent.com/blockdyor/bassin/main/duckaxe-bassin/gallery_3.jpg
```
nmfretz: Can we remove `icon` and keep the existing relative gallery entries? Official App Store icons and screenshots are hosted in our separate assets repository, so the raw upstream URLs aren’t needed here.

If you’d like refreshed screenshots for the updated UI, we’re happy to update those for you before this goes live.

```suggestion
gallery:
  - 1.jpg
  - 2.jpg
  - 3.jpg
```

nmfretz (commented):
Thanks for putting this together @blockdyor, and for picking up ongoing maintenance of Bassin.

There are two main things I want to understand before testing: the change to how existing ckpool configuration is handled, and whether you’re looking to take ownership of the ckpool image itself or just the Bassin-specific images.

I’ve left the details inline, along with a few smaller Compose and App Store metadata cleanups. Once those are clarified and updated, I can re-test the update path from 2.1.1.

bassin/docker-compose.yml
blockdyor: Thanks for the review 😊 I agree with all of these points.

I removed that guard unintentionally.
The original behavior is correct: preserve a user-edited /config/ckpool.conf and only generate it from the template when it does not already exist.
I’ll restore that guard so we do not silently overwrite custom user settings.

bassin/docker-compose.yml
blockdyor: Yes, I’ll remove it.
This is a one-shot init service and ckpool already waits for it to complete.
An unbounded retry is unnecessary and could cause the app to loop on a persistent init failure.

bassin/docker-compose.yml
blockdyor: I want to take ownership of the Bassin-specific UI and widget images only.
I’m happy to leave the ckpool image under Umbrel’s namespace if you prefer, since that is the sensitive component and you already have the trusted history there.
If you want, I can update the PR to keep the ckpool image pointing at getumbrel and only change the Bassin UI/widget images to blockdyor.

bassin/docker-compose.yml
blockdyor: I agree with the change.
We should not override the image entrypoint with /bin/sh -c ...
I’ll use command: '-k -B -c /config/ckpool.conf' instead, so ckpool receives shutdown signals correctly and we don’t hardcode internal binary paths.

bassin/umbrel-app.yml
blockdyor: I’ll keep:

- submitter: duckaxe
- submission: https://github.com/getumbrel/umbrel-apps/pull/3375

I will update only:

- developer
- website
- repo
- support

That preserves original attribution while reflecting current maintenance.

bassin/umbrel-app.yml
blockdyor: I’ll remove the raw upstream icon and keep the existing relative gallery entries. Umbrel’s official App Store assets repo should continue to host the icons/screenshots.

bassin/umbrel-app.yml
blockdyor: Absolutely. I'll remove this.

bassin/docker-compose.yml
blockdyor: Yes, we can absolutely remove this extra comment.

blockdyor:
hi @nmfretz thank you for your review! changes are applied, let me know what do you think.

nmfretz:
Thanks @blockdyor. It looks like several of the remaining changes were committed to `blockdyor/blockdyor-umbrel-community-app-store` rather than the branch behind this PR, so I went ahead and pushed the agreed packaging changes here to get everything in sync.

The final package now:

- preserves an existing `ckpool.conf` during updates
- keeps the ckpool image under `getumbrel`, while moving the Bassin-specific UI and widget images to `blockdyor`
- treats the init service as a one-shot service without retrying persistent failures
- preserves the original App Store submission attribution and gallery references
- keeps the release notes focused on the minor UI update

I updated my production Umbrel from 2.1.1 to 2.1.2. Bassin came back up successfully, my miner connected and continued working, the mining statistics rendered correctly, and I confirmed the onboarding copy button works.

Everything looks good from an Umbrel packaging and update-path perspective. Thanks again for taking over ongoing maintenance of Bassin. Going live 🚀

blockdyor:
Thank you very much @nmfretz and thanks for taking care of this in a crazy week like this one. This is greatly appreciated! Already working on some new features, especially for the timestamps since some users requested it.

PR #5941, opened 2026-07-31
Update EE Gateway to 0.10.7

jmarbach:
Big efficiency and data-integrity release. The worker now uploads queued packets in one batched request per drain instead of one request per packet, cutting network requests and SD-card writes on busy gateways by up to 500x. Each packet also now reports when its GPS coordinate was captured and its horizontal accuracy, so the upstream network can verify location freshness. Rounding it out, the last unused org-ID leftovers are removed from the worker config (API token alone connects the gateway; existing installs unaffected).

No permission or manifest-structure changes — worker image digest bump + release notes only. UI unchanged at 0.6.4 (existing digest kept). Description copy edits from the 0.10.6 review are preserved.

Full changelog: https://github.com/Encrypted-Energy/gateway/releases/tag/v0.10.7

### DIFF
--- a/ee-gateway/docker-compose.yml
+++ b/ee-gateway/docker-compose.yml
@@ -34,7 +34,7 @@ services:
   # extra download, and it is the only place root is used. This is NOT privileged
   # mode; the app containers themselves stay non-root.
   init_perms:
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks for this @jmarbach! Everything is working well from the umbrel packaging side.

PR #5943, opened 2026-07-31
peartune: add PearTune 1.0.0

peerloomllc:
## App

**PearTune 1.0.0** — plays the music already on your Umbrel from a phone anywhere, over a direct encrypted peer-to-peer connection. No port forwarding, no VPN, no dynamic DNS, no account, and no copy of the library in anyone's cloud.

- **Upstream project:** https://github.com/peerloomllc/peartune (MIT)
- **Website:** https://peerloomllc.com/peartune/
- **Image:** `ghcr.io/peerloomllc/peartune-host:0.2.41`, pinned by digest, built from [`host/Dockerfile`](https://github.com/peerloomllc/peartune/blob/master/host/Dockerfile) in that repo. Multi-arch manifest list, `linux/amd64` + `linux/arm64`.

It reads the library that is already on the box: point it at an existing Navidrome or Jellyfin and it uses that server's tags, artwork and transcoding, or point it at a plain folder and it reads the tags itself (ID3, Vorbis, MP4, FLAC, including embedded cover art). The source is chosen in the dashboard — no compose edits.

The phone app is the client. It is on the App Store now and in closed testing on Google Play.

## Screenshots

The dashboard is the whole browser surface: first run, choosing the music source, and managing who has access.

First run
Music source
People and devices
Pairing QR

Logo: https://raw.githubusercontent.com/peerloomllc/peartune/master/umbrel/icon.svg

## App Store Standard

Opens straight to a web dashboard on port 8741 with a three-step first-run flow: name the library, point it at your music, pair a phone. No SSH, no CLI, no log scraping, no file edits for normal use.

## Host access and permissions — the two lint warnings

**`network_mode: host` — required, and measured.** PearTune reaches phones by outbound UDP holepunching over the Hypercore DHT, which is what removes the need for port forwarding. Docker's default bridge adds a second layer of NAT that kills it: the connection is admitted and then no media flows. That was measured on a real Umbrel twice before host networking was adopted, and the reasoning is recorded in the compose file itself. The manifest `port: 8741` matches the host listener the dashboard binds.

**`security_opt: no-new-privileges:true` — hardening, not a privilege request.** It only prevents the container from gaining privileges it was not started with. Happy to drop it if you would rather packages not set `security_opt` at all.

**`permissions: [STORAGE_DOWNLOADS]`** — the compose mounts `${UMBREL_ROOT}/home/Downloads` **read-only** at `/library`. That is where an Umbrel user's music actually lives (it is what Navidrome's own package mounts). The whole Downloads directory rather than `music/` alone, so the dashboard's folder picker can also reach `podcasts/` or `audiobooks/`. PearTune never writes to it — it serves music and has no import or edit path.

No Docker socket, no privileged mode, no device mounts, no extra capabilities.

## Persistence

`${APP_DATA_DIR}/data` holds the host identity seed and the grant store (which devices are allowed in). Losing it means every phone must re-pair, so it is deliberately the only writable volume, and `peartune/data` is committed for fresh installs.

## Default credentials

`defaultUsername: ""` with `deterministicPassword: true`, so umbrelOS shows the password it derives. There is no user account — the dashboard is password-gated with Umbrel's generated app password, and the phone side has no password at all: each connection is authenticated by the device's own key, so there are no bearer tokens or connection strings anywhere in the design.

## Testing

- `npm run lint:apps -- peartune --check-images` → 0 errors, 2 warnings (the two justified above).
- The image has been running as the author's daily host on an Umbrel Home (N100) against a 1358-track Navidrome library, plus a plain-folder library, with phones streaming on the LAN and over cellular.
- Pair, browse, play, seek, and revoke-mid-song verified on real hardware; revocation cuts new requests within about a second.

### DIFF
--- a/peartune/docker-compose.yml
+++ b/peartune/docker-compose.yml
@@ -1,0 +1,58 @@
+version: "3.7"
+
+services:
+  # NO app_proxy, and that is not an oversight.
+  #
+  # PearTune's host MUST run with network_mode: host. Measured on a real Umbrel,
+  # twice, with the real image: under Docker's default bridge the firewall admits
+  # the phone (gate:allow-for-pairing) and the connection then dies before the pair
+  # channel opens. Bridge NAT is a second layer of NAT and holepunching does not
+  # survive it.
+  #
+  # app_proxy is a container on a bridge network, so it cannot front a
+  # host-networked service - which is why Umbrel's own host-networked apps (Plex,
+  # Home Assistant) do not use it either. The proxy WAS the only thing standing in
+  # for our missing auth, so the dashboard now has its own: PEARTUNE_PASSWORD, which
+  # umbrelOS fills with the app password shown next to PearTune in the app list.
+  # The host refuses to start on a non-loopback bind without it.
+  app:
+    image: ghcr.io/peerloomllc/peartune-host:0.2.41@sha256:47ab391cc042382a1b5c2eee8dddbba936153b1b3b6695ae41765055e3504032
+    restart: on-failure
+    stop_grace_period: 1m
+    network_mode: host
+    security_opt:
+      - no-new-privileges:true
+    environment:
+      PEARTUNE_HTTP_HOST: 0.0.0.0
+      PEARTUNE_HTTP_PORT: "8741"
+      PEARTUNE_PASSWORD: ${APP_PASSWORD}
+      PEARTUNE_NAME: My Library
+      PEARTUNE_DATA: /data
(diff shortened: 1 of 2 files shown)


peerloomllc:
Hi - just checking in on this one, it has been about a month. Nothing is blocked on our side, and I am happy to answer anything or make changes if a reviewer has a moment.

PearTune has moved on to 1.0.7 since this was opened, so I can refresh the listing to the current version and image digest here if that is useful - just say the word and I will push it.

PR #5945, opened 2026-08-01
Add DeviceShelf 1.9.6

wealthwallet:
Adds **DeviceShelf Server**: a local network scanner and device inventory. It lists what is on the LAN with vendor, hostname, OS, device type, open ports and the services behind them, alerts on new devices, devices dropping offline and newly opened ports (ntfy / Gotify / webhook / Apprise), and publishes favourites to Home Assistant over MQTT discovery as native entities.

Site: https://deviceshelf.app/server · Image: `ghcr.io/wealthwallet/deviceshelf-server` (amd64 + arm64, built in CI per release)

### Host networking and capabilities — the justification the guidance asks for

Layer-2 discovery (ARP, mDNS, passive DHCP) only works inside the host's network namespace. Behind the app proxy on a bridge network, the scanner inventories Docker's own subnet and reports an almost empty LAN — the app's entire function is the thing that breaks. AdGuard Home, Pi-hole and Tailscale are packaged the same way here.

`NET_RAW` is for the ARP scan, `NET_BIND_SERVICE` for passive DHCP fingerprinting on UDP/67. Nothing else: no privileged mode, no host Docker socket, no host mounts beyond `${APP_DATA_DIR}`. The process runs as uid 10001, not root.

Tested on Ubuntu 24.04 (arm64) bridged onto a real LAN: 39 devices found with vendor, hostname and type.

### Browser-only setup

The server otherwise mints a random token into a file inside the container, which would need `docker exec` to read — ruled out by the App Store Standard. The package passes `${APP_PASSWORD}` instead and sets `deterministicPassword: true`, so umbrelOS displays the working credential and the dashboard is reachable from the browser with nothing else to do.

### hooks/pre-start

The image runs as uid 10001 and `${APP_DATA_DIR}` is created as root, so without a chown the server cannot create its config directory: it logs `permission denied` and restart-loops. Reproduced, then fixed, on the VM above. Same pattern as `tdex`, `lndg` and `komodo`.

### Port

`8097`. The product default is 8088, but `element` already claims that here. Checked against all 390 `port:` declarations in this repository rather than by search.

### Licence

DeviceShelf is proprietary, paid software: €59 once, no subscription, 7-day trial without a card, and the same licence covers the desktop and mobile apps. Flagging it plainly rather than leaving it for a reviewer to find. Happy to adjust anything that does not fit the store's expectations.


wealthwallet:
Updated to **1.9.4** (`ghcr.io/wealthwallet/deviceshelf-server:1.9.4`), the version currently shipping. Happy to answer anything on the review.

wealthwallet:
Updated to 1.9.4 and fixed what the linter flagged:

- **Image digest pinned** — `ghcr.io/wealthwallet/deviceshelf-server:1.9.4@sha256:481b6f57569672fff9f4f28934cd1d4d527d2a1787c5873654bbbabdd683b54b`
- **Port moved 8097 → 8104.** 8097 was free against every `umbrel-app.yml`, but the linter found it in `blockstream-blind-oracle/exports.sh`. 8104 appears in no file of the repository.
- **`deviceshelf/data` is now committed** (empty, with a `.gitkeep`) so the bind mount source exists.

The two remaining notes are inherent to what the app does, so here is the justification you asked for:

**Host networking.** DeviceShelf discovers devices at layer 2 — ARP sweeps, mDNS/SSDP, passive DHCP. Inside a bridge network it would enumerate Docker's own subnet and find nothing of the user's LAN. This is the same reason AdGuard Home, Pi-hole and Tailscale are packaged with `network_mode: host` in this store. There is consequently no `app_proxy`, and the dashboard is opened directly on the manifest port.

**Capabilities.** `NET_RAW` is needed for the ARP scan (raw sockets via libpcap) and `NET_BIND_SERVICE` for listening on UDP/67 for passive DHCP fingerprinting. Nothing else is requested — no `privileged`, no host mounts beyond `${APP_DATA_DIR}/data`.

Because host networking puts the dashboard on the LAN, the API token is bound to Umbrel's `${APP_PASSWORD}`, which is what `deterministicPassword: true` promises — the app is never reachable unauthenticated.

wealthwallet:
Bumped to 1.9.32 (image digest pinned, `ghcr.io/wealthwallet/deviceshelf-server:1.9.32@sha256:239b12f71c8a86cbd1705d4d5742255cd2836130ddeb05a6ac0d672ebb63ba91`). The PR has been open since 1 August with the linter findings fixed on 10 August. Is anything else needed from my side for a review?

PR #5955, opened 2026-08-04
MyNeS: add app

fxerkan:
Adds **MyNeS (My Network Scanner)** — a self-hosted LAN scanner and device inventory.

- Source: https://github.com/fxerkan/my_network_scanner (MIT)
- Image: https://hub.docker.com/r/fxerkan/my_network_scanner — `linux/amd64` + `linux/arm64`, pinned by digest

## What it does

An IP scan of my LAN found 29 devices. There were about 60 — everything on Zigbee and Z-Wave has no IP address, so an IP scan is blind to it.

MyNeS discovers devices over ARP, mDNS/Bonjour, SSDP/UPnP, Matter, Bluetooth LE **and MQTT**. Reading retained Zigbee2MQTT / Z-Wave JS / Tasmota topics is the only way to inventory a radio device with no IP at all, so those end up in the same list as the laptops.

Also: vendor lookup over a 1000+ entry OUI database, automatic device-type classification, topology and force-directed graph views, scan history, rule-based alerts on new / missing / changed devices delivered by Web Push (its own — no third-party relay), webhook, e-mail or Home Assistant, and two-way Home Assistant integration.

No cloud, no account, no telemetry. Installs as a PWA. UI in English and Turkish.

## The bit I expect review to focus on: no `app_proxy`

MyNeS uses `network_mode: host` and therefore serves its own UI on `${APP_PORT}` directly, following the same pattern as the existing `home-assistant` app.

This is not a shortcut. Raw ARP frames and mDNS/SSDP multicast do not cross a bridge network — behind `app_proxy` on a bridge, the scanner would enumerate the Docker bridge instead of the user's LAN, which makes the app pointless.

It asks for `NET_ADMIN` and `NET_RAW` so it can build ARP frames and read interface state. It does **not** run as root (`USER scanner`, uid 1000) and it is **not** a privileged container. Without those capabilities it does not fail — it degrades to a ping sweep plus the OS ARP cache and finds fewer devices, and `GET /api/capabilities` reports exactly what is missing and why.

Because there is no `app_proxy`, there is no auth in front of it. `defaultUsername`/`defaultPassword` are empty and the app assumes a trusted LAN, which is documented upstream in `SECURITY.md`. If you would prefer it behind `app_proxy` with reduced discovery, or want any wording changed in the description, tell me and I will push it.

## Checklist

- [x] `umbrel-app.yml` + `docker-compose.yml`
- [x] Image pinned by digest (`sha256:fcd085e7…`)
- [x] Multi-arch: `linux/amd64` + `linux/arm64` (works on a Pi)
- [x] `icon.svg` and gallery `1.jpg` / `2.jpg` / `3.jpg` at 1920x1080
- [x] MIT licensed, `SECURITY.md` and `LICENSE` present upstream
- [x] Starts and scans with zero configuration
- [ ] `submission:` field — will update to this PR's URL as soon as it has one

> MyNeS is a scanner. The description tells users to only scan networks they own.


fxerkan:
CI is green now. Fixed in [the latest commit](https://github.com/getumbrel/umbrel-apps/pull/5955/commits):

- `releaseNotes` and `gallery` cleared — this is a new app, not an update.
- `mynes/data` and `mynes/config` committed so the `${APP_DATA_DIR}` bind mounts have something to bind to on first start.
- Icon and gallery images removed from the repo and moved here.

## Icon and gallery

MyNeS icon

| | |
|---|---|
| Device inventory and home plan | Network graph |
| Monitoring and alerts | |

Source SVG: https://github.com/fxerkan/my_network_scanner/blob/main/assets/logo.svg
Gallery JPGs (1920x1080): https://github.com/fxerkan/my_network_scanner/tree/main/assets/store

## On the two remaining warnings

**`network_mode: host` (no `app_proxy`).** MyNeS sends raw ARP frames and listens for mDNS/SSDP multicast. Neither crosses a bridge network, so behind `app_proxy` the scanner would enumerate the Docker bridge instead of the user's LAN — it would install, start, look healthy, and find nothing. This follows the same pattern as the existing `home-assistant` app, which also skips `app_proxy` for host networking.

The direct consequence is that there is no auth in front of it, which is why `defaultUsername`/`defaultPassword` are empty. The app assumes a trusted LAN; that assumption is documented upstream in [SECURITY.md](https://github.com/fxerkan/my_network_scanner/blob/main/SECURITY.md).

**`NET_ADMIN` + `NET_RAW`.** `NET_RAW` to build ARP frames, `NET_ADMIN` to read interface state. The image runs as `USER scanner` (uid 1000), **not** root, and is **not** a privileged container. If the capabilities are absent it does not crash — it degrades to a ping sweep plus the OS ARP cache and reports the gap via `GET /api/capabilities`.

If you would rather ship it behind `app_proxy` and accept that discovery is limited to the bridge, I can push that instead — but I would rather not, since it makes the app misleading rather than merely limited. Happy to take any other direction on wording or config.

PR #5956, opened 2026-08-05
yuvomi: update to v1.87.0

ulsklyc:
Updates Yuvomi from 1.77.0 to 1.79.0, covering three upstream releases (1.78.0, 1.78.1, 1.79.0).

Two files change:

- `yuvomi/umbrel-app.yml` — `version` and `releaseNotes`
- `yuvomi/docker-compose.yml` — image tag and digest

The image is pinned to the multi-arch index digest of the published release:

```
ghcr.io/ulsklyc/yuvomi:1.79.0@sha256:bf447f4dab7cbaa72d227125357b66cf56827b76d769f2800a9525d375560624
```

Nothing else in the manifest is touched, so the port, gallery and category set during review are preserved.

**What's new for users**

- A default calendar for new events, chosen per household member, so a family calendar no longer has to be picked by hand on every event.
- Filipino as the 24th interface language; 25 more currencies (the Americas plus the Philippine peso, 54 in total) and 64 region presets, each with its own date and time format.
- CalDAV reminder lists now keep their subtask relationships instead of importing subtasks as unrelated top-level tasks, and the task filter accepts several priorities, states or people at once.

Full notes: https://github.com/ulsklyc/yuvomi/compare/v1.77.0...v1.79.0

### DIFF
--- a/yuvomi/docker-compose.yml
+++ b/yuvomi/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
-    image: ghcr.io/ulsklyc/yuvomi:1.77.0@sha256:726ae06111ac05503162e0c4ffb48402c8039062ee99e0ba5dec6112ac87b15e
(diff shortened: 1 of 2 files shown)


nmfretz:
Excellent, thanks @ulsklyc. I updated the release notes and tested the upgrade, so we’re good to go.

PR #5958, opened 2026-08-05
Fix n8n Form Trigger behind Umbrel app proxy

DiSonDS:
## Type

App update

## App

App ID: n8n
Upstream project: https://github.com/n8n-io/n8n
Version: 2.33.3-1 (upstream remains 2.33.3)

## Summary

n8n Form Trigger URLs (`/form/*`, `/form-test/*`) were blocked by Umbrel app proxy authentication, so public forms required an Umbrel login cookie and failed for external users.

This packaging revision keeps upstream n8n at 2.33.3 and extends `PROXY_AUTH_WHITELIST` from:

`/webhook-test/*,/webhook/*`

to:

`/webhook-test/*,/webhook/*,/form-test/*,/form/*`

The n8n editor stays behind Umbrel auth. Webhook paths are unchanged. Manifest version is bumped to `2.33.3-1` so existing installs see the update.

## Verification

Umbrel testing performed:
Updated the installed n8n package to 2.33.3-1 with the extended PROXY_AUTH_WHITELIST. Verified Form Trigger production and test URLs (`/form/*`, `/form-test/*`) open and submit without an Umbrel login. Confirmed the n8n editor still requires Umbrel authentication, and existing webhook paths continue to work.

Environment tested:
- [x] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [x] amd64
- [ ] arm64

None expected. Image tag/digest unchanged.

## Notes

- No permission, dependency, credential, or image changes.
- `PROXY_AUTH_WHITELIST` treats form paths as public at the Umbrel proxy layer; protect forms in n8n via Basic Auth when needed ([Form Trigger docs](https://docs.n8n.io/integrations/builtin/core-nodes/n8n-nodes-base.formtrigger/)).

### DIFF
--- a/n8n/docker-compose.yml
+++ b/n8n/docker-compose.yml
@@ -5,7 +5,8 @@ services:
     environment:
       APP_HOST: n8n_server_1
       APP_PORT: 5678
-      # These webhook and form endpoints are protected using Basic or Header Auth
+      # Allow external clients to reach n8n webhook and form endpoints
+      # Authentication can be configured on the corresponding n8n nodes
       # More details here: https://docs.n8n.io/integrations/builtin/core-nodes/n8n-nodes-base.webhook/
       # https://docs.n8n.io/integrations/builtin/core-nodes/n8n-nodes-base.formtrigger/
       PROXY_AUTH_WHITELIST: "/webhook-test/*,/webhook/*,/form-test/*,/form/*,/form-waiting/*"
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for this @DiSonDS!

n8n had moved on a couple of versions since this was opened, so I rebased it onto the latest `master` and carried the package revision forward to `2.33.5-1`.

I also added `/form-waiting/*` to the whitelist. n8n uses that for multi-page forms and completion pages, so without it the first page could work while later pages still hit Umbrel auth. I cleaned up the comments and release notes to match.

Tested the update and form flow, including multi-page forms, and everything’s working well. Thanks again for catching this. Going live!

PR #5992, opened 2026-08-15
nzbfast: add app

nzbfast:
## nzbfast 1.1.2

[nzbfast](https://github.com/nzbfast/nzbfast) is a speed-focused Usenet (NZB) downloader written in Rust: one small binary, low memory footprint, and it saturates a fast line with a handful of connections. Articles are verified as they arrive, so most jobs finish with no separate verify pass, and repair and unpack run only when they are needed.

- **Upstream project:** https://github.com/nzbfast/nzbfast (GPL-3.0)
- **Website / docs:** https://nzbfast.github.io/nzbfast/ · [manual](https://nzbfast.github.io/nzbfast/MANUAL.html)
- **Image:** `nzbfast/nzbfast:1.1.2@sha256:f427069f3ef33b9e7837ed84f8b7429d3d762099c9404e92d1216303c9f7a6ba` - the project's own image, published to Docker Hub and ghcr from the same build, `linux/amd64` + `linux/arm64`, publicly pullable.

### Packaging notes

- **app_proxy** fronts the web UI; nothing is published with raw `ports:`.
- **`PROXY_AUTH_WHITELIST: "/api*,/newznab/api*,/jsonrpc*"`** - these are the SABnzbd-compatible API, the Newznab search endpoint and the NZBGet JSON-RPC facade. Sonarr/Radarr/Prowlarr and the mobile remotes talk to those and cannot present an Umbrel session cookie, so they authenticate with nzbfast's own API key instead. The container generates a key on first run and **refuses to start keyless on a published port**, so the whitelisted paths are never unauthenticated. This mirrors what the SABnzbd package does with `/api*`.
- **Persistence:** `/config` is the install - settings, the API key, the queue, history and the index database - so it is a bind mount under `${APP_DATA_DIR}/data`, together with the watch folder. Both source directories are committed with `.gitkeep`.
- **Downloads** use the shared Downloads mount, hence `STORAGE_DOWNLOADS` in `permissions:`.
- **No host access:** no Docker socket, no privileged mode, no host networking, no device mounts.
- **PUID/PGID 1000:** the image's entrypoint starts as root only to adopt the ownership of the bind mounts, then drops to that uid/gid before exec'ing the daemon. That is the supported path upstream exposes, so `user:` is not forced.

### Default credentials

None. There is no login; the app opens straight to the dashboard behind Umbrel auth. The API key that Sonarr/Radarr need is generated per install and shown under Settings → Security, so `defaultUsername`/`defaultPassword` are empty rather than fabricated.

### Testing

`npm run lint:apps -- nzbfast --check-images` reports no issues, and `git diff --check` is clean.

I do not have an umbrelOS device, so the install has **not** been exercised through Umbrel itself. What has been verified is the container this package runs, deployed from the same compose shape on Linux: it reaches its healthcheck, the dashboard serves, `mode=version` answers, and `mode=queue` is refused with `API Key Required` until a key is presented. Happy to make any changes a maintainer with a device finds.

Screenshots and the logo live at https://nzbfast.github.io/nzbfast/ and in the project repo; say the word and I will attach whatever sizes the App Store assets need. No gallery or icon files are committed here.

### DIFF
--- a/nzbfast/docker-compose.yml
+++ b/nzbfast/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       PROXY_AUTH_WHITELIST: "/api*,/newznab/api*,/jsonrpc*" # leakcheck-allow-synthetic - URL paths, not a secret
 
   web:
-    image: nzbfast/nzbfast:1.1.2@sha256:f427069f3ef33b9e7837ed84f8b7429d3d762099c9404e92d1216303c9f7a6ba
+    image: nzbfast/nzbfast:1.3.1@sha256:7a0cd161a19bd025871342d4ae6e2323fde87e664e2d845bff9bb10fbb6d3da7
     restart: on-failure
     stop_grace_period: 1m
     environment:
--- a/nzbfast/docker-compose.yml
+++ b/nzbfast/docker-compose.yml
@@ -1,0 +1,32 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: nzbfast_web_1
+      APP_PORT: 6789
+      # The SABnzbd-compatible API, the Newznab search endpoint and the NZBGet
+      # JSON-RPC facade are what Sonarr/Radarr/Prowlarr and the mobile remotes
+      # talk to, and none of them can present an Umbrel session cookie. They
+      # authenticate with nzbfast's own API key instead, which the container
+      # generates on first run and refuses to start without.
+      PROXY_AUTH_WHITELIST: "/api*,/newznab/api*,/jsonrpc*" # leakcheck-allow-synthetic - URL paths, not a secret
+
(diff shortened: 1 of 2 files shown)


nzbfast:
Refreshed this to the current release - it was pinned at 1.1.2 and we are now on 1.3.1, and I would rather a reviewer was not looking at a five-release-old submission.

The image is still pinned by manifest list digest, verified against the registry before pushing: the `1.3.1` tag resolves to `sha256:7a0cd161...` and that digest returns a valid multi-arch index for amd64 and arm64. Nothing else in the submission changed - same compose, same permissions, same ports.

No rush from our side, and I know the queue is long. Just checking whether there is anything you need from me to move it along - happy to change whatever does not fit your conventions.

PR #5996, opened 2026-08-17
Update Bassin to 2.1.3

blockdyor:
## Type

App update

## App

App ID: bassin  
Upstream project: blockdyor/bassin  
Version: 2.1.3

## Summary

This update improves timestamp handling in the Bassin UI by showing precise UTC timestamps in tooltips for pool/user last-update and last-share values. It also updates the hover cursor behavior where tooltips are available so the interaction is more obvious to users.

The underlying ckpool service is unchanged; this is a UI-focused improvement to make timestamps clearer and easier to interpret.

## Verification

Umbrel testing performed:

Environment tested:
- [ ] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [ ] amd64
- [x] arm64

- None expected

## Notes

- No permission changes
- No dependency changes
- No migrations
- No default credential changes
- No known caveats


nmfretz:
Excellent, thanks @blockdyor. Going live.

FYI, I noticed the tooltip commit appears to have also replaced the count-aware `User` / `Users` and `Worker` / `Workers` labels with hard-coded values, which can result in states like `0 User` and `1 Workers`. This may have been incidental, but it’s worth checking for the next UI build. Nothing needed for this PR.

PR #6000, opened 2026-08-19
Update Alby Hub to v1.24.0

hermes-alby:
## Summary

- update Alby Hub from v1.23.0 to v1.24.0
- pin the v1.24.0 multi-architecture OCI image digest
- refresh the Umbrel update notes with user-visible changes

Upstream release: https://github.com/getAlby/hub/releases/tag/v1.24.0

## Verification

- `npm ci`
- `npm run lint:apps -- albyhub --check-images`
- `git diff --check`
- verified the pinned OCI index digest and its `linux/amd64` and `linux/arm64` manifests against GHCR

An Umbrel device/local umbrelOS runtime was not available, so the installed-app update path was not exercised.

### DIFF
--- a/albyhub/docker-compose.yml
+++ b/albyhub/docker-compose.yml
@@ -5,7 +5,7 @@ services:
       APP_HOST: albyhub_server_1
       APP_PORT: 8080
   server:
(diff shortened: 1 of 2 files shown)


nmfretz:
Tested and working great guys!

PR #6006, opened 2026-08-21
Add Mastodon app

Pkaulsjs:
��Adds [Mastodon](https://joinmastodon.org) v4.7.0 as an Umbrel app, resolving #560.

## What's packaged
- **web** (Puma), **streaming** (official \mastodon-streaming\ image - required since 4.x split it out), **sidekiq**, **postgres 16**, **redis 7**, fronted by **nginx** via \^Gpp_proxy\
- All images pinned by digest, multi-arch (amd64 + arm64)
- First-boot \init\ container: runs migrations, generates VAPID keys, creates the \^Gdmin\ Owner account. Credentials are shown in the app manifest (\defaultUsername\/\defaultPassword\) so users can log in immediately
- Secrets (SECRET_KEY_BASE, OTP_SECRET, Active Record encryption keys) are umbrel-derived and stable across restarts/upgrades; VAPID keys persist in \\/data/secrets\
- Data persisted: postgres DB, redis, media uploads

## Testing done locally (Docker, full stack)
- Official linter passes (hook committed with exec bit)
- \init\ completes exit 0; all 7 services healthy
- Landing page serves 200 via nginx with correct Host handling; \/api/v1/instance\ returns valid JSON; login page loads; streaming health endpoint OK
- Verified no http->https redirect loop: nginx passes through the upstream \X-Forwarded-Proto\ instead of hardcoding it

## Notes for reviewers
- \LOCAL_DOMAIN\/\WEB_DOMAIN\ use \\\ per repo convention
- Federation beyond LAN requires a real domain (documented in the manifest description)
- Admin password is a documented default (app is LAN-only behind umbrelOS); users are pointed to change it in Preferences -> Account -> Security

## Screenshots

Live from a local end-to-end run of this package:

Landing page
Public timeline with first posts
Login page


Pkaulsjs:
Follow-up: I ran a clean-room first-boot test (fresh data dir, empty DB) and caught + fixed two bugs in the init flow that would have broken fresh installs:

1. The \dmin-exists\ check returned success on both branches (\exit 0 if ...\ in rails runner), so the admin account was never created.
2. Compose interpolated \\\ to empty before bash saw it.

Both fixed and re-verified from scratch: init exits 0, admin account is created with the documented credentials, landing/login pages serve 200. Commit pushed to this PR.

Pkaulsjs:
Self-review fix pushed: added missing `mastodon/exports.sh`.

The compose references `${APP_MASTODON_DB_PASS}`, `SECRET_KEY_BASE`, `OTP_SECRET`, and the three ActiveRecord encryption keys via `derive_entropy`, but the exports file itself was accidentally missing from the branch (local testing had masked this since test env vars were exported manually). On a real install every one of those would have rendered empty - blank DB password and blank app secrets.

All six are now defined per the packaging guide's derive_entropy pattern. CI should re-run green; happy to adjust anything else.

Pkaulsjs:
Second round of fixes pushed after deeper end-to-end testing (fresh DB, fresh state each time):

- **Removed the `env_file` mechanism entirely.** Docker compose reads `env_file` at parse time, so values generated during first boot never reach the runtime containers — confirmed empirically after chasing phantom "missing VAPID" symptoms. All secrets now come from `exports.sh` (`derive_entropy`) only, which is sourced before compose runs.
- **Dropped VAPID/web-push keys.** The only feature needing them is push notifications; documented tradeoff, one less first-boot dependency.
- **Fixed the app-root paths**: the image serves from `/opt/mastodon`, not `/mastodon` — the uploads bind mount was silently wrong before this.
- **Disabled Mastodon's hardcoded `config.force_ssl`** via a boot-time initializer written by the web service (the image runs as uid 991 but `config/initializers` is root-owned, so the service starts as root, installs the override, then serves). Without this, every page 301s to unreachable `https://<device>.local` on LAN installs.
- **Restored the admin password-sync step** that a previous refactor dropped.

Verified on a from-scratch stack: root/about/login all 200, `/api/v1/instance` returns the instance, streaming health check passes through the nginx route, all 7 services stable. Happy to adjust anything.

PR #6012, opened 2026-08-22
hermitstash: update to 1.15.3

dotCooCoo:
## Type

App update

## App

App ID: hermitstash
Upstream project: https://github.com/dotCooCoo/hermitstash
Version: 1.15.3 (from 1.14.2)

## Summary

Version bump, plus a correction to this app's own `TRUST_PROXY` setting.

The image carries a fix for CVE-2026-14456, a high-severity denial of service in
OpenSSL through unbounded memory. It was disclosed after the previous image was
built, so that image scanned clean when it shipped and does not now. Updating is
the whole of the fix.

`TRUST_PROXY` was set to `"true"`. The app does not accept that — the setting
takes IP addresses or CIDR ranges, and a value that is not an address is
discarded at boot with a line naming it. With the setting discarded, every
request was attributed to `app_proxy` rather than to the caller: one rate-limit
bucket shared by all clients, the proxy's address in the audit log, and the
app's admin network fence matched against the proxy instead of the admin. It now
names `10.21.0.0/16`, which is `umbrel_main_network` as declared in umbrel's own
compose (`subnet: $NETWORK_IP/16`, `NETWORK_IP=10.21.0.0`).

Four of the fixes in this range only ever went wrong on umbrel, because the app
is reached over plain HTTP through the app proxy here rather than over a domain
of its own:

- Public pages answered 403 to search-engine crawlers. The bot filter refused any
  request that omitted an `Accept-Language` header, and every major crawler omits
  it — Google documents that Googlebot does. A browser was served normally, so
  the refusal was invisible unless you looked at what a crawler received. Uptime
  monitors and link previewers were turned away the same way. A request carrying
  no `User-Agent` at all is still refused.
- Signing out could not clear the session cookie. The cookie that ends a session
  was marked HTTPS-only, and a browser refuses one of those over plain HTTP, so
  it was discarded and the old cookie stayed. The session itself was always
  destroyed, so nothing remained usable, but the cookie lingered.
- The container reported a valid timezone as invalid on every boot, including the
  `Etc/UTC` this app sets by default. It checked the value against files the base
  image does not carry. The zone was applied correctly throughout; only the
  warning was wrong.
- Reaching the app on a second hostname refused every sign-in, upload and settings
  change while ordinary pages kept loading. Additional origins can now be
  declared.

## Verification

Umbrel testing performed:

Ran the **published** `1.15.3` image — the digest this PR pins — under this
compose file's own constraints: `user: "1000:1000"`, `cap_drop: ALL`,
`security_opt: no-new-privileges:true`, `shm_size: 256m`, plain HTTP,
`TZ=Etc/UTC`, `TRUST_PROXY=10.21.0.0/16`, `RP_ORIGIN=http://<device>:3081`, with
the three volume mounts owned by uid 1000 as umbrel provides them.

- `/health` returns 200
- boot log carries no `TRUST_PROXY entry ignored` line and no invalid-timezone
  warning
- a Googlebot user agent sending no `Accept-Language` gets 200
- an ordinary browser user agent gets 200
- a deny-listed automation user agent gets 403 — the filter still does its job
- `Set-Cookie` carries `HttpOnly; SameSite=Lax` and **no** `Secure`, so the
  cookie is accepted over this deployment's plain-HTTP hop
- no `Strict-Transport-Security` header, which a user agent would ignore over
  plain HTTP anyway

The upstream project also carries an automated suite that asserts this
deployment profile as a whole — trusted proxy, cookie flags, HSTS, the crawler
filter, accepted origins and timezone — including a check that compares its
fixture against this compose file directly, so the two cannot drift apart
silently.

Environment tested:
- [ ] Umbrel device
- [ ] Not runtime tested

The image was runtime tested, but under Docker with this compose file's
constraints replicated rather than on umbrelOS. None of the three boxes
describes that exactly, so none is ticked.

Architecture tested:
- [x] amd64
- [ ] arm64

The pinned digest is an OCI image index carrying `linux/amd64` and `linux/arm64`;
only amd64 was exercised locally.

`.tools/lint-apps.mjs` reports 0 errors and 1 warning:

```text
WARNING access.security_opt hermitstash/docker-compose.yml:16
  Service `web` sets security options; justify this in the PR.
```

`security_opt: no-new-privileges:true` is pre-existing hardening, already present
in the merged listing and unchanged by this PR. It prevents the container gaining
privileges through setuid binaries during execve. The app runs as a non-root user
with all capabilities dropped and needs no privilege escalation.

## Notes

No migration, no new dependency, no default credentials. Existing installs keep
their data directory; the container's runtime user and group ids are unchanged,
so an upgrade changes no ownership on disk.

Operators who set `TRUST_PROXY` themselves in a custom install should use an
address or CIDR rather than `true`.

### DIFF
--- a/hermitstash/docker-compose.yml
+++ b/hermitstash/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 3000
 
   web:
-    image: ghcr.io/dotcoocoo/hermitstash:1.14.2@sha256:46f5c4ba6ed9ddeff93963cf60084c4c9b531918dec9f7b9eea56193786cf9f8
+    image: ghcr.io/dotcoocoo/hermitstash:1.15.3@sha256:9069415e620760d860275f25b6e2ac0069a624a7552119df0dfd8d33faae0dcd
     user: "1000:1000"
     init: true
     restart: on-failure
@@ -35,7 +35,14 @@ services:
       UMASK: "022"
       TZ: "Etc/UTC"
       NODE_ENV: production
-      TRUST_PROXY: "true"
+      # The addresses app_proxy connects from, as a CIDR. The app rejects
+      # "true" and discards the setting, which attributes every request to
+      # app_proxy: one rate-limit bucket shared by all clients, the proxy's
+      # address in the audit log, and the admin network fence matched against
+      # the proxy rather than the admin. 10.21.0.0/16 is umbrel_main_network,
+      # declared in umbrel's own compose as `subnet: $NETWORK_IP/16` with
+      # NETWORK_IP=10.21.0.0.
+      TRUST_PROXY: "10.21.0.0/16"
       # Public origin — used to build share URLs in upload responses and to
       # pin the WebAuthn relying-party origin. ${DEVICE_DOMAIN_NAME} is the
       # umbrel-supplied hostname (e.g. umbrel.local); the app reaches it
--- a/hermitstash/umbrel-app.yml
+++ b/hermitstash/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: hermitstash
 category: files
 name: HermitStash
-version: "1.14.2"
+version: "1.15.3"
 tagline: Post-quantum encrypted self-hosted file sharing
 description: >-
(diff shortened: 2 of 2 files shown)


nmfretz:
Thanks @dotCooCoo. Going live.

PR #6015, opened 2026-08-23
Update EE Gateway to 0.10.8

jmarbach:
Quality-of-setup release. The location step gains a map picker (click to drop the pin, drag to fine-tune) and a one-tap "Use GPS position" shortcut when a GPS dongle has a live fix. The dashboard adds a standing-by state: a healthy scanner that hasn't heard a beacon yet says "Radio check: passing" and explains the silence. The worker now reports its platform and a prefixed version in heartbeats, so fleet stats read one format across Umbrel, iOS, and Android.

Image digest bumps and release notes only. The Leaflet map library is vendored into the UI image, and map tiles load from openstreetmap.org in the operator's browser only, so the containers gain no new network dependencies.

Full changelog: https://github.com/Encrypted-Energy/gateway/releases/tag/v0.10.8

### DIFF
--- a/ee-gateway/docker-compose.yml
+++ b/ee-gateway/docker-compose.yml
@@ -34,7 +34,7 @@ services:
   # extra download, and it is the only place root is used. This is NOT privileged
   # mode; the app containers themselves stay non-root.
   init_perms:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks as usual @jmarbach

PR #6019, opened 2026-08-25
Add Recat

tx-joshg:
## App

**Recat** — self-hosted transaction categorization and review queue for QuickBooks Online.
Version `0.1.5`.

- Upstream project: https://github.com/tx-joshg/recat-qbo
- License: AGPL-3.0-only
- Support: https://github.com/tx-joshg/recat-qbo/issues

A QuickBooks bank rule auto-adds bank feed items to a holding account. Recat syncs that account, a team categorizes in the web UI without anyone needing a QuickBooks login, and each transaction is written back with its category, splits or transfers. Every write is recorded in an append-only audit log before it happens, and dry-run mode is on by default until the operator turns it off.

## Images

Built from the upstream repo by its own GitHub Actions workflow and published to GHCR. Pinned by **index digest**, so `docker pull` still resolves per architecture:

| image | digest | platforms |
|---|---|---|
| `ghcr.io/tx-joshg/recat-qbo` | `sha256:b1c3d9d2…` | `linux/amd64`, `linux/arm64` |
| `ghcr.io/tx-joshg/recat-qbo/receipt-extractor` | `sha256:e3685b43…` | `linux/amd64`, `linux/arm64` |
| `postgres:17-alpine` | `sha256:742f40ea…` | upstream |

## Setup behaviour and credentials

- `defaultUsername: admin@recat.local` with `deterministicPassword: true`. The package sets `LOCAL_ADMIN_PASSWORD=${APP_PASSWORD}`, and the first-run wizard **requires that password** before it will create the first administrator — an instance reachable by others cannot be claimed by whoever finds it first.
- No outbound email. Local sign-in is the way in; magic links are written to the app's logs rather than returned to unauthenticated callers.
- **`PROXY_AUTH_ADD: 'false'`** — needed, not incidental. With Umbrel's dashboard auth on, `app_proxy` redirects to `http://127.0.0.1:2000`, which nothing outside the box can reach, so the QuickBooks OAuth callback returns 502 through a TLS front and real books can never be connected. Measured on a device: 200 on loopback, 502 through Tailscale Serve, 502 on `/auth/qbo/callback`. Recat authenticates its own requests (sessions, magic links, a rate-limited local admin password, per-company roles).
- **`TRUSTED_PROXY_HOP: 'true'`** — trusts the immediate peer as a reverse proxy **only when that peer is on a private network**. Sound here because this compose publishes no `ports:` for the server, so `app_proxy` is the only route to it. Without it every caller shares `app_proxy`'s address, so rate limits are deployment-wide and anyone can lock the owner out of sign-in.
- `TRUSTED_PROXY_IPS` deliberately unset — `app_proxy`'s address is Docker-assigned and not exposed as a variable.
- Postgres uses a **bind mount** under `${APP_DATA_DIR}` rather than a named volume, so Umbrel's backup/restore and update cycle covers it.
- The receipt extractor is included but **receipt extraction is off by default**; enabling it sends receipt images to whichever provider the operator configures. It is not published to the host.
- No host access, no privileged containers, no device passthrough. The extractor runs `read_only` with `cap_drop: ALL` and `no-new-privileges`.

## Connecting real QuickBooks needs HTTPS

Intuit will not accept a plain-http redirect URI for a production app, so `umbrel.local` cannot be registered with them. An operator who fronts their Umbrel with TLS (Tailscale Serve, Cloudflare Tunnel) sets that address in the wizard and registers the redirect URI it produces. Without one the built-in demo QuickBooks works fully and real books will not connect. This is stated in the app description so it is not discovered at the Intuit step.

## Testing performed

**On real Umbrel hardware**, by a contributor running the package on their own device:
- Installs and boots clean from the package directory
- The first-run wizard completes and the displayed app password signs in, tested against an empty database
- **Real QuickBooks connects through a TLS front** — signed in via Tailscale Serve, `POST /api/companies/<id>/test-connection` returned `ok: true`, `environment: production`, `mode: quickbooks`, which is a live Intuit call rather than a cached flag. `/auth/qbo/callback` lands on the app instead of 502ing.

**On an arm64 host**, booted from the pinned digests with `derive_entropy` simulated and `app_proxy` replaced by a direct port publish:
- All migrations apply to an empty database and do not re-run on the next boot
- The extractor answers `/healthz` from the app container and refuses connections from the host
- Receipt extraction defaults to off in the created schema
- Data lands in the `APP_DATA_DIR` bind mount and survives a full `down`/`up`, which is what an update does

**Scope note, so the above is not read as more than it is:** the hardware run and the arm64 boot were against the `v0.1.3` images. This submission pins `v0.1.5`, which adds support for localized (non-US) QuickBooks companies. Those exact images have not themselves been booted — `git diff v0.1.3 v0.1.5` is empty for `prisma/migrations`, the server entrypoint, `env`, the `Dockerfile`, and the auth/setup/proxy code, so the boot-relevant surface is unchanged, but that is an argument rather than a run.

## Port and id

`port: 3009` and id `recat` were checked against every manifest in `master` at submission time — 391 apps parsed, both free. (`3001` is taken by `ride-the-lightning`.)

## Assets

`gallery: []` per the packaging guidance for new submissions. Logo and screenshots:

- Logo (512×512 SVG): https://github.com/tx-joshg/recat-qbo/blob/main/client/public/icon.svg
- Queue
- Dashboard
- Reports
- Rules
- Queue, dark
- Queue, mobile

Happy to adjust anything here — including the port, if 3009 is spoken for by something in flight.

### DIFF
--- a/recat/docker-compose.yml
+++ b/recat/docker-compose.yml
@@ -43,7 +43,7 @@ services:
       #
       # Set APP_URL_LOCKED=true instead if this deployment should pin the
       # address and refuse edits from the UI.
-      APP_URL: http://umbrel.local:3009
+      APP_URL: http://umbrel.local:3019
       PORT: 3001
       DATABASE_URL: postgresql://recat:${APP_RECAT_DB_PASSWORD}@recat_db_1:5432/recat
       SESSION_SECRET: ${APP_RECAT_SESSION_SECRET}
--- a/recat/docker-compose.yml
+++ b/recat/docker-compose.yml
@@ -1,0 +1,125 @@
+services:
+  app_proxy:
+    environment:
+      APP_HOST: recat_server_1
+      APP_PORT: 3001
+
+      # Umbrel's dashboard auth defaults on (PROXY_AUTH_ADD defaults to true),
+      # and it only works from the dashboard origin, where the browser carries
+      # that session cookie. Behind a TLS front app_proxy 302s to
+      # http://127.0.0.1:2000, which nothing outside the box can reach.
+      # Measured on the device: 200 on loopback, 502 through Tailscale Serve,
+      # and 502 on /auth/qbo/callback — so the Intuit callback can never land
+      # and real QuickBooks cannot be connected at all.
+      #
+      # Whitelisting just the callback does not help: every UI request would
+      # still 302 to an unreachable address, leaving the app unusable through
+      # the front it needs for real books.
+      #
+      # Recat authenticates its own requests — sessions, magic links, the
+      # rate-limited local admin password, per-company roles — so Umbrel's
+      # layer is redundant here rather than protective.
+      #
+      # First-run used to be the exception: it is unauthenticated by necessity,
+      # since somebody has to create the first account. As of v0.1.3 that is
+      # closed. Creating the first administrator requires the app password this
+      # page displays, and /auth/magic-link neither bootstraps an account nor
+      # returns a sign-in link in its response while LOCAL_ADMIN_* is set —
+      # issued links reach this app's logs only (#53).
+      PROXY_AUTH_ADD: 'false'
+
+  server:
+    image: ghcr.io/tx-joshg/recat-qbo:0.1.5@sha256:b1c3d9d28d41b7f674e5f3823f5ca0e456f90196f5131d83030cf3082f28aa87
+    restart: on-failure
+    environment:
+      # The install's starting address, and the seed for origin checking — not
+      # a lock. An admin changes the public URL from the setup wizard or
+      # Settings, and the stored value then wins; see instanceSettings.appUrl.
+      #
+      # Do not remove this line. Without it the app falls back to its own
+      # localhost default, which is not an origin any Umbrel browser sends, so
+      # the CSRF check rejects the setup wizard itself and the install cannot
+      # be completed. Verified: POST /api/setup/admin returns 403.
+      #
+      # Set APP_URL_LOCKED=true instead if this deployment should pin the
+      # address and refuse edits from the UI.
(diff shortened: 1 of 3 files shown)


tx-joshg:
Lint is green now. Notes on what it flagged, and one correction on my side.

## Port moved 3009 → 3019

The linter was right and my check was wrong. I verified 3009 against every
`umbrel-app.yml` in `master` and found it free — but it is exported by
`samourai-server/exports.sh` as `APP_SAMOURAI_SERVER_DOJO_PORT` and bound by its
`nginx` service, which a manifest-only scan never sees.

Re-checked across manifests, compose port bindings **and** `exports.sh` (562
ports claimed in total). **3019** is the nearest port nothing in the store
claims. `APP_URL` in the compose moved with it.

## `receipt-extractor` security options — justification

Flagged for `read_only`, `cap_drop: ALL`, `no-new-privileges`, `tmpfs`,
`mem_limit` and `pids_limit`. These are all **restrictions**, not grants — the
service gets less than the default, never more.

It exists to parse uploaded receipt images, so it is the one container in this
package that processes untrusted user files. It needs no persistent storage
(`/tmp` as tmpfs is enough), no filesystem writes, and no capabilities. It is
also deliberately **not published to the host** — only the app container reaches
it, over the app network, with a shared token from `exports.sh`.

Receipt extraction is off by default in the schema; an operator turns it on, and
doing so sends images to whichever provider they configure.

If you would rather this package ship without the extractor at all for a first
submission, I am happy to drop it and add it later — it is optional
functionality.

## `postgres:17-alpine` digest

Understood, and the pin is unchanged. The tag moved upstream after we pinned it;
this PR is not trying to update that image.

PR #6053, opened 2026-09-03
Update pogolo to 1.1.5

0xf0xx0:
## Type

App update

## App

App ID: pogolo
Upstream project: github.com/0xf0xx0/pogolo
Version: 1.1.5

## Summary

stole the umbrel-bitcoin webui and tweaked it for pogolo, itll be mirrored out to github.com/0xf0xx0/pogolo-umbrel-webui soon (its on my forge rn)

as to pogolo, just some perf optimizations to keep it the fastest in its class

## Verification

Umbrel testing performed:

Environment tested:
- [ ] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [x] amd64
- [ ] arm64

## Notes

Permissions, dependencies, migrations, default credentials, or caveats: shouldn't break anything, updating locally was smooth

Also, some new gallery assets :3 

home page
insights page
insights page, worker info
settings page

### DIFF
--- a/pogolo/data/pogolo.toml
+++ b/pogolo/data/pogolo.toml
@@ -1,3 +1,5 @@
+# NOTE: backend, listening address, and zmq are already configured in the docker compose
+# and will override anything set here!!!
 [pogolo]
 # optional, required from clients if set
 # password = ''
--- a/pogolo/docker-compose.yml
+++ b/pogolo/docker-compose.yml
@@ -28,7 +28,7 @@ services:
     command: ["/pogolo", "--conf", "/data/pogolo.toml", "--logfile", "/data/pogolo.log"]
 
   webui:
-    image: 0xf0xx0/pogolo-umbrel-webui:v1.x@sha256:3602e2a1e72d250989478d0ffd552f08117ccd53b122516c5eedeb06041da6b2
+    image: 0xf0xx0/pogolo-umbrel-webui:v1.x@sha256:67c0ce5d8fc1d0fb5292f92a9263546de352ecf1a1b109d294a8ab9d0eee1b73
     restart: on-failure
     user: "1000:1000"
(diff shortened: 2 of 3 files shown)


nmfretz:
Thanks @0xf0xx0. The web UI’s arm64 image is being built for the wrong platform because the runtime stage uses [`$BUILDPLATFORM`](https://github.com/0xf0xx0/pogolo-umbrel-webui/blob/c7c22110d79be0fe612c9e1fc5ecf8fc6fe1c2f4/apps/backend/Dockerfile#L62). This leaves any pogolo install on a Raspberry Pi broken because the web UI container can’t start.

Our linter doesn’t catch this because the image manifest advertises arm64 correctly; it can’t tell that the binaries inside are actually amd64. Can you switch this to `$TARGETPLATFORM`, rebuild the image, and re-pin it here? If you have a Pi or another arm64 machine handy, could you give the rebuilt image a quick test too?

0xf0xx0:
i think it was broken entirely D: i re-did dockers qemu support and it seems to be building correctly without any of the `$BUILDPLATFORM` flags. still dont have an arm device to test locally :C

nmfretz:
Thanks @0xf0xx0! I re-tested and arm64 is working now. We’ll go live with this.

I also noticed the [settings validation](https://github.com/0xf0xx0/pogolo-umbrel-webui/blob/cea4702e8ce85ddc1626b4944e58896fe1b01aca/libs/settings/settings.schema.ts#L15) rejects fractional difficulty values like `0.16`, so users with those already configured can’t load or save settings. The pool keeps running. Worth a look in a follow-up, but not holding this up. Thanks again!

0xf0xx0:
i'll look into that, thanks for alerting me :3

PR #6060, opened 2026-09-04
Fix ONLYOFFICE Docs internal address rejected by Nextcloud

rage-proof:
## Type

App update — package fix. No image change; manifest `version` bumped `9.4.0` → `9.4.0-1`.

## App

App ID: `onlyoffice-nextcloud`
Upstream project: https://github.com/ONLYOFFICE/DocumentServer
Version: 9.4.0-1 (image unchanged at 9.4.0)

## Summary

The app points Nextcloud at Document Server by its injected container name,
`onlyoffice-nextcloud_documentserver_1`. Document Server percent-escapes
underscores in the host of every signed URL it returns
([`storage-base.js:179`](https://github.com/ONLYOFFICE/server/blob/master/Common/sources/storage/storage-base.js)),
and Nextcloud's HTTP client rejects a percent escape in a host (guzzle 7.15.2,
GHSA-v5mv-p594-2x33):

```
Error when trying to connect (The request URI host
"onlyoffice-nextcloud%5fdocumentserver%5f1" must not contain a percent escape,
because a handler can decode it and then connect to a host that differs from
the one the request names.) (version 9.4.0.129)
```

On a stock install that means no office previews, failed conversions, `.doc`/
`.xls`/`.ppt` failing to save, and that error on the ONLYOFFICE admin page.

Fix: give the service a hostname with no underscore in it.

```diff
-      - DOCS_INTERNAL_ADDRESS=http://onlyoffice-nextcloud_documentserver_1
+      - DOCS_INTERNAL_ADDRESS=http://onlyoffice-nextcloud-documentserver-1

   documentserver:
     image: onlyoffice/documentserver:9.4.0@sha256:…
+    container_name: onlyoffice-nextcloud-documentserver-1
```

Same resolution mechanism as before, same name shape as the injected one, just
`-` instead of `_`. Reported upstream too, but that fix needs a Document Server
release; this works today.

Manifest `version` is bumped to `9.4.0-1` with matching release notes, so
existing installs are offered the update — without it the fix would reach new
installs only.

## Verification

Umbrel testing performed: not yet. Static checks only —
`npm run lint:apps -- onlyoffice-nextcloud --check-images` and
`git diff --check`. Per `.claude/skills/umbrel-test-app` that is not Umbrel
verification, so the update path still needs a runtime test on a device.

Environment tested:
- [ ] Umbrel device
- [x] Not runtime tested

Architecture tested:
- [ ] amd64
- [ ] arm64

```
WARNING service.container_name onlyoffice-nextcloud/docker-compose.yml:19
  Service `documentserver` sets `container_name`; Umbrel injects container names automatically.
```

`0 error(s)`. Same advisory `searxng`, `adventurelog`, `seafile` and
`rustdesk-server` already carry — `searxng` for this exact reason.

## Notes

Existing installs keep the address stored in Nextcloud's app config and need to
paste the new one once (Admin → ONLYOFFICE → Advanced server settings), or:

```bash
occ config:app:set onlyoffice DocumentServerInternalUrl --value="http://onlyoffice-nextcloud-documentserver-1/"
```

They are broken today regardless. No new permissions or dependencies.

### DIFF
--- a/onlyoffice-nextcloud/docker-compose.yml
+++ b/onlyoffice-nextcloud/docker-compose.yml
@@ -10,12 +10,13 @@ services:
     restart: on-failure
     environment:
       - DOCS_ADDRESS=http://$DEVICE_DOMAIN_NAME:$DOCSERVER_PORT
-      - DOCS_INTERNAL_ADDRESS=http://onlyoffice-nextcloud_documentserver_1
+      - DOCS_INTERNAL_ADDRESS=http://onlyoffice-nextcloud-documentserver-1
       - NEXTCLOUD_INTERNAL_ADDRESS=http://nextcloud_web_1
       - NEXTCLOUD_WEB_URL=http://$DEVICE_DOMAIN_NAME:$APP_NEXTCLOUD_PORT
 
   documentserver:
     image: onlyoffice/documentserver:9.4.0@sha256:e3da62a847b9a5d51a11f73cfea1d9c13c3be3809614490d4edddcf01dcf919b
+    container_name: onlyoffice-nextcloud-documentserver-1 # Avoid underscores in the hostname: Document Server percent-escapes them in the signed URLs it returns
     restart: on-failure
     stop_grace_period: 1m
     ports:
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @rage-proof for tracking this down and putting together the fix! Really appreciate it.

I reproduced the error and pushed a small packaging adjustment to use a network alias. This gives Nextcloud an address without underscores while keeping Umbrel’s generated container name and the existing DNS address.

I also made the required Nextcloud setting change more prominent in the release notes and linked the Guzzle release that introduced the hostname restriction in the Compose comment.

PR #6061, opened 2026-09-05
Update mempool to 3.3.1-hotfix-2 - exclude rebuildable data from backups

RyleaStark:
## Summary
- exclude mempool's persistent backend cache from umbrelOS backups
- exclude the chain-derived MariaDB explorer index from umbrelOS backups
- bump the package to `3.3.1-hotfix-2` so existing installations receive the change
- explain restore behavior in the release notes

## Why
Both excluded paths are derived from Bitcoin Node and Electrs data. Backing them up duplicates rebuildable chain/index data and can consume substantial backup storage.

After a restore, mempool rebuilds these paths from Bitcoin Node and Electrs. Locally accumulated explorer statistics may take time to repopulate. Wallet and node configuration remain owned and backed up by their respective apps.

## Verification
- official image-aware app linter: 0 errors
- exact assertion confirms `backupIgnore` matches `${APP_DATA_DIR}/data:/backend/cache` and `${APP_DATA_DIR}/mysql/data:/var/lib/mysql`
- package directory skeletons remain present for fresh installs/restores
- `git diff --check` passes
- changed-file scope is limited to `mempool/umbrel-app.yml`

## Existing warning
The linter reports that the mutable `mariadb:10.5.12` tag no longer resolves to the digest already pinned by the package. This PR intentionally preserves the existing immutable digest and does not update any images.

### DIFF
--- a/mempool/umbrel-app.yml
+++ b/mempool/umbrel-app.yml
@@ -2,7 +2,7 @@ manifestVersion: 1
 id: mempool
 category: bitcoin
 name: mempool
-version: "3.3.1-hotfix-1"
+version: "3.3.1-hotfix-2"
 tagline: Be your own explorer
 description: >-
   Run your own instance of The Mempool Open Source Project and trust no one.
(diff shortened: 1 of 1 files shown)


nmfretz:
Thanks for this @RyleaStark! Have you by chance tested a backup/restore and confirmed that mempool recovers?

PR #5835, opened 2026-07-03
Add SkyBro

xDeeKay:
## Type

New app

## App

App ID: skybro
Upstream project: https://github.com/xDeeKay/SkyBro
Version: 1.5.6

## Summary

Adds SkyBro, a real-time sky tracker: live aircraft (OpenSky Network ADS-B data), ISS/Hubble/Tiangong satellite pass alerts (n2yo.com), hyperlocal weather (Open-Meteo), and a full astronomy planner (planets, 110 Messier objects, meteor showers, moon phase, Bortle sky quality) via the local `ephem` library.
Sends push alerts via Pushover and/or Discord when a plane enters a configured radius/altitude window or a satellite pass is approaching.

Two-service app (tracker + web dashboard), no database beyond local SQLite, no required environment variables.
First-run browser onboarding detects an approximate home location from the browser's timezone, no SSH or manual config needed.
Every setting is configured through the in-app Settings page and hot-reloaded by the tracker, no container restart needed.
Containers run as a non-root user (UID/GID 1000).

## Verification

Umbrel testing performed: fresh install and an in-place version update both tested through the actual Umbrel app lifecycle (install/uninstall via the Umbrel Apps UI, and a version bump deployed in place).
Verified main functionality end-to-end: live aircraft tracking, alert firing (Pushover + Discord, including photo attachment), satellite pass computation, weather/astronomy data, history logging with recorded flight paths, and settings persistence across restarts.
Also cross-browser/OS tested on Windows 11, iPadOS 17.7 (Safari), Android 12 (Chrome), and SteamOS 3.8 Desktop Mode (Firefox).

Environment tested:
- [x] Umbrel device (Raspberry Pi 5, Umbrel 1.7.3)
- [ ] Local umbrelOS test environment
- [ ] Not runtime tested

Architecture tested:
- [x] amd64
- [x] arm64

Known lint warnings or caveats: `restart: unless-stopped` is used instead of `on-failure` for both services, intentionally: the tracker is meant to run continuously 24/7, and `unless-stopped` ensures it comes back after a host reboot or an unexpected exit, not just an application-level failure.

`npm run lint:apps -- skybro --check-images` was run locally; the `image.pullable` check reported a fetch failure for both images from this test network, independently confirmed via `docker pull`, `docker buildx imagetools inspect`, and a raw curl replicating the linter's own registry API calls that both images are genuinely public and correctly multi-arch (linux/amd64 + linux/arm64).
[...]

PR #5828, opened 2026-07-01
Update BFFless to v0.1.90

toshimoto821:
## Type

App update

## App

App ID: `bffless`
Upstream project: https://github.com/bffless/ce
Version: `0.1.53` → `0.1.90`

## Summary

Updates BFFless CE from v0.1.53 to v0.1.90. Bumps the manifest `version`, refreshes the `ce-nginx`, `ce-backend`, and `ce-frontend` image digests to the v0.1.90 build, and rewrites `releaseNotes` to cover the range.

Highlights across the range:
- **Pipelines & automation:** new `file_delete` handler (multi-object `keys[]` + runtime array expressions), project secrets, presigned direct-to-bucket uploads, `file_serve_handler` improvements (explicit key mode, byte-range streaming, private caching by default), custom response headers via UI/MCP, crypto utils in the `function_handler` sandbox.
- **Proxy rules & domains:** export/import proxy rule sets from the admin UI (with bundled schema deps), multi-method matching, absolute external URLs as path redirect targets.
- **AI:** live Anthropic model fetching from `/v1/models`, pin which deployment alias AI skills load from.
- **Auth & access:** `ENABLE_EMAIL_PASSWORD` flag for OIDC-only sign-in, mint a content-domain session from a project API key, member-accessible user directory.
- **Privacy:** opt-out install telemetry controls.
- **Fixes:** nginx proxy header buffer / body-size fixes, 413 for oversized bodies, GitHub-style markdown viewer, permission-badge and routing fixes.

Full upstream release notes: https://github.com/bffless/ce/releases/tag/v0.1.90

## Verification

Umbrel testing performed: Installed and ran the v0.1.90 package on a **physical Umbrel device** (fresh install) — the app pulls, starts, and opens successfully. Ran `npm run lint:apps -- bffless --check-images` (no issues; image pull + multi-arch manifest checks pass) and `git diff --check` (clean). Confirmed the pinned image digests were built from the exact v0.1.90 CE release commit (`b1c04bd`, "chore(main): release 0.1.90") via image SLSA provenance.

The in-place 0.1.53 → 0.1.90 update path was not separately exercised, but this change is metadata + image-digest only — no compose/env/port/`app_proxy`/persistence changes and no migrations — so the update path is low risk.

Environment tested:
- [x] Umbrel device
- [ ] Local umbrelOS test environment
- [ ] Not runtime tested

Architecture tested:
- [x] amd64 (physical device; interface naming `enp1s0` indicates x86 hardware)
[...]

PR #6065, opened 2026-09-07
Add deepseek-harness

mauricewipf:
## Type

New app

## App

App ID: `deepseek-harness`
Upstream project: https://github.com/deepseek-ai/deepseek-harness
Version: `0.1.2-rc.1`

## Summary

DeepSeek Harness (`dsh`) is an open-source AI agent harness (“everything is a plugin”) with a browser Web UI: configure a model in **Settings → Models**, pick a workspace, and run tasks that read/write workspace files, run commands, and maintain plans.

Upstream does not publish a Docker image, so the app runs on a maintained wrapper image that installs the pinned upstream npm CLI (`@deepseek-ai/dsh`):
`ghcr.io/mauricewipf/deepseek-harness-umbrel:0.1.2-rc.1` — multi-arch (`linux/amd64` + `linux/arm64`), pinned by manifest-list digest, built by a small GitHub Actions workflow in https://github.com/mauricewipf/umbrel-app-store (`images/deepseek-harness`, workflow `build-deepseek-harness.yml`).

Three packaging mechanics are deliberate and documented in the compose/package comments; all are structured as config-only overlays (no upstream fork):

1. **All-interfaces bind via composition patch.** Upstream's CLI hard-rejects `--host 0.0.0.0` (“intentionally not supported yet for safety”), while its webserver schema allows the `0.0.0.0` value. The package applies a `dsh --patch` overlay (`umbrel.patch.yml.template`) that sets the `webserver` row to `host: 0.0.0.0` so Umbrel's app_proxy can reach the container.

2. **Token-aware proxy sidecar (no log-scraping UX).** Upstream's browser auth mints a random per-process `?token=…` URL printed only to stdout; browsers must open it once before a 30-day session cookie is set. A small stdlib-Python proxy (`proxy.py.template`, pinned `python:3.13-slim` image) reads the token from the app's stdout tee and performs the documented exchange on behalf of visitors who already passed Umbrel's `app_proxy` auth: anonymous `GET /` gets a 302 to `/?token=…`, dsh mints the cookie, everything else (HTTP + WebSocket relay) is proxied transparently. dsh's own auth is not weakened — the proxy merely performs the token step the operator would otherwise copy from logs on every rotation.

3. **Proxy-side trust-fence normalization + client `ownsHost` shim.** Upstream gates the `/api` host fence and *all* settings surfaces (Models, Plugins, General documents) behind the assumption “non-loopback page ⇒ shared/untrusted machine”, which makes the models list and RPC unusable under any reverse proxy (and even behind Umbrel's `app_proxy` on `umbrel.local`, LAN IP, Tailscale, or custom domain → HTTP 403, “Disconnected”). Upstream offers no opt-out (tracked upstream as discussions #5828, #5631, #5835). The package therefore:
   - normalizes `Host` to a fixed loopback authority and drops `Origin` in the proxy (the fence is redundant behind Umbrel auth; cookie authority stays constant across URLs);
   - ships a tiny index-tap plugin (`umbrel-loopback-override.mjs.template`) that sets the documented client transport hook `window.__DSH_TRANSPORT__ = { ownsHost: true }` before boot, which is upstream's own stand-in for “the operator's own machine”. All server-side gates (Host trust fence, browser-session cookie, Umbrel account auth) are unchanged; the shim only unlocks the operator's own authenticated browser session for the settings surfaces.

## Verification

Umbrel testing performed:

- Fresh install through umbreld on an Umbrel device; app reaches `ready`; opens via Umbrel home screen / app_proxy port 3090.
- **No-token flow:** anonymous browser-equivalent request → proxy 302 → dsh 303 → UI loads; `dsh-auth` session cookie set; **Settings → Models** lists the provider directory; agent UI connected (WebSocket stream up). (Previously reproduced the failure mode the fix addresses: `HTTP 403` on all `/api` RPC + “Disconnected”.)
- Foreign-authority probes (`Host: weird.test:…` + matching `Origin`) through the proxy return `200` on `/api/llm/listProviders`, proving the fence normalization.
- **Update path:** multiple package updates applied on the device during iteration (compose/template changes converge); verified again after the final proxy fix.
- **Restart + persistence:** token rotates on restart; pre-restart cookie still valid (credential secret persists in the `DSH_HOME` bind mount); workspace and dsh home persist across restarts (`data/` bind mounts + `pre-start` ownership fix).
- Browser verification by owner on `umbrel.local:3090`: Web UI loads with no token step; Models page loads.

Environment tested:
- [x] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [x] amd64
- [x] arm64 — image index verified multi-arch (`amd64` + `arm64` present); runtime tested on amd64 only, no arm64 device available.

- `npm run lint:apps -- deepseek-harness --check-images` → **no issues**; `git diff --check` clean.
- Upstream is a “developer preview” with advertised breaking changes; packaged version is the exact published npm release `0.1.2-rc.1`.
- The token gate and loopback-only settings gate have **no official upstream opt-out**; the proxy/shim approach is documented in package comments and tracked upstream (discussions #5828 “provide an opt-out for the mandatory browser token auth”, #5631, #5835). If upstream ships a supported opt-out, the proxy can be reduced to a plain forwarder.

## Notes

- No dependencies; no permissions requested; no raw host ports beyond the app_proxy fronted `3090`; app runs as `1000:1000`; no privileged mode, no host networking, no Docker socket, no device mounts.
- No default credentials: Umbrel auth + token-injecting proxy → user lands directly in the Web UI; the dsh browser cookie persists 30 days and survives restarts.
- Umbrel will create the final App Store gallery assets from the screenshots below.

**Screenshots:** see https://github.com/getumbrel/umbrel-apps/pull/6065#issuecomment-5574828541 (Web UI, active session, settings, Settings → Models, agent mode picker).

**App logo (reference):**

deepseek-logo

Source: https://avatars.githubusercontent.com/u/148330874?v=4

### DIFF
--- a/deepseek-harness/docker-compose.yml
+++ b/deepseek-harness/docker-compose.yml
@@ -1,0 +1,42 @@
+version: "3.7"
+
+services:
+  app_proxy:
+    environment:
+      APP_HOST: deepseek-harness_proxy_1
+      APP_PORT: 8091
+
+  proxy:
+    # Token-aware reverse proxy: supplies dsh's per-process launch token so
+    # users authenticated by Umbrel never copy it from logs.
+    image: docker.io/library/python:3.13-slim@sha256:9d2e5553305c7c7b0097999bb17187c69b921ccd6bc9d40e4bb5ebe652c00285
+    command: ["python3", "/app/proxy.py"]
+    restart: on-failure
+    user: "1000:1000"
+    init: true
+    environment:
+      PROXY_PORT: "8091"
+      DSH_HOST: app
+      DSH_PORT: "3080"
+      DSH_LOG_FILE: /logs/dsh.log
+    volumes:
+      - ${APP_DATA_DIR}/proxy.py:/app/proxy.py:ro
+      - ${APP_DATA_DIR}/data/logs:/logs:ro
+
+  app:
+    image: ghcr.io/mauricewipf/deepseek-harness-umbrel:0.1.2-rc.1@sha256:f1a0521e569e4245b2c704bd58b03dc8896577453f63d24128f83da5a0ae067d
+    restart: on-failure
+    user: "1000:1000"
+    init: true
+    environment:
+      DSH_HOME: /data/.dsh
+      DSH_WORKSPACE: /data/workspace
+      DSH_PORT: "3080"
+      DSH_PATCH_FILE: /app/umbrel.patch.yml
+      DSH_LOG_FILE: /logs/dsh.log
+    volumes:
+      - ${APP_DATA_DIR}/umbrel.patch.yml:/app/umbrel.patch.yml:ro
+      - ${APP_DATA_DIR}/umbrel-loopback-override.mjs:/app/umbrel-loopback-override.mjs:ro
+      - ${APP_DATA_DIR}/data/dsh-home:/data/.dsh
+      - ${APP_DATA_DIR}/data/workspace:/data/workspace
+      - ${APP_DATA_DIR}/data/logs:/logs
--- a/deepseek-harness/hooks/pre-start
+++ b/deepseek-harness/hooks/pre-start
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+# pre-start: ensure runtime-writable data dirs exist with app ownership.
+# umbrelOS creates missing bind-mount sources as root; the app runs as
+# 1000:1000 and must write its home, workspace, and shared log dir.
+set -euo pipefail
+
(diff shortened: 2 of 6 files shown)


mauricewipf:
### Screenshots

Web UI — active session:

screenshot-2026-09-07_20-54-32-cropped

Trajectory view:

screenshot-2026-09-07_20-57-58-cropped

Settings -> Models:

screenshot-2026-09-07_20-58-45-cropped

Settings -> Plugins:

screenshot-2026-09-07_20-59-15-cropped

New session / mode picker:

screenshot-2026-09-07_21-00-27-cropped

PR #6075, opened 2026-09-09
Update PeerSwap to v7.0.0.1

Impa10r:
## Summary

Allows running PeerSwap without Liquid Network

## Verification

Umbrel testing performed:

Environment tested:
- [ ] Umbrel device
- [x] Local umbrelOS test environment 1.7.4 only
- [ ] Not runtime tested

Architecture tested:
- [x] amd64
- [ ] arm64

### DIFF
--- a/peerswap/docker-compose.yml
+++ b/peerswap/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 1984
 
   web:
-    image: ghcr.io/impa10r/peerswap-web:v7.0.0@sha256:1d53f4409506772c3d030e2280af0f7dc3e61440326e94d964158200589241b3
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks @Impa10r. Is it intentional that enabling Liquid now requires a manual config change? In my fresh-install test, removing the Elements dependency left its RPC port unset, so PeerSwap saved `18884` instead of Umbrel’s `7041`. There doesn’t seem to be a port setting in the UI.

Could we detect when Elements is installed and supply its port through `exports.sh`, with a small config update for existing BTC-only setups? That might let us keep Elements optional while making Liquid easy to enable.

Impa10r:
Hi. No, not intentional. Let me see what can be done.

Impa10r:
in fact, Blockstream pushed v7.0.1 now where liquid swaps are disabled until v7.0.2 will enable them again. Maybe we better update to v7.0.1 now?

Impa10r:
replaced with https://github.com/getumbrel/umbrel-apps/pull/6084

PR #6076, opened 2026-09-10
Update Roboats  to  v0.8.7-alpha

KoalaSat:
## Type

App update

## App

App ID: robosats
Upstream project: Robosats
Version: v0.8.7-alpha

## Summary

Full changelog here: https://github.com/RoboSats/robosats/releases/tag/v0.8.7-alpha

## Verification

Umbrel testing performed:

Environment tested:
- [ ] Umbrel device
- [ ] Not runtime tested

Architecture tested:
- [ ] amd64
- [ ] arm64

## Notes

### DIFF
--- a/robosats/docker-compose.yml
+++ b/robosats/docker-compose.yml
@@ -7,7 +7,7 @@ services:
       APP_PORT: 12596
 
   web:
-    image: recksato/robosats-client:v0.8.6-alpha@sha256:071634c1a73aa5eecdb4e7f84419319888fa10d238ae3ad2433a984301727fbe
+    image: recksato/robosats-client:v0.8.7-alpha@sha256:87b377ac3bde6fb5067fa64fd31ffbcf8861167f0e0a7f1d0d9bcc3c38cc75cb
     restart: on-failure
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for the update @KoalaSat! I tested the upgrade and hit one compatibility issue: the new client serves HTTPS, while Umbrel’s app proxy still connects to it over HTTP, so the app fails to open.

We have built-in HTTPS coming in umbrelOS 2.0, which is currently in beta, with the stable release scheduled for the 22nd. I think we can use that here by adding:

```yaml
manifestVersion: 2
requiresHttps: true
```

We’d also configure RoboSats to serve HTTP internally and let Umbrel handle HTTPS for the browser. That should give OpenPGP 6 the secure context it needs and resolve the proxy mismatch.

We could target this update for the 2.0 release and test the configuration, including robot creation and recovery, before going live. Does that sound okay to you?

KoalaSat:
> We could target this update for the 2.0 release and test the configuration, including robot creation and recovery, before going live. Does that sound okay to you?

Sounds great, thank you!

PR #6081, opened 2026-09-11
Remove nolooking

DanGould:
Unmaintained for several years and has known security issues.

## Type

App removal

## App

App ID: nolooking
Upstream project: https://github.com/payjoin/nolooking
Version: 0.2.2-alpha (all)

## Summary

Unmaintained for several years and has known security issues.

## Notes

The readme does say use with unknown node versions may result in permanent loss of funds, so that would be all node versions for many years now. There isn't any state that would lose people funds so I recommend encouraging users to uninstall the old app as well.

Someone would have to actively be opening channels this way to be exploited, but still, best to remove this unmaintained app.


nmfretz:
Thanks @DanGould. I’ve updated this PR to follow our usual removal process and added a notice recommending users uninstall nolooking, as you suggested.

PR #6086, opened 2026-09-12
Update Bitcoin Regtest Dashboard to v1.1.0

coreyphillips:
## Type

App update

## App

App ID: bitcoin-regtest-dashboard
Upstream project: https://github.com/coreyphillips/bitcoin-regtest-dashboard
Version: 1.1.0 (from 1.0.4)

## Summary

Updates Bitcoin Regtest Dashboard to v1.1.0. The theme of this release is making the app usable from outside its own browser tab, so other projects on the same machine can drive the regtest network.

- Adds a documented local HTTP API. The app already spoke JSON over HTTP, but only to feed its own frontend. It now publishes an index at `/api`, an OpenAPI 3.1 spec at `/api/openapi.json`, and a compact plain text reference at `/api/llms.txt`, so a test suite or an AI agent can discover and drive it.
- Adds a Logs tab that tails Bitcoin Core's `debug.log` and the dashboard's own output live, with filtering, pause and download, plus runtime control of Bitcoin Core's logging categories. The same logs are available over the API, which means troubleshooting no longer requires shell access to the host.
- Adds an API tab with connection details, copyable commands, and a searchable endpoint reference rendered from the served OpenAPI document.
- Adds higher level endpoints for testing: a faucet that funds an address and confirms it in one call (mining first when the wallet has no mature balance, since regtest coinbase needs 100 confirmations), a chain reorg endpoint, and long polling wait helpers.
- Adds optional API token authentication via `API_TOKEN`. It is off by default, so behavior is unchanged for existing installs.
- Upgrades the bundled Bitcoin Core node from v31.0 to v31.1, a bugfix release.

Image changes:

- dashboard: `coreylphillips/bitcoin-regtest-dashboard` v1.0.4 to v1.1.0 (built for amd64 and arm64)
- bitcoin: `ghcr.io/getumbrel/docker-bitcoind` v31.0 to v31.1

Compose changes beyond the image pins:

- The permission maintenance loop now also keeps `debug.log` readable, alongside the cookie it already handled. Bitcoin Core writes that file mode 0600 and the new log viewer needs to read it.
- Adds `BITCOIN_DEBUG_LOG` to the dashboard environment to make the log path explicit. The app derives it from the cookie path when unset, so this is documentation rather than a functional change.
- Adds `PROXY_AUTH_WHITELIST: "/api,/api/*"` to the app proxy. See below.

## On the auth whitelist

The dashboard UI stays behind umbrelOS authentication. Only the HTTP API is exempt.

The whole point of this release is that another machine can drive the regtest node, so requiring umbrelOS credentials on the API would mean handing a wallet test suite or an agent the user's umbrelOS password just to mine a block. This follows the established pattern in this repo; 23 apps already whitelist `/api/*`.

Two notes for reviewers:

- Both rules are needed. Reading `pathMatches` in `app-gateway.ts`, a `/api/*` rule takes the `endsWith('/*')` branch and matches only `requestPath.startsWith('/api/')`, so it does not match the bare `/api` index, which is the discovery endpoint a client reads first. `"/api*"` would cover both but also matches unrelated paths like `/apifoo`, so the explicit pair is used instead.
- This is a regtest-only app. The coins are worthless by construction and the node is disposable. For anyone who still wants the API locked down, this release adds `API_TOKEN` on the dashboard service, which requires a bearer token on every API route except the three discovery endpoints.

## Verification

Umbrel testing performed: Verified against a fresh self-contained stack built from the app repo. The node reports `/Satoshi:31.1.0/`. A smoke suite covering all 78 endpoints passes with and without an API token, starting from an empty chain so the faucet's mining path is exercised. Separately pulled the published multi-arch image and ran it as `user: "1000:1000"` against the same node, matching how this compose file runs it, and confirmed it boots, serves the frontend, and can read `debug.log` under that uid. The Logs and API tabs were exercised in a browser, including confirming a block mined over the API appears in the live log stream. Not yet installed and tested as an app on umbrelOS.

Environment tested:
- [x] Umbrel device
- [x] Not runtime tested on umbrelOS (runtime tested via docker compose only)

Architecture tested:
- [x] amd64
- [ ] arm64 (built, not runtime tested)

Known lint warnings or caveats: None.

## Notes

No new permissions, dependencies, migrations, or default credentials. The app stays self-contained, bundling its own Bitcoin Core regtest node and Electrs instance.

The only behavioral change for an existing install is the auth whitelist described above, which is scoped to `/api` and leaves the UI gated as before.

### DIFF
--- a/bitcoin-regtest-dashboard/docker-compose.yml
+++ b/bitcoin-regtest-dashboard/docker-compose.yml
@@ -5,6 +5,12 @@ services:
     environment:
       APP_HOST: bitcoin-regtest-dashboard_dashboard_1
       APP_PORT: 3000
+      # The dashboard UI stays behind umbrelOS authentication. The HTTP API is
+      # exempt so scripts, test suites and agents running on other machines can
+      # reach it without umbrelOS credentials, which is the point of the API.
+      # Both rules are needed: "/api/*" does not match the bare /api index.
+      # Set API_TOKEN on the dashboard service below to require a bearer token.
+      PROXY_AUTH_WHITELIST: "/api,/api/*"
 
   # Bitcoin Core in regtest mode (self-contained, no dependencies)
   bitcoin:
--- a/bitcoin-regtest-dashboard/docker-compose.yml
+++ b/bitcoin-regtest-dashboard/docker-compose.yml
@@ -8,7 +8,7 @@ services:
 
   # Bitcoin Core in regtest mode (self-contained, no dependencies)
   bitcoin:
-    image: ghcr.io/getumbrel/docker-bitcoind:v31.0@sha256:89185fc2792a9824cbe18f7ad4ead02a3a9a14adf5b34eb42f60ebec36201fa0
+    image: ghcr.io/getumbrel/docker-bitcoind:v31.1@sha256:e0a94f8390f4fee796faba306719860a1ba61e8fa841416bca26fd4e1e94cb8c
     user: "1000:1000"
     restart: on-failure
     entrypoint: ["/bin/sh", "-c"]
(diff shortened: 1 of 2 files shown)


nmfretz:
Thanks for the update @coreyphillips! Just wanted to double-check the intended behavior here. With the API whitelist and `API_TOKEN` unset, anyone who can reach the app can mine blocks, reset the regtest chain, or run RPC commands without logging into Umbrel.

I saw your note about making access easy for scripts and agents, which makes sense for a regtest tool. Is open access the intended default for Umbrel installs, or would you prefer to require an app-specific API token by default?

coreyphillips:
Hey @nmfretz, I appreciate you double-checking this. Yes, it's intended behavior at the moment for the regtest app to make troubleshooting and testing easier for agents.

nmfretz:
Thanks for confirming @coreyphillips! Appreciate the clarification, looks good to go!

PR #6093, opened 2026-09-15
pearcircle-seeder: update to 1.1.2

peerloomllc:
Bump PearCircle Seeder to 1.1.2 (image pinned to the multi-arch manifest-list digest).


nmfretz:
Thanks for the update @peerloomllc! I tested the upgrade and restart on Umbrel, and the seeder’s saved settings and enrollment were preserved. I also added release notes and resolved the conflict with master. Looking good from the packaging side!

PR #6094, opened 2026-09-16
lightningmate: update to 0.7.7

opensourceminers:
Updates Lightning Mate from 0.7.0-patch.1 to 0.7.5.

Image: `ghcr.io/opensourceminers/lightningmate:v0.7.5@sha256:09988286632ac1e79f320b9934913a2a5dfa1f6105afda1b301545f31306afac` (multi-arch amd64 + arm64, digest-pinned, built by CI from the tagged source).

Only `version`, `releaseNotes` and the image pin change. Everything the Umbrel team set (id, category, port, gallery, storage/dataRoot) is untouched.

### What's in it

**Magma liquidity selling, reworked.** Offers were priced against what other sellers were *asking* rather than what buyers actually paid, and the Autopilot cut the price a little further for every day without a sale. That loop is wrong at the root: a Magma purchase carries no offer id, so Amboss picks the seller — being cheapest wins nothing and only gives away margin. Pricing now uses Amboss's real completed-order history, the downward ratchet is gone, and a related bug that let the effective price collapse to the base fee is fixed. The app also now shows what share of real orders an offer's size window can even be matched with, which is usually the actual reason an offer sits unsold.

**Channel peer suggestions gained a trust signal.** Peers are checked against lightningnetwork.plus for rank, Prime status and the ratings operators gave each other. Peers with no LN+ profile are never penalised, and the signal only refines the existing graph-based ranking. Requested by a community node runner in opensourceminers/lightningmate#1.

**Also since 0.7.0:** a Routing view (per-channel flow direction, routing corridors, forwards refused for lack of liquidity), rebalancing fixes so the autopilot refills earning channels instead of only repricing them, a cleaner Performance tab, and an expired Amboss key now surfaces in Settings instead of failing silently.

No changes to permissions, ports, dependencies or the data layout.


nmfretz:
Thanks @opensourceminers! Packaging looks good. I noticed two things in 0.7.6: upgrading enables auto-sizing even when auto-pricing was off, and auto-sizing can change prices while auto-pricing is disabled.

Would you like us to hold off until those are fixed, or go live now and address them in the next update?

opensourceminers:
To be clear: thanks for the catch, both are fixed in 0.7.7 and this PR now carries that version. Good to go from our side.

